PR #9, opened 2018-09-19
Complete documentation

### LINKED ISSUE #6: Finish documenting code in rocketpyAlpha.py
The code in [nbks/rocketpyAlpha.py](https://github.com/giovaniceotto/RocketPy/blob/master/nbks/rocketpyAlpha.py) is not fully documented. Methods in classes Motor, Rocket and Flight are missing docstring documentation.

### LINKED ISSUE #7: Return Behavior
Some methods in [rocketpyAlpha ](https://github.com/giovaniceotto/RocketPy/blob/master/nbks/rocketpyAlpha.py)return **_None_**, while others return **_self_**. This behavior should be consistent so that its predictable. 

I see no problems in all methods which edit an object either returning the object itself or valuable information relating to the edit.

giovaniceotto:
Pull request for:

- Functions are now fully documented in RocketpyAlpha.py (fixes #6) 
- Return behavior of functions is now predictable (fixes #7)

### DIFF
--- a/nbks/rocketpyAlpha.py
+++ b/nbks/rocketpyAlpha.py
@@ -1453,7 +1453,89 @@ class Environment:
 
 class Motor:
     """Class to specify characteriscts and useful operations for solid
-    motors."""
+    motors.
+    
+    Class attributes:
+
+    Geometrical attributes:
+    Motor.nozzleRadius : float
+        Radius of motor nozzle outlet in meters.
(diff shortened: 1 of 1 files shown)


giovaniceotto:
Documentation seems good, although improvements can be made in the future to make it more uniform, using the same terms through out the document etc.

But this is definitely good enough for now.

PR #37, opened 2020-07-28
Minor changes for py3.8 to comply with issue34850

andre-geldenhuis:
There are some warnings generated when using python 3.8 due to [issue34850](https://bugs.python.org/issue34850).

I made some very minor syntax changes to address these warnings.

### DIFF
--- a/rocketpy/Flight.py
+++ b/rocketpy/Flight.py
@@ -787,7 +787,7 @@ class Flight:
                     # print('\t\t\tAltitude: ', phase.solver.y[2])
                     # print('\t\t\tEvals: ', self.functionEvaluationsPerTimeStep[-1])
                     # Check for out of rail event
-                    if self.outOfRailState is 0 and (
+                    if self.outOfRailState == 0 and (
                         self.y[0] ** 2
                         + self.y[1] ** 2
                         + (self.y[2] - self.env.elevation) ** 2
@@ -869,7 +869,7 @@ class Flight:
                         phase.timeNodes.addNode(self.t, [], [])
                         phase.solver.status = "finished"
                     # Check for apogee event
-                    if self.apogeeState is 0 and self.y[5] < 0:
+                    if self.apogeeState == 0 and self.y[5] < 0:
                         # print('\nPASSIVE EVENT DETECTED')
                         # print('Rocket Has Reached Apogee!')
                         # Apogee reported
@@ -2016,7 +2016,7 @@ class Flight:
             print(name + " Parachute Inflated at Height of: {:.3f} m".format(altitude))
 
         # Print impact conditions
-        if not (self.impactState is 0):
+        if self.impactState != 0:
             print("\n\nImpact\n")
             print("X Impact: {:.3f} m".format(self.xImpact))
             print("Y Impact: {:.3f} m".format(self.yImpact))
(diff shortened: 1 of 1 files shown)


andre-geldenhuis:
The variables outOfRailState, apogeeState, impactState are sometimes boolean and sometimes an array.  With python 3.8 this causes an error as the truth value of an array with more than one element is ambiguous.

I initialised the variables as single element arrays containing 0 and testing if all the values inside are 0 as this keeps the logic similar to before.  However, I'm not sure that's the clearest approach.  It might be clearer to test the array length?

giovaniceotto:
Thanks for bringing up this issue!

I agree with you, perhaps checking the length of the array would be the best option to solve the issue. I ran some speed tests to verify if there is any significant difference between them in terms of execution time and this is the result:

```[python]
>>> timeit.timeit(stmt='a==0', setup='a=0') 
0.047009199999997975

>>> timeit.timeit(stmt='a is 0', setup='a=0') 
<timeit-src>:2: SyntaxWarning: "is" with a literal. Did you mean "=="?
<timeit-src>:6: SyntaxWarning: "is" with a literal. Did you mean "=="?
0.04326839999999521

>>> timeit.timeit(stmt='len(a) == 0', setup='import numpy as np; a = np.array([])')
0.08204170000000488

>>> timeit.timeit(stmt='(a==0).all()', setup='import numpy as np; a = np.array([0])') 
3.8522727000000145
```

As we can see, it appears that using `(a==0).all()` is significantly more time consuming. Doing some quick calculations, estimating that each flight simulation can reach up to 10k iterations, checking two if conditions like this, using `(a==0).all()` instead of `len(a) == 0` could increase simulation time by 100 ms.

What do you think @andre-geldenhuis?

rocketpy/Flight.py
```diff
@@ -787,7 +787,7 @@ def __init__(
                     # print('\t\t\tAltitude: ', phase.solver.y[2])
                     # print('\t\t\tEvals: ', self.functionEvaluationsPerTimeStep[-1])
                     # Check for out of rail event
-                    if self.outOfRailState is 0 and (
+                    if (self.outOfRailState == 0).all() and (
```
giovaniceotto, line 822:
```suggestion
                    if len(self.outOfRailState) == 1 and (
```

rocketpy/Flight.py
```diff
@@ -869,7 +869,7 @@ def __init__(
                         phase.timeNodes.addNode(self.t, [], [])
                         phase.solver.status = "finished"
                     # Check for apogee event
-                    if self.apogeeState is 0 and self.y[5] < 0:
+                    if (self.apogeeState == 0).all() and self.y[5] < 0:
```
giovaniceotto, line 905:
```suggestion
                    if len(self.apogeeState) == 1 and self.y[5] < 0:
```

rocketpy/Flight.py
```diff
@@ -2016,7 +2016,7 @@ def info(self):
             print(name + " Parachute Inflated at Height of: {:.3f} m".format(altitude))
 
         # Print impact conditions
-        if not (self.impactState is 0):
+        if (self.impactState != 0).all():
```
giovaniceotto, line 2113:
```suggestion
        if len(self.impactState) != 0:
```

PR #47, opened 2020-09-23
Airfoil implementation

brunosorban:
The commit brings the implementation to use airfoiled fins simulation using a .csv file as input for the lift coefficient. At _Flight_, two minor changes were made. The first was to move the access to lift coefficient at Flight class to the end because _Udot()_ function initializes without the _compAttackAngle_ variable. Lift force calculus were also changed because _cLift_ is the lift coefficient and not the lift coefficient derivative anymore (the name were also changed). 

At Rocket class there are major changes. _addNonecone()_, _addTail()_ and _AddFins()_ don’t return a derivative anymore at _aerodynamicSurfaces[1][1]_. Now, at the position, an object of Function class is returned instead with a function that describes the component’s lift by attack angle. At nosecone and tail, it is a linear function with derivative _clalpha_. At fins, it either be a linear function using previous implementation of _clalpha_, or be a function of the .csv input linearly interpolated. Minor changes at _evaluateStaticMargin()_ and _allInfo()_ were also made to make those functions compatible with the new lift coefficient implementation.


rocketpy/Flight.py
```diff
@@ -1315,14 +1314,14 @@ def uDot(self, t, u, postProcessing=False):
                 compStreamVzBn = compStreamVzB / compStreamSpeed
                 if -1 * compStreamVzBn < 1:
                     compAttackAngle = np.arccos(-compStreamVzBn)
+                    cLift = abs(aerodynamicSurface[1](compAttackAngle))
```
giovaniceotto, line 1317: This seems perfectly correct, but I am curious about the use of `abs` here. Isn't `compAttackAngle` always positive?

giovaniceotto: On second thought, I think your decision to use `abs` makes the code more robust, making sure the lift force direction is not affected by `cLift`.

rocketpy/Rocket.py
```diff
@@ -420,9 +423,10 @@ def addTail(self, topRadius, bottomRadius, length, distanceToCM):
 
         # Calculate clalpha
         clalpha = -2 * (1 - r ** (-2)) * (topRadius / rref) ** 2
+        self.cldata = Function(lambda x: clalpha*x, 'Alpha (rad)', 'Cl', interpolation='linear')
```
giovaniceotto, line 426: I am not sure this works as you intended. Using `self.cldata` sets the Rocket's attribute `cldata`, not `cldata` for the tail only.

giovaniceotto (changes requested):

rocketpy/Flight.py
brunosorban: I was in doubt if `compAttackAngle` were always positive and, as the coefficient must always be positive, abs was added. But at principle it isnt't really necessary

rocketpy/Rocket.py
brunosorban: You are right, that explains why the corrections is so strong. I'll fix it.

rocketpy/Flight.py
brunosorban: What do you suggest?

giovaniceotto (approved):
Excelent!

PR #50, opened 2020-10-11
Finally NASA and RocketPy working together

Gui-FernandesBR:
Hello everyone, 
I made some changes in the Environment class and now RocketPy can handle with earth Topography. Of course this is not a final version yet, but I think it will be helpful in the future. 

The main source of elevation data that I found was the NASA Digital Model, described [here](https://earthdata.nasa.gov/esds/competitive-programs/measures/nasadem). I could import those data in a netCDF4 format, which I think is a great because the Environment has always handled the netCDF4 package. For more specific details, check this [wesite](https://cmr.earthdata.nasa.gov/search/concepts/C1546314436-LPDAAC_ECS.html).

The functions (or methods) that I implemented are the following:
1. setTopographicProfile: 
         Now we can define an "topographic profile", very similar to the usual wind or pressure profiles that we have in Environment Class.
2. getElevationFromTopograghicProfile:
         Once we define  the topographic profile, we are able to get elevation information of any point inside the covered area
3. geodesicToUtm:
         Function to convert lat/lon to UTM
4. utmToGeodesic:
         Function to convert UTM to lat/lon
5. calculateEarthRadius:
         The Earth Radius varies with the latitude, as described [here](https://en.wikipedia.org/wiki/Earth_radius), and I implemented these calculations in this function
6. decimalDegressToArcSeconds:
          Simple function to convert decimal degrees to deg/arc-min/arc-sec
7. printEarthDetails:
          I create this function for printing all information about Earth models in the futere, for now we only have

I would like to ask help by reviewing these improvements, please, @giovaniceotto .


giovaniceotto:
Looks like a stroke of genius @Gui-FernandesBR! It also seems quite complex. Perhaps a 30 minute talk over discord can help me with this review. Shall we?

Gui-FernandesBR:
Of course, @giovaniceotto ! 

I finally will have the chance of meeting you for the first time. I heard you look like a wizard, is that true?

Let's schedule this talk for the next Saturday **(october 17th) at 16 UTC,** is that ok for you?

Or better, let's make this way:
`talk = Environment.setDate(2020, 10, 17, 16)`

PR #51, opened 2020-10-11
New TO-DO list

Gui-FernandesBR:
Hey, now we have a column for "current stage" so we can say if the task was already started or not. Hope ypu like it!


giovaniceotto (approved):
Really liked the new tasks added, along with the new column. Makes me thing if we can make better use of GitHub Projects to organized all of this.

PR #73, opened 2021-08-03
Bug/mixed calendars

Lucas-KB:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`python setup.py install`) was run locally and any changes were pushed
- [x] Lint (`black rocketpy`) has passed locally and any fixes were made for failures

## Pull request type
 
 

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes

The code throws an error when the user tries to set the atmospheric model to anything that uses netCDF4 with a time interval of less than a day from now.

## What is the new behavior?

The code uses only the Gregorian calendar.

- [ ] Yes
- [x] No

Special thanks to @ompro07 @lucasfourier and @FranzYuri


giovaniceotto:
@Lucas-KB Thanks for the fix! Out of curiosity, what caused the error to happen exactly? What made netCDF4 use a different calendar then expected?

Lucas-KB:
I am not entirely sure. According to the netCDF4 documentation, if the calendar type is not specified, it will use the same as the input data, which would be correct, but apparently it doesn't work for some reason. We tried specifying it manually and it apparently works.

giovaniceotto:
Any chance some input data can use a different calendar then Gregorian?

giovaniceotto (approved):

Gui-FernandesBR:
I finally have tested the proposed code and I believe it is working very well.
However, @Lucas-KB , I wonder if we can get some troubles dealing with different files in the future (I say this based on @giovaniceotto's last comment). 
I mean, it seems you made an assumption that every file imported by rocketpy through netCDF4 will be specified in gregorian calendar. Would not be safer if we first read the file, then determine which type of calendar we need to use?
For instance:
- The user specify a .nc file containing environment data;
- RocketPy reads the .nc (using the netCDF4 library) and try to find the "calendar" attribute;
- The next methods will use only this same "calendar" type, even it is gregorian or not.

I know it may sounds a little "unexciting" to-do, but I think it would prevent us to get at this same issue again in the fututre.
Maybe we can approve this PR with current version and write a warning in documentation for future analysis.

Anyway, I appreciate your work, that's a masterpiece on working with the netCDF4/RocketPy interface.

Lucas-KB:
@Gui-FernandesBR I have researched some more into this matter and as far as I could understand, setting the calendar type to Gregorian means that the dates inputted by the user are considered to be Gregorian, not the dates from the files imported through netCDF4. I found the documentations a bit confusing, but the conclusion I drew was that the calendar of the file is determined by the units parameter of the time variable of the dataset and the date inputted by the user is compared with the one in the dataset based on the calendar selected.

References:

- [https://unidata.github.io/netcdf4-python/](url)
- [https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch04s04.html](url)
- [https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch07s04.html](url)

Gui-FernandesBR:
That's good @Lucas-KB . 
It seems you're saying that "as long as the user input the date in a specific format compatible with the gregorian calendar, we probably will not have problems with calendar attributes anymore."

If that's right, I trust your statements. And of course, I have to say that I appreciate your answers.

So I hope you can merge this one @giovaniceotto , please.

Gui-FernandesBR (approved):

giovaniceotto (approved):
It is worth noting that Python uses the _proleptic Gregorian_ calendar, which might not be standard.

If we were only talking about dates after the year 1582, I wouldn't worry. However, our code uses the function `date2num`, which converts a date to a number. This number can represent different things. For GFS, it represents 'days since 1-1-1 00:00:0.0', which is the unit used by GFS. Notice that this means computations involving dates before 1582 are done.

GFS does not declare which calendar it uses in the .nc file we retrieve via OpenDAP, which it should (at least I could not find it). It seems to be _proleptic Gregorian_ as well since things just work.

I am not entirely sure what will be the behavior in each case, so let's attempt to create good unit tests to make sure this works properly. For now, I am confident given what @Lucas-KB said that it works in most use cases.

giovaniceotto:
@Lucas-KB can you solve the conflicts and merge?

Lucas-KB:
@giovaniceotto Yes, I can do it

PR #80, opened 2021-08-08
Unit test motor class

PatrickSampaioUSP:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`python setup.py install`) was run locally and any changes were pushed
- [x] Lint (`black rocketpy`) has passed locally and any fixes were made for failures

## Pull request type
 
 

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [x] Other (please describe): 

- [ ] Yes
- [x] No

## Other information

Added 95% coverage unit test for SolidMotor class. In general lines the tests are validating the values calculated by the SolidMotor classes, but not verifying every single value, but doing a more analytical assert to check if the results make sense.


rocketpy/SolidMotor.py
```diff
@@ -689,7 +711,7 @@ def exportEng(self, fileName, motorName):
         )
 
         # Write thrust curve data points
-        for item in self.thrust.source[:-1, :]:
+        for item in self.thrust.source[:-1, :][1:]:
```
PatrickSampaioUSP, line 714: @giovaniceotto Could you validate this? The previous code was adding two lines with 0.0 0.0 in the .eng file

giovaniceotto: Just checked here and I couln't reproduce it. Trying the following block of code with your implementation, I get a file missing the initial 0.0, 0.0 point.

## Code
```python
from rocketpy import SolidMotor

Pro75M1670 = SolidMotor(
    thrustSource="data/motors/Cesaroni_M1670.eng",
    burnOut=3.9,
    grainNumber=5,
    grainSeparation=5/1000,
    grainDensity=1815,
    grainOuterRadius=33/1000,
    grainInitialInnerRadius=15/1000,
    grainInitialHeight=120/1000,
    nozzleRadius=33/1000,
    throatRadius=11/1000,
    interpolationMethod='linear'
)

Pro75M1670.exportEng('test.csv', 'Pro75M1670')
```

## Result file test.csv
```csv
Pro75M1670 66.0 625.0 0 2.96 2.96 PJ 
0.0550 100.000
0.0920 1500.000
0.1000 2000.000
0.1500 2200.000
0.2000 1800.000
0.5000 1950.000
1.0000 2034.000
1.5000 2000.000
2.0000 1900.000
2.5000 1760.000
2.9000 1700.000
3.0000 1650.000
3.3000 530.000
3.4000 350.000
3.9000 0.000
```

PatrickSampaioUSP: You are correct. 

What motivated me to make this change is that the `importEng` method already initialize the `dataPoints` variable with a `[[0, 0]]`. Which results in our exported file to be read with two rows with zeros, what I thinks should not be ideal. What do you think it is the best solutions for this issue?

I see that we either remove the initial `[0, 0]` from our exported file or we do not assume the `[0, 0]` in the `dataPoints` variable. A third option would be to check if the input file already has a `[0, 0]` row.

PatrickSampaioUSP:
@giovaniceotto We are going to save the .eng file without the initial 0,0 and the importEng will add the initial 0,0 in the datapoints, right? The code is already producing this behaviour.

giovaniceotto (approved):
Seems like everything is working as expected!

PR #84, opened 2021-08-17
Evaluate Travis-CI build failure reports

giovaniceotto:
Trying to simulate conditions which will cause tests to fail in order to check if the notification shown by Travis-CI facilitates debugging.


giovaniceotto:
As this was just a test PR, I'm closing it for now.

PR #108, opened 2021-10-29
Enh/parachute

FranzYuri:
## Pull request type

## Pull request checklist

Currently there is no place to keep functions that are useful but do not belong to any of the 5 core classes in Rocketpy.
One of these functions is the compute_cds_from_drop_test, that calculates the CdS of a parachute from a drop test.

## What is the new behavior?

With the module utilities.py, it is possible to save functions like the compute_cds_from_drop_test.

- [ ] Yes
- [X] No

## Other information

Enter text here...

### DIFF
--- a/rocketpy/utilities.py
+++ b/rocketpy/utilities.py
@@ -22,9 +22,10 @@ from matplotlib import cm
 
 from rocketpy import Environment, Rocket, SolidMotor, Flight
 
-def compute_cds_from_drop_test(final_air_density, final_speed, rocket_mass, gravity):
+def compute_cds_from_drop_test(final_speed, final_air_density, rocket_mass, gravity):
     """Returns the parachute's Cds calculated through its final speed, air 
-    density, the rocket's mass and the force of gravity in the landing point.
+    density of the landing point, the rocket's mass and the force of gravity
+    in the landing point.
 
     Parameters
     ----------
--- a/rocketpy/utilities.py
+++ b/rocketpy/utilities.py
@@ -22,8 +22,9 @@ from matplotlib import cm
 
 from rocketpy import Environment, Rocket, SolidMotor, Flight
 
+
 def compute_cds_from_drop_test(final_speed, final_air_density, rocket_mass, gravity):
-    """Returns the parachute's Cds calculated through its final speed, air 
+    """Returns the parachute's Cds calculated through its final speed, air
     density in the landing point, the rocket's mass and the force of gravity
     in the landing point.
 
@@ -44,6 +45,6 @@ def compute_cds_from_drop_test(final_speed, final_air_density, rocket_mass, grav
         Number equal to drag coefficient times reference area for parachute.
 
     """
-        
+
     cds = 2 * rocket_mass * gravity / ((final_speed ** 2) * final_air_density)
     return cds
--- a/rocketpy/utilities.py
+++ b/rocketpy/utilities.py
@@ -5,32 +5,31 @@ __copyright__ = "Copyright 20XX, Projeto Jupiter"
 __license__ = "MIT"
 
 
-
-def compute_CdS_from_drop_test(terminal_velocity, rocket_mass, air_density=1.225, gravity=9.80665):
-
-    """Returns the parachute's Cds calculated through its final speed, air
+def compute_CdS_from_drop_test(
+    terminal_velocity, rocket_mass, air_density=1.225, gravity=9.80665
+):
+    """Returns the parachute's CdS calculated through its final speed, air
     density in the landing point, the rocket's mass and the force of gravity
     in the landing point.
 
     Parameters
     ----------
-    final_speed : float
-        Rocket's speed in m/s^2 when landing.
-    final_air_density : float, optional
-        Air density, in kg/m^3, right before the rocket lands. Default value is 1.225.
-
+    terminal_velocity : float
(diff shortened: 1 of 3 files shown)


Lucas-KB:
@FranzYuri, I would suggest you run black to reformat your code:
  1. Install black running `pip install black`
  2. Run `black rocketpy` when in the repository folder
 
The second thing is that it may be necessary to add something like `from .utilities import *` or `from .utilities import compute_cds_from_drop_test` to the `__init__.py` file, so the import process can be achieved using `from rocketpy import *`.

Gui-FernandesBR:
> Currently there is no place to keep functions that are useful but do not belong to any of the 5 core classes in Rocketpy. One of these functions is the compute_cds_from_drop_test, that calculates the CdS of a parachute from a drop test.

Can you guys please indicate other functions with that same behavior in the code?

Lucas-KB:
@Gui-FernandesBR for now I haven't personally identified any function that applies. But I imagine we could use this file in the future to gather functions to help design Rockets/Parachutes/Motors etc. For example functions related to rocket parameter optimizations (ideal mass for a certain motor or something similar). I think the suggestion to create `utilities.py` came from @giovaniceotto and @FranzYuri, maybe they had different ideas.

Gui-FernandesBR:
Tks for explaning that, @Lucas-KB . If I'm understamdimg this correctly, I have some candidates to be included at utilities.py, let's see:
> **Environment:**
> def decimalDegressToArcSeconds(self, angle):
> 
> **Flight:**
> def calculateFinFlutterAnalysis(self, finThickness, shearModulus):
> def exportPressures(self, fileName, timeStep):

Those could help to give more "body" to this new .py file, and they could be moved in another PR I guess. Please let me know if these suggestions are good or not.

Also, I would recommend creating a simple example usage for these functions. At least for me, it is not easy to visualize how I could use the new capabilities.

giovaniceotto:
Excellent suggestions @Gui-FernandesBR! I hope you are ok with me creating a new issue regarding your ideas so that we can come back to them in a future PR. The issue is here #111.

rocketpy/utilities.py
```diff
@@ -0,0 +1,50 @@
...
+from collections import namedtuple
+
+import numpy as np
+from scipy import integrate
+from scipy import linalg
+import matplotlib.pyplot as plt
+from mpl_toolkits.mplot3d import Axes3D
+from matplotlib import cm
+
+from rocketpy import Environment, Rocket, SolidMotor, Flight
```
giovaniceotto, line 23: I believe these import statements are not needed, so I would suggest removing them.

```suggestion
```

rocketpy/utilities.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def compute_cds_from_drop_test(final_speed, final_air_density, rocket_mass, gravity):
```
giovaniceotto, line 26: I may be wrong, but is it more common to find `CdS` in the literature, instead of `cds`?
Would `terminal_velocity` be a more technical term when compared with `final_speed`?
Would it be a bad idea to add default values for the air density and gravity?

```suggestion
def compute_CdS_from_drop_test(terminal_velocity, rocket_mass, air_density=1.225, gravity=9.80665):

```

rocketpy/utilities.py
```diff
@@ -0,0 +1,50 @@
...
+    Parameters
+    ----------
+    final_speed : float
+        Rocket's speed in m/s^2 when landing.
+    final_air_density : float
+        Air density, in kg/m^3, right before the rocket lands.
+    rocket_mass : float
+        Rocket's mass in kg.
+    gravity : float
+        force of gravity in the landing point in newtons.
```
giovaniceotto, line 40:
```suggestion
        Gravitational acceleration experienced by the rocket and parachute during descent in m/s^2. Default value is the standard gravity, 9.80665.

```

rocketpy/utilities.py
```diff
@@ -0,0 +1,50 @@
...
+    """Returns the parachute's Cds calculated through its final speed, air
+    density in the landing point, the rocket's mass and the force of gravity
+    in the landing point.
+
+    Parameters
+    ----------
+    final_speed : float
+        Rocket's speed in m/s^2 when landing.
+    final_air_density : float
+        Air density, in kg/m^3, right before the rocket lands.
```
giovaniceotto, line 36:
```suggestion
    final_air_density : float, optional
        Air density, in kg/m^3, right before the rocket lands. Default value is 1.225.

```

giovaniceotto (changes requested):
Looks awesome! Just a few details to iron out and then we can merge this PR, which marks a new path for RocketPy functionalities.

Gui-FernandesBR:
Don't need my permission @giovaniceotto ... Thanks for opening the 111, I appreciate it

rocketpy/utilities.py
FranzYuri: Indeed CdS is the most precise way to write, I just wrote "cds" because I thought it would be best to use only letters in lowercase.

FranzYuri:
@giovaniceotto I think it is alright now, can you take a look?

giovaniceotto (approved):

giovaniceotto:
@FranzYuri, my bad, I had reviewed your changes but forgot to approve it!

Looks perfect now. Would you do the honors of merging the pull request?

PR #110, opened 2021-11-09
Enh/utc time

FranzYuri:
## Pull request type

## Pull request checklist

Currently the output dates on RocketPy are only available in UTC time zone.

## What is the new behavior?

Now RocketPy can print dates on local time zones.

- [ ] Yes
- [X] No

## Other information

Tanks to @ompro07 for all help, especially with the contributions at <https://github.com/Projeto-Jupiter/Hackathon/pull/97/files>


giovaniceotto:
Haven't taken a complete look at this PR just yet, but it seems like the tests are failing because of the `pytz` library.

@FranzYuri, can you modify the `requirements.txt` file in the root folder and add `pytz` as a required module?

Let me know if you have any doubts.

Gui-FernandesBR:
@giovaniceotto , as requested for you, the requirements.txt are now updated. Can you please confirm that everything is in the right format? This file is not familiar for me, but I think that maybe we need specify package version, right?

giovaniceotto:
I am sorry I haven't had time to review this yet. I will need a little more time since this uses a library that I am not familiar with. 
Help from @Lucas-KB and @ompro07 would be awesome.

Gui-FernandesBR:
@Lucas-KB , could you please take a look at this pull request for us?

rocketpy/Environment.py
```diff
@@ -347,6 +348,12 @@ def __init__(
         else:
             self.date = None
 
+        # Save local date
+        self.local_date = None
+
+        # Save local time zone
+        self.time_zone = None
+
```
giovaniceotto, line 356: Please add `Environment.local_date` and `Environment.time_zone` to the class docstring near line 85. Example:

```
...
Environment.date : datetime
            Date time of launch in UTC.
Environment.local_date : datetime
            Date time of launch in the local time_zone, defined by Environment.time_zone.
Environment.time_zone : string
            Local time zone specification.

...
```

rocketpy/Environment.py
```diff
@@ -378,14 +385,17 @@ def __init__(
...
-        date dependent atmospheric model is used.
+        date dependent atmospheric model is used. To see all time zones use
+        print(pytz.all_timezones).
 
         Parameters
         ----------
         date : Date
             Date object specifying launch date and time.
+        time_zone : string, optional
+            Name of the time zone.
```
giovaniceotto, line 398:
```suggestion
            Name of the time zone in which the date argument is given. Default is None, in which case UTC is used as the time_zone. To see all time zones available, import pyzt and run print(pytz.all_timezones).
```

rocketpy/Environment.py
```diff
@@ -378,14 +385,17 @@ def __init__(
 
         return None
 
-    def setDate(self, date):
+    def setDate(self, date, time_zone=None):
```
giovaniceotto, line 388: Can we make it so that the default entry for the time_zone argument is "UTC" instead of `None`? If so, I believe we wouldn't need the `if` statement in line 407, making the code cleaner. (I don't know if it would work as expected...)

rocketpy/Environment.py
```diff
@@ -394,6 +404,11 @@ def setDate(self, date):
         # Store date
         self.date = datetime(*date)
 
+        if time_zone != None:
+            self.time_zone = time_zone
+            tz = pytz.timezone(self.time_zone)
+            self.local_date = self.date.replace(tzinfo=pytz.UTC).astimezone(tz)
+
```
giovaniceotto, line 411: I am not sure that this is working as I expected. I expected that the user inputs the time in the local time zone, specifies which time zone is he using and the RocketPy internally converts this time to UTC in order to work properly. 

These lines seem to be doing the oposite. The user needs to input the datetime in UTC and specify a time zone. Then these lines are calculating what is the local time in such time zone based on the UTC time given. And then this local time is printed in the info methods. Although this is useful, I believe letting the user enter the local time is better than having to input time always in UTC.

giovaniceotto (changes requested):
Took a while, but I finally got to reviewing this masterpiece! Magnificent work @FranzYuri and @ompro07! 

I made some comments, suggestions and raised a question. Let me know what you think. If needed, we can discuss further via discord.

giovaniceotto:
@FranzYuri, I chipped in and did some minor code refactoring. Here are the most significant changes I have made:

- Introduced the `time_zone` argument in the `Environment.__init__` constructor. I did this since it is also possible to set the datetime in this constructor, so I believe it would be important to have the time zone there as well.
- Changed the way that the local time is converted to UTC. Before, we had
`self.date = self.localDate.replace(tzinfo=pytz.UTC)` which I modified to `self.date = self.localDate.astimezone(pytz.UTC)`. This converts the local date to UTC, which is what RocketPy needs to correctly interact with weather models.
- Minor changes to how the time is printed. Now, both time zones are shown in the same line.
- Minor code refactoring, converting snake_case to camelCase since we are still using this standard. This should be reverted when we release v1.0.0.

I believe there are only two things left:

- Create a test for this new feature. 
- Add an example of this new feature to the notebook at `RocketPy/docs/notebooks/environment_class_usage.ipynb`.

Do you wish to work on these items on this PR? Otherwise, we can merge this PR after you review my changes and then create a new one focused on the two items left.

giovaniceotto (approved):

Gui-FernandesBR:
Hey, I'd like to recommend merging this PR and then start a new one including the examples. 

@FranzYuri do u agree?

FranzYuri:
> Hey, I'd like to recommend merging this PR and then start a new one including the examples.
> 
> @FranzYuri do u agree?

Yes! Sorry for the delay...

PR #112, opened 2021-11-19
WIP: Enh/estimate drag coefficients

Lucas-KB:
## Pull request type

## Pull request checklist

The user has to provide a drag curve as a function of the Mach number.

## What is the new behavior?

The user can supply additional rocket parameters instead of the curves themselves, but the old behavior can still be achieved.

- [ ] Yes
- [X] No


rocketpy/Rocket.py
```diff
@@ -278,8 +309,138 @@ def __init__(
...
+        Returns
+        -------
+        self.viscous_friction_coefficient : Function
+            Function of mach number expressing the viscous friction coefficient,
+            defined as in the literature.
+        """
+
+        def calculations(mach):
+            reynolds = (
+                (self.air_density ** 2)
```
giovaniceotto, line 370: @Gui-FernandesBR pointed out that this should be:

```suggestion
            reynolds = (
                (self.air_density)
```

rocketpy/Rocket.py
```diff
@@ -165,8 +165,12 @@ def __init__(
         radius,
         distanceRocketNozzle,
         distanceRocketPropellant,
-        powerOffDrag,
-        powerOnDrag,
+        powerOffDrag=None,
+        powerOnDrag=None,
+        rocket_length=None,
+        speed_of_sound=340.293988026089,
```
Gui-FernandesBR, line 171: Is it really necessary using so many decimal places? I mean, what a precise number you got here... hahahahahaha

rocketpy/Rocket.py
```diff
@@ -207,11 +211,30 @@ def __init__(
             information. If int or float is given, it is assumed constant. If
             callable, string or array is given, it must be a function of Mach
             number only.
-
+        rocket_length : int, float, optional
+            Rocket length used in the expressions for estimating drag coefficients
+            in meters (m).
+            Air density used in the expressions for estimating drag coefficients
```
Gui-FernandesBR, line 217: Lines 216 and 217 can be deleted, they are redudant here

rocketpy/Rocket.py
```diff
@@ -278,8 +309,138 @@ def __init__(
         # Evaluate static margin (even though no aerodynamic surfaces are present yet)
         self.evaluateStaticMargin()
 
+        self.evaluateDragCoefficient()
+
         return None
 
+    def evaluateDragCoefficient(self):
+        """Calculates and returns the drag coefficient as the sum of
+        of the individual drag coefficients at zero angle of attack
```
Gui-FernandesBR, line 318: the word 'of' is duplicated here!

rocketpy/Rocket.py
```diff
@@ -278,8 +309,138 @@ def __init__(
...
+        self.viscous_friction_coefficient : Function
+            Function of mach number expressing the viscous friction coefficient,
+            defined as in the literature.
+        """
+
+        def calculations(mach):
+            reynolds = (
+                (self.air_density ** 2)
+                * self.speed_of_sound
+                * self.rocket_length
```
Gui-FernandesBR, line 372: Is there a specific reason for using rocket_length as the reference length instead of diameter? I believe different literatures has different opinons on that, but I'm not an expert tough

rocketpy/Rocket.py
```diff
@@ -278,8 +309,138 @@ def __init__(
...
+        as a function of the Mach number. The angle of attack is
+        assumed to be zero.
+
+        Parameters
+        ----------
+        None
+
+        Returns
+        -------
+        self.viscous_friction_coefficient : Function
```
Gui-FernandesBR, line 328: I think the return is outdated here

rocketpy/Rocket.py
```diff
@@ -278,8 +309,138 @@ def __init__(
...
+            Function of mach number expressing the viscous friction coefficient,
+            defined as in the literature.
+        """
+        self.evaluateViscousFrictionCoefficient()
+        self.evaluateForebodyDragCoefficient()
+
+        self.drag_coefficient_estimate = self.forebody_drag_coefficient + np.sum(
```
Gui-FernandesBR, line 335: One problem that I saw is that when I run "Rocket.allInfo()", the graphs of Cd x Mach has a title of "Scalar x Scalar". I believe some changes on Function objects could adjust that behavior.

Gui-FernandesBR:
Hey @Lucas-KB and @lucasfourier , great to have such an interesting contributions from you guys.

I made some commetns in order to help your review before pulling the code to the develop branch, can't wait to see that happening!

rocketpy/Rocket.py
lucasfourier: The literature we are following explicitly states recommends that the reference length to be used is the rocket length. If you want to take a look, we were following the document "Estimating the dynamic and aerodynamic paramters of passively controlled high power rockets for flight simulaton", by Simon Box, Christopher M. Bishop and Hugh Hunt. The document can also be found in the following link http://cambridgerocket.sourceforge.net/AerodynamicCoefficients.pdf

rocketpy/Rocket.py
Lucas-KB: We reviewed our references and I believe there is the square, because of an algebraic manipulation. The deduction can be seen below:

image

image

Substituting the values yields:

image

What do you think, @Gui-FernandesBR @giovaniceotto?

Gui-FernandesBR: Reynolds is a dimensionless quantity... Just change "kinematic" for "dyanamic" in the first image you've sent and it's going to work then, it's a honest and simple mistake.

https://en.wikipedia.org/wiki/Reynolds_number

rocketpy/Rocket.py
giovaniceotto: I agree using length makes more sense in this case, since we are interested in the turbulent behavior along the rocket axis, rather than relative to its diameter.

rocketpy/Rocket.py
giovaniceotto: Yes, I still believe there is a slight error here. I hope this source, which has both formulas (using dynamic and kinematic viscosity) can help settle the debate: [https://www.engineeringtoolbox.com/reynolds-number-d_237.html](https://www.engineeringtoolbox.com/reynolds-number-d_237.html)

Lucas-KB: Ok, I understand now what is going on now, apparently our main reference had a typo on the definition of the Reynolds number and we just trusted it. Thank you both for pointing it out and helping us understand the problem. It will be fixed soon.

tests/test_solidmotor.py
```diff
@@ -174,7 +188,7 @@ def tests_export_eng_asserts_exported_values_correct(solid_motor):
         "0",
         "{:2.3}".format(grain_mass),
         "{:2.3}".format(grain_mass),
-        "RocketPy"
+        "RocketPy",
```
Gui-FernandesBR, line 191: Should that comma exist? I believe it was created automatically, but I'm concerned if this going to create a bug or not. @Lucas-KB @lucasfourier

rocketpy/Rocket.py
```diff
@@ -649,18 +842,35 @@ def cnalfa1(cn):
...
-            # Takes an approximation to an angular coefficient
-            clalpha = cldata.differentiate(x=0, dx=1e-2)
+        # Calculate Cd
+        if drag_coefficient is None and thickness is not None:
+            exposed_area = (1 / 2) * (rootChord + tipChord) * span
+            total_area = exposed_area + self.radius * rootChord
+
+            # In the formula, instead of span, it should be the distance between the
+            # midpoint of the root and the tip of the fin, but we are using
+            # span for now.
```
Gui-FernandesBR, line 852: If you consider a right-trapezoid, it should be `sqrt(span² + (Rc - tc)²/4)`. If you want to be more abstract, could get the angle bettwen span and rocket axis and use it on the law of cosines.

In my opinion, you don't need to make this correction on the code, I'm approving the pull request with the current approach. My point here is that it's not so hard to determine the distance you said, I think it's just about geometry. Reccomend save this idea for later.

Gui-FernandesBR (approved):
There are some comments on some files, but I believe it accomplished with your initial goals so congratulations for this development. Hope you can commit again soon

giovaniceotto:
Quick question. Is this implementation compatible with rockets which have multiple "tails" (which may actually be diameter variation) and multiple fin sets?

Gui-FernandesBR:
As I just made with #110, I suggest merging this PR and then start another to (1) create the examples and (2) better documation, **if** the plan is still to do do 1 and 2 before merging.

@giovaniceotto sorry, I took a look but I can't answer your last question precisally. @Lucas-KB may know it.

Gui-FernandesBR:
TODO (@Lucas-KB  & @lucasfourier):
- Start varying angle of attack during the evaluation of drag coefficient
- Standardize the reference area (the problem of using rocket area to adjust the Cd)
- Correct geometric parameters for the fins
- Implement Rail Button (launch lug) drag

* Study decorated functions

Gui-FernandesBR:
I'm quoting @MateusStano here : (automatic translated through google translate)

> Guys, just talking about the PR of the Drag Coefficient calculation. Basically what had been done there were some functions to calculate the drag. Functions are basically implementations of formulas. Furthermore, the functions are implemented in a provisional way and in the current state of the code it is not possible to define a Rocket object. Another thing is that the results are different from what was expected (comparing to rasAero for example).
> I would say that since we know the results of the drag calculations are not satisfactory (or different from rasAero) and since we don't know why, PR is not ready for Merge. For it become ready for merge I would say that we would have to do a review of the whole theory so we can start to understand the old code and find what is wrong (or not) in it. Honestly, I would say that the best thing would be to just redo the PR from scratch and that would take a long time. I don't think it's feasible to merge PR agr and maybe the best thing is just to close it

Gui-FernandesBR:
> Getting this PR back online will be no easy task. That being said, I do not believe closing it as a viable option.
> 
> We need someone to take over the great work left by @Lucas-KB and @lucasfourier. I truly believe it is more a challenge of manpower rather then technical dificulties. So let's go find someone who could help!

Yeah agreed on that. I think it would be good if we could convert it to a draft PR since the code is not finished yet. But I'm afraid only @Lucas-KB can perform such conversion.

Also, I think we need not only new people to work on aerodynamic developments but also need to improve our development and integration process. Similar to Environment Analysis, we should list all implementations before start cododing, and try our best to separate everything into small pull requests so it can be better managed.

Gui-FernandesBR:
Thinking again, there's no problem if we close the PR since github will still remain information regarding everything. There's a difference between closing and deleting.

The only thing we must me carefull is that we should not delete head branch!

Therefore I believe we should go ahead and close the PR in the next few days

PR #117, opened 2022-01-04
WIP: Airfoil fins normal force coefficient derivative calculation.

MateusStano:
## Pull request type

## Pull request checklist

The current calculation of the normal force coefficient derivative for airfoil fins assumes that the term where the derivative of the lift coefficient is multiplied by alpha is a constant value. However, for curves that are not linear that simplification does not apply due to another intermediate derivative term.

## What is the new behavior?

The calculations now use a generic formula of the lift coefficient derivative based on thin airfoil theory, which is more in the lines of what Barrowman showed in his paper.  This calculation works for any fin shape, with airfoil or not. It is not as exact as the formula that is currently used for trapezoidal fins but is a good approximation.

- [ ] Yes
- [x] No

## Other information

Since the new calculation of the lift coefficient derivative takes into account the number of Mach I had to change a bit more of the code than just what relates to the fins. The changes are nothing major but they can be useful when adding supersonic calculations. I also made a pdf with more in-depth documentation.


Gui-FernandesBR:
A masterpiece!!!!!!!!!!!!!!!!!!!!!!!!! @MateusStano thanks for showing up... Great contribution

" It is not as exact as the formula that is currently used for trapezoidal fins but is a good approximation."
- Do you have the equations with you? Could you share with us (if it is of easy access)?

MateusStano:
> * Do you have the equations with you? Could you share with us (if it is of easy access)?

The equations are documented here: docs/technical/aerodynamics/Fins_Lift_Coefficient.pdf

I wrote as an explanation/documentation.  I don't know if its any good but everything I used should be in there

Gui-FernandesBR:
Great Stano, really appreciate your expertise writing this document, it will definetely help a lot!

I think the best candidates to revivew the calculations here are @Lucas-KB and @lucasfourier , as they were working on aerodynamic. Perhaps in the next week they will be able to review and merge.

Otherwise, I hope @giovaniceotto could hep earlier, or evene @brunosorban

MateusStano:
> * it does not make sence to set interpolation and extrapolation methods.

Ah good to know! Will change that right now

> One question: how different are equations 6 and 15 of the theory you submitted? In termos of relative change, what's the percentage difference for a typical fin?

The difference seems to be that equation 6 gives a result 63% smaller then equation 15. In other words, the lift generated by the airfoils is 63% bigger than with trapezoildal plane fins.

Here is a comparison with the compareFunction method:
image

The yellow line is the airfoiled one.

giovaniceotto:
To be honest, I am quite intrigued by the difference. I would like to first review this difference before proceeding to merging this PR.

PR #42, opened 2020-08-11
SolidMotor_class_usage

lucasfourier:
Finished the SolidMotor_class_usage python notebook with examples and explanations.

PR #132, opened 2022-02-13
doc: setting inputs and discrete intervals for Function objects on SolidMotor

Gui-FernandesBR:
## Pull request type

## Pull request checklist

As described on #42 , some plots of SolidMotor.allInfo() are currently being set as Scalar x functions functions instead of Time (s) x Variable functions.

## What is the new behavior?
Now all graphs from SolidMotor.allInfo() have the 'x' axis well defined by the setInputs method

- [X] No

## Other information

Please verify if setDiscrete is not breaking other parts of our code. As far as I understand, the Function.setDiscrete converts functions into lists, so perhaps it will create a bug that I didn't find locally.


Gui-FernandesBR:
Well, everything look fine here, I performed basic simulations using the modified files for checking.

@MrGribel could ypu please review this as you better understand SolidMotor class?

MrGribel:
@Gui-FernandesBR could you clarify a bit what you're trying to solve here? Those changes don't make much sense to me at a first glance. The altered functions already seem to display the correct axes names on `develop`. This should have been changed at the branch which PR #42 provides from. Also, since all the changed plots already provide in some form from `thrust` or `grainInnerRadius`, which are already list-defined rocketpy Functions, applying the `setDiscrete` over them seems redudant.

Let me know if there is something I am missing here.

Gui-FernandesBR:
See Function.py to correct problems

giovaniceotto:
@Gui-FernandesBR, I did some digging here and strongly believe that all the issues that this PR attempts to solve are fixed with the simple 1 line PR #141.

It was a simple mistake I made a while back, in the Function class... my fault.

Issue #55, opened 2020-11-08
Rocket.addFins() airfoil argument should support floats and lists


giovaniceotto:
**Use case**
Try to add fins to a rocket object using a list as the airfoil argument:

```python
cl_data = [
    [0.00, 0.0],
    [0.20, 1.05],
    [1.00, 1.05],
    [1.57, 0.0],
]

FinSet = TestRocket.addFins(
    n=3, 
    rootChord=0.077, 
    tipChord=47/1000, 
    span=0.104,
    distanceToCM=-0.99586,
    airfoil=cl_data
)
```

**Expected result**
Fins are added properly with no errors  or warnings.
Airfoil argument should support entries in list format, just like other similar RocketPy entries (powerOnDrag, powerOffDrag, thrust, etc.):
- **Lists** should be converted into functions using **Function**.
- **Floats** could be interpreted as a lift coefficient derivative.

**What happens instead**

```shell
Exception has occurred: AttributeError 'list' object has no attribute 'split'
  File "rocketpy\Rocket.py", line 607, in addFins
    read = genfromtxt(airfoil, delimiter = ',')
  File "test.py", line 41, in <module>
    FinSet = TestRocket.addFins(
```

**Applies to versions:**
Master branch only, no releases yet.

Gui-FernandesBR:
@Projeto-Jupiter/back-end

Gui-FernandesBR:
Fixed bt @MateusStano on #140 . The release v9.10.0 will modify what is needed on the code.

Awesome work!!

PR #140, opened 2022-02-16
BUG: Airfoil fins normal force coefficient derivative calculation.

MateusStano:
Continuing #117 that was accidentally closed

## Pull request type

## Pull request checklist

The current calculation of the normal force coefficient derivative for airfoil fins assumes that the term where the derivative of the lift coefficient is multiplied by alpha is a constant value. However, for curves that are not linear that simplification does not apply due to another intermediate derivative term.

## What is the new behavior?

The calculations for the lift coefficient now use a more generalized equation based on thin airfoil theory, which is more in the lines of what Barrowman showed in his paper. The differences in calculating the lift with and without airfoils is only considered in the Cnalpha0 (see equation 3-1 in Barrowman). For planar fins, Cnalpha0 is considered to be `2π / beta`. For airfoil fins, Cnalpha0 comes from the derivation of a lift coefficient curve that must be given as an input to the `addFins` method.

- [ ] Yes
- [x] No

## Other information

The addition of the `beta` parameter makes it so the calculations for fins lift coefficient take into account the variation of the Mach number, which in turn can change a few other behaviors in the simulation (i.e. the static margin). This is worth noting since the other aerodynamic surfaces (nosecone and tail) don't consider the Mach number (yet).


MateusStano:
Hey, one more thing. I failed to create a test for the airfoils because I couldn't figure out a way to reach the airfoil lift curves in the tests. Some weird errors kept popping up.
I left the test commented in test_rocket.py so someone can check it out.

rocketpy/Rocket.py
```diff
@@ -564,11 +564,16 @@ def addFins(
...
+        airfoil : tuple, null
+            Airfoil's lift curve. If Null, fins will be considered
+            planar. If tuple, fins will be considered airfoiled.
+            The tuple must contain as its first item a .csv or .txt
+            file with the aifoil's lift coefficient by angle of attack
+            curve. The file must contain no headers and the first column
+            must specify the angle of attack, while the second column
+            must specify the lift coefficient. The tuple must contain as
+            its second item the unit of the angle of attack, accepting
+            either "radians" or "degrees".
```
giovaniceotto, line 576: Please update the argument description to include the fact that the first item can also be a list or a callable (function), or even a Function actually (no need to mention this, since a Function is callable). See the description of the source argument of the `Function.__init__` method for more details. This is important to handle #55.

Also, make sure to flag this parameter as optional: `airfoil : tuple, optional`

rocketpy/Rocket.py
```diff
@@ -689,74 +694,44 @@ def finNumCorrection(n):
...
+            clalpha2D = Function(lambda mach: 2 * np.pi / beta(mach))
+        else:
+            # Defines clalpha2D as the derivative of the
+            # lift coefficient curve for a specific airfoil
+            airfoilCl = Function(
+                airfoil[0], interpolation="linear", extrapolation="natural"
             )
+            if airfoil[1] == "degrees":
+                for i in range(len(airfoilCl)):
+                    airfoilCl[i][0] = np.radians(airfoilCl[i][0])
```
giovaniceotto, line 708: Current implementation does not support Functions based on callables (Functions not created with lists or csv, but a lambda function or similar). This simple change can make it compatible with everything. What do you thing?

```suggestion
            if airfoil[1] == "degrees": airfoilCl *= np.pi/180 # Convert to radians
```

rocketpy/Rocket.py
```diff
@@ -689,74 +694,44 @@ def finNumCorrection(n):
...
-            )  # Function of mach number
-
-            # Calculates clalpha * alpha
-            cl = Function(
-                lambda alpha, mach: alpha * clalphaMultipleFins(mach),
-                ["Alpha (rad)", "Mach"],
-                "Cl",
+            # Differentiating at x = 0 and correctign for compressible flow
+            clalpha2D = Function(
+                lambda mach: airfoilCl.differentiate(x=0, dx=0.1) / beta(mach)
```
giovaniceotto, line 712: What do you think about using `x = 1e-3` (which means `x = dx`) instead of `x=0`? Since the `.differentiate` method uses [central diferences](https://en.wikipedia.org/wiki/Finite_difference), this makes sure that `airfoilCl` will not be evaluated at `alpha < 0`, as it happens with the current implementation.

I noticed that you added `extrapolation="natural"` when initializing `airfoilCl`, but this may not be useful if the source is a callable (example: `airfoil = (lambda alpha = 2*np.pi/abs(alpha), 'radians')`).

```suggestion
                lambda mach: airfoilCl.differentiate(x=1e-3, dx=1e-3) / beta(mach)
```

tests/test_rocket.py
```diff
@@ -76,3 +76,89 @@ def mainTrigger(p, y):
...
+    Drogue = test_rocket.addParachute(
+        "Drogue",
+        CdS=1.0,
+        trigger=drogueTrigger,
+        samplingRate=105,
+        lag=1.5,
+        noise=(0, 8.3, 0.5),
+    )
+
+    assert test_rocket.allInfo() == None
```
giovaniceotto, line 164: I believe it would be quite important to assert the static margin, for both cases, with and without airfoils. What do you think?

giovaniceotto (changes requested):
Finally! Finally, a perfect airfoil implementation! Great work @MateusStano, we really needed this!

I am suggesting a few simple changes, see what you think about them!

rocketpy/Rocket.py
MateusStano: This actually doesn't work for .csv or .txt. I added a `if callable(airfoil[0]):` make callables work

tests/test_rocket.py
MateusStano: I added an `assert static_magin<2` in both the tests in the rocket_tests.py which I think is fair but maybe there is a better way to do it. Any suggestions?

rocketpy/Rocket.py
giovaniceotto: - Really? It should work! What is the error that occurs? Must be a bug in the Function class.

MateusStano: No errors, just weird values.
Using only `if airfoil[1] == "degrees": airfoilCl *= np.pi/180` :

image

Checking with callable:

image

tests/test_rocket.py
MateusStano: Aah that is smart. I changed the interval to 0.01, I think it is more reasonable

Gui-FernandesBR:
image

Am I doing something wrong or the getting_started_colab.ipynb is no longer working on this branch's Pull Request ?

MateusStano:
Uhh i think something is wrong. It is not running the new code. That error shows the old airfoil implementation

Gui-FernandesBR:
OK I fixed my code here.

In my opinion it's brillant, we can move forward. Congratulations @MateusStano really easy to understand the code.

Sugestion of creating good plots of airfoils data in the future, embedded on allInfo()

giovaniceotto:
> Sugestion of creating good plots of airfoils data in the future, embedded on allInfo()

Can we add this to our action log?

giovaniceotto:
I'll finish my last review by tonight so that we can merge it!

rocketpy/Rocket.py
```diff
@@ -689,74 +696,50 @@ def finNumCorrection(n):
...
+            if airfoil[1] == "degrees":
+                if callable(airfoil[0]):
+                    airfoilCl *= np.pi / 180
+                else:
+                    for i in range(len(airfoilCl)):
+                        airfoilCl[i][0] *= np.pi / 180
+
+            # Differentiating at x = 0 and correctign for compressible flow
+            clalpha2D = Function(
+                lambda mach: airfoilCl.differentiate(x=1e-3, dx=1e-3) / beta(mach)
```
giovaniceotto, line 720: The current implementation can be inefficient, since every time `clalpha2D` is called, the function `airfoilCl` is differentiated. Since the result will always be the same, it makes sense to do the differentiation only once.

Furthermore, we can also improbe the current degrees to radians correction to all values of the `airfoilCl` Function. What if we differentiate first, and then convert only the derivative result, instead of multiple values?

Here is my suggestion:
```suggestion
            # Differentiating at x = 0 to get cl_alpha
            clalpha2D_Mach0 = airfoilCl.differentiate(x=1e-3, dx=1e-3)

            # Convert to radians if needed
            if airfoil[1] == "degrees": clalpha2D_Mach0 *= 180/np.pi

            # Correcting for compressible flow
            clalpha2D = Function(
                lambda mach: clalpha2D_Mach0 / beta(mach)
```

giovaniceotto (changes requested):
Three more simple, but important, suggestions! If you accept them all, we can merge it. But check if you really agree with them.

Two of them are simple ones regarding documentation. However, the third one is a bit more complex and important.

rocketpy/Rocket.py
MateusStano: I really like this! Seems to work perfectly too

giovaniceotto (approved):

giovaniceotto:
@Gui-FernandesBR, when you approve, feel free to merge!

Gui-FernandesBR (approved):

PR #141, opened 2022-02-17
Bug/function class magic operations

giovaniceotto:
## Pull request type

## Pull request checklist

  - [x] Tests for the changes have been added (No new tests needed)
  - [x] Docs have been reviewed and added / updated if needed (No docs change needed)

As described in #42, some plots of SolidMotor.allInfo() are currently being shown with the wrong input. 

This is caused because inside the Function class, **the inputs set, in this case `Time (s)`, are replaced by the name of the argument of the callable used as the source argument, in most cases, `x`**. The same happens when the Function object is initialized with a float as the source argument, since internally, this float is converted into a lambda.

This bug is caused by an incorrect check done by the Function class. It should check if the inputs have already been set and only use the name of the source arguments in case no input name has been given, in which case the default "Scalar" is set. For some weird reason, instead of checking for "Scalar" to see if the default name is set, it is checking for "Time (s)".

## What is the new behavior?

The Function class checks correctly if the input is set to the default `Scalar` when a callable or int/float is used as the source argument. In such case, it replaces the input name `Scalar` by the name of the argument of the callable or `X` when int or float. 

This solves the problem described in #42.

- [ ] Yes
- [x] No


Gui-FernandesBR:
Shall we merge into develop instead of master branch?

PR #146, opened 2022-02-23
ENH: Export Flight Data

giovaniceotto:
## Pull request type

## Pull request checklist

Exporting data from a flight simulation often requires other libraries and custom code. No easy option exists to export a csv file with all relevant data.

## What is the new behavior?

A new method has been introduced, `Flight.exportData`, which can export any attribute of the Flight class which is a Function of time.

- [ ] Yes
- [x] No

## Other information

This PR features a substantial number of new lines due to two new fixture files used for testing. The code implementation itself is probably less than 50 lines.


giovaniceotto:
The new test seems to be failing since in each computer, the simulation result will be slightly off. I am not sure if there is a good solution.

giovaniceotto:
@MrGribel, just letting you know that I have fixed the test and this PR is ready for review. Let me know if you have any questions.

MrGribel (approved):
This a very useful and practical feature @giovaniceotto.
The implementation seems solid as well. A couple of comments:
- It would be nice to have an equivalent method to export data from the other classes, since the current method can't reach their attribute dictionaries. This is a matter for another set of PRs, though.
- This is very specific, but if you include the illegal Windows character `:` in the filename string argument, the method runs without erros, but writes an empty file. Other forbidden characters return the appropriate `OSError`. I think it has to do with [this Windows feature](https://stealthbits.com/blog/ntfs-file-streams/), since the resultant filename is only what's to the left of the colon.

You might want to look into that last point, but since it's such a small detail, I approved this for merging anyway. Perhaps a simple addition to the method docstrings warning of illegal characters should be enough.

giovaniceotto:
Thanks a lot for the review @MrGribel!

As suggested, I'll add a note in the docstring regarding forbidden characters and `:` in windows.

I'll merge this PR afterwards.

Gui-FernandesBR:
Thanks @MrGribel and @giovaniceotto , really well done

PR #148, opened 2022-02-24
WIP: Pressure Thrust corrections

MrGribel:
## Pull request type

## Pull request checklist

Rocketpy currently does not take into account the thrust additions due to pressure differences between nozzle exit and external pressure.

## What is the new behavior?
A few changes have been added:

1. - A correction is now applied to given thrust data to correct any thrust differences originated from static fire tests made at altitude.
2. - Pressure thrust is calculated at each instant of Flight and added to the total thrust of the Rocket, according to trajectory.

- [ ] Yes
- [x] No

## Other information
Reference for equations:
https://www.grc.nasa.gov/WWW/k-12/airplane/thrsteq.html

Preliminary tests were made. Code works, but results must be further investigated.
Bella Lui had a ~20m apogee increase.

## TO-DOs:
- Pressure thrust must be also calculated during rail phase
- A few Flight Functions must be altered to account for these changes in plots
- Create a test for this feature
- Alter `getting_started` notebook to reflect changes

### DIFF
--- a/rocketpy/Flight.py
+++ b/rocketpy/Flight.py
@@ -1230,7 +1230,7 @@ class Flight:
             Mt = self.rocket.motor.mass.getValueOpt(t)
             # Pressures
             exitPressure = self.rocket.motor.exitPressure
-            freestreamPressure = self.env.pressure(z)
+            freestreamPressure = self.env.pressure.getValueOpt(z)
             # Check for optimal expansion
             calculatePT = not self.rocket.motor.altitudeCompensatingNozzle
             # Thrust
--- a/rocketpy/Flight.py
+++ b/rocketpy/Flight.py
@@ -833,7 +833,7 @@ class Flight:
                         self.y[0] ** 2
                         + self.y[1] ** 2
                         + (self.y[2] - self.env.elevation) ** 2
-                        >= self.effective1RL ** 2
+                        >= self.effective1RL**2
                     ):
                         # Rocket is out of rail
                         # Check exactly when it went out using root finding
@@ -847,7 +847,7 @@ class Flight:
                         # Get points
                         y0 = (
                             sum([self.solution[-2][i] ** 2 for i in [1, 2, 3]])
-                            - self.effective1RL ** 2
+                            - self.effective1RL**2
                         )
                         yp0 = 2 * sum(
                             [
@@ -858,7 +858,7 @@ class Flight:
                         t1 = self.solution[-1][0] - self.solution[-2][0]
                         y1 = (
                             sum([self.solution[-1][i] ** 2 for i in [1, 2, 3]])
-                            - self.effective1RL ** 2
+                            - self.effective1RL**2
                         )
                         yp1 = 2 * sum(
                             [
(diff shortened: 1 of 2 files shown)


giovaniceotto:
@PatrickSampaioUSP, we need your review in this draft PR! Can you help us with your knowledge of propulsion systems?

PatrickSampaioUSP (approved):
Don't have much to add besides that maybe would be nice to add some unti test to this new feature, showing how it changed the behaviour when a value is provided and that it work properly when no value is provided on those new optional parameters

rocketpy/Flight.py
```diff
@@ -1228,8 +1228,19 @@ def uDot(self, t, u, postProcessing=False):
...
+            calculatePT = not self.rocket.motor.altitudeCompensatingNozzle
             # Thrust
-            Thrust = self.rocket.motor.thrust.getValueOpt(t)
+            pressureThrust = (
+                calculatePT
+                * (exitPressure - freestreamPressure)
+                * np.pi
+                * self.rocket.motor.nozzleRadius ** 2
+            )
+            Thrust = self.rocket.motor.thrust.getValueOpt(t) + pressureThrust
```
giovaniceotto: In terms of design, is this implementation better than having `motor.thrust` include the pressure term? If so, how?

(In such implementation, the thrust curve supplied by the `Motor` to the `Flight` class would need to be a function of both time and pressure. The code at the `Flight` class would not change significantly.)

MrGribel:
I've had some problems with select atmospheric model options with regards to the `Environment` pressure function having undefined values below a certain, seemingly random, altitude threshold. Setting the extrapolation argument of this function to "natural" worked well enough for now, but it's not perfect.

Here's a couple plots illustrating the new behaviour in a standard simulation for Bella Lui, without any additional arguments.
plot3

Note that because none of the new optional arguments were passed, it assumes a sea-level optimized motor, with an exit pressure of 1 atm. Since the rocket is launched from ~400m ASL, it already starts with a pressure differential (leading to a ~20N extra thrust), which increases further as the rocket climbs.

rocketpy/Flight.py
MrGribel: That's an interesting point which I've thought about quite a bit during the implementation.

My reasoning was that the momentum thrust (currently represented by `motor.thrust`) is an inherent function of the motor itself, while the pressure thrust part will also be dependent on the external environment and on a given trajectory - so it is a bit easier to keep them separate. 

We know exactly how the momentum thrust will behave over time due to thrust curve being an input, while the best we can do for pressure thrust **while still inside motor class** is to the create it as function-defined Function, with no known boundaries. We could create a new `totalThrust` attribute which is a sum of both and a function of two parameters, yet it would inherit the issues that pressureThrust has inside of `Motor`, namely the lack of information about pressure(t). Perhaps it would be more interesting to create it as an attribute of `Flight`.

Another point is that momentum thrust (currently `motor.thrust`) is used on a few other calculations (such as `massDot`) - so it probably should be its own thing (this doesn't prevent a new totalThrust attribute from existing of course).

giovaniceotto: Thanks for the clarifications @MrGribel. It makes a lot of sense.

I agree with you, perhaps a `Flight.totalThrust` and a `Flight.pressureThrust` attribute (as a Function of time) would be usefull.

Gui-FernandesBR:
@MrGribel why is the `altitudeCompensatingNozzle` attribute default false?

I mean, if I don't know at which altitude the motor was tested, shouldn't be better assuming there will be none pressure thrust variation?

I believe not every rocket designer knows the "exitPressure" value, which I understand is resulted from your motor design (correct me if I'm wrong but usually you don't know the target exitPressure, specially if you are just buying a motor).

By setting this to True by default, you don't modify flight results of those rockets that you don't have any ideia about static fire ambient pressure/altitude or about your exitPressure, right?

Please let me know if any of this makes sense to you

MrGribel:
Some interesting questions there, @Gui-FernandesBR.
>I mean, if I don't know at which altitude the motor was tested, shouldn't be better assuming there will be none pressure thrust variation?

Probably not. Pressure thrust effects will be present regardless, unless the motor has [a nozzle that compensates for it](https://en.wikipedia.org/wiki/Altitude_compensating_nozzle) (pretty rare for sounding and HPR applications). By providing an imprecise value for exit pressure, the pressure differential curve might be displaced by some factor, yet this is certainly a better guess than neglecting the effect entirely.

>I believe not every rocket designer knows the "exitPressure" value, which I understand is resulted from your motor design

This might be a problem for commercial engines, where you don't have control over many of the project variables. Regarding motor design, there is a parameter known as the nozzle's **ideal expansion ratio**, which is the area ratio required to [reduce the exhaust pressure to ambient pressure](https://en.wikipedia.org/wiki/Rocket_engine_nozzle#Atmospheric_use). Well-designed motors will usually adjust the nozzle's exit area as to match this value, leading to optimal expansion to the desired external pressure (usually close to or at sea-level).

>By setting this to True by default, you don't modify flight results of those rockets that you don't have any ideia about static fire ambient pressure/altitude or about your exitPressure, right?

That's correct. The dilemma here boils down to whether RocketPy current behavior of indifference regarding pressure variation effects is preferable to using the new implementation with uncertain inputs. Should we be conservative concerning historical simulation results? Very hard to find out without lots of data for validation, since the resulting difference is such small.

Gui-FernandesBR:
OK @MrGribel , understood.

Can we proceed with the merge of this pull request? Also, I believe we should raise an issue to accomplish with the problem of elevation definition inside of the environment class.

Could you make that please?

Gui-FernandesBR:
Edit: With the addition of Motor.py file and therefore the end of SolidMotor.py file, this PR will be deprecated.

Code lines additions must be moved to a new PR, created on a branch based on the updated develop

Gui-FernandesBR:
Must be re-created on top of new Motor.py file

PR #149, opened 2022-02-27
Tests: fix test warnings

giovaniceotto:
## Pull request type

## Pull request checklist

  - [x] Tests for the changes have been added: no new test needed in this case

The main goal of this PR was to clean up a couple of things in the code so that tests could run without throwing any warnings. Currently, 24 warnings are raised, as can be confirmed [here](https://github.com/Projeto-Jupiter/RocketPy/runs/5318012374?check_suite_focus=true), for example.

## What is the new behavior?

No more warnings! All tests should now be passing without raising warnings. The following code modifications were needed:

- Converted some strings to [raw strings](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals) to handle latex in matplotlib and search expressions in regex. This removes warnings related to invalid escape sequences.
- Remove matplotlib max open figures limit.
- Changed net `netCDF4.date2index` `select=nearest` argument to `select=before` when processing forecasts, reanalysis and ensembles. `nearest` would be the best option, but this was causing a bug reported here: https://github.com/Unidata/netcdf4-python/issues/1155. 
  - Even though this bug is caused by `netCDF4`, it was also a RocketPy bug since it caused the wrong time to be chosen from the weather file, even when the correct time existed. As described, this has been fixed by using the `before` option rather than `nearest`, since `before` works fine.

- [ ] Yes
- [x] No


Gui-FernandesBR (approved):
Everything works fine here, thanks for your contribution @giovaniceotto !

PR #153, opened 2022-03-08
Enh/dimensionality tests

PatrickSampaioUSP:
## Pull request type

## Pull request checklist

## What is the new behavior?

Added dimension tests for the Rocket class


tests/conftest.py
```diff
@@ -47,6 +48,50 @@ def rocket(solid_motor):
...
 
 
+@pytest.fixture
+def m():
+    return numericalunits.m
+
+
+@pytest.fixture
+def kg():
+    return numericalunits.kg
```
giovaniceotto, line 58: Should there by a fixture for seconds as well?

```suggestion
@pytest.fixture
def m():
    return numericalunits.m


@pytest.fixture
def kg():
    return numericalunits.kg


@pytest.fixture
def s():
    return numericalunits.s
```

tests/conftest.py
```diff
@@ -47,6 +48,50 @@ def rocket(solid_motor):
...
+
+
+@pytest.fixture
+def dimensioneless_solid_motor(kg, m):
+    example_motor = SolidMotor(
+        thrustSource="data/motors/Cesaroni_M1670.eng",
+        burnOut=3.9,
```
giovaniceotto, line 65: Should the burnOut time be given in seconds?

```suggestion
def dimensioneless_solid_motor(kg, m, s):
    example_motor = SolidMotor(
        thrustSource="data/motors/Cesaroni_M1670.eng",
        burnOut=3.9*s,
```

giovaniceotto (commented):
Had a quick look and noticed no time dimensions are checked. Did we forget it or is this the intended behavior?

I suspect wonder if there are any calculations which involve a change in time dimensions (such as time derivatives or integration) in the Rocket class.

giovaniceotto (commented):
From what I understand, when an equation does contain a dimensional error, an assert will fail. However, the assert can fail for multiple reasons, and a dimensional error is only one of the possibilities.

I wonder if we should implement such tests in a way that we have exclusive tests to guarantee dimensional homogeneity. This means that some tests would check only dimensionality and would fail if, and only if, there is a dimensional error. Another test would be needed verify if the function is producing a correct numerical answer.

The advantage of doing so is that the developer would have more information to understand where the error is and what it is.

tests/conftest.py
Gui-FernandesBR: Sure thing! But I think we use input based on time only on addParachutes method (the lag input) so in that case I think I agree using only mass and length for now

Gui-FernandesBR (approved):
I'm approving since it's OK for me, however Giovani's comments should be addressed before merging.
Great work btw, I think it's the beggining of a second level unit tests, PR&O is becoming expert on such things

PatrickSampaioUSP:
@Gui-FernandesBR I think we should wait a little, I'm working on something regarding that comment

PatrickSampaioUSP:
@giovaniceotto Could you take a look again, please?

giovaniceotto (approved):

PR #142, opened 2022-02-20
WIP: Front-End Integration

luimot:
## Pull request type

Please check the type of change your PR introduces:

- [x] Code base additions (bugfix, features)
- [ ] Code maintenance (refactoring, formatting, renaming, tests)
- [ ] ReadMe, Docs and GitHub maintenance
- [ ] Other (please describe):

## Pull request checklist

Please check if your PR fulfills the following requirements, depending on the type of PR:

- Code base additions (for bug fixes / features):

  - [ ] Tests for the changes have been added
  - [ ] Docs have been reviewed and added / updated if needed
  - [x] Lint (`black rocketpy`) has passed locally and any fixes were made
  - [x] All tests (`pytest --runslow`) have passed locally

## What is the current behavior?
All information is passed through method arguments or class instance attributes and there's currently no method returning all available simulation information as to easily handle these outputs.

## What is the new behavior?
Work In Progress but the inputs and outputs are more easily handled by an external server.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information
First of all, an introduction to why things are being changed the way they are, knowing a GUI integration of RocketPy is underway and in order to accompany it some of the RocketPy library must be changed.
The GUI is being developed in Flutter, a framework running on Dart and due to its (still) small community and very recent rising in popularity the best way to communicate Python with it is through [Flask](https://flask.palletsprojects.com/en/2.0.x/), an HTTP python microframework that enables bidirectional communication as to create an input/output interface for the user with a small delay if the server is local.
A simple diagram of how it works can be seen below:

image

Therefore, the way I found to solve this was to create two methods on each ```rocketpy``` class (excluding the Function class) named ```allInfoReturned``` and ```allPlotInfoReturned``` , those methods return the data as Python dicts, making it a really easy task parsing the information into [JSON](https://www.json.org/json-en.html) format, another great way to pass information to and from the python server. I believe all the @Projeto-Jupiter/front-end team have made the best project decisions to contribute to the project the best way possible but if you have any suggestion be more than welcome and we shall add to this code as well as our own's.
There was also an addition to the ```Flight``` class with the ```timeBeforeParachuteEvent()``` method and the ```parachuteEventTimeCalculated``` attribute  in order to clean the previously cited methods and other plot relevant methods in the file but those need to be revised by someone with more knowledge of the overall code.

PR #157, opened 2022-03-17
WIP: Front-End Integration

luimot:
## Pull request type

## Pull request checklist

All information is passed through method arguments or class instance attributes and there's currently no method returning all available simulation information as to easily handle these outputs.

## What is the new behavior?
Work In Progress but the inputs and outputs are more easily handled by an external server.

- [ ] Yes
- [x] No

## Other information
First of all, an introduction to why things are being changed the way they are, knowing a GUI integration of RocketPy is underway and in order to accompany it some of the RocketPy library must be changed.
The GUI is being developed in Flutter, a framework running on Dart and due to its (still) small community and very recent rising in popularity the best way to communicate Python with it is through [Flask](https://flask.palletsprojects.com/en/2.0.x/), an HTTP python microframework that enables bidirectional communication as to create an input/output interface for the user with a small delay if the server is local.
A simple diagram of how it works can be seen below:

image

Therefore, the way I found to solve this was to create two methods on each ```rocketpy``` class (excluding the Function class) named ```allInfoReturned``` and ```allPlotInfoReturned``` , those methods return the data as Python dicts, making it a really easy task parsing the information into [JSON](https://www.json.org/json-en.html) format, another great way to pass information to and from the python server. I believe all the @Projeto-Jupiter/front-end team have made the best project decisions to contribute to the project the best way possible but if you have any suggestion be more than welcome and we shall add to this code as well as our own's.
There was also an addition to the ```Flight``` class with the ```timeBeforeParachuteEvent()``` method and the ```parachuteEventTimeCalculated``` attribute  in order to clean the previously cited methods and other plot relevant methods in the file but those need to be revised by someone with more knowledge of the overall code.

This is another version of #142 where the commits are cherry-picked so it is easier to review, also there is, at first, outdated code since errors were fixed in upcoming commits, the first ``` Environment.py ``` and ``` Flight.py ``` files can be reviewed so the concept is more easily grasped by the reviewer


Gui-FernandesBR (approved):
In the future it will be nice to implement Launch site coordinates as an information of allInfo() method (currently it is implemented only on info() method). Such feature will facilitate users to locate the lauch point based on UTM zones.

But anyway, great work @luimot , I can't wait for the following Pull Requests from your side!

See example below.

image

rocketpy/Environment.py
```diff
@@ -3221,6 +3221,117 @@ def allInfo(self):
...
+        grid = np.linspace(self.elevation, self.maxExpectedHeight)
+        plotInfo = dict(
+            grid=[i for i in grid],
+            windSpeed=[self.windSpeed(i) for i in grid],
+            windDirection=[self.windDirection(i) for i in grid],
+            speedOfSound=[self.speedOfSound(i) for i in grid],
+            density=[self.density(i) for i in grid],
+            windVelX=[self.windVelocityX(i) for i in grid],
+            windVelY=[self.windVelocityY(i) for i in grid],
+            pressure=[self.pressure(i) / 100 for i in grid],
```
giovaniceotto, line 3245: This `/100` really bugs me.... Is there any reason for it to be here? Are the pressure values being returned in the dict in units different from Pa?

If so, I am not sure this is good practice, since RocketPy strives to use only Pa for consistency throughout the code.

rocketpy/Environment.py
```diff
@@ -3221,6 +3221,117 @@ def allInfo(self):
...
+        Parameters
+        ----------
+        None
+
+        Returns
+        ------
+        plotInfo : Dict
+            Dict of data relevant to plot externally
+        """
+        grid = np.linspace(self.elevation, self.maxExpectedHeight)
```
giovaniceotto, line 3236: `np.linspace` returns 50 evenly spaced samples by default. Is this good enough for all cases? 

These Functions, such as `windSpeed` and `density` can sometimes be defined as arrays, meaning they are already sampled data. In such cases, could the `grid` be just `windSpeed.source[:, 0]`?

rocketpy/Environment.py
```diff
@@ -3221,6 +3221,117 @@ def allInfo(self):
...
+        info = dict(
+            grav=self.g,
+            launch_rail_length=self.rL,
+            elevation=self.elevation,
+            modelType=self.atmosphericModelType,
+            modelTypeMaxExpectedHeight=self.maxExpectedHeight,
+            windSpeed=self.windSpeed(self.elevation),
+            windDirection=self.windDirection(self.elevation),
+            windHeading=self.windHeading(self.elevation),
+            surfacePressure=self.pressure(self.elevation) / 100,  # in hPa
```
giovaniceotto, line 3310: I am not really sure departing from SI will do us any good 😟

giovaniceotto:
I know this has already been merged, but I was curoius to look through the code and ended up adding some comments which can aid in future implementations and in similar PRs reagarding GUI integration. I hope they can help!

rocketpy/Environment.py
luimot: Since the ```allInfoReturned``` is based off the ```allInfo``` method, the pressure on the plots were given in hPa hence the ```/100``` but it can be implemented in hPa at the FE code and ```rocketpy``` stay in SI uniformly.

giovaniceotto: You are right, we do plot in hPa to avoid scientific notation, which doesn't always look good on plots. Still, I am not sure storing information in hPa would be a good practice. 

I worry that there is no documentation behind it saying that it is in hPa, so this might cause some doubts for devs in the future.

luimot: Totally comprehensible and I agree with you, I will make the changes necessary at FE and use only SI in my changes

rocketpy/Environment.py
luimot: This has the same explanation as the previous question about ```self.pressure(i) /100``` and the solution will be the same, implement in FE and maintain ```rocketpy``` as SI

PR #163, opened 2022-03-27
Enh: Adding latitude and longitude attributes to Flight class

Gui-FernandesBR:
## Pull request type

## Pull request checklist

Currently, the user sets latitude and longitude coordinates for the launch pad, however, we use this information for nothing but catching environment details to initialize. And it is really boring the fact you never know the landing point coordinates in lat/lon although you've used the same system to pass launch pad coordinates.

## What is the new behavior?
I'm adding two new attributes: `Flight.latitude` and `Flight.longitude` to the Fligt class, they are Fucntion objects and represent both the lat and lon coordinates over the flight, including the final landing point. The evaluation is based on the famous Haversine equation, which consider the distance along a sphere surface.

- [X] No

## Other information

I'm currently working on a simple test for this addition, but I believe it's time to submit this Pull Request so we can check if the attributes are working properly


Gui-FernandesBR:
An example of this implementation working fine, please notice the lauch pad coordinates are latitude= -21.996, longitude= -47.4268.

See the flight path performed by the rocket:
image

Now let's check the variation of coordinate x:
image

And then compare with Flight.longitude attribute:
image


We can see the same analysis considering y and latitude coordinates
image

image

Gui-FernandesBR:
Ok now I added a simple test for this implementation.

The LintBlack test is failling but I have formatted the code offline as you can see in the commit history.
So I believe we are ready to review, fix erros, and then merge

@MateusStano or @MrGribel  would 1 of you please review this assignment as you guys really understand about Flight class and pytests? tks

giovaniceotto (changes requested):
Amazing work @Gui-FernandesBR. What a fast response to such an issue.

I'll let @MateusStano do the heavy lifting in this review, as per your request. I have, however, looked into your implementation and liked it a lot.

I will only ask you to add the two new attributes, `Flight.latitude` and `Flight.longitude` to the `Flight.__init_post_process_variables()` method. As the name implies, this method initializes all variables which will later be calculated during the post process. This eliminates *attribute not found* errors and is also needed so that MATLAB can access these variables.

Since both attributes you created are a Function, you can just add them as:

```python
def __init_post_process_variables(self):
    ...
    self.latitude = Function(0)
    self.longitude = Function(0)
    ...
```

Regarding the test you implemented, I also really liked it. Here are just a couple of suggestions:
 -  I am not sure `assert test_flight.allInfo() == None` is needed. With have other tests with these, so it is better to not retest this, avoiding the same error appearing in multiple tests.
 - Perhaps you could be more specific in your test, by checking if the latitude and longitude values match expected values: `assert test_flight.latitude(t) - expected_value < 1e-6`.

Let me know if you have any questions.

Lucas-KB:
@Gui-FernandesBR, very interesting functionality, I am sure it will be very useful. Just something I would like to suggest, wouldn't it be more organized and reusable if the `Environment` class (maybe the `utilities` module) had a function to convert from our ENU coordinate system to LLA? I imagine it could be handy if someone is simulating some kind of control system or sensor fusion that uses GPS data (I am sure there are more use cases, but I couldn't think of them right now).

Gui-FernandesBR:
@giovaniceotto the `assert test_flight.allInfo() == None` is currently being used just in order to call the postProcess method and then start the other asserts. I could simply call postProcess instead of using allInfo, would be a good idea?

Regarding the expected_value ideia, well I'm not a big fan of this introduction as in the future this method can be improved in order to evaluate better results for final latitude and longitude coordinates. So how can I say I have an expected value if I know that with coulpe more refinements future devs could calculate values better than nowadays? For now, I believe the best is just ensure the rocket is going to the correct direction based on its initial heading. I can add couple more verifications, specially for heading like 0°, 90°, 180° and 270°, which can usually be bottlenecks during these calculations

Gui-FernandesBR:
> @Gui-FernandesBR, very interesting functionality, I am sure it will be very useful. Just something I would like to suggest, wouldn't it be more organized and reusable if the `Environment` class (maybe the `utilities` module) had a function to convert from our ENU coordinate system to LLA? I imagine it could be handy if someone is simulating some kind of control system or sensor fusion that uses GPS data (I am sure there are more use cases, but I couldn't think of them right now).

definetely a great ideia @Lucas-KB , I will understand if it would be better implementing whis right now or during future pull requests.

Currently we have `geodesicToUtm(self.lat, self.lon, self.datum)`and `geodesicToUtm(self.lat, self.lon, self.datum)` methods on the Environment class that could be useful

giovaniceotto:
> @giovaniceotto the `assert test_flight.allInfo() == None` is currently being used just in order to call the postProcess method and then start the other asserts. I could simply call postProcess instead of using allInfo, would be a good idea?

Yes.

> Regarding the expected_value ideia, well I'm not a big fan of this introduction as in the future this method can be improved in order to evaluate better results for final latitude and longitude coordinates. So how can I say I have an expected value if I know that with coulpe more refinements future devs could calculate values better than nowadays? For now, I believe the best is just ensure the rocket is going to the correct direction based on its initial heading. I can add couple more verifications, specially for heading like 0°, 90°, 180° and 270°, which can usually be bottlenecks during these calculations.

I like your idea very much. And keep in mind that tests can also change in future implementations.

rocketpy/Flight.py
```diff
@@ -2091,6 +2097,114 @@ def postProcess(self, interpolation="spline", extrapolation="natural"):
...
+                        + np.cos(lat1) * np.sin(distance[i] / R) * np.cos(bearing[i])
+                    )
+                )
+                lon2.append(
+                    (180 / np.pi) * lon1
+                    + np.arctan2(
+                        np.sin(bearing[i]) * np.sin(distance[i] / R) * np.cos(lat1),
+                        np.cos(distance[i] / R) - np.sin(lat1) * np.sin(lat2[i]),
+                    )
+                )
```
MateusStano, line 2196: If I understood this correctly it might be nicer (and maybe faster) to do something like this:

```suggestion
        # Applies the haversine equation to find final lat/lon coordinates
        for i in range(len(distance)):
            # Please notice that for distances lower than 1 centimeter the difference on latitude or longitude too small
            if abs(self.x[i][1]) < 1e-2:
                lon2.append(self.env.lon)
            else:
                lon2.append(
                    (180 / np.pi) * lon1
                    + np.arctan2(
                        np.sin(bearing[i]) * np.sin(distance[i] / R) * np.cos(lat1),
                        np.cos(distance[i] / R) - np.sin(lat1) * np.sin(lat2[i]),
                    )
                )
            if abs(self.y[i][1]) < 1e-2:
                lat2.append(self.env.lat)
            else:
                lat2.append(
                    (180 / np.pi)
                    * np.arcsin(
                        np.sin(lat1) * np.cos(distance[i] / R)
                        + np.cos(lat1) * np.sin(distance[i] / R) * np.cos(bearing[i])
                    )
                )
```

rocketpy/Flight.py
```diff
@@ -2091,6 +2097,114 @@ def postProcess(self, interpolation="spline", extrapolation="natural"):
...
+
+        latitude = []
+        longitude = []
+        for i in range(len(self.solution)):
+            latitude.append([self.solution[i][0], lat2[i]])
+            longitude.append([self.solution[i][0], lon2[i]])
```
MateusStano, line 2202: We can use some list comprehension here too:

```suggestion
        latitude = [[self.solution[i][0], lat2[i]] for i in range(len(self.solution))]
        longitude = [[self.solution[i][0], lon2[i]] for i in range(len(self.solution))]
```

MateusStano (changes requested):
Hey! I proposed some changes to make the code clearer/faster but I had some problems with the simulations. I ran the getting_started notebook with a 'StandardAtmosphere' and I don't think the results are correct.

First I got a warning in the postProcess():
image

The plots for TestFlight.x and TestFlight.longitude seems fine:
image

The plots for TestFlight.y and TestFlight.latitude doesn't:
image

I think it stopped appending points to Flight.latitude list once the warning was raised.

Running the simulations with an environment with forecast also has some problems.
It is fine for lon:
image

But not for lat:
image

It is actually just a peek, if I zoom in the plot the curve is the correct shape. From 1s to 200s:
image

Probably some numerical errors with all the trigonometric functions. Could you look into it @Gui-FernandesBR? 

Great work btw!!

rocketpy/Flight.py
```diff
@@ -2091,6 +2097,114 @@ def postProcess(self, interpolation="spline", extrapolation="natural"):
...
+            elif self.x[i][1] * self.y[i][1] < 0 and self.x[i][1] < 0:
+                bearing.append(
+                    3 * 3.14159265359 / 2
+                    + np.arctan(abs(self.y[i][1]) / abs(self.x[i][1]))
+                )
+                continue
+
+        # Store values of distance and bearing using approriate units
+        # self.distance = distance      # Must be in meters
+        # self.bearing = bearing        # Must be in radians
```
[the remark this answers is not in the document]
Gui-FernandesBR: Yeah sure

Gui-FernandesBR:
I'll take a look on these errors by tomorrow, definetely, 

thanks a lot for testing it, Stano!!

Gui-FernandesBR:
I finally added all the requirements you've made here, Giovani and Stano. Tks for the solid review.

So, @MateusStano could please you re-run the tests again and merge if everything is alright?
Once we have lat/lon attributes added to flight class, I will be pulling the kml file generator function and then we will close an important issue

MateusStano:
Everything seems good now! Great addition!!

MateusStano (approved):

giovaniceotto (approved):

PR #176, opened 2022-05-29
Enh/environment analysis units

giovaniceotto:
## Pull request type

## Pull request checklist

The EnvironmentAnalysis class, just like the rest of RocketPy, only supports SI units. This can make some outputs challenge to interpret, since SI units may not be the preferred unit in all cases. Example: Temperature data is seldom
interpreted in Kelvins for rocket flight analysis.

## What is the new behavior?

The EnvironmentAnalysis class (and only it) now supports a new initialization argument: `unit_system` which the user can specify as `SI`, `metric` or `imperial`. In the future, it is possible to allow the user to specify the unit system that he wants.

A unit system is just a dictionary with some dimensions and the units that should be used to represent it. Take the `metric` system for example:

```python
metric_system = {
    "length": "km",
    "velocity": "m/s",
    "acceleration": "g",
    "mass": "kg",
    "time": "s",
    "pressure": "hPa",
    "temperature": "degC",
    "angle": "deg",
    "precipitation": "mm",
    "wind_speed": "m/s",
}
```

The default unit system is `metric` and it is initialized by the `EnvironmentAnalysis.___init_unit_system` method. Then, these units are saved as a dictionary in the `EnvironmentAnalysis.unit_system` attribute.

Besides this dictionary, another new dictionary is also important: the `EnvironmentAnalysis.current_units` attribute. This is initialized by the `EnvironmentAnalysis__init_data_parsing_units` method, and, as the name suggests, defines the dimensions that each parsed variable uses. Here is how it looks:

```python
self.current_units = {
    "height_ASL": "m",
    "pressure": "hPa",
    "temperature": "K",
    "windDirection": "deg",
    "windHeading": "deg",
    "windSpeed": "m/s",
    "windVelocityX": "m/s",
    "windVelocityY": "m/s",
    "surface100mWindVelocityX": "m/s",
    "surface100mWindVelocityY": "m/s",
    "surface10mWindVelocityX": "m/s",
    "surface10mWindVelocityY": "m/s",
    "surfaceTemperature": "K",
    "cloudBaseHeight": "m",
    "surfaceWindGust": "m/s",
    "surfacePressure": "Pa",
    "totalPrecipitation": "m",
}
```

As the name `self.current_units` suggests, this dictionary is only used as presented here to parse the netCDF files and then it changes once the parsed data is converted to the units of `self.unit_system`. This conversion takes place in two methods whose names are self-explanatory.

```python
# Convert units
self.convertPressureLevelData()
self.convertSurfaceData()
```

These run for the first time right after the parsing functions. But they can run other times if the user decides to change the units after initializing the `EnvironmentAnalysis` object. This is done by using the`EnviornmentAnalysis.set_unit_system(unit_system="metric")` method.

A couple of other methods were created and are mainly used inside `convertPressureLevelData` and `convertSurfaceData`. These are:

```python
__convert_units(self, variable, from_unit, to_unit, axis=1)           # General method
__convert_temperature(self, variable, from_unit, to_unit)             # Only for temperatures
__convert_units_Functions(self, variable, from_unit, to_unit, axis=1) # Only for Functions
__conversion_factor(self, from_unit, to_unit)                         # Holds all conversion factors, except for temperatures
```

This last method has a quite interesting dictionary created with the help of GitHub Copilot and worth sharing here:
```python
units_conversion_dict = {
    # Units of length. Meter "m" is the base unit.
    "mm": 1e3,
    "cm": 1e2,
    "dm": 1e1,
    "m": 1,
    "dam": 1e-1,
    "hm": 1e-2,
    "km": 1e-3,
    "ft": 1 / 0.3048,
    "in": 1 / 0.0254,
    "mi": 1 / 1609.344,
    "nmi": 1 / 1852,
    "yd": 1 / 0.9144,
    # Units of velocity. Meter per second "m/s" is the base unit.
    "m/s": 1,
    "km/h": 3.6,
    "knot": 1.9438444924406047,
    "mph": 2.2369362920544023,
    # Units of acceleration. Meter per square second "m/s^2" is the base unit.
    "m/s^2": 1,
    "g": 1 / 9.80665,
    "ft/s^2": 1 / 3.2808399,
    # Units of pressure. Pascal "Pa" is the base unit.
    "Pa": 1,
    "hPa": 1e-2,
    "kPa": 1e-3,
    "MPa": 1e-6,
    "bar": 1e-5,
    "atm": 1.01325e-5,
    "mmHg": 1 / 133.322,
    "inHg": 1 / 3386.389,
    # Units of time. Seconds "s" is the base unit.
    "s": 1,
    "min": 1 / 60,
    "h": 1 / 3600,
    "d": 1 / 86400,
    # Units of mass. Kilogram "kg" is the base unit.
    "mg": 1e-6,
    "g": 1e-3,
    "kg": 1,
    "lb": 2.20462,
    # Units of angle. Radian "rad" is the base unit.
    "rad": 1,
    "deg": 1 / 180 * np.pi,
    "grad": 1 / 200 * np.pi,
}
```

To finish off with implementation details, the label of some plots and also print statements have been modified to support the unit system. Here is an example:

```python
print("Gust Information")
print(f"Global Maximum wind gust: {self.maximum_wind_gust} {self.unit_system['wind_speed']}")
print(f"Average maximum wind gust: {self.average_max_wind_gust} {self.unit_system['wind_speed']}")
print("Temeprature Information")
print(f"Global Maximum temperature: {self.record_max_temperature} {self.unit_system['temperature']}")
```

I hope this gives a good overall explanation of how the implementation of units was done. I am at the reviewer's disposal for more information if needed and an interative live review session if desired.

- [x] Yes
- [ ] No

The breaking change caused is relative to other code in the `enh/environment_analysis` branch. The current code does not
convert the attribute `EnvironmentAnalysis.elevation`. Therefore, it is always in meters while the rest is in other units (km or ft).
`EnvironmentAnalysis.elevation` is usually used to set plot labels. I believe another PR would be needed to improve the implementation of `EnvironmentAnalysis.elevation`, getting the data directly from the surface data file instead of user input. When done like that, the conversion would make much more sense in this new framework.

## Other information

Not having the units 'attached' to the values is quite frustrating to work with. When getting raw values, it may be hard to know in which units they are expressed. This is where `EnvironmentAnalysis.current_units` is your best friend. However, in the future, I do believe a solution using `astropy.units` may be the best way to go!

### DIFF
--- a/rocketpy/EnvironmentAnalysis.py
+++ b/rocketpy/EnvironmentAnalysis.py
@@ -54,7 +54,43 @@ class EnvironmentAnalysis:
         surfaceDataFile=None,
         pressureLevelDataFile=None,
         timezone=None,
+        unit_system="metric",
     ):
+        """Constructor for the EnvironmentAnalysis class.
+        
+        Parameters
+        ----------
+        start_date : datetime.datetime
+            Start date and time of the analysis. When parsing the weather data
+            from the source file, only data after this date will be parsed.
+        end_date : datetime.datetime
+            End date and time of the analysis. When parsing the weather data
+            from the source file, only data before this date will be parsed.
+        latitude : float
+            Latitude coordinate of the location where the analysis will be
+            carried out.
+        longitude : float
+            Longitude coordinate of the location where the analysis will be
+            carried out.
+        elevation : float
+            Elevation of the location where the analysis will be carried out.
+        surfaceDataFile : str, optional
+            Path to the netCDF file containing the surface data.
+        pressureLevelDataFile : str, optional
+            Path to the netCDF file containing the pressure level data.
+        timezone : str, optional
+            Name of the timezone to be used when displaying results. To see all
+            available time zones, import pytz and run print(pytz.all_timezones).
+            Default time zone is the local time zone at the latitude and
+            longitude specified.
+        unit_system : str, optional
+            Unit system to be used when displaying results. Default is metric.
+            Options are: SI, metric, imperial. Default is metric.
+
+        Returns
+        -------
+        None
+        """
         # Save inputs
         self.start_date = start_date
         self.end_date = end_date
@@ -64,8 +100,11 @@ class EnvironmentAnalysis:
         self.surfaceDataFile = surfaceDataFile
         self.pressureLevelDataFile = pressureLevelDataFile
         self.prefered_timezone = timezone
+        self.unit_system_string = unit_system
 
-        # Manage timezones
+        # Manage units and timezones
+        self.__init_unit_system()
+        self.__init_data_parsing_units()
         self.__find_prefered_timezone()
         self.__localize_input_dates()
 
@@ -74,6 +113,10 @@ class EnvironmentAnalysis:
         self.surfaceDataDict = {}
         self.parsePressureLevelData()
         self.parseSurfaceData()
+
+        # Convert units
+        self.convertPressureLevelData()
+        self.convertSurfaceData()
 
         # Initialize result variables
         self.average_max_temperature = 0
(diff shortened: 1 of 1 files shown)


Gui-FernandesBR:
I'm getting problems on objects that don't have tz.info information

image

Gui-FernandesBR:
I tried some following codes, but nothing is working...

    def __localize_input_dates(self):
        try:
            self.start_date.tzinfo
        except:
            self.start_date = self.prefered_timezone.localize(self.start_date)
        try:
            self.end_date.tzinfo
        except:
            self.end_date = self.prefered_timezone.localize(self.end_date)

or 

    def __localize_input_dates(self):
        self.start_date = self.prefered_timezone.localize(self.start_date)
        self.end_date = self.prefered_timezone.localize(self.end_date)

Gui-FernandesBR:
And I'm using the following code to teste:

```
from rocketpy.EnvironmentAnalysis import EnvironmentAnalysis
  import datetime
  
  env = EnvironmentAnalysis(
      start_date= (2019, 1, 1),
      end_date=   (2021, 1, 1),
      latitude=32.990148,
      longitude=-106.969333,
      elevation=1400,
      surfaceDataFile="data\weather\spaceport_america_single_level_reanalysis_2002_2021.nc",
      pressureLevelDataFile="data\weather\spaceport_america_pressure_levels_reanalysis_2002_2021.nc",
      # timezone=,
      unit_system="metric",
  )
```

I also tryed `from datemie import datetime` as an alternative.

@giovaniceotto do you have any idea of what could be happening?

rocketpy/EnvironmentAnalysis.py
```diff
@@ -290,6 +333,216 @@ def __find_prefered_timezone(self):
...
+            "nmi": 1 / 1852,
+            "yd": 1 / 0.9144,
+            # Units of velocity. Meter per second "m/s" is the base unit.
+            "m/s": 1,
+            "km/h": 3.6,
+            "knot": 1.9438444924406047,
+            "mph": 2.2369362920544023,
+            # Units of acceleration. Meter per square second "m/s^2" is the base unit.
+            "m/s^2": 1,
+            "g": 1 / 9.80665,
```
PatrickSampaioUSP, line 424: I didn't get why 1/9.8065

PatrickSampaioUSP (approved):
It's perfect! The only thing is that I would try to move the functions responsible for converting units to a other file, because other classes may use this in the future, and this code doesn't need to be inside the environment analysis (i.e __conversion_factor)

Gui-FernandesBR (approved):
Everything is working fine for me. I've tested some scenarios and the only problem I saw was on the "plot_wind_gust_distribution" function, which were not finding a good fit for some cases, but I understand the problem was due to low range of dates, it means that increasing the range between start date e end date solved everything.

Btw, it was an awesome implementation by you @giovaniceotto !! Let's follow Patrick's idea of using this unitsystem conversion on other classes.

giovaniceotto:
Interesting link: https://joss.theoj.org/papers/10.21105/joss.00809

PR #178, opened 2022-06-01
Revert "ENH: Environment Analysis - Units"

giovaniceotto:
Reverts Projeto-Jupiter/RocketPy#176


giovaniceotto:
This needs to be reverted since I found major bugs in the code (it pretty much does not work.)

Gui-FernandesBR:
What happened @giovaniceotto ? Can you show an example of what bug you found?

giovaniceotto:
When converting units, it does not convert altitude properly. Furthermore, for all variables, it only converts the first hour of the first day.

PR #180, opened 2022-06-01
ENH: General Refactoring - Introducing Motor class

MrGribel:
## Pull request type

## Pull request checklist

RocketPy has only a SolidMotor class, which doesn't allow much flexibility in it's methods to adapt them to other engine types.

## What is the new behavior?
This pull request aims to implement an abstraction of the Motor class from which different motor classes can be derived, such as the old `SolidMotor` and the now under-development `HybridMotor`. 

This was achieved with the help of the **ABC** (abstract base class) python module. 

Some of the old `SolidMotor` more general attributes and methods are now fully defined in the abstract class, while other, more specific ones, are left for each child class to provide their own implementation.

This shouldn't introduce a breaking change in most example code since the old `SolidMotor` class is still imported in the init file which defines rocketpy as a module. If, however, for some reason this code tries a relative import of the likes of:
`from .SolidMotor import SolidMotor`, it will not work since `SolidMotor.py` doesn't exist anymore.

- [ ] Yes
- [x] No


Gui-FernandesBR (approved):
Well done!! @MrGribel tks for your addition!

rocketpy/Motor.py
```diff
@@ -0,0 +1,1937 @@
...
+
+        Parameters
+        ----------
+        None
+
+        Returns
+        -------
+        yCM : Function
+            Position of the center of mass as a function
+            of time.
```
giovaniceotto, line 358: In RocketPy, we usually refer to the rocket's axial axis as `z` instead of `y`. Can calling this attribute `yCM` cause confusion?

rocketpy/Motor.py
```diff
@@ -0,0 +1,1937 @@
...
+        """
+
+        pass
+
+    def evaluateMass(self):
+        """Calculates and returns the total propellant mass curve by
+        numerically integrating the MassDot curve, calculated in
+        evaluateMassDot. Numerical integration is done with the
+        Trapezoidal Rule, given the same result as scipy.integrate.
```
giovaniceotto, line 367: `given` -> `giving`

rocketpy/Motor.py
```diff
@@ -0,0 +1,1937 @@
...
+            * (self.grainOuterRadius**2 - self.grainInitialInnerRadius**2)
+        )
+        self.grainInitialMass = self.grainDensity * self.grainInitialVolume
+        self.propellantInitialMass = self.grainNumber * self.grainInitialMass
+        # Dynamic quantities
+        self.evaluateMassDot()
+        self.evaluateMass()
+        self.evaluateGeometry()
+        self.evaluateInertia()
+        self.evaluateCenterOfMass()
```
giovaniceotto, line 875: This contains a lot of repeated code, which can make maintaining it a hard task (you may fix it in one place, while the other place stays wrong). How much of it could be substituted by calling `super().__init__()`?

rocketpy/Motor.py
```diff
@@ -0,0 +1,1937 @@
...
+
+        Returns
+        -------
+        self.exhaustVelocity : float
+            Constant gas exhaust velocity of the motor.
+        """
+        pass
+
+    @abstractmethod
+    def evaluateMassDot(self):
```
giovaniceotto, line 326: Should this method and similar ones be converted to abstract properties just as was done for exhaust velocity?

rocketpy/Motor.py
```diff
@@ -0,0 +1,1937 @@
...
+        self.massDot : Function
+            Time derivative of total propellant mas as a function
+            of time.
+        """
+        pass
+
+    @abstractmethod
+    def evaluateCenterOfMass(self):
+        """Calculates and returns the time derivative of motor center of mass.
+        The result is a function of time, object of the Function class, which is stored in self.yCM.
```
giovaniceotto, line 348: Does it compute the time derivative or the relative position? Furthermore, what is this position relative to?

rocketpy/Motor.py
```diff
@@ -0,0 +1,1937 @@
...
+
+        # Inertia Z Dot
+        self.inertiaZDot = (1 / 2.0) * self.massDot * (
+            self.grainOuterRadius**2 + self.grainInnerRadius**2
+        ) + self.mass * self.grainInnerRadius * self.burnRate
+        self.inertiaZDot.setOutputs("Propellant Inertia Z Dot (kg*m2/s)")
+
+        return [self.inertiaI, self.inertiaZ]
+
+    def exportEng(self, fileName, motorName):
```
giovaniceotto, line 1145: Interesting... why was there a need to overwrite this method from the base class?

rocketpy/Motor.py
```diff
@@ -0,0 +1,1937 @@
...
+
+        # Write last line
+        file.write("{:.4f} {:.3f}\n".format(self.thrust.source[-1, 0], 0))
+
+        # Close file
+        file.close()
+
+        return None
+
+    def info(self):
```
giovaniceotto, line 1191: Also wondering why this method was overwritten...

rocketpy/Motor.py
```diff
@@ -0,0 +1,1937 @@
...
+        self.Kn()
+        self.inertiaI()
+        self.inertiaIDot()
+        self.inertiaZ()
+        self.inertiaZDot()
+
+        return None
+
+
+class HybridMotor(Motor):
```
giovaniceotto, line 1300: Could you describe the technology readiness level of this class?

giovaniceotto (commented):
I know this PR has already been merged, but since it is such an important PR, I felt I needed to do a thorough review even so.

I have commented on simple things (such as typos), but also brought up more relevant questions regarding implementation details and practices. I would love to discuss my comments further through an online conference.

And just one final question! What do you believe needs to change in the Motor abstract class in the future for it to support motors like: liquid motors, cold gas thrusters and water rockets?

rocketpy/Motor.py
MrGribel: Perhaps it should? Seems like a matter of preference to me. Up until now I've been defining as properties simple calculations which require 1-2 lines of code, and methods for everything else. Do you have any specific criteria we could use in mind?

rocketpy/Motor.py
MrGribel: This method is only a work-in-progress for now. One of the next few PR's regarding hybrids will tackle it specifically, as well as improve this docstring which I agree it's lacking and confusing. 

The method calculates the position of the center of mass of the total propellant, relative to the **center of mass of the solid propellant**, which is the new reference point for all-things solids and hybrids. 

One of the ways it will do so is by solving an initial value problem of a differential equation (hence the time derivative of the motor's center of mass).

rocketpy/Motor.py
MrGribel: Oops! I had indeed included a `super().__init__()` to the `SolidMotor` and `HybridMotor` child classes (see  549e6f539ce8b3b93fa6f6beae0b83c0510b261f), but it got lost in the refactoring process where we separated this PR from other changes. I'll be sure to add it back ASAP.

rocketpy/Motor.py
MrGribel: Good catch! Can't see why these shouln't go on the base class, quite frankly.

rocketpy/Motor.py
MrGribel: Kept this one only on the base class as well. 

Regarding `allInfo`, on the other hand, it seemed better to be kept class-dependent, since there will be lots of motor-type specific info displayed by the method.

PR #172, opened 2022-05-19
ENH: Elliptical fins added to Rocket class

KrWanderley:
## Pull request type

Please check the type of change your PR introduces:

- [x] Code base additions (bugfix, features)
- [ ] Code maintenance (refactoring, formatting, renaming, tests)
- [ ] ReadMe, Docs and GitHub maintenance
- [ ] Other (please describe):

## Pull request checklist

Please check if your PR fulfills the following requirements, depending on the type of PR:

- ReadMe, Docs and GitHub maintenance:

  - [x] Spelling has been verified
  - [x] Code docs are working correctly 

- Code base maintenance (refactoring, formatting, renaming):

  - [x] Docs have been reviewed and added / updated if needed
  - [x] Lint (`black rocketpy`) has passed locally and any fixes were made
  - [ ] All tests (`pytest --runslow`) have passed locally

- Code base additions (for bug fixes / features):

  - [ ] Tests for the changes have been added
  - [x] Docs have been reviewed and added / updated if needed
  - [x] Lint (`black rocketpy`) has passed locally and any fixes were made
  - [ ] All tests (`pytest --runslow`) have passed locally

## What is the current behavior?

RocketPy only accept trapezoidal fins.

## What is the new behavior?

Now, RocketPy accept elliptical fins. To do that, Barrowman equations and some deductions were used to define center of mass and lift coeficients. All the equations can be found on the PDF: https://github.com/Projeto-Jupiter/RocketPy/blob/a4778ba969f044ff669bc55cf35a572f2927b61f/docs/technical/aerodynamics/Elliptical_Fins.pdf

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

It's a breaking change because the way fins are added on addFins method was changed: before that, fins were defined by the 3 input parameters and now it's defined by a dictionary, which breakes older codes.

## Other information 

Due to the significant changes in the way fins are added in the addFins method, several tests will have to be updated. Furthermore, new tests for elliptical fins still must be created. Moreover, all the Getting Started notebooks must be updated.

PR #206, opened 2022-07-29
ENH: Refactor Rocket.addFins to avoid breaking changes

giovaniceotto:
## Pull request type

## Pull request checklist

The proposed implementation of Rocket.addFins (#172 ) will cause breaking changes.

## What is the new behavior?

The new proposed implementation creates two methods: `Rocket.addTrapezoidalFins` and `Rocket.addEllipticalFins` to avoid breaking changes. Furthermore, it marks `Rocket.addFins` as a future deprecation to be removed in version 2.0.0 (to be discussed).

- [ ] Yes
- [x] No

## Other information
Work to fix notebooks and docs that mention `addFins` must be worked on to reflect theses changes.

### DIFF
--- a/docs/notebooks/getting_started.ipynb
+++ b/docs/notebooks/getting_started.ipynb
@@ -345,18 +345,19 @@
          "source": [
             "NoseCone = Calisto.addNose(length=0.55829, kind=\"vonKarman\", distanceToCM=0.71971)\n",
             "\n",
-            "FinSet = Calisto.addFins(type=\"trapezoid\",\n",
+            "FinSet = Calisto.addTrapezoidalFins(\n",
             "                         n=4,\n",
-            "                         finParameters={\"span\": 0.100, \n",
-            "                                        \"rootChord\": 0.120,\n",
-            "                                        \"tipChord\": 0.040},\n",
+            "                         rootChord=0.120,\n",
+            "                         tipChord=0.040,\n",
+            "                         span=0.100,\n",
             "                         distanceToCM=-1.04956,\n",
-            "                         radius=0,\n",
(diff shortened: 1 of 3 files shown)


MateusStano:
Hey! This seems to fix the breaking change, I just fixed a few small bugs regarding the radius parameter

Just wondering about some repeated code on both new functions, like the functions `def beta(mach)` and `def finNumCorrection(n)` or the `cl` calculations. Wouldn't it be better split up de addFins functions into even more functions? Or create a class maybe?

giovaniceotto:
@MateusStano, I completely agree that it is about time to create a class to manage fins and perhaps even the remaining aerodynamic surface.

How about something like this in the future?

```python
from rocketpy import Rocket
from rocketpy.aerodynamics import BoatTail, TrapezoidalFins, ElipticalFins, NoseCone

TestRocket = Rocket(...)
TestFins = TrapezoidalFins(...)

TestRocket.addAerodynamicSurface(TestFins)
```

With something like this, both TrapezoidalFins and ElipticalFins could be a class inherited from the same Fins class, which could contain the repeated code. Furthermore, the base Fins class could isnheret from a even more basic and abstract AerodynamicSurface class.

What do you think?

The only issue is that this causes breaking changes. So I would advise releasing the ElipticalFins implementation first and then create theses changes.

MateusStano (approved):

MateusStano:
I really like the approach for the classes. I think it would make the code simpler to understand and easier to expand. Maybe instead of using inheriting abstract calss would fit better for teh Aero Surfaces class for example

I guess the best way then is to merge this PR and leave the Aerodynamic Surface and Fin class to some other time

PR #221, opened 2022-09-10
BUG: Power on and power off drag inverted

MateusStano:
## Pull request type

`powerOnDrag` and `powerOffDrag` are inverted in the flight class, so the simulation is using `powerOffDrag` while the motor is on an `powerOnDrag` when the motor is off. 

Also, both drag curves are used to set a Function object with `'spline'` interpolation. Sometimes this can cause problems with clustered data archives. Changed it to `'linear'` so it is more compatible.

### DIFF
--- a/rocketpy/Flight.py
+++ b/rocketpy/Flight.py
@@ -1368,7 +1368,7 @@ class Flight:
 
         # Determine aerodynamics forces
         # Determine Drag Force
-        if t > self.rocket.motor.burnOutTime:
+        if t < self.rocket.motor.burnOutTime:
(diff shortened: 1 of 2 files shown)


Gui-FernandesBR:
Insanely good catch, @MateusStano ! I'm approving since it was easy to review and, also, all the checks have passed.

However, I believe it could be merged into master directly, what do you think?

We should prepare a new release version as well.

Gui-FernandesBR (approved):
Awesome!

giovaniceotto (approved):

PR #229, opened 2022-09-13
ENH: automate dispersion inputs with csv

FranzYuri:
## Pull request type

## Pull request checklist

  - [X] Spelling has been verified
  - [X] Code docs are working correctly 

Currently we need to write a dictionary for the dispersion

## What is the new behavior?

now it is possible to get a dictionary for the dispersion through an .csv file from excel

- [ ] Yes
- [X] No

## Other information

I also added a .csv file as an example at /data/dispersion csv files/juninho.


Gui-FernandesBR:
Good one @FranzYuri !!

Quick questions before reviewing:
- [ ] The csv file seems to be not posslished, with some empty columns at the begining, I'd suggest cleaning it before merging. Also, I don't think it's safe idea using `sep=";"` in the code, as the .csv file could also have "," (comma) as separator. 
- [ ] Could we use Calisto's parameters on the example .csv file so we keep in the same page as the current documentation? I believe the rocket you described on your file belongs to a specific rocket design team... 
- [ ] Is it possible to you adding a simulation example using this method? This could be in an jupyter notebook if you prefer

MateusStano:
Hey, two commentes here:

- Could we make it work with other file types besides .csv? Like using the native excel file formats such as .xlsx

- I think another important part of the automation is maybe saving some relevant results back in the original file. This would make the results easy to analyze and compare if the simulations are done on a large scale

Gui-FernandesBR:
> Hey, two commentes here:
> 
> - Could we make it work with other file types besides .csv? Like using the native excel file formats such as .xlsx
> 
> - I think another important part of the automation is maybe saving some relevant results back in the original file. This would make the results easy to analyze and compare if the simulations are done on a large scale

Just remember to ensure that the code doesn't modify the input data on original file. It is possible to open a new file if it's needed.

Gui-FernandesBR:
This will be understood as a colum separator if the user set "," as sepator.
image

I will try to fix it

rocketpy/utilities.py
```diff
@@ -110,7 +108,7 @@ def check_constant(f, eps):
 
         Parameters
         ----------
-        f : array, list
+        f : array, row
```
Gui-FernandesBR, line 111:
```suggestion
        f : array, list

```

Gui-FernandesBR:
> Hey, two commentes here:
> 
> * Could we make it work with other file types besides .csv? Like using the native excel file formats such as .xlsx
> * I think another important part of the automation is maybe saving some relevant results back in the original file. This would make the results easy to analyze and compare if the simulations are done on a large scale

@MateusStano , the ".xlsx" is a binary file, therefore it's not recommended to upload/use it on git. I think that it's correctly doing when asking for .csv files. When it is the case the user has a .xlsx file, he/she should also be able to easily convert it to a .csv file (by using MS excel or others).

But yes, the idea of supporting different files is good one. I think the same with .txt and .json files format.

Gui-FernandesBR (approved):
Great work @FranzYuri . The code was not perfect, but after some minor adjustments I think it is working properly now.

Please take a look at the #242 issue that I just raised based on Stano's comment.

PR #230, opened 2022-09-14
ENH: Simple Initial Solution

giovaniceotto:
## Pull request type

## Pull request checklist

Using the `initialSolution` argument of the Flight class is quite inconvenient, specially when you retrieve from a previous Flight instance. The argument only supports an array as input.

## What is the new behavior?

`Flight` instances can now be used as input to the `initialSolution` argument of the Flight class. 

- [ ] Yes
- [x] No

## Other information

One option that we have is: should we initialize the solution monitors by copying the data from the Flight instance submited in the `initialSolution` argument? If we do this, the resulting Flight will have trajectory data from both Flights. Is this a convinience or an inconvience?

### DIFF
--- a/rocketpy/Flight.py
+++ b/rocketpy/Flight.py
@@ -616,29 +616,7 @@ class Flight:
 
         # Flight initialization
         self.__init_post_process_variables()
-        # Initialize solution monitors
-        self.outOfRailTime = 0
-        self.outOfRailState = np.array([0])
-        self.outOfRailVelocity = 0
-        self.apogeeState = np.array([0])
-        self.apogeeTime = 0
-        self.apogeeX = 0
-        self.apogeeY = 0
-        self.apogee = 0
-        self.xImpact = 0
-        self.yImpact = 0
-        self.impactVelocity = 0
-        self.impactState = np.array([0])
-        self.parachuteEvents = []
-        self.postProcessed = False
-        self.latitude = 0  # Function(0)
-        self.longitude = 0  # Function(0)
-        # Initialize solver monitors
-        self.functionEvaluations = []
-        self.functionEvaluationsPerTimeStep = []
-        self.timeSteps = []
-        # Initialize solution state
-        self.solution = []
+        self.__init_solution_monitors()
         self.__init_flight_state()
         
         self.tInitial = self.initialSolution[0]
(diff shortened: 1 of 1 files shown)


Gui-FernandesBR:
`One option that we have is: should we initialize the solution monitors by copying the data from the Flight instance submited in the initialSolution argument? If we do this, the resulting Flight will have trajectory data from both Flights. Is this a convinience or an inconvience?`

Both options of anwser can be right. I mean, there will be moments where the user wants to copy data from previous flight, but also there will be moments qhere you want to start a totally new flight. Should we be flexible? that old mergeFlights idea doesn't sound weird here.

Another question: Would be good idea allowing the user to set the time where to get the initial solution state? I mean, is that for sure that the last state vector is always the best vector to be used in the initialization of a second flight?

All in all, just have to say thank you for this PR, @giovaniceotto , this will definitely improve our development!

rocketpy/Flight.py
```diff
@@ -539,15 +539,18 @@ def __init__(
...
-                                w1Init, w2Init, w3Init].
-            If None, the initial solution will start with all null values,
-            except for the euler parameters which will be calculated based
+            initialSolution = []
+                self.tInitial,
+                xInit, yInit, zInit,
+                vxInit, vyInit, vzInit,
+                e0Init, e1Init, e2Init, e3Init,
+                w1Init, w2Init, w3Init
+            ].
```
MateusStano, line 550:
```suggestion
            initialSolution = [
                self.tInitial,
                xInit, yInit, zInit,
                vxInit, vyInit, vzInit,
                e0Init, e1Init, e2Init, e3Init,
                w1Init, w2Init, w3Init
            ].
```

MateusStano:
I like what @Gui-FernandesBR said! I also think having at least one test would be really important for this one since some changes in flight could make for some extreme initial conditions act weirdly. For now though seems to be a really good implementation

giovaniceotto:
Thanks for the reviews. I will make some quick updats to the implementation considering the feedback.

I agree that tests are needed. I will add a couple of them.

giovaniceotto:
Thanks for the reviews. I will make some quick updats to the implementation considering the feedback.

I agree that tests are needed. I will add a couple of them.

Gui-FernandesBR:
@giovaniceotto please just write down what tasks you have in mind and we can separate ourselves to tackle everything together if this is faster

Gui-FernandesBR:
> Thanks for the reviews. I will make some quick updats to the implementation considering the feedback.
> 
> I agree that tests are needed. I will add a couple of them.

Hey @giovaniceotto really sorry for that, but I need to move forward today with the "complex simulations" development. We might be using the codes from this PR with current state.
You can add your ideas as TODOs and this is going to be already good, better than coding.

Gui-FernandesBR (approved):
Nice one, tks

PR #233, opened 2022-09-17
ENH: Liquid Motors

giovaniceotto:
Co-authored-by: @phmbressan
Co-authored-by: @MrGribel
Co-authored-by: @MateusStano
Co-authored-by: @ompro07

## Pull request type

## Pull request checklist

RocketPy does not support Rockets powered by Liquid Rocket Engines.

## What is the new behavior?

A `LiquidMotor` class is created based on the `Motor` abstract class. It makes use of other new classes: `Tank`, `UllageBasedTank`, `MassBasedTank`, `MassFlowRateBasedTank` and `Fluid`.

- [x] Yes
- [ ] No

`from rocketpy.Motor import SolidMotor` does not work anymore. Instead, one will need to use `from rocketpy import SolidMotor` or `from rocketpy.motors import SolidMotor`.

## Other information
Other Pull Requests will be created to complement this one before this can be approved and merged.

## TODO List

- [x] Refactor Motor class so as to remove its dependencies from Solid Motors (make it truly abstract). 
- [x] Refactor Rocket class (motor and inertia).
- [x] Review bugs in new Equations of Motion from Flight class.
- [x] Implement Funcify to methods in LiquidMotor.
- [x] Add https://github.com/Space-Enterprise-at-Berkeley/LiquidRocketPy enhancements.


Gui-FernandesBR:
Task linked: [CU-864dp0fn5 Liquid Motors](https://app.clickup.com/t/864dp0fn5)

giovaniceotto:
In [ebe458f](https://github.com/RocketPy-Team/RocketPy/pull/233/commits/ebe458f1c6ad40832d14b620eda60f7dfacbc7d9), I have defined methods for all components of the inertia tensor, for the motor classes only, with the exception of the EmptyMotor class.

Here are a couple of observations:

- The `GenericMotor` class is missing a definition for the abstract method `mass`. @phmbressan, were you aware of this?
- I did not like having to repeat the docstring so often... It only occupies space. Could we leave the docstring for these methods in the abstract methods only? Python naturally uses the docstring of the abstract method if the derived methods doesn't have any. However, the `funcify` decorator seems to mess this up.
- In the future, I want to derive the EmptyMotor class from the GenericMotor class.

phmbressan:
@giovaniceotto Regarding your observations:

- I believe the `GenericMotor` should inherit the `mass` method from the generic motor. I will check if this what is actually happening.

- I believe we can just leave the abstract methods docstrings. However, it is important to keep in mind that there are a few methods that only exists in the concrete class, I think that would be weird to just docstring them.

- Can't the `EmptyMotor` be an instance of `GenericMotor` with no thrust?

rocketpy/motors/Tank.py
```diff
@@ -0,0 +1,524 @@
...
+from rocketpy.Function import Function, funcify_method
+from rocketpy.utilities import except_negative
+
+
+class Tank(ABC):
+    def __init__(self, name, geometry, gas, liquid=0):
+        self.name = name
+        self.geometry = geometry
+        self.gas = gas
+        self.liquid = liquid
```
MateusStano, line 18: The argument name "mass" makes it seem like it is the tank mass, but it is only the mass of the tank's contents. I think we should change this name for something a bit more descriptive

rocketpy/motors/Tank.py
```diff
@@ -0,0 +1,524 @@
...
+    def gasHeight(self, t):
+        fluid_volume = self.gasVolume + self.liquidVolume
+        gasHeight = self.geometry.inverse_volume(fluid_volume(t))
+        if gasHeight <= self.geometry.top:
+            return gasHeight
+        else:
+            raise ValueError(
+                f"The tank {self.name}, is overfilled"
+                f"with gas height {gasHeight} at time {t}"
+            )
```
MateusStano, line 371: This overfilling check needs to be done in `liquidHeight` too.

rocketpy/motors/Tank.py
```diff
@@ -0,0 +1,524 @@
...
```
MateusStano, line 524: This overfilling check needs to be done in `liquidHeight` too.

rocketpy/motors/HybridMotor.py
```diff
@@ -0,0 +1,526 @@
...
+        grainInitialInnerRadius,
+        grainInitialHeight,
+        grainSeparation=0,
+        nozzleRadius=0.0335,
+        nozzlePosition=0,
+        throatRadius=0.0114,
+        reshapeThrustCurve=False,
+        interpolationMethod="linear",
+        coordinateSystemOrientation="nozzleToCombustionChamber",
+    ):
```
MateusStano, line 130: What is up with the optional argument values of `nozzleRadius` and `throatRadius`? I know this comes from SolidMotor but I think it is time to change it.

MateusStano (changes requested):
Hey hey, just made a few commits that fixed a few small issues. I think the commits are self-explanatory. Also made a first review of the overall code, here are some comments:

-  I think, for optimization purposes, we should make CylindricalTank and Spherical tank implementation completely algebraic. The code is more elegant when using only Functions but much slower. It is an easy change and the approach using Function can still be used in a "arbitrary tank geometry" class. What I'd suggest is that we still create a "CustomTankGeometry" that uses the Function class extensively

- The `@except_negative` decorator has an issue with numerical errors due to integration. This needs to be further tested to guarantee it won't cause problems with user inputs

- The error in the image below keeps popping up because of the `@except_negative` decorator. This happens if you call `Pressure_tank.liquidMass()` in `SEBLM.ipynb`. It seems that the attribute `liquid_mass` is already a Function object, and for some reason this breaks when using the `@funcify_method` with the `@except_negative`.

image

- In `SEBLM.ipynb`, the `Pressure_tank.gasHeight` values are blowing up.

- The Hybrid Motor class seems to be working fine, but still needs further testing with actual motor data

- `allInfo` method for LiquidMotors needs to be improved. `allInfo` for Hybrid seems to be missing some info regarding the tank added

rocketpy/motors/Tank.py
giovaniceotto: Agreed.

rocketpy/motors/HybridMotor.py
giovaniceotto: In fact, we had already fixed this in the `SolidMotor` class. I'll update the `HybridMotor` constructor to follow the same pattern: no default value for `grainSeparation` and `nozzleRadius`. The `throatRadius` keeps a default value of `0.01` due to its importance only for computing Kn.

giovaniceotto:
Getting the following test errors, but they seem unrelated to the subject of this PR. Any change they came from beta v1 branch?

```
FAILED tests/test_rocket.py::test_add_nose_assert_cp_cm_plus_nose[0.6666666666666666-conical] - assert -7.146806469828009 ± 7.1e-12 == -7.1522240775789525 ±...
FAILED tests/test_rocket.py::test_add_nose_assert_cp_cm_plus_nose[0.466-ogive] - assert -8.02893397638969 ± 8.0e-12 == -8.034351584140632 ± 8.0e-12
FAILED tests/test_rocket.py::test_add_nose_assert_cp_cm_plus_nose[0.563-lvhaack] - assert -7.602523503948745 ± 7.6e-12 == -7.607941111699688 ± 7.6e-12
FAILED tests/test_rocket.py::test_add_nose_assert_cp_cm_plus_nose[0.5-default] - assert -7.879470511822761 ± 7.9e-12 == -7.884888119573703 ± 7.9e-12
FAILED tests/test_rocket.py::test_add_nose_assert_cp_cm_plus_nose[0.5-not a mapped string, to show default case] - assert -7.879470511822761 ± 7.9e-12 == -7...
FAILED tests/test_rocket.py::test_add_tail_assert_cp_cm_plus_tail - assert 9.61377233165358 ± 9.6e-12 == 9.608354723902638 ± 9.6e-12
FAILED tests/test_rocket.py::test_add_fins_assert_cp_cm_plus_fins - assert 8.682852322822908 ± 8.7e-12 == 8.677434715071966 ± 8.7e-12
```

MateusStano:
> Getting the following test errors, but they seem unrelated to the subject of this PR. Any change they came from beta v1 branch?

Seems it has not come from v1 branch. Apparently, the aero surfaces tests are breaking only in the `dimensionless_rocket` assertions. I couldn't really find out what changed the dimensionless_rocket though

.DS_Store
Gui-FernandesBR, line 1: Why is this file being created here?

rocketpy/utilities.py
```diff
@@ -15,6 +15,20 @@
 from .AeroSurfaces import TrapezoidalFins
 
 
+def except_negative(method):
```
Gui-FernandesBR, line 18: I believe this functions should goes into the `tools.py` file.
I'm opened to discussions

phmbressan:
> Getting the following test errors, but they seem unrelated to the subject of this PR. Any change they came from beta v1 branch?
> 
> ```
> FAILED tests/test_rocket.py::test_add_nose_assert_cp_cm_plus_nose[0.6666666666666666-conical] - assert -7.146806469828009 ± 7.1e-12 == -7.1522240775789525 ±...
> FAILED tests/test_rocket.py::test_add_nose_assert_cp_cm_plus_nose[0.466-ogive] - assert -8.02893397638969 ± 8.0e-12 == -8.034351584140632 ± 8.0e-12
> FAILED tests/test_rocket.py::test_add_nose_assert_cp_cm_plus_nose[0.563-lvhaack] - assert -7.602523503948745 ± 7.6e-12 == -7.607941111699688 ± 7.6e-12
> FAILED tests/test_rocket.py::test_add_nose_assert_cp_cm_plus_nose[0.5-default] - assert -7.879470511822761 ± 7.9e-12 == -7.884888119573703 ± 7.9e-12
> FAILED tests/test_rocket.py::test_add_nose_assert_cp_cm_plus_nose[0.5-not a mapped string, to show default case] - assert -7.879470511822761 ± 7.9e-12 == -7...
> FAILED tests/test_rocket.py::test_add_tail_assert_cp_cm_plus_tail - assert 9.61377233165358 ± 9.6e-12 == 9.608354723902638 ± 9.6e-12
> FAILED tests/test_rocket.py::test_add_fins_assert_cp_cm_plus_fins - assert 8.682852322822908 ± 8.7e-12 == 8.677434715071966 ± 8.7e-12
> ```

After a bit of digging, I found out that they are related to small innacuracies in motor CoM evaluation due to the `solve_ivp` termination event checker. This is not the first issue due to this fact and is exacerbated by the really tight testing tolerances.

- The "easy" fix would be to reduce tests tolerances, however it would be a massive reduction _e.g._ `1e-5`. I still need to experiment a bit more, but it has been quite hard bringing this number lower;
- Think in a better way to stop integration when the grain burns out completely that the `solve_ivp` termination event checker.

Gui-FernandesBR:
Ok, this should be our last week working on this PR (god bless us please!). Here a quick list of TO-DOs:
- [ ] @phmbressan , finish the burn_time introduction and merge it 
- [ ] @giovaniceotto , solve conflicts with linear algebra
- [ ] @MateusStano , solve the current errors with motor mass PR and merge it
- [ ] @phmbressan , Double check (complete) docstrings
- [ ] @MateusStano , Double check (complete) notebooks (each motor class having at least 1 simple example)
- [ ] @Gui-FernandesBR , review the entire PR, great!

Some other tasks will need to be done after this PR to be merged, i.e. in different new PRs:
- [ ] @Gui-FernandesBR refactor info() and allInfo() methods to keep the consistency with the rest of the code. This is synonym of "moving plots and prints"
- [ ] Ensure a good code coverage

rocketpy/motors/Motor.py
```diff
@@ -0,0 +1,1373 @@
...
+
+        return thrust
+
+    @staticmethod
+    def clipThrust(thrust, new_burn_time):
+        """Clips the thrust curve data points according to the new_burn_time
+        parameter. If the burn_time range does not coincides with the thrust
+        dataset, their values are interpolated.
+
+        Parameters
```
Gui-FernandesBR, line 909: Please add `new_burn_time` parameter here.
I don't want to guess what is it.

rocketpy/motors/Motor.py
```diff
@@ -0,0 +1,1373 @@
...
+        None
+
+        Return
+        ------
+        None
+        """
+        pass
+
+
+class GenericMotor(Motor):
```
Gui-FernandesBR, line 1113: Add docstring here because I honestly don't have any idea why this thing exists.

rocketpy/motors/Tank.py
```diff
@@ -0,0 +1,626 @@
...
+from rocketpy.tools import tuple_handler
+
+
+class Tank(ABC):
+    """Abstract Tank class that defines a tank object for a rocket motor, so
+    that it evaluates useful properties of the tank and its fluids, such as
+    mass, volume, fluid flow rate, center of mass, etc.
+    """
+
+    def __init__(self, name, geometry, flux_time, gas, liquid, discretize=100):
```
Gui-FernandesBR, line 19: So, maybe the name argument should be optional? 
In the AeroSurface's children classes, the name is always optional, since it is not so crucial to the simulation.

What do you think that is better?

rocketpy/motors/Tank.py
```diff
@@ -0,0 +1,626 @@
...
+        """
+        Returns the volume of the gas as a function of time.
+
+        Returns
+        -------
+        Function
+            Volume of the gas as a function of time.
+        """
+        pass
+
```
Gui-FernandesBR, line 128: What do you think about this little guy?

```suggestion

@property
    def netVolume(self):
        """
        Returns the net volume of the tank as a function of time. Net
        volume is the volume of the gas plus the volume of the liquid.

        Returns
        -------
        Function
            Net volume of the tank as a function of time.
        """
        return self.gasVolume + self.liquidVolume
        
```

rocketpy/motors/Tank.py
```diff
@@ -0,0 +1,626 @@
...
+        Returns
+        -------
+        Function
+            Inertia tensor of the tank's fluids as a function of time.
+        """
+        return self.liquidInertiaTensor + self.gasInertiaTensor
+
+
+class MassFlowRateBasedTank(Tank):
+    def __init__(
```
Gui-FernandesBR, line 333: Docstring please 🥲

Gui-FernandesBR: At a first look, I wanted to understand what is the actual difference of using one tank model or another. Should they always give me the same results?

rocketpy/motors/Tank.py
```diff
@@ -0,0 +1,626 @@
...
+
+class LevelBasedTank(Tank):
+    def __init__(
+        self,
+        name,
+        geometry,
+        flux_time,
+        liquid,
+        gas,
+        liquid_height,
```
Gui-FernandesBR, line 512: This one was a bit hard to understand, way more than what it looks like.

rocketpy/motors/Tank.py
```diff
@@ -0,0 +1,626 @@
...
+        discretize=100,
+    ):
+        super().__init__(name, geometry, flux_time, gas, liquid, discretize)
+
+        # Define liquid height
+        self.liquid_height = Function(
+            liquid_height, "Time (s)", "volume (m³)", "linear"
```
Gui-FernandesBR, line 519: Is it volume or height?

rocketpy/motors/TankGeometry.py
```diff
@@ -0,0 +1,376 @@
...
+    def average_radius(self):
+        """
+        The average radius of the tank.
+
+        Returns
+        -------
+        float
+            Average radius of the tank.
+        """
+        return self.radius.average(self.bottom, self.top)
```
Gui-FernandesBR, line 92: mmmmmm I'm not sure if the result here is what you expect.

This seems to be dependent on the discretization used.
If you set more points to define the geometry near the caps, the average radius will be lower than what it really looks like.

Maybe we should integrate the transversal area and divide it by the height?

rocketpy/motors/TankGeometry.py
```diff
@@ -0,0 +1,376 @@
...
+        the domain of the geometry.
+
+        Returns
+        -------
+        float
+            Top coordinate of the tank.
+        """
+        return max(self._geometry.keys())[1]
+
+    @cached_property
```
Gui-FernandesBR, line 120: this calculation is so simple, why cache it?

rocketpy/motors/TankGeometry.py
```diff
@@ -0,0 +1,376 @@
...
+        -------
+        float
+            Total height of the tank.
+        """
+        return self.top - self.bottom
+
+    @funcify_method("Height (m)", "Area (m²)", extrapolation="zero")
+    def area(self):
+        """
+        The area of the tank cross section as a function of height.
```
Gui-FernandesBR, line 135: You should define what is the cross section. 
I understand that it is the area parallel to the nozzle exit, but not everyone will understand the same.

Gui-FernandesBR (changes requested):
Partial review only. I checked only 37/71 files so far.
I updated some notebooks too, and in the next days I will take a look at the others.

Everything seems simple in my review, that's because I didn't check the inertia moments yet XD

Great work here btw, some outputs were really fantastic!!!

rocketpy/Flight.py
```diff
@@ -1563,6 +1575,258 @@ def uDot(self, t, u, postProcessing=False):
...
+        # r = Vector([x, y, z])               # CDM position vector
+        v = Vector([vx, vy, vz])  # CDM velocity vector
+        e = [e0, e1, e2, e3]  # Euler parameters/quaternions
+        w = Vector([omega1, omega2, omega3])  # Angular velocity vector
+
+        # Retrieve necessary quantities
+        rho = self.env.density.getValueOpt(z)
+        total_mass = self.rocket.totalMass.getValueOpt(t)
+        total_mass_dot = self.rocket.totalMass.differentiate(t)
+        total_mass_ddot = self.rocket.totalMass.differentiate(t, order=2)
```
Gui-FernandesBR, line 1614: Did you consider to calculate the total_mass_dot inside the Rocket class, storing it as a Function attribute, and then simply using `getValueOpt()` to retrieve the value at each time it is needed? 

I wonder if the differentiate function could be slowing down the code since the derivative is being calculated inside the simulation loop.

Plus, I also think that the user could want to see the total_mass_dot and total_mass_ddot before running the flight simulation.

rocketpy/Flight.py
```diff
@@ -1563,6 +1575,258 @@ def uDot(self, t, u, postProcessing=False):
...
+        R3 += -0.5 * rho * (freestreamSpeed**2) * self.rocket.area * (dragCoeff)
+
+        ## Off center moment
+        M1 += self.rocket.cpEccentricityY * R3
+        M2 -= self.rocket.cpEccentricityX * R3
+
+        # Get rocket velocity in body frame
+        vB = Kt @ v
+        # Calculate lift and moment for each component of the rocket
+        for aeroSurface, position in self.rocket.aerodynamicSurfaces:
```
Gui-FernandesBR, line 1708: This loop is being repeated in the different uDot functions. 
Maybe a generic function should be created to perform this loop separatly?

I know it could slow down the code a little bit, but it would also keep everything more organized in this file.

rocketpy/motors/Motor.py
```diff
@@ -0,0 +1,1277 @@
...
+        self.dry_I_23 = inertia[5]
+
+        # Handle .eng file inputs
+        if isinstance(thrustSource, str):
+            if thrustSource[-3:] == "eng":
+                _, _, points = Motor.importEng(thrustSource)
+                thrustSource = points
+
+        # Evaluate raw thrust source
+        self.interpolate = interpolationMethod
```
Gui-FernandesBR, line 278: Duplicated with line 257

```suggestion
```

rocketpy/motors/SolidMotor.py
```diff
@@ -0,0 +1,736 @@
...
+    def exhaustVelocity(self):
+        """Exhaust velocity by assuming it as a constant. The formula used is
+        total impulse/propellant initial mass.
+
+        Returns
+        -------
+        self.exhaustVelocity : rocketpy.Function
+            Gas exhaust velocity of the motor.
+        """
+        return self.totalImpulse / self.propellantInitialMass
```
Gui-FernandesBR, line 372: Even if it a constant value, I think you should return a rocketpy.Function object.

This would improve the consistency with other motor classes

rocketpy/Rocket.py
```diff
@@ -428,7 +489,167 @@ def evaluateStaticMargin(self):
...
+        the rocket's center of mass. The inertias are saved and returned
+        in units of kg*m².
+
+        Parameters
+        ----------
+        None
+
+        Returns
+        -------
+        self.dry_I_11 : float
```
Gui-FernandesBR, line 506: is it really a float?

Please check it because it smells like rocketpy.Function again

rocketpy/Rocket.py
```diff
@@ -428,7 +489,167 @@ def evaluateStaticMargin(self):
...
+        """
+        # Compute axes distances
+        noMCM_to_CDM = self.center_of_mass_without_motor - self.centerOfDryMassPosition
+        motorCDM_to_CDM = (
+            self.motor_center_of_dry_mass_position - self.centerOfDryMassPosition
+        )
+
+        # Compute dry inertias
+        self.dry_I_11 = (
+            self.I_11_without_motor
```
Gui-FernandesBR, line 552: This is confusing, but it is working!

We should create specific tests to these calculations in the future.

rocketpy/Flight.py
```diff
@@ -1563,6 +1575,258 @@ def uDot(self, t, u, postProcessing=False):
...
+        rail.
+
+        Parameters
+        ----------
+        t : float
+            Time in seconds
+        u : list
+            State vector defined by u = [x, y, z, vx, vy, vz, q0, q1,
+            q2, q3, omega1, omega2, omega3].
+        postProcessing : bool, optional
```
Gui-FernandesBR, line 1591: Not related to this PR but... When are we going to depracate `postProcessing` argument?

rocketpy/prints/rocket_prints.py
```diff
@@ -45,8 +45,12 @@ def inertia_details(self):
         print(
             "Rocket Mass: {:.3f} kg (With Propellant)".format(self.rocket.totalMass(0))
         )
-        print("Rocket Inertia I: {:.3f} kg*m2".format(self.rocket.inertiaI))
-        print("Rocket Inertia Z: {:.3f} kg*m2".format(self.rocket.inertiaZ))
+        print("Rocket Inertia 11: {:.3f} kg*m2".format(self.rocket.dry_I_11))
```
Gui-FernandesBR, line 48: 1. "Rocket Inertia" is soemthing quite ambiguous here as the user will not know if it considers the motor (without propellant) or no. I suggest letting it more explicit.
2. In the future, we should create a matrix plot here, just to facilitate things.

Gui-FernandesBR:
Some simulations were not adjusted yet, and therefore they can't be ran without errors, currently. These simulations are:
- tests\fixtures\acceptance\EPFL_Bella_Lui\bella_lui_flight_sim.py
- tests\fixtures\acceptance\NDRT_2020\ndrt_2020_flight_sim.py
- tests\fixtures\acceptance\PJ_Valetudo\valetudo_flight_sim.py
- docs\notebooks\compare_flights_usage.ipynb
- docs\notebooks\deployable_payload_example.ipynb
- docs\notebooks\dispersion_analysis\dispersion_analysis.ipynb
- docs\notebooks\dispersion_analysis\parachute_drop_from_helicopter.ipynb

However, they can easily be modified in future PRs.

rocketpy/Flight.py
MateusStano: I think it is a great idea as long as it does not make the simulation slower, which it probably won't. I think this will definitely be changed when we go through with optimizing the speed of the new EOMs

rocketpy/Flight.py
MateusStano: A `calculateAerodynamicExternalReactions` function could solve this. I doubt that will slow anything down

rocketpy/Rocket.py
MateusStano: Yes, all dry inertias should be floats

rocketpy/Flight.py
MateusStano: I guess the v1 release is the best moment to do so. We can also remove all the other deprecated functions

rocketpy/motors/HybridMotor.py
```diff
@@ -0,0 +1,565 @@
...
+        return self.totalImpulse / self.propellantInitialMass
+
+    @funcify_method("Time (s)", "Mass (kg)")
+    def propellantMass(self):
+        """Evaluates the total propellant mass of the motor as the sum
+        of each tank mass and the grains mass.
+
+        Parameters
+        ----------
+        t : float
```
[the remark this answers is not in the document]
phmbressan: After pondering for a bit, I agree with you. Solved by 33c46c8b333d3fbde74423088d4f761000d66e69.

rocketpy/motors/Motor.py
```diff
@@ -0,0 +1,1373 @@
...
+        total impulse/propellant initial mass.
+
+        Parameters
+        ----------
+        None
+
+        Returns
+        -------
+        self.exhaustVelocity : float
+            Constant gas exhaust velocity of the motor.
```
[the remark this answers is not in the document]
phmbressan: Good catch. Solved by 33c46c8b333d3fbde74423088d4f761000d66e69.

rocketpy/motors/LiquidMotor.py
```diff
@@ -0,0 +1,390 @@
...
+        Notes
+        -----
+        The e_3 direction is assumed to be the axial direction of the rocket
+        motor.
+
+        References
+        ----------
+        .. [1] https://en.wikipedia.org/wiki/Moment_of_inertia#Inertia_tensor
+        """
+        return 0
```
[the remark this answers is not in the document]
phmbressan: Personally, I believe the main reasoning behind the `funcify` was to avoid making this explicit `Function` "cast". I prefer not returning a `Function` only to apply `funcify` immediately after. Seems a bit redundant.

rocketpy/motors/Motor.py
phmbressan: Solved by 33c46c8b333d3fbde74423088d4f761000d66e69.

rocketpy/motors/Motor.py
phmbressan: Added on 33c46c8b333d3fbde74423088d4f761000d66e69.

rocketpy/motors/Tank.py
phmbressan: I prefer keeping it mandatory so that to encourage the user to label tanks in a explicit way. The naming is used in various exceptions and is fundamental to distinguish, for instance, which of the added Tanks overfilled.

Even if is not mandatory for simulation I think the parameter should always be set. In the future this might be revised if necessary.

rocketpy/motors/Tank.py
phmbressan: Great suggestion. I have only made the implementation concrete since some tanks can have this value computed in specific ways that are more accurate. See 97c0b160fbd1f2c432e121ebff0ab1763946d78a.

rocketpy/motors/Tank.py
phmbressan: Docstrings added on 80daa46310121108ea6557f44e5c4663c2892ef3. The difference and assumptions behind each Tank were succinctly described in the added docstrings. 

To sum up, the UllageBasedTank and LevelBasedTank are less accurate than the others, since they assume uniform gas distribution filling all the portion that is not occupied by liquid. Therefore, these tanks can only model the tank state until the liquid runs out.

rocketpy/motors/Tank.py
phmbressan: I hope the docstrings added on 80daa46310121108ea6557f44e5c4663c2892ef3 made it clearer. If it is still confusing, do not hesitate in continuing this thread.

rocketpy/motors/Tank.py
phmbressan: Fixed on 80daa46310121108ea6557f44e5c4663c2892ef3.

rocketpy/motors/TankGeometry.py
phmbressan: Could you elaborate a bit more your concerns with this implementation. Just to make sure we are on the same page, the formula for average used is the `average` method of `Function` class.

This method integrates the function and divides it by the domain length.

rocketpy/motors/TankGeometry.py
phmbressan: That was a mistake, fixed on 80daa46310121108ea6557f44e5c4663c2892ef3.

rocketpy/motors/TankGeometry.py
phmbressan: Solved by expanding docstrings on feb8166690df4330c4c3f8f817f118cd102da25b.

rocketpy/motors/SolidMotor.py
phmbressan: The same comment that I made for your earlier suggestion applies here. The value is constant, but is turned into a Function by funcify.

Explicitly calling the Function is a bit redundant.

rocketpy/Rocket.py
phmbressan: Absolutely!

rocketpy/motors/Tank.py
Gui-FernandesBR: @phmbressan , any opinion in this one?

rocketpy/motors/Tank.py
Gui-FernandesBR: duplicated

rocketpy/motors/Tank.py
Gui-FernandesBR: @phmbressan , any comment?

rocketpy/motors/LiquidMotor.py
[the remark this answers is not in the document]
Gui-FernandesBR: Ok. 
Since you guarantee no problem will be generated by doing this, let's go.

rocketpy/motors/TankGeometry.py
Gui-FernandesBR: I was thinking you were simply taking the average of the values, not doing the integral.
You guys were more prepared than expected hahahah

Well done!

rocketpy/motors/Tank.py
Gui-FernandesBR: Agreed based on your reasons.
This is a different case of AeroSurface then.

rocketpy/Flight.py
Gui-FernandesBR: Let's keep an eye on it for future PRs!

rocketpy/Flight.py
Gui-FernandesBR: For future reference: 
- https://peps.python.org/pep-0387/
- https://peps.python.org/pep-0702/#deprecation-policy

rocketpy/prints/rocket_prints.py
Gui-FernandesBR: PR #387 will address this one

rocketpy/motors/Tank.py
phmbressan: Good points. I believe the name `fluidMass` is more descriptive.

rocketpy/motors/Tank.py
phmbressan: This is a good catch. To be clear, my earlier reasoning for not doing this check was that, in code, `gasHeight >= liquidHeight`.

Therefore, if the liquidHeight verfilled, the gasHeight exception would trigger. However, this may not happen, if, for some reason, the property `liquidHeight` is called alone.

This is indeed a mistake.

rocketpy/motors/Tank.py
Gui-FernandesBR: Solved here: [c810266](https://github.com/RocketPy-Team/RocketPy/pull/233/commits/c81026638be688442ae06f8056b4c96552b422e7)

rocketpy/motors/Tank.py
Gui-FernandesBR: Solved here: [3cb8e8d](https://github.com/RocketPy-Team/RocketPy/pull/233/commits/3cb8e8d81b49c0e1925ab80f0a4d6159952b7a90)

PR #226, opened 2022-09-11
WIP: ENH: Env Analysis Euroc 2022

Gui-FernandesBR:
## Pull request type
- [X] Code base additions (bugfix, features)

## Pull request checklist

- Code base maintenance (refactoring, formatting, renaming):

  - [x] Docs have been reviewed and added / updated if needed
  - [x] Lint (`black rocketpy`) has passed locally and any fixes were made
  - [x] All tests (`pytest --runslow`) have passed locally

- Code base additions (for bug fixes / features):

  - [ ] Tests for the changes have been added
  - [x] Docs have been reviewed and added / updated if needed
  - [x] Lint (`black rocketpy`) has passed locally and any fixes were made
  - [x] All tests (`pytest --runslow`) have passed locally

## What is the current behavior?

The EnvironmentAnalysis class was an important project that lead us to our last milestone [Environment Analysis SACup 2022](https://github.com/RocketPy-Team/RocketPy/milestone/6).
However, it is time to work hard again in order to fill some blank spaces we left last time, and also to help reaching the newest milestone of our team: [RocketPy at EuroC 2022](https://github.com/RocketPy-Team/RocketPy/milestone/7)

### Features that are being required:
- [x] 1 - Create a function to export environment object to be used as customAtmosphere in other simulations
- [x] 2 - Create a function to export environment_analysis to be used as customAtmosphere in Environment
- [x] 3 - Catch Reanalysis Data for EuroC's site and save it on data folder (#238 )
- [x] 4 - Run EnvAnalys example for EuroC's site and export to a proper format (needs debug pacman problem)

### Additional features and tasks:
- [x] Create Wind Heading profiles (#197) (Need discussion over "wrap 0-360º problem")
- [x] Save EnvironmentAnalysis processed data so it can be loaded later in order to speed up by skipping the parse steps (#237) 
- [x] Compare EnvAnalysis against Environement Forecasts (#244 )

### Future developments (outside this milestone)
- Check if we can aggregate both surface and pressure level data from environment analysis into a flight simulation
- Automate data input by using OpenDap protocol (check if it is possible with Reanalysis)
- Add tests to EnvironmentAnalysis Class

## Does this introduce a breaking change?

- [ ] Yes
- [X] No

PR #237, opened 2022-09-21
ENH: Refactor Env Analysis .json conversions

Gui-FernandesBR:
## Pull request type

## Pull request checklist

## What is the new behavior?

I'm fixing a problem regarding the PR #226 tasks list.
Now it's finally possible to save the already-processed data from EnvironmentAnalysis into a .json file and then re-use it to initialize other Environment Analysis objects 

- [ ] Yes
- [X] No

## Other information
I'm creating this PR only to have theses steps documented in the future, reviewing it is important, but please keep in mind that others PR are coming soon and they will need these implementations already merged

### DIFF
--- a/docs/user/requirements.rst
+++ b/docs/user/requirements.rst
@@ -27,6 +27,7 @@ The following packages are needed in order to run RocketPy:
 - timezonefinder
 - simplekml
 - ipywidgets >= 7.6.3
+- jsonpickle
 
  
 All of these packages, with the exception of netCDF4, should be automatically installed when RocketPy is installed using either ``pip`` or ``conda``.
--- a/docs/user/requirements.rst
+++ b/docs/user/requirements.rst
@@ -49,6 +49,7 @@ The packages needed can be installed via ``pip`` by running the following lines 
     pip install pytz
     pip install timezonefinder
     pip install simplekml
+    pip install jsonpickle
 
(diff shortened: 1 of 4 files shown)


Gui-FernandesBR:
Currently facing the following problem: 
- Some values of 'cloudBaseHeight' are masked
image

- When converting to json, this is what we get:
image

- I could replace theses values with NaN, but the ideal scenario would be replacing values with maskes **before** saving the .json file. Could anyone help with this?

Gui-FernandesBR:
I'm attaching the files that I've used to test the current code. 

The implementation saved significantly amount of time, **reducing from 61 to 2 seconds** the Environment Analsis performance in the case example
[environment_analysis_class_usage.zip](https://github.com/RocketPy-Team/RocketPy/files/9617728/environment_analysis_class_usage.zip)

PatrickSampaioUSP:
Sorry for not remembering this before! But for complexes classes, like EnvAnalysis, we should use `jsonpickle` to save it! It can serialize the class into a string and after you load it using jsonpickle and it will return you the class as you saved, so we can accomplish this with 2 lines of code (therefore don't needing tests). I think we should go in this direction, because this method will solve the problem even when we change the attributes of the class, and it's a perfect serializer. I'm doing a pull request to this branch implementing this idea, and you can review if it really does what we need.

PR #249, opened 2022-09-29
ENH: Introducing properties in Flight class

Gui-FernandesBR:
## Pull request type

## Pull request checklist

The `Flight.postProcess()` method usually takes too many time to run, creating a huge step from 

## What is the new behavior?

- Most of the old attributes (e.g. `Flight.x`, `Flight.angleOfAttack`, `Flight.speed`) are now defined by using properties decorators at the Flight class. Therefore they can be accessed without previous running `postProcess()`, meaning that the property object will be evaluated only when it is called.
- I have used cached_properties since memory usage was not a problem at the time, but perhaps in the future it can be modified. This way we guarantee that. Mutating properties can be a particular trick part here, I'm not an expert to explain, but we should be aware errors similar to: "Attribute Error: can't set attribute ..."
- The old `Flight.postProcess()` was divided into different smaller parts, allowing the user to call specific attributes instead of running everything everytime. See the new `def __operation(self)` methods to get the idea.

Perhaps more attributes could be transformed in properties in the future. But right now, I have converted all that I believe that are feasible to be properties.

Just go ahead and try running:
> TestFlight = Flight(...)
> TestFlight.speed()

- [X] No

## Other Information:
I tested the code and everything seems to be working. I ask help to imagine different scenarios
No extra test is needed since this is just a refactor Pull Request, the changes were made on background only.

### DIFF
--- a/rocketpy/Flight.py
+++ b/rocketpy/Flight.py
@@ -6,6 +6,7 @@ __author__ = (
 __copyright__ = "Copyright 20XX, RocketPy Team"
 __license__ = "MIT"
 
+from logging import warning
 import math
 import time
 from functools import cached_property
@@ -16,6 +17,7 @@ import simplekml
 from scipy import integrate
 
 from .Function import Function
+import warnings
 
 
 class Flight:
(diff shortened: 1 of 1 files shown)


Gui-FernandesBR:
Hey! It was really bothering me the current design of postProcess. I hope this way we can improve Dispersion Simulations, Controlled Flights, and code performance as well.

I'll wait for your review, @giovaniceotto @MateusStano , and of course I'm accepting suggestions with regards to all the @properties architecture

rocketpy/Flight.py
```diff
@@ -1596,190 +1612,216 @@ def uDotParachute(self, t, u, postProcessing=False):
...
+        return self.speed[maxSpeedTimeIndex, 0]
+
+    @cached_property
+    def maxSpeed(self):
+        return self.speed(self.maxSpeedTime)
+
+    # Accelerations
+    @cached_property
+    def acceleration(self):
+        if self.retrievedAcceleration is not True:
```
PatrickSampaioUSP, line 1769: what is the purpose of using this condition when you already have the @cached_property in place?

Gui-FernandesBR: Thank you for the review, it pushed me to get rid of those booleans, I'm adding new changes and asking a re-review

Gui-FernandesBR:
@Patrick can you test and review please?
- Everything is a property now
- I've tested under different scenarios, code looks faster and I addressed the bugs that I found
- Tests are passing locally. If Environment tests doesn't work, it has nothing related with my changes on Flight.py

Be free to directly commit suggestions if you want to.

I'm already using those implementations on the Dispersion Class as they are important to calculate some parameters.

PatrickSampaioUSP (approved):
Awesome usage of the @cached_property

Gui-FernandesBR:
Just emphasizing that getting started notebooks are working as expected in this version of Flight.py, let's move forward

PR #255, opened 2022-10-09
Enh/liquid motors ullage tank

phmbressan:
## Pull request type

## Pull request checklist

## What is the new behavior?

Add support for tanks that are based on ullage inputs.

- [ ] Yes
- [x] No

## Other information

Enter text here...


rocketpy/motors/LiquidMotor.py
```diff
@@ -181,15 +194,36 @@ def centerOfMass(self, t):
...
-        bottomCapMass = self.liquid.density * self.bottomCap.filled_volume
-        cylinderMass = self.liquid.density * self.cylinder.filled_volume
-        upperCapMass = self.liquid.density * self.upperCap.filled_volume
+        bottomCapLiquidMass = self.liquid.density * self.bottomCap.filled_volume
+        cylinderLiquidMass = self.liquid.density * self.cylinder.filled_volume
+        upperCapLiquidMass = self.liquid.density * self.upperCap.filled_volume
+
+        bottomCapGasMass = self.gas.density * self.bottomCap.empty_volume
+        cylinderGasMass = self.gas.density * self.cylinder.empty_volume
+        upperCapGasMass = self.gas.density * self.upperCap.empty_volume
```
giovaniceotto, line 203: Genius idea to have an `empty_volume` attribute!

rocketpy/supplement.py
```diff
@@ -96,10 +116,30 @@ def filled_volume(self, volume):
...
             self._filled_height = 0
 
+    @property
+    def empty_volume(self):
+        return self.volume - self._filled_volume
+
+    @property
+    def empty_centroid(self):
+        if self.empty_volume == 0:
+            return 0
```
giovaniceotto, line 126: I know this will not change any results, but shouldn't it return `self.height` if we were to consider the limit of the centroid poisition as the empty volume becomes zero?

giovaniceotto (approved):

rocketpy/supplement.py
phmbressan: Yes, I believe you are correct. As you said it does not affect mathematically the results since this only happens when `empty_volume = 0`.

That if clause likely was not needed, it was more of a extra check so that nothing weird happens when the tank is completely filled.

Issue #46, opened 2020-09-21
Alpha and Beta for Cd assumptions?


nbrophy3:
I noticed that your Cd is purely from (I assume Rasaero) Power on and Power off cd data. With Missile Datcom readers and other simulations, either Alpha or both Alpha and Beta values are plotted along with Mach in order to yield a Cd at a certain moment in time. Do you make any adjustments to your Cd in order to account for this difference (like some kind of offset)? Or does your code make the assumption of small angle of attack and symmetric about an axis to avoid using this information? It's a good assumption tbh, if that's what you're doing, but how do your results compare to something like Rasaero or Open Rocket?

giovaniceotto:
@Gui-FernandesBR

Gui-FernandesBR:
I'm sorry @giovaniceotto, but you are probably the best person to answer that question. But I will try...

@nbrophy3 , in fact, Rocketpy considers the drag force (R1) as always parallel to the rocket's axis. We know that the most common literature assumes that the drag acts in a direction opposite to the movement of the aircraft, but there is also a reason for our approach.

I **think** it is more common to consider the drag force towards the "aircraft" axis (as RocketPy does) when we are working with **rockets**, but this is not a very certain answer. I'm sure @giovaniceotto can give us a good reason for that assumption.

Gui-FernandesBR:
The complete resolution for this question can be read on [RocketPy Article](https://ascelibrary.org/doi/abs/10.1061/%28ASCE%29AS.1943-5525.0001331), See Flight and Rocket Model for better compreend it.

_"Aerodynamic forces acting on rockets are usually decomposed into the axial force A and the normal force N"_

PR #263, opened 2022-10-23
ENH: Adding AeroSurfaces classes

Gui-FernandesBR:
## Pull request type

## Pull request checklist

It is impossible to access arguments of an exisiting aerodynamic surface.

## What is the new behavior?

Following the Parachute.py file example, I'm changing the addTail, addFins, addNose to be working with the help of separate classes. This way, we can better capture data from each object. 

Also, no more dictionaries to save aerodynamic surfaces data, instead let's keep everything inside classes! `getattr(obj, attribute)` will make our life easier.

- [ ] Yes
- [X] No

## Other information
I'm doing this just to make our life easier on the future dispersion class


giovaniceotto:
What a great surprise @Gui-FernandesBR! Since this exceeds the limit of 200 line changes, I request a discord call so that we can discuss the work that has been done in this PR before review.

How about 2022/10/24, from 22:00 UTC until 22:30 UTC?

@MateusStano is, of course, also invited as he is a reviewer and the back-end manager.

Feel free to suggest another option if this does not work for you.

Gui-FernandesBR:
> What a great surprise @Gui-FernandesBR! Since this exceeds the limit of 200 line changes, I request a discord call so that we can discuss the work that has been done in this PR before review.
> 
> How about 2022/10/24, from 22:00 UTC until 22:30 UTC?
> 
> @MateusStano is, of course, also invited as he is a reviewer and the back-end manager.
> 
> Feel free to suggest another option if this does not work for you.

This time doens't work for me, let's try another way, I will contact you.
Please don't worry about the size of this PR, you should be worried about what comes next (the dispersion class!). 💔
For now, surviving tests and running notebooks without breaking the code are the easiest confirmations I think you could try

MateusStano:
Oh, one more thing. The `getting_started.ipynb` notebook is set with elliptical fins instead of trapezoidal like it used to be. Is this intended?

Gui-FernandesBR:
I will try to review this one by today, but I am already a fan of your commits.

I don't see any reason to use elliptical fins on the getting started notebook.

Indeed, omce the Calisto rocket presented trapezoidal fins, let's keep with those...

rocketpy/Flight.py
```diff
@@ -1412,7 +1412,8 @@ def uDot(self, t, u, postProcessing=False):
                 compStreamVzBn = compStreamVzB / compStreamSpeed
                 if -1 * compStreamVzBn < 1:
                     compAttackAngle = np.arccos(-compStreamVzBn)
-                    cLift = aerodynamicSurface["cl"](compAttackAngle, freestreamMach)
+                    cLift = aerodynamicSurface.cl(compAttackAngle, freestreamMach)
+                    cLift = aerodynamicSurface.cl(compAttackAngle, freestreamMach)
```
giovaniceotto, line 1416:
```suggestion
                    cLift = aerodynamicSurface.cl(compAttackAngle, freestreamMach)
```

rocketpy/Flight.py
```diff
@@ -1428,8 +1429,8 @@ def uDot(self, t, u, postProcessing=False):
                     M1 -= (compCp + a) * compLiftYB
                     M2 += (compCp + a) * compLiftXB
             # Calculates Roll Moment
-            if aerodynamicSurface["name"] == "Fins":
-                Clfdelta, Cldomega, cantAngleRad = aerodynamicSurface["roll parameters"]
+            if aerodynamicSurface.name == "Fins":
+                Clfdelta, Cldomega, cantAngleRad = aerodynamicSurface.rollParameters
```
giovaniceotto, line 1433: Should we apply EAFP here rather than LBYL?

https://realpython.com/python-lbyl-vs-eafp/

giovaniceotto:
There are still some improvements to be made before this PR gets merged... but it is of mangnificent importance to RocketPy in the long run! Can't wait for this!

rocketpy/Flight.py
Gui-FernandesBR: Yes, I think we should, let me make a suggestion here, and you can say if it is clear enough or not: 

```suggestion
            try:
                Clfdelta, Cldomega, cantAngleRad = aerodynamicSurface.rollParameters
            except AttributeError:
                pass
```

However, I think we now should do something different, cheking the type of the aerodynamic surface, what do you think? And for that *we should import Fins class in the beggining of the file*

```suggestion
            if isinstance(aerodynamicSurface, Fins):
                Clfdelta, Cldomega, cantAngleRad = aerodynamicSurface.rollParameters
````

Gui-FernandesBR:
Hello @MateusStano , this is getting intersting... Really thank you for pushing such good features to these classes.

Describing what I did after your 1st review:
- [x] Added allInfo methods for both Tail and Nose 
- [x] Updated all the docstrings, being ready to go for a merge
- [x] Modify a single at Function to properly handle the 2D plots

My review comments are the following (6):

1. Can we analyze the name of some methods? Here is a list of what I saw we could change:
- `Fins.setAttribute` can create a confusion with `__setattr__` method, I suggest `modifyAttribute` or similar
- `evaluateRollCoefficients` could be `evaluateRollParameters`, since it mainly creates the rollParameters attribute
- `liftInfo` could be `aerodynamicInfo`?
- Also, any reason for keeping a single underscore at these methods? `_beta` and `_finNumCorrection`
- And just out of curiosity... In the future, will we use this beta function to other aero surfaces as well? 

3. I'd like to understand if these are correct: 
`self.rollParameters[0]()  # TODO: improve this plot`
`self.rollParameters[1]()  # TODO: improve this plot`
image
image

4. The Fin.draw() method is good, but I think we can move forward without this functionality for Tail and Nose. In the future we might want to do a overkill draw method that takes the axes from each of theses drawings and create the rocket draw at once.
5. Regarding the NoseCone, why don't we take into considerations the nose radius/diameter? It took me years to realize that this is not an input for nose definition.
6. Maybe we should add these new allInfo methods to our getting started, or do you think we can accelerate and deliver a rocket_class_usage with this detailed information? Probably would be too many information at the getting started!

Gui-FernandesBR:
Also, I always like to give some demos here, so people can easily preview what is going on, so let's go:
- First, NoseCone.allInfo() method:
image

- Now, take a look on what Fins.allInfo can do:
image
image
image
image
image
image


- Finally, Tail.allInfo() is here:
image

Gui-FernandesBR:
Question: 
- Should we implement drag coefficient based on both Mach and alpha? As far as I know, both openrocket and Rasaero can create that. This made me to rethink my answer on this issue: https://github.com/RocketPy-Team/RocketPy/issues/46

Gui-FernandesBR:
TODO: 
- transform def Fins(self) in an abstract class 
- AerodynamicSurface() also having an abstract class

- Add setAttribute ...
-

MateusStano (approved):

PR #279, opened 2022-11-05
FIX: dropping cached_properties functionality

Gui-FernandesBR:
## Pull request type

Please check the type of change your PR introduces:

- [x] Code maintenance (refactoring, formatting, renaming, tests)

## Pull request checklist

- Code base maintenance (refactoring, formatting, renaming):

  - [x] Docs have been reviewed and added / updated if needed
  - [x] Lint (`black rocketpy`) has passed locally and any fixes were made
  - [x] All tests (`pytest --runslow`) have passed locally

## What is the current behavior?

cached_properties were introduced by #249 , but apperantly this functionality is available only on python 3.8 or higher. 
As Google Colaboratory currently applies python 3.7 as default, we think it's not a good time to drop python 3.7 support.

## What is the new behavior?

Simple search and replace `@cached_property` --> `@property @functools.lru_cache()`

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

This only happened cause the github tests were seted to run using python 3.8, which was fixed by #262

PR #283, opened 2022-11-12
Fix: Cached Property support for Python 3.6 and 3.7

giovaniceotto:
## Pull request type

## Pull request checklist

Python 3.6 and 3.7 does not include `functools.cached_property`, which is currently being used in the develop branch causing errors.
Since RocketPy does support Python 3.6+ according to our [docs](https://docs.rocketpy.org/en/latest/user/requirements.html#python-version), this needs to be fixed.

## What is the new behavior?

@Gui-FernandesBR had proposed a solution for this issue in #279, however, I was not comfortable with it since it made use of `functools.lru_cache`. It is my understanding that such implementation uses much more memory since it stores a copy of the class instance to make sure that the input argument `self` matches before using the cached result.

Since I have a keen interest in special class-based decorators such as these, given similar applications in the Liquid Motors milestone, I did some research to learn how could this be improved.

Based on other implementations and also the [oficial `functools.cached_property` implementation](https://github.com/python/cpython/blob/main/Lib/functools.py#L965), I have created a simplified version to be used whenever the official implementation is not available. This version is faster and does not consume extra memory during caching.

In summary, the solution was:

- Create a new decorator named `cached_property` in a new auxiliary file named `tools.py`.
- In the Flight module, try to import `functools.cached_property` and if it fails, import `tools.cached_property` instead.

These are the only modifications thus far.

- [ ] Yes
- [x] No

## Other information

A couple of things to discuss:

- The `tools.py` file should probably be named `utils.py` and contain all auxiliary decorators, classes and functions that can be used by the rest of the core code but has no interesting use to end-users.
- The `utilities.py` should probably be renamed to `tools.py` as it contains a series of tools for the end-user.
- All Flight methods that have the `@cached_property` decorator also contain other arguments, such as `interpolation="spline", extrapolation="natural"`. This is **broken**. Properties cannot have any other argument besides self. Although this does not cause any errors during execution, it is currently impossible to use and change these arguments from their default value. **This does not seem right.**

### DIFF
--- a/rocketpy/tools.py
+++ b/rocketpy/tools.py
@@ -13,6 +13,9 @@ class cached_property:
     def __get__(self, instance, owner=None):
         if instance is None:
             return self
+        if self.attrname is None:
+            raise TypeError(
+                "Cannot use cached_property instance without calling __set_name__ on it.")
         cache = instance.__dict__
         val = cache.get(self.attrname, _NOT_FOUND)
         if val is _NOT_FOUND:
--- a/rocketpy/tools.py
+++ b/rocketpy/tools.py
@@ -15,7 +15,8 @@ class cached_property:
             return self
         if self.attrname is None:
             raise TypeError(
-                "Cannot use cached_property instance without calling __set_name__ on it.")
+                "Cannot use cached_property instance without calling __set_name__ on it."
+            )
         cache = instance.__dict__
         val = cache.get(self.attrname, _NOT_FOUND)
         if val is _NOT_FOUND:
--- a/rocketpy/tools.py
+++ b/rocketpy/tools.py
@@ -1,0 +1,21 @@
+_NOT_FOUND = object()
+
+
+class cached_property:
+    def __init__(self, func):
+        self.func = func
+        self.attrname = None
+        self.__doc__ = func.__doc__
+
+    def __set_name__(self, owner, name):
+        self.attrname = name
+
+    def __get__(self, instance, owner=None):
+        if instance is None:
+            return self
+        cache = instance.__dict__
+        val = cache.get(self.attrname, _NOT_FOUND)
+        if val is _NOT_FOUND:
+            val = self.func(instance)
+            cache[self.attrname] = val
+        return val
--- a/rocketpy/Flight.py
+++ b/rocketpy/Flight.py
@@ -9,7 +9,6 @@ __license__ = "MIT"
 import math
 import time
 import warnings
-from functools import cached_property
 
 import matplotlib.pyplot as plt
 import numpy as np
@@ -17,6 +16,11 @@ import simplekml
 from scipy import integrate
 
 from .Function import Function
+
+try:
+    from functools import cached_property
+except ImportError:
+    from .tools import cached_property
 
 
 class Flight:


Gui-FernandesBR:
Good to read this PR, not that trivial though.

> - The tools.py file should probably be named utils.py and contain all auxiliary decorators, classes and functions that can be used by the rest of the core code but has no interesting use to end-users.

Agreed. I never liked the "utilities.py" name.

> The utilities.py should probably be renamed to tools.py as it contains a series of tools for the end-user.

Not entirely sure about tools.py name, but I get the idea and agree again.


> All Flight methods that have the @cached_property decorator also contain other arguments, such as interpolation="spline", extrapolation="natural". This is broken. Properties cannot have any other argument besides self. Although this does not cause any errors during execution, it is currently impossible to use and change these arguments from their default value. This does not seem right.

My bad, I didn't know that when I implemented.. can we elaborate some options?

giovaniceotto:
> Not entirely sure about tools.py name, but I get the idea and agree again.

I am eager to hear more name suggestions. Or perhaps, other options to organize the material currently in this file.

> My bad, I didn't know that when I implemented.. can we elaborate some options?

I don't believe there are many options to be honest, assuming we want the user to be able to control interpolation and extrapolation. The only option I can think of is kind of a global setting that the user gives the flight class to control interpolation and extrapolation of all Function objects.

However, if we just remove these arguments for now, the user always has the option to change interpolation and extrapolation later. Example:

```python
TestFlight.x.setInterpolation('linear')
```

giovaniceotto:
I have a lot to thank you @Gui-FernandesBR. Inspired by the bug you noticed and the solution you brought forward, I dug a little deeper and found a way to accomplish something I have wanted for a long time but had just recently given up on... a `funcify_method` decorator which converts methods into Function instances.

In fact, it may even be useful as a solution here. Take a look at the working example below:

```python
class Test():
    @funcify_method(inputs=["Time (s)"], outputs=["Time**2 (s**2)"])
    def analytical_func(self):
        """Function which calculates the square of the time."""
        f = lambda t: t**2
        return f
    
    @funcify_method(inputs=["Time (s)"], outputs=["Time**2 (s**2)"], interpolation="polynomial", extrapolation="natural")
    def experimental_func(self):
        """Function which calculates the square of the time."""
        f = [
            [0, 0],
            [1, 1],
            [2, 4],
        ]
        return f

t = Test()
print(t.analytical_func)
```
\>\> Function from R1 to R1 : (Time (s)) → (Time**2 (s**2))
```python
print(t.analytical_func.__doc__)
```
\>\> Function which calculates the square of the time.
```python
print(t.analytical_func(10))
```
\>\> 100
```python
print(t.experimental_func)
```
\>\> Function from R1 to R1 : (Time (s)) → (Time**2 (s**2))
```python
print(t.experimental_func(10))
```
\>\> 100.0
```python
t.analytical_func.plot()
t.experimental_func.plot()
```


89c802d9-2594-4020-9088-3ae24c8008e0

bb2c3c2f-5b4c-45db-983e-5574de2ea6bf

*Note that the `funcify_method` decorator also caches the result.

Gui-FernandesBR:
>  a funcify_method decorator which converts methods into Function instances.

Ok, that's something cool! Could you point some examples of how do you see that being used inside the flight class?

giovaniceotto:
> > a funcify_method decorator which converts methods into Function instances.
> 
> Ok, that's something cool! Could you point some examples of how do you see that being used inside the flight class?

Sure! Here is an example.

Instead of:
```python
@cached_property
def x(self, interpolation="spline", extrapolation="natural"):
    return Function(
        np.array(self.solution)[:, [0, 1]],
        "Time (s)",
        "X (m)",
        interpolation,
        extrapolation,
    )
```

We could have:
```python
@funcify_method("Time (s)", "X (m)", "spline", "natural")
def x(self):
    return np.array(self.solution)[:, [0, 1]]
```

*Docs have been omitted for clarity. They would still be present.

This would apply to almost all, if not all, methods which current return Function objects in the Flight class... and that is a lot.

Gui-FernandesBR:
> I am eager to hear more name suggestions. Or perhaps, other options to organize the material currently in this file.

Here is my suggestion, and it considers other pull requests that are currently going on: 

**Suggestions 1:** 

    rocketpy
    └───
    |    __init__.py
    |    └─── motor
    |    |    ....
    |    └─── utils    // All the auxiliary files
    |    |     _decorators.py  // non-public!
    |    |     units.py
    |    |     function.py
    |    |     geometry.py    // any geometric manipulation
    |    |     coordinates.py    // Haversine, UTM, ref. sistem and simlar
    |    └─── plots
    |    |     environment_plots.py
    |    |     flight_plots.py
    |    |     rocket_plots.py
    |    |    ...
    |    └───design
    |    |    environment_analysis. py
    |    |    flight_analysis.py    // Receive one or more flight objects, create cool stuffs!
    |    environment.py
    |    rocket.py
    |    aerosurfaces.py
    |    parachute.py
    |    flight.py

**Suggestions 2:** 

    rocketpy
    └───
    |    __init__.py
    |    └─── motor
    |    |    ....
    |    └─── utils    // All the auxiliary files
    |    |     _decorators.py  // non-public!
    |    |     units.py
    |    |     function.py
    |    |     geometry.py    // any geometric manipulation
    |    |     coordinates.py    // Haversine, UTM, ref. sistem and simlar
    |    └─── plots
    |    |     environment_plots.py
    |    |     flight_plots.py
    |    |     rocket_plots.py
    |    |    ...
    |    └───environment
    |    |    environment.py
    |    |    environment_analysis. py
    |    └───flight
    |    |    flight.py    
    |    |    flight_analysis.py    // Receive one or more flight objects, create cool stuffs!
    |    └───rocket
    |    |     rocket.py
    |    |    aerosurfaces.py
    |    |    parachute.py
Obs.: Please imagine dispersion.py as well

Gui-FernandesBR:
So, my idea is to keep any name you want in this PR. But right after that, modify all names and pull to `beta/v1.0.0`

Those are all my comments for now.

I'll wait for @funcify decorator commit before reviewing and testing. All in all I believe we should give the property with any standard interpolation name and just ensure we describe in our examples/docs how to modify that if needed.

giovaniceotto:
I like both directory layout ideas @Gui-FernandesBR. I agree they would have to be discussed before v1.0.0, so we need to add this to the milestone.

Regarding the `@funcify` decorator, I prefer to leave it at another PR. The custom `@cached_property` decorator will be needed anyways for several methods and thus needs to be implemented as in this PR.

Please review and merge this PR. Once it has been merged, I will submit another PR implementing the `@funcify` decorator and refactoring many of the Flight class methods.

giovaniceotto:
@Gui-FernandesBR, if you want a more technical review of the decorator, I am sure @phmbressan can help.

Gui-FernandesBR:
@phmbressan could you review the cached_property implementation in this PR? Only 20 code lines and everything seems good to be, however, I am not confortable enough to say that nothing is wrong (the implementation is not exactly identic to functools' core code, therefore I cannot evaluate it so well).

Besides that, the code is running properly in my machine. That would be enough for my approve/review, let's see what @phmbressan says.

@giovaniceotto I think you can continue with you're code using the current code in the `fix/cached_property_py37`. Openning a draft PR may already give us a preview of implementations you're planning at Flight Class.

rocketpy/tools.py
```diff
@@ -0,0 +1,21 @@
...
+        self.attrname = None
+        self.__doc__ = func.__doc__
+
+    def __set_name__(self, owner, name):
+        self.attrname = name
+
+    def __get__(self, instance, owner=None):
+        if instance is None:
+            return self
+        cache = instance.__dict__
```
phmbressan, line 16: I prefer keeping the exception here as in `functools.cached_property` source code, since it prevents the decorator being used on a method that is called _before_ `__set_name__`.

```
if self.attrname is None:
            raise TypeError(     
    "Cannot use cached_property instance without calling __set_name__ on it.")
```

rocketpy/tools.py
```diff
@@ -0,0 +1,21 @@
...
+        self.func = func
+        self.attrname = None
+        self.__doc__ = func.__doc__
+
+    def __set_name__(self, owner, name):
+        self.attrname = name
+
+    def __get__(self, instance, owner=None):
+        if instance is None:
+            return self
```
giovaniceotto, line 15:
```suggestion
        if instance is None:
            return self
        if self.attrname is None:
            raise TypeError(
                "Cannot use cached_property instance without calling __set_name__ on it.")
```

rocketpy/tools.py
giovaniceotto: I have accepted and committed your suggestion @phmbressan. I have also requested another final review.

Out of curiosity, can you share an example of when such a thing can happen? The decorator being used on a method that is called before __set_name__? Would it be like trying to use the decorator in __init__?

phmbressan (approved):
I really like this implementation for it allows a simple fix to support older python versions. Only a few things to keep in mind:

1 - `functools` implementation is naturally more robust since it has checks and exception handling for parallel cache processing. This is not a concern for now, but it may become relevant if multi-threading is used in the future.

2 - As it is in `functools.cached_property` there is a difference in setter behavior when compared to python standard property: the former allows implicit setting whilst the latter raises an error. Hence, when dealing with attributes that must not be changed by the user, manual setter protection have to be employed.

giovaniceotto:
Thanks for the review @phmbressan! Indeed, you raise two great points. Here is my take:

1. I have simplified the `functools` implementation since I really believe we do not need to worry about parallel cache processing for RocketPy right now. When it comes to parallel processing in this case, I believe it makes much more sense to have different simulations running in different threads rather than having a single simulation running on multiple threads.

2. You are right. Currently, we have no protection against anything overriding these cached properties. If this happens inside the codebase, it can be quite a challenge to find such a bug right now. I do not think this is a problem for this PR, but it is something that we need to keep in mind and hopefully fix soon. Do you know any neat way to do that?

Finally, we are ready to merge this PR after your approval!

rocketpy/tools.py
phmbressan: I have thought a bit and I don't think it is quite possible to make a method that is called before `__set_name__` without too cumbersome implementations. I believe that what may trigger the exception is bypassing the `__set_name__` method as shown [here](https://stackoverflow.com/questions/62160411/pythons-new-functools-cached-property-bug-or-limitation).

phmbressan:
@giovaniceotto I have the same understanding regarding parallel processing: it will likely occur for seperate simulations. Thus the caching problem will not arise.

About the second topic, I also do not know any good solution for setter protection. I have to check what is the most common "good-practice" in this case, but it is quite a challege. I have seen something alike using dummy "public" setters while the property itself is "private". However this seems more of a workaround than a solution.

PR #287, opened 2022-11-12
ENH: Refactor and replace plots and outputs + Compare plots

MateusStano:
## Pull request type

Please check the type of change your PR introduces:

- [x] Code base additions (bugfix, features)
- [x] Code maintenance (refactoring, formatting, renaming, tests)
- [ ] ReadMe, Docs and GitHub maintenance
- [ ] Other (please describe):

## What is the current behavior?  What is the new behavior?

Currently, every class has a set of methods that are used for plotting useful information. These methods add up to thousands of lines of code that could be better placed somewhere else.

The changes this PR brings can be summarized in

- create `rocketpy/plots` folder with separate files for the plot code for every class. Each of these files defines a class whose methods displays relevant plots for that class:
   -  `flight_plots.py` that holds code for `class FlightPlots()` 
   -  `environment_plots.py` that holds code for `class EnvironmentPlots()` 
   -  `rocket_plots.py` that holds code for `class RocketPlots()` 

- create `rocketpy/prints` folder with separate files for the code that prints important information for every class. Each of these files defines a class whose methods print relevant information regarding an object of the class:
   -  `flight_prints.py` that holds code for `class FlightPrints()` 
   -  `environment_prints.py` that holds code for `class EnvironmentPrints()` 
   -  `rocket_prints.py` that holds code for `class RocketPrints()` 

- Each class (`Flight`, `Environment` and `Rocket`) will have two new attributes that will objects of their respective `plot` and `print` classes.

- `rocketpy/plots/compare_plots.py` will be created. This file will contain the classes (or class) that will be responsible for comparing plots of different objects. So, if it receives two Flight objects, it will be able to plot eveything in an Flight.allInfo() but with information regarding both Flight objects. This could eventually become a way to compare simulated flights with telemetry data or with simulations from other software

## Workflow

See: https://github.com/RocketPy-Team/RocketPy/pull/287#issuecomment-1371390646

## Does this introduce a breaking change?

- [x] Yes
- [ ] No

## Other information

It is important to note that this PR will not deal with any plots in the Motor classes since they are currently being reworked. The Motor plots will need to be refactored later.

PR #290, opened 2022-11-13
ENH: Moved Flight plots to flight_plots.py

MateusStano:
## Pull request type

## Pull request checklist

---

This PR is related to the `1st PR: moving plots from Flight.py to flight_plots.py` check box in https://github.com/RocketPy-Team/RocketPy/pull/287


rocketpy/plots/flight_plots.py
```diff
@@ -1,8 +1,936 @@
...
+            Instance of the Flight class
+
+        Returns
+        -------
+        None
+        """
+        self.flight = flight
+        return None
+
+    # Start definition of 'basic' plots methods, the traditional RocketPy plots
```
Gui-FernandesBR, line 34:
```suggestion
```

Gui-FernandesBR: I'd remove that comment, it does not make so many sense

rocketpy/plots/flight_plots.py
```diff
@@ -1,8 +1,936 @@
...
+        ----------
+        None
+
+        Return
+        ------
+        None
+        """
+        # Post-process results
+        if self.flight.postProcessed is False:
+            self.flight.postProcess()
```
Gui-FernandesBR, line 49: I believe this is no longer necessary

Gui-FernandesBR: Please notice that this part repeats 9 times in this current file

rocketpy/plots/flight_plots.py
```diff
@@ -1,8 +1,936 @@
...
+
+        # Get index of time before parachute event
+        if len(self.flight.parachuteEvents) > 0:
+            eventTime = (
+                self.flight.parachuteEvents[0][0]
+                + self.flight.parachuteEvents[0][1].lag
+            )
+            eventTimeIndex = np.nonzero(self.flight.x[:, 0] == eventTime)[0][0]
```
Gui-FernandesBR, line 202: This is an unused variable in the code

```suggestion
```

rocketpy/plots/flight_plots.py
```diff
@@ -1,8 +1,936 @@
...
+        else:
+            eventTime = self.flight.tFinal
+            eventTimeIndex = -1
```
Gui-FernandesBR, line 205:
```suggestion
```

Unused

rocketpy/plots/flight_plots.py
```diff
@@ -1,8 +1,936 @@
...
+        if self.flight.postProcessed is False:
+            self.flight.postProcess()
+
+        # Get index of time before parachute event
+        if len(self.flight.parachuteEvents) > 0:
+            eventTime = (
+                self.flight.parachuteEvents[0][0]
+                + self.flight.parachuteEvents[0][1].lag
+            )
+            eventTimeIndex = np.nonzero(self.flight.x[:, 0] == eventTime)[0][0]
```
Gui-FernandesBR, line 273:
```suggestion
```

Unused

rocketpy/plots/flight_plots.py
```diff
@@ -1,8 +1,936 @@
...
+        else:
+            eventTime = self.flight.tFinal
+            eventTimeIndex = -1
```
Gui-FernandesBR, line 276:
```suggestion
```

Unused

MateusStano:
Changed all the things you commented, @Gui-FernandesBR

Gui-FernandesBR:
Hey, @MateusStano ,I just found a little bug here: 

    TestFlight.plots.all()

output: 

image

I am using a standard rocket available at the getting_started notebook. See what I could see when running:

    dir(TestFlight.rocket.parachutes[0])

Output: 

    ['CdS',
     '__class__',
     '__delattr__',
     '__dict__',
     '__dir__',
     '__doc__',
     '__eq__',
     '__format__',
     '__ge__',
     '__getattribute__',
     '__gt__',
     '__hash__',
     '__init__',
     '__init_subclass__',
     '__le__',
     '__lt__',
     '__module__',
     '__ne__',
     '__new__',
     '__reduce__',
     '__reduce_ex__',
     '__repr__',
     '__setattr__',
     '__sizeof__',
     '__str__',
     '__subclasshook__',
     '__weakref__',
     'cleanPressureSignal',
     'lag',
     'name',
     'noiseBias',
     'noiseCorr',
     'noiseDeviation',
     'noiseFunction',
     'noiseSignal',
     'noisyPressureSignal',
     'samplingRate',
     'trigger']

And if I expose their types, we get:

    cleanPressureSignal <class 'list'>
    lag <class 'float'>
    name <class 'str'>
    noiseBias <class 'int'>
    noiseCorr <class 'tuple'>
    noiseDeviation <class 'float'>
    noiseFunction <class 'function'>
    noiseSignal <class 'list'>
    noisyPressureSignal <class 'list'>
    samplingRate <class 'int'>
    trigger <class 'function'>

**What is the problem?**
The `noisyPressureSignal`, `noiseSignal` and `cleanPressureSignal` attributes are lists, but we also need to create the `noiseSignalFunction`, `noisyPressureSignalFunction`, and `cleanPressureSignalFunction` in order to be converted to Function objects, to accomplish with our current documentation [here](https://github.com/RocketPy-Team/RocketPy/blob/master/rocketpy/Parachute.py#L59)

**How to solve it?**
I would suggest we create another PR to accomplish these changes directly into develop. There two files that may/could be modified: Parachutes.py and Flight.py.

Honestly, I do not see enough sustainable reasons to keep both `cleanPressureSignal` and `cleanPressureSignalFunction` in the parachute objects, that's redudant, if we stick with Function I believe it's better. However, everything is used by the Flight class in an certain way, so we need to verify how to implement those corrections properly.

Can you confirm here it is indeed a bug?

Gui-FernandesBR:
Also, and I believe this is my last comment here, during the execution of the `_FlightPlots.all()` method, we can find the "Get index of time before parachute event" approach repeated several times:

image

Maybe, we could get all important index and events inside the `__init__` method, and all the other methods simply call this attribute, for instance, by using `self.first_event_time`. What do you think?

If you think this is not the place to do that, since the PR name is "...Moved...", we can left opened to the next PR.

rocketpy/Flight.py
```diff
@@ -4466,31 +3744,31 @@ def allInfo(self):
         self.info()
```
Gui-FernandesBR, line 3744: Shall we use `self.prints.all()` and `self.plots.all()` instead of invking each method separately?

MateusStano:
> **How to solve it?** I would suggest we create another PR to accomplish these changes directly into develop. There two files that may/could be modified: Parachutes.py and Flight.py.
> 
> Honestly, I do not see enough sustainable reasons to keep both `cleanPressureSignal` and `cleanPressureSignalFunction` in the parachute objects, that's redudant, if we stick with Function I believe it's better. However, everything is used by the Flight class in an certain way, so we need to verify how to implement those corrections properly.
> 
> Can you confirm here it is indeed a bug?

I think the problem is actually something else:

    @cached_property
    def __calculate_pressure_signal(self):
        """Calculate the pressure signal from the pressure sensor.
        It creates a SignalFunction attribute in the parachute object.
        Parachute works as a subclass of Rocket class.

        Returns
        -------
        None
        """
        # Transform parachute sensor feed into functions
        for parachute in self.rocket.parachutes:
            parachute.cleanPressureSignalFunction = Function(
                parachute.cleanPressureSignal,
                "Time (s)",
                "Pressure - Without Noise (Pa)",
                "linear",
            )
            parachute.noisyPressureSignalFunction = Function(
                parachute.noisyPressureSignal,
                "Time (s)",
                "Pressure - With Noise (Pa)",
                "linear",
            )
            parachute.noiseSignalFunction = Function(
                parachute.noiseSignal, "Time (s)", "Pressure Noise (Pa)", "linear"
            )

        return None

This is defined inside the Flight class and doesn't seem to be correct. From my understanding, a `@cached_property`  need to return whatever it is being cached. I don't know how to make this work since what we are doing here is defining attributes of another object, and we access each object inside a for loop.

We could do as you say and use only Function for these attributes,  I believe we would need to change where the `cleanPressureSignal`, `noiseSignal` and `noisyPressureSignal` lists get appended in Flight.py (lines 776 to 784)

    ...
    for parachute in node.parachutes:
        # Calculate and save pressure signal
        pressure = self.env.pressure.getValueOpt(self.ySol[2])
        parachute.cleanPressureSignal.append([node.t, pressure])
        # Calculate and save noise
        noise = parachute.noiseFunction()
        parachute.noiseSignal.append([node.t, noise])
        parachute.noisyPressureSignal.append([node.t, pressure + noise])
        if parachute.trigger(pressure + noise, self.ySol):
    ...

So instead of doing `parachute.cleanPressureSignal.append([node.t, pressure])`, we would do `parachute.cleanPressureSignal.source.append([node.t, pressure])`.

The problem with this is that you can't define a list based Function with an empty list... also, doing this might be slower so I don't  know

MateusStano:
> If you think this is not the place to do that, since the PR name is "...Moved...", we can left opened to the next PR.

I think might as well change it right now! Just commited the changes

rocketpy/Flight.py
MateusStano: `Flight.allInfo()` doesn't show all the plots, I didn't use `self.plots.all()` for that reason

Gui-FernandesBR: > `Flight.allInfo()` doesn't show all the plots, I didn't use `self.plots.all()` for that reason

Mmmmm that's weird... why not? And don't worry, we don't need to get constrained by the current Flight.allInfo() behavior, the breaking change is already going to happen anyway.
I believe `self.plots.all()` would be more easy to understand since it applies the same idea of `self.prints.all()`, do you agree? Otherwise I don't see the point of having a all() method in the _FlightPlots class, makes sense?

Gui-FernandesBR:
Apparently, the `from .tools import cached_property` broke the code in the last commit [46a4397](https://github.com/RocketPy-Team/RocketPy/pull/290/commits/46a4397ead5c6dcc7af37d3db8c235ee8375b670) by Stano.

@MateusStano could you take a look at it please?

Gui-FernandesBR:
> > **How to solve it?** I would suggest we create another PR to accomplish these changes directly into develop. There two files that may/could be modified: Parachutes.py and Flight.py.
> > Honestly, I do not see enough sustainable reasons to keep both `cleanPressureSignal` and `cleanPressureSignalFunction` in the parachute objects, that's redudant, if we stick with Function I believe it's better. However, everything is used by the Flight class in an certain way, so we need to verify how to implement those corrections properly.
> > Can you confirm here it is indeed a bug?
> 
> I think the problem is actually something else:
> 
> ```
> @cached_property
> def __calculate_pressure_signal(self):
>     """Calculate the pressure signal from the pressure sensor.
>     It creates a SignalFunction attribute in the parachute object.
>     Parachute works as a subclass of Rocket class.
> 
>     Returns
>     -------
>     None
>     """
>     # Transform parachute sensor feed into functions
>     for parachute in self.rocket.parachutes:
>         parachute.cleanPressureSignalFunction = Function(
>             parachute.cleanPressureSignal,
>             "Time (s)",
>             "Pressure - Without Noise (Pa)",
>             "linear",
>         )
>         parachute.noisyPressureSignalFunction = Function(
>             parachute.noisyPressureSignal,
>             "Time (s)",
>             "Pressure - With Noise (Pa)",
>             "linear",
>         )
>         parachute.noiseSignalFunction = Function(
>             parachute.noiseSignal, "Time (s)", "Pressure Noise (Pa)", "linear"
>         )
> 
>     return None
> ```
> 
> This is defined inside the Flight class and doesn't seem to be correct. From my understanding, a `@cached_property` need to return whatever it is being cached. I don't know how to make this work since what we are doing here is defining attributes of another object, and we access each object inside a for loop.
> 
> We could do as you say and use only Function for these attributes, I believe we would need to change where the `cleanPressureSignal`, `noiseSignal` and `noisyPressureSignal` lists get appended in Flight.py (lines 776 to 784)
> 
> ```
> ...
> for parachute in node.parachutes:
>     # Calculate and save pressure signal
>     pressure = self.env.pressure.getValueOpt(self.ySol[2])
>     parachute.cleanPressureSignal.append([node.t, pressure])
>     # Calculate and save noise
>     noise = parachute.noiseFunction()
>     parachute.noiseSignal.append([node.t, noise])
>     parachute.noisyPressureSignal.append([node.t, pressure + noise])
>     if parachute.trigger(pressure + noise, self.ySol):
> ...
> ```
> 
> So instead of doing `parachute.cleanPressureSignal.append([node.t, pressure])`, we would do `parachute.cleanPressureSignal.source.append([node.t, pressure])`.
> 
> The problem with this is that you can't define a list based Function with an empty list... also, doing this might be slower so I don't know

This function does not need a @cached_property decorator actually.

I think we should:
1. Remove the decorator from Flight.py
2. When calling the `parachute.cleanPressureSignalFunction` attribute, use try/except. In case the attribute doesn't exist, call the `self.__calculate_pressure_signal()` and try again.

Extra:
3. Probably good idea to initialize the functions at Parachute.py with `self.cleanPressureSignalFunction = Function(0)`, just to accomplish with documentation. In that case the step 2. would stop working correctly, so we should go and always call the `self.__calculate_pressure_signal()` method instead

rocketpy/Flight.py
MateusStano: Agreed, I'll add self.plots.all() to allInfo

Gui-FernandesBR (commented):
#292 presents a cleaner implementation of this commit, by implementing `Flight.outOfRailTimeIndex` and similars

Using a boolean is a digress from the properties utilization, i.e., we would start doing the same idea behind postProcess again. I'd make an effort here to no rely on booleans again.

MateusStano:
Hey! Comming back to this one

Think I fixed everything that was remaining. After rebasing this branch to the changes that are now in develop to use the decorators it was quite easy. I also made the Rocket Pressure at Altitude to always plot instead of how it was before where it was only plotted if there were no parachutes added to the rocket.

Gui-FernandesBR (approved):
@MateusStano thank you for leveraging the quality of these functions. Everything is running as expected locally, both tests and getting started nbks.

The only warning that I got is related to the number of plots opened. But I don't think this should be a concern now, let's handle it later. 
image

The tests are running locally, but idk what is happening with github actions lately.
Btw what is going on with `timezonefinder`? It's on the actions.

Anyway, @MateusStano , feel free to merge it!

MateusStano:
Will merge as soon as #313 is done.

PR #291, opened 2022-11-13
ENH: Function class - Funcify Decorator & setDiscreteBasedOnModel

giovaniceotto:
## Pull request type

## Pull request checklist

  - [x] Tests for the changes have been added (no tests added since the Function class still does not have unit tests)

When working on refactoring some of the methods of the flight class and the liquid motor class, as discussed in #283, I noticed the need for a decorator to convert ordinary methods into `rocketpy.Function` instances. Inspired by `functools.cached_property`, I saw how to make such a decorator a reality.

Furthermore, I also noticed that regularly we needed to change the grid on which a list-based Function was defined. So seized the opportunity to create a better way to do this.

## What is the new behavior?

- `rocketpy.Function.funcify_method` has been created as a decorator factory to allow the conversion of any method into a `rocketpy.Function` instance saved as a cached property.
- `rocketpy.Function.setDiscreteBasedOnModel` has been created to allow the discretization of a Function instance based on another one, effectively copying the grid, interpolation method and extrapolation method from a model Function.

- [ ] Yes
- [x] No

## Other information

This PR needs to be approved before some refactoring of the Flight class methods can be accomplished. Such refactoring depends on the code implemented here to improve readability of several methods, while also decreasing significantly the number of lines of code in the Flight class.


giovaniceotto:
@phmbressan and @MateusStano, could you both review this PR by today?

The additions are quite straight forward, but I am sure that your feedback can help improve the code so that we can merge it and keep going with further PRs.

giovaniceotto:
@phmbressan and @MateusStano, could you both look at this PR by today?

The additions are quite straight forward, but I am sure that your feedback can help improve the code so that we can merge it and keep going with further PRs.

rocketpy/Function.py
```diff
@@ -473,6 +473,92 @@ def setDiscrete(
             self.__interpolation__ = "shepard"
         return self
 
+    def setDiscreteBasedOnModel(self, modelFunction, oneByOne=True):
+        """This method transforms function defined Functions into list
+        defined Functions. It evaluates the function at certain points
+        (sampling range) and stores the results in a list, which is converted
+        into a Function and then returned. The original Function object is
+        replaced by the new one.
```
MateusStano, line 481: I'm a little lost on this method. What is it supposed to do? And where can this be used?

Also, I think the docs here need to be changed, it is exactly the same as the `setDiscrete` method

rocketpy/Function.py
```diff
@@ -473,6 +473,92 @@ def setDiscrete(
...
+    def reset(
+        self,
+        inputs=None,
+        outputs=None,
+        interpolation=None,
+        extrapolation=None,
+    ):
+        """This method resets the Function object to its initial state, with the
+        possibility of resetting all of its initialization arguments besides the
+        source.
```
MateusStano, line 533: Does it reset to the original state or does it just reset to what is inputed when calling the method?

What is the difference between using this instead of `setInputs`,  `setOutputs` and the other `set` methods?

MateusStano (commented):
Just a suggestion here, shouldn't de decorator be called  `funcfy` instead of `funcify`? I think the `i` there is a little weird

MateusStano:
So i got this error:

image


Am I doing something? Never used decorators much before

rocketpy/Function.py
```diff
@@ -2030,3 +2118,101 @@ def differentiate(self, x, dx=1e-6):
...
+
+    Returns
+    -------
+    decorator : function
+        Decorator function to wrap callables as Function objects.
+
+    Examples
+    --------
+    >>> class Test():
+    ...     @Funcify(inputs=['x'], outputs=['y'])
```
phmbressan, line 2142: Docstring still references old naming `Funcify` instead of `funcify_method`.

rocketpy/Function.py
```diff
@@ -2030,3 +2118,101 @@ def differentiate(self, x, dx=1e-6):
...
+    ...     def func(x):
```
phmbressan, line 2143: The implementation seems to raising an error when the method has arguments, as found by @MateusStano and I could reproduce it. Still searching for a fix. For methods that implicitly define a function the decorator worked just fine, _e.g._:

```
class A:
        def __init__(self) -> None:
            pass

        @funcify_method(inputs="t", outputs="t2")
        def square(self):
            return lambda t: t**2

    a = A()

    print(a.square(3))
```

rocketpy/Function.py
```diff
@@ -2030,3 +2118,101 @@ def differentiate(self, x, dx=1e-6):
...
+            self.attrname = name
+
+        def __get__(self, instance, owner=None):
+            if instance is None:
+                return self
+            cache = instance.__dict__
+            try:
+                val = cache[self.attrname]
+            except KeyError:
+                source = self.func(instance)
```
phmbressan, line 2204: Found a not so elegant fix with the following try clause:

```
try:
    source = self.func(instance)
except TypeError:
    source = lambda *_: self.func(instance, *_)
```

giovaniceotto:
> Just a suggestion here, shouldn't de decorator be called `funcfy` instead of `funcify`? I think the `i` there is a little weird

Quick google search shows that `funcify` is more popular than `funcfy`. Furthermore, this name was inspired by [`sympy.utilities.autowrap.ufuncify`](https://docs.sympy.org/latest/modules/utilities/autowrap.html#sympy.utilities.autowrap.ufuncify), while other examples also exist in other languages, such as [funcify.js](https://www.npmjs.com/package/funcify)

rocketpy/Function.py
giovaniceotto: I will update the docstring and add usage examples. Thanks for letting me know it is not at all clear!

rocketpy/Function.py
giovaniceotto: It is identical to using `setInputs`, `setOutputs`, `setInterpolation` and `setExtrapolation`. The benefits of having such a method are two:

- Being able to call a single method to set multiple things. Example: `total_mass.reset('Times (s)', 'Total Mass(kg)')`. This avoids the need for the famous two lines of code setting inputs and outputs throughout RocketPy.
- Making the code used by the decorator `funcify_method` faster and easier to read.

@MateusStano, do you have a suggestion of other names we could use besides `reset` which can cause such confusion? I am struggling to find other options.

rocketpy/Function.py
giovaniceotto: Thanks for noticing! I'll fix this right away.

rocketpy/Function.py
giovaniceotto: Great suggestion! I could not find anything better really, so I restructure the code in a single try except block that handles all possible scenarios. Seems elegant to me.

giovaniceotto: Mark as resolved if you are satisfied.

rocketpy/Function.py
giovaniceotto: I have updated the docs. Mark as resolved if you are satisfied.

giovaniceotto:
Ready for a re-review @phmbressan and @MateusStano. I must say that your review has forced me to significantly improve docs, tests, and implementation. 

I hope the current state satisfies you both. But let me know of any further modification that would be nice to add.

rocketpy/Function.py
```diff
@@ -473,6 +473,160 @@ def setDiscrete(
             self.__interpolation__ = "shepard"
         return self
 
+    def setDiscreteBasedOnModel(self, modelFunction, oneByOne=True):
+        """This method transforms a Function instance defined from callables into a
+        Function instance defined by a list of discrete points.
```
MateusStano, line 478: The method also works with discrete based Functions right?

So it also transforms the mesh of a Function instance defined by discrete points into the mesh of the modelFunction

phmbressan (approved):
Excellent additions, the decorator makes it much more _pythonic_ the conversion of methods into Functions. The most recent changes of docstrings made it much clearer the utility and usage of both methods.

rocketpy/Function.py
```diff
@@ -473,6 +473,160 @@ def setDiscrete(
...
+
+        >>> g.setDiscreteBasedOnModel(f)
+        Function from R1 to R1 : (Scalar) → (Scalar)
+        >>> h = f * g
+        >>> h.source
+        array([[ 0.,  0.],
+               [ 1.,  1.],
+               [ 2.,  8.],
+               [ 3., 27.],
+               [ 4., 64.]])
```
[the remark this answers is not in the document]
giovaniceotto: I do believe there is good use of not converting both Functions to the same domain before algebraic operations are carried out. I you will evaluate the result in just a few points, this is more efficient then using the `setDiscreteBasedOnModel`.

Since the current implementation allow for both options (converting the domain and not converting), I prefer to keep it like this for now.

rocketpy/Function.py
giovaniceotto: Perfect observation, I have adjusted in the last commit!

giovaniceotto:
Seems to me like we are ready to merge. Thanks @MateusStano and @phmbressan for all the improvements!

(I will wait for tests to finish and merge it.)

PR #293, opened 2022-11-13
ENH: adding name as a karg in Flight class

Gui-FernandesBR:
## Pull request type

## Pull request checklist

I cannot easily differentiate one flight object from other.

## What is the new behavior?
New argument and attribute "name" added to Flight class, allowing if to easily label/name your object. It was inspired in the recent AeroSurfaces implementation. This will help a lot on compare plots functions.

- [ ] Yes
- [x] No

## Other information
Could apply to other classes as well, if evrybody thinks

### DIFF
--- a/rocketpy/Flight.py
+++ b/rocketpy/Flight.py
@@ -586,7 +586,7 @@ class Flight:
         verbose : bool, optional
             If true, verbose mode is activated. Default is False.
         name : str, optional
-            Name of the flight. Default is "".
+            Name of the flight. Default is "Flight".
 
         Returns
         -------
--- a/rocketpy/Flight.py
+++ b/rocketpy/Flight.py
@@ -62,6 +62,8 @@ class Flight:
             Original rail length minus the distance measured from nozzle exit
             to the lower rail button. It assumes the nozzle to be aligned with
             the beginning of the rail.
+        Flight.name: str
+            Name of the flight.
 
 
         Numerical Integration settings:
@@ -527,6 +529,7 @@ class Flight:
         atol=6 * [1e-3] + 4 * [1e-6] + 3 * [1e-3],
         timeOvershoot=True,
         verbose=False,
+        name="",
     ):
         """Run a trajectory simulation.
 
@@ -582,6 +585,8 @@ class Flight:
             time in some cases. Default is True.
         verbose : bool, optional
             If true, verbose mode is activated. Default is False.
+        name : str, optional
+            Name of the flight. Default is "".
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


Gui-FernandesBR:
Flight class now receives 'name' as argument, so you can give a name for your flight and I can do `TestFlight.name` without problem.

1. Does it make sense?
2. Is that good approach?
3. Do you know any better approach?

rocketpy/Flight.py
```diff
@@ -527,6 +529,7 @@ def __init__(
         atol=6 * [1e-3] + 4 * [1e-6] + 3 * [1e-3],
         timeOvershoot=True,
         verbose=False,
+        name="",
```
MateusStano, line 532: To make it more standardized with what is in AeroSurfaces I think the default for the `name` parameter should be `name = "Flight"`

MateusStano:
> Flight class now receives 'name' as argument, so you can give a name for your flight and I can do `TestFlight.name` without problem.
> 
> 1. Does it make sense?
> 2. Is that good approach?
> 3. Do you know any better approach?

1. yes
2. i think so
3. no

I just don't like how many things we are adding to develop branch when all additions that aren't bug fixes should be going to beta/v1.0. Every addition to develop is another merge/rebase we have to do in beta/v1.0

rocketpy/Flight.py
Gui-FernandesBR, line 532:
```suggestion
        name="Flight",
```

rocketpy/Flight.py
```diff
@@ -582,6 +585,8 @@ def __init__(
             time in some cases. Default is True.
         verbose : bool, optional
             If true, verbose mode is activated. Default is False.
+        name : str, optional
+            Name of the flight. Default is "".
```
Gui-FernandesBR, line 589:
```suggestion
            Name of the flight. Default is "Flight".
```

Gui-FernandesBR:
> > Flight class now receives 'name' as argument, so you can give a name for your flight and I can do `TestFlight.name` without problem.
> > 
> > 1. Does it make sense?
> > 2. Is that good approach?
> > 3. Do you know any better approach?
> 
> 1. yes
> 2. i think so
> 3. no
> 
> I just don't like how many things we are adding to develop branch when all additions that aren't bug fixes should be going to beta/v1.0. Every addition to develop is another merge/rebase we have to do in beta/v1.0

I've updated the PR then, thank you for reviewing

MateusStano:
Seems good to me then. Feel free to merge

PR #243, opened 2022-09-24
ENH: A more intuitive deployable payload flight example

Gui-FernandesBR:
## Pull request type

- [X] Code base additions (bugfix, features)

## Pull request checklist

- Code base additions (for bug fixes / features):

  - [ ] Tests for the changes have been added
  - [ ] Docs have been reviewed and added / updated if needed
  - [ ] Lint (`black rocketpy`) has passed locally and any fixes were made
  - [ ] All tests (`pytest --runslow`) have passed locally

## What is the current behavior?

Currently we don't have examples of simulations for flying rockets with medium-high complexity. It's really bothers me that we all know how would it be to perform such simulations, but we have never done and shared before. Let's change that scenario.

## What is the new behavior?

Here a list of simulations that we are lacking in terms of example usage:
- [x] Add example simulation of a rocket with Deployable Payload Flight

## Does this introduce a breaking change?
- [ ] Yes
- [x] No

PR #295, opened 2022-11-14
ENH: Introducing CompareFlights class, a comparison of flight plots

Gui-FernandesBR:
## Pull request type

## Pull request checklist

## What is the new behavior?
Comparison of different flight attributes, allowing us to compare different

- [x] Yes - I'm changing the compare classes and creating a compare submodule. @MateusStano let me know if the final implementation makes sense to you!

# Other information:
This relies on #293 , please review that one first. Otherwise we need to insert some code lines to handle the Flight objects names.


Gui-FernandesBR:
@MateusStano ready for review! But unfortunately, I'll have to ask you an extra hand here. Some of the plots are not eorking as expected. Really hard to get the perfect plot for each function, actually. No need to be a perfeccionist here, but a double check to ensure the information is correct is very welcome.

I had to commit the `compare_flights_usage.ipynb` notebook, otherwise it would be really hard for you to test the methods.
If lucky, we can use it as documentation when we get at master.

Please notice that the `CompareFlights` class can be used in the future to compare different simulators, we will just need another script to create flight objects from predetermined results, instead of simulating a flight.

Please let me know if you have any questions. A quick meeting may be helpful as well.

MateusStano:
Very interesting additions! 

A quick look and I found a few bugs:
1- Currently the `compare_flights_usage.ipynb` is using inline plots with `%matplotlib inline`. If instead we use `%matplotlib widget` the plots won't show. This is due to the `plt.close()` call at the end of the plot methods. Removing this line seems to fix this

2- Seem the problem with most of the plots not working has to do with limits in the y axis. Using `set_ylim` will probably be necessary, we just need to find a good to do this for different plots

This was just a quick look, not a propper review. I'm leaving @ompro07 to review this one

MateusStano:
Hey, took the liberty to change some things on this one, since I thought it would be faster than just adding a review. Here is a list of the things I changed:

 - Created `compare.py` that has `class Compare`. The idea here was to set things up for when we create the other compare classes. So all the other compare classes inherit from this one and then use `create_comparison_figure` for most plots.
 - Adapted `create_comparison_figure` to accept different limits to both the horizontal and vertical axis and accept different attributes for both the horizontal and vertical axis. Again, this makes it usable with other future compare classes.
 - Adapted all `compare_flights.py` methods to work with the changes made in `create_comparison_figure`.
 - Added the key word/string `"apogee"` to be used for the x_lim argument. This sets the limit to the maximum apogee_time of all the flights being compared

### Issues

- Found an issue with the plots when using `%matplotlib widget`. The legend and title disappear because they are positioned "outside" of the figure. This seems to be a known bug of jupyter: https://github.com/microsoft/vscode-jupyter/discussions/10470
- Some of the plots can look weird because of the axis limits. This happens especially with ballistic simulations. To counter act this I added arguments for `x_lim` and `y_lim` as stated before. Unfortunately I don't think there is a better way to do this without taking away the ability to see the whole plot

### Remaining changes

I think these are the only remaining changes that still need to be made:

 - [ ] Finish `report` method
 - [ ] Finish `trajectories_2d` method
 - [ ] Tidy up `compare_flights_usage.ipynb`

Gui-FernandesBR:
Well done @MateusStano ! I really think it was good idea using a sup class for comparisons. Here are some comments I have:
* Regarding the `%matplotlib widget` issue, it doesn't concern me at all since the user can still use `%matplotlib inline`. Perhaps in the future we can further investigate that, but no need for now IMO.
* I'm giving up the `report` method in this PR, I think there will be a moment in the future when we you need it more, right now it just seems 'out of context' for this PR
* "Finish trajectories_2d method", done! ✅
*  "Tidy up compare_flights_usage.ipynb", also done! ✅

Issues: 
* Currently, the `rail_buttons_forces()` method is not working. I think that it is due to the array length that rail button forces attributes have. I've tried to debug but could not find an easy solution. **Could you help here too, please? ** If things get too complicate, I suggest just letting this for later. All in all, comparing forces on rail buttons is not so interesting as the other functionalities we are adding. So, I'd agree releases without implementing this `rail_buttons_forces()` and then fix that later.

Perhaps a viable solution could be adding `outOfRailTime`, `maxVelocityTime` attributes, just like you did for `apogee`. What do you think?

Gui-FernandesBR:
Btw this will significantly help us on #243 , therefore I'm mentioning it here

Gui-FernandesBR:
Also, an alternative to the `flight_list` argument that is currently being used is to use *args. 
This was the user would no longer need to define a list with different flight objects.

No need to do this now, but I'll save the link here for later.

https://realpython.com/python-kwargs-and-args/#ordering-arguments-in-a-function

MateusStano:
> * **Could you help here too, please? **

All done! Just had to use `y_attribute` to get the data for the horizontal axis and the vertical axis.

MateusStano (approved):

Gui-FernandesBR:
Tks again, everything LGTM.

Hopefully in the future we discover why the buttonForces issues appeared, and with that we will resolve the following comment in the code:
``` 
# Needs to be manually done beacuse of specific time array length
```

I'm going ahead merging this one since you've approved, and everything is running ok.

PR #301, opened 2022-11-17
ENH: implement retrieval of atmospheric data from Windy API

Gui-FernandesBR:
## Pull request type

## Pull request checklist

## What is the new behavior?
@giovaniceotto is bringing windy API to improve our Environment models!

- [ ] Yes
- [x] No

## Other information
Can we skip the tests and examples additions? We can raise an issue saying that in the future we should describe, in environment calss usage noteook , how to use wind and, also, create unit tests


giovaniceotto:
I do not agree with skipping changes to the Environment class notebook in this case. We already have a lot of issues opened. If we can, let's fix them before they are created.

Gui-FernandesBR:
This is the only one lacking to go to develop before releasing new version. I can work on this by tomorrow. But please consider moving forward without this one if needed

MateusStano:
Hey! I added the windy example to the usage notebook and changed a few things in the process. 

Just to be more in lines of what we were using in the code, I changed the inputs for the `file` argument to be able to be all capitalized. So before it had to be e.g. `ecmwf`, now it can also be `ECMWF`. In the documentation, I left the capitalized form as the standard.

Also, a big problem. The `iconEu` doesn't seem to be working, I couldn't really find the solution. Since we are not using all the atmospheric models available on windy, I think we could just remove `iconEu` if you are okay with it.

giovaniceotto:
Quick note about iconEu, it only works if the latitude and longitude coordinates are within Europe, otherwise, the request returns `Out of model bounds`. Perhaps we can catch this error and notify the user?

Gui-FernandesBR:
> Quick note about `ICON-EU`, it only works if the latitude and longitude coordinates are within Europe, otherwise, the request returns `Out of model bounds`. Perhaps we can catch this error and notify the user?

I think that's the best approach here

MateusStano:
> Quick note about `ICON-EU`, it only works if the latitude and longitude coordinates are within Europe, otherwise, the request returns `Out of model bounds`. Perhaps we can catch this error and notify the user?

Done

Gui-FernandesBR (commented):
This is my review, I think we are ready to merge this one. I cannot approve the PR since I was the author. @MateusStano or @giovaniceotto please one of you approve and they merge, 

tks for helping

rocketpy/Environment.py
```diff
@@ -1428,6 +1438,154 @@ def windSpeed(h):
...
+        if model[-1] == "u":  # case iconEu
+            model = "".join([model[:4], model[4].upper(), model[4 + 1 :]])
+        # Load data from Windy.com: json file
+        url = f"https://node.windy.com/forecast/meteogram/{model}/{self.lat}/{self.lon}/?step=undefined"
+        try:
+            response = requests.get(url).json()
+        except:
+            if model == "iconEu":
+                raise ValueError(
+                    "Could not get a response from Icon-EU by windy. Maybe latitude and longitude coordinates are out of Europe?",
```
giovaniceotto, line 1463:
```suggestion
                raise ValueError(
                    "Could not get a valid response for Icon-EU from Windy. Check if the latitude and longitude coordinates set are inside Europe.",
```

giovaniceotto (approved):

giovaniceotto:
I approved this PR, but I found a single mistake which can easily be solved before merging:

image

It says that it will print the data for ECMWF but actually prints for ICON-EU

MateusStano:
> I approved this PR, but I found a single mistake which can easily be solved before merging:
> 
> image
> 
> It says that it will print the data for ECMWF but actually prints for ICON-EU

Oh, good catch. Already fixed, going to merge now

PR #304, opened 2022-11-19
TST: Fix static margin related tests

MateusStano:
## Pull request type

## What is the current behavior? What is the new behavior?

In tests where we checked static margin values we were calling the `rocket.staticMargin(-1)` and `rocket.centerOfMass(-1)` when trying to get the final values of these objects.
This actually returns the value of that Function for a input value of `-1`. The correct way to get the final values is with `rocket.staticMargin(np.inf)` and `rocket.centerOfMass(np.inf)`

- [ ] Yes
- [x] No


giovaniceotto (approved):
Elegant and simple fix.

giovaniceotto:
I have changed it to develop again because of all the other commits that were appearing. I think we are ready to merge.

PR #313, opened 2022-12-29
BUG: Flight.plotPressureSignals and "StandardAtmosphere" environment pressure calculation

MateusStano:
## Pull request type

## Description

When working on #290 I found that the `Flight.plotPressureSignals` method did not work. This was just because the `__calculate_pressure_signal` method was not being called. Furthermore, when using `"StandardAtmosphere"` the `Pressure at Rocket's Altitude` plot raised an error. This error comes from calling `Environment.pressure` with an array as such: 

image

Using a list instead works fine:

image

I am pretty sure that a Function object should be able to receive an np.array so I think this not working as intended.

One more thing, looking at how the pressure Function is calculated for `StandardAtmosphere`. Lines 2832 and 2833 are incoherent. I don't know much about pressure profiles though, so I might be wrong. Maybe @Gui-FernandesBR or @giovaniceotto know more.

image

## What was changed

The only thing changed currently in this PR is that the plot for `Pressure at Rocket's Altitude` now converts the np.array into a list when getting the pressure values and also calls `__calculate_pressure_signal`, which was not doing before so it just did not work. This fixes the plot but there may be more things wrong here. 

- [ ] Yes
- [x] No

### DIFF
--- a/rocketpy/Flight.py
+++ b/rocketpy/Flight.py
@@ -2470,7 +2470,6 @@ class Flight:
 
         return F11, F12, F21, F22
 
-    @cached_property
     def __calculate_pressure_signal(self):
         """Calculate the pressure signal from the pressure sensor.
         It creates a SignalFunction attribute in the parachute object.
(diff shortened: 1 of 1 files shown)


Gui-FernandesBR (approved):
Nice one! I verified locally and the code is running as expected.

Regarding line 2833, it was not modified since the first addition of Environment.py file, made 3 years ago by @giovaniceotto . It seems wrong to me, at least one of the definition of the `h` variable should be deleted. Please notice this problem also happen in line 2815 of the same file.

@giovaniceotto , what do you think? Can we fix that in another PR?

@MateusStano , I believe you can already merge this one 👌

MateusStano:
Closing this. I opened issue #318 to continue the discussion there.

Issue #314, opened 2022-12-31
Significant Dependency Issue: Timezonefinder does not support Python 3.7


giovaniceotto:
**Describe the bug**

[Timezonefinder](https://github.com/jannikmi/timezonefinder) no longer supports python 3.7. This is a significant issue since RocketPy officially supports Python 3.7 but depends on `timezonefinder`.

**To Reproduce**

Create an environment with Python 3.7.9 on Windows and attempt to install RocketPy. The following error should occur (tested locally and via GitHub Actions):

```shell
Best match: timezonefinder 6.1.9
Processing timezonefinder-6.1.9.tar.gz
error: Couldn't find a setup script in C:\Users\ghceo\AppData\Local\Temp\easy_install-byqibg5q\timezonefinder-6.1.9.tar.gz
```

**Expected behavior**

RocketPy should install without any issues, but it fails since a required library cannot be installed.

**Additional context**

This started to be a problem only recently. Requiring an earlier version of `timezonefinder` (<6.1) would solve the issue. However, this is dangerous for many reasons since we can end up with out of dated time zone delimitations.

giovaniceotto:
Did a little digging and found out that `timezonefinder` dropped support for version 3.7 once Numpy started supporting versions 3.8 and above.

I believe we should do the same for release v.1.0.0: drop support for Python 3.7, support Python 3.8 - Python 3.11.

Gui-FernandesBR:
If numpy dropped 3.7, I agree dropping as well.

On the other hand, why not just dropping timezonefinder then? This is a not so light weight library, and probably there are other options available.

Gui-FernandesBR:
I think @phmbressan may have an opinion as well

giovaniceotto:
> On the other hand, why not just dropping timezonefinder then? This is a not so light weight library, and probably there are other options available.

There is no other viable alternative to `timezonefinder` available, in my opinion. If we drop it, we will drop the feature that automatically detects the timezone based on latitude and longitude.

giovaniceotto:
To add to the discussion of dropping this library, or making it optional, it is currently not implemented in the `Environment` class. It is only used in the `EnvrionmentAnalysis` class. In fact, it is only used briefly in one method: https://github.com/RocketPy-Team/RocketPy/blob/master/rocketpy/EnvironmentAnalysis.py#L346

It can easily be made an optional dependecy!

giovaniceotto:
[...]

PR #315, opened 2022-12-31
ENH: Make timezonefinder an optional dependency

giovaniceotto:
## Pull request type

## Pull request checklist

1. RocketPy fails to install when using Python 3.7 on Windows, since the `timezonefinder` package does not support Python 3.7 anymore. See #314.
2. The `timezonefinder` package is rarely used and not required for any major RocketPy functionality. It is currently a required package inspite of this. And to make things worse, it requires a download larger than 40 MB.

## What is the new behavior?

The `timezonefinder` is made an optional requirement. By default, when installing RocketPy, it will not be installed and thus RocketPy will remain compatible with Python 3.7. The total downloads needed to install RocketPy will be reduced by about 40 MB. If the user desires, he/she can install the optional requirement using:

```shell
pip install rocketpy[timezonefinder]
```

Note, however, that this only works with Python 3.8+.

The only feature that is not supported when `timezonefinder` is not installed is automatic timezone detection based on latitude and longitude when initializing the `EnvironmentAnalysis` class. The user needs to specify the timezone to be used for reporting. If the user does not specify it, the following error will be raised:

```
ImportError: The timezonefinder package is required to automatically determine local timezone based on lat,lon coordinates. Please specify the desired timezone using the `timezone` argument when initializing the EnvironmentAnalysis class or install it with `pip install timezonefinder`.
``` 

- [ ] Yes
- [x] No

## Other information

This PR should solve the failing GitHub Actions that has been haunting us for a couple of months now.

### DIFF
--- a/.github/workflows/test_pytest.yaml
+++ b/.github/workflows/test_pytest.yaml
@@ -22,11 +22,10 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements.txt
         pip install -r requirements_test.txt
     - name: Build RocketPy
       run: |
-        python setup.py install
+        pip install -e .
     - name: Test with pytest
       run: |
         pytest
--- a/docs/user/requirements.rst
+++ b/docs/user/requirements.rst
@@ -7,9 +7,10 @@ But read further if this is not your case!
 Python Version
 --------------
 
-RocketPy was made to run on Python 3.6+.
+RocketPy supports Python 3.7 and above.
+Support for Python 3.11 is still limited by some dependencies.
 Sorry, there are currently no plans to support earlier versions.
-If you really need to run RocketPy on Python 3.5 or earlier, feel free to submit an issue and we will see what we can do!
+If you really need to run RocketPy on Python 3.6 or earlier, feel free to submit an issue and we will see what we can do!
 
 Required Packages
 -----------------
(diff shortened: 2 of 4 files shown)


Gui-FernandesBR (approved):
Everything seems to be working very well. 
Questions: 
* Do you think we should keep timezonefinder in the requirement.txt file?
* In the documentation you're modifying python version (from 3.6 to 3.7). Does this mean that we are officially dropping python 3.6 support for RocketPy? If so, should that be a big deal?

Great fix btw. In terms of coding, there's no change to ask. Only the above questions as curiosity.

MateusStano (approved):

giovaniceotto:
> * Do you think we should keep timezonefinder in the requirement.txt file?

Yes. The requirements.txt file is used for development only, it is not included in the packaged served through PyPI.

> In the documentation you're modifying python version (from 3.6 to 3.7). Does this mean that we are officially dropping python 3.6 support for RocketPy? If so, should that be a big deal?

Support for Python 3.6 was dropped a while ago. Only 3.7+ is supported, that docs page was wrong. Python 3.6 stopped being supported by the Python organization in Dec 2021: https://endoflife.date/python

giovaniceotto:
This PR has been approved and all tests are passing, so I will merge it now.

Quick note: I have modified how RocketPy is installed by the GitHub actions before tests are performed. I switched from `python setup.py install` to `pip install -e .`, basically because pip can handle dependency version conflicts much better, and is also what users make use of.

Issue #281, opened 2022-11-10
MNT: Cleaning the data folder 


Gui-FernandesBR:
**What I propose is: (soft suggestions)**
- [x] Create a "rockets" folder and save "caldene", "calisto", "euporia", "valetudo" folders inside that
- [x] Move "jiboia", "keron", "mandioca" to "motors" folder
- [x] Clean the "weather" folder by removing the majority of files and letting only the crucial ones.
- [x] Do not duplicate files in tests/fixtures. The best place to save data is the data folder!
- [x] Organize motors in different subfolders if needed

Gui-FernandesBR:
This issues requires at least a little discussion before starting applying new changes. @giovaniceotto could we help with some insights here?

giovaniceotto:
Great suggestions @Gui-FernandesBR.

I agree that the data folder may no longer be needed. We can move some of the files used for tests to the `tests` folders as fixtures, and create an sepecific folder for complete example cases, including comparisons with real flight data.

The only point I disagree with is creating a separate repository for examples. While this would make this repository lighter, it would be a nightmare to manage. Imagine having to sync two separate repositories so that the examples can always run with the latest RocketPy version. I do not believe this is worth the effort.

Gui-FernandesBR:
>I agree that the data folder may no longer be needed.

Nooo, did I say that? sorry hahahaha
Data folder is indeed important. Lot of notebooks are using powerOnDrag='data/calisto/....csv' file, for example. The problem is that the folder is currently a mess.

What I think we could do is:
1. Delete the not so useful files in the data folder (#316)
2. Keep only necessary files in the tests/fixtures, not duplicating them when the file already exists in the data folder.
3. Comparisons, examples cases may be used mainly as documentation. Thus we should add them as notebooks and refer in our docs. The currently `tests\fixtures\acceptance\EPFL_Bella_Lui\bella_lui_flight_sim.py`  file, for instance, could be converted into a jupyter notebook, where we can convert to .rst and add to our docs.

aureliobarbosa:
To give a context: I discussed a little bit with your colleagues on PythonBR 2024 and they welcome contributions. I also mentioned that I would be mostly interested in CI, lib infrastructure and code maintenance in general... 

Regarding the data folder, why I think it is a serious problem? 

I started to clone the project on wifi and it took forever on a slow broadband (depending on the context it will make possible contributors to run away). In my home setup it was necessary to plug in a cable to download the project!

[...]

PR #316, opened 2023-01-01
MAINT: Cleaning up some repo files

Gui-FernandesBR:
## Pull request type

- [x] GitHub maintenance

## Pull request checklist

When trying to download (i.e. to clone) the RocketPy repository, usually the user also downloads several files that he/she will probably not use. Some of the files were added in the first upload of this repo and have never been used since then.

## What is the new behavior?

38 data files are being deleted from this repo. I'm saving them in our organizational gdrive account, so we never lost it.
Cloning the repo should be easier now.

- [ ] Yes
- [x] No


Gui-FernandesBR:
This is part 1 of issue #281

Gui-FernandesBR:
Kinda bothers me how the thrust file for the keron motor is repeated 3 times in the repo (1 for tests,1 for data/motors, and 1 for dispersion analysis), even if it is a small file, it is annoying.
Accepting suggestions if they are easy to implement, otherwise let's just move forward.

MateusStano:
> Kinda bothers me how the thrust file for the keron motor is repeated 3 times in the repo (1 for tests,1 for data/motors, and 1 for dispersion analysis), even if it is a small file, it is annoying. Accepting suggestions if they are easy to implement, otherwise let's just move forward.

I think it would be better if every data files were only found inside the data folder and everything has to access to them from there. 

For the tests to work are the data files required to be inside the tests folder? If so then I would suggest that we leave the keron file inside the tests folder, which is better than repeating the file more than once.

If the tests don't require the files to be inside the tests folder then I think is better to create a standard and leave all data-related files inside the data folder and remove them from the fixtures folder

MateusStano (approved):

Gui-FernandesBR:
> For the tests to work are the data files required to be inside the tests folder? 

Not, it is not. Indeed, some of the tests already uses the files available at data folder.
I agree on moving data from tests to data folder. Let's see what @giovaniceotto think about it, since he was the one who created the fixtures.

Also, I think that the data folder should contain only raw data files, I mean, mainly .csv or .json file.
Any code (.py or .ipnb) should be at examples or docs.

giovaniceotto:
Sorry guys... Was this branch merged into master without any tests passing?

Gui-FernandesBR:
> Sorry guys... Was this branch merged into master without any tests passing?
> 
> 

If I understood correctly, the problem with tests were related to the timezonefinder issue.
Codes were running locally.
None of the deleted files were being used by tests of notebooks.
None of the tests or codes were modified.

Nevertheless, I restored the branch if you want to further investigate.

PR #321, opened 2023-01-06
MAINT: refactor flight class __init__ method

Gui-FernandesBR:
## Pull request type

## Pull request checklist

## What is the current behavior vs what is the new behavior?

Four issues are being solved (each one is a different commits):
1 - When using initialSolution, the time array gets a mismatch with other properties. This is a very small issue that can annoy sometimes. Not pulling into develop since it is minor.
2 - Simulation only runs when a set of rail button is added, I'm changing it. In the future we probably will need to better discuss the role of "effectiveRailXX" attributes.
3 - Extracting two initialization methods from the 
4 - Allow Flight objected being passed as arguments to initial solution; 

- [ ] Yes
- [x] No

## Other information
I'm bringing by cherry-pick from #243 PR, which will be refactored.
Credits to @giovaniceotto  for most of the commits.

### DIFF
--- a/rocketpy/Flight.py
+++ b/rocketpy/Flight.py
@@ -609,13 +609,6 @@ class Flight:
         self.timeOvershoot = timeOvershoot
         self.terminateOnApogee = terminateOnApogee
 
-        # Modifying Rail Length for a better out of rail condition
-        upperRButton = max(self.rocket.railButtons[0])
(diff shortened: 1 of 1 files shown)


MateusStano (approved):
Very straight forward. Everything seems good!

PR #323, opened 2023-01-10
REL: Welcome to v0.13.0

Gui-FernandesBR:
## Pull request type

- [x] Other (please describe): Release!

## Pull request checklist

This is a release PR, all changes were already approved previously ;)
Tests are passing, of course, and docs were updated.

The current master branch is outdated. A lot of changes have been made to develop branch and, finally, I believe that we are able to merge everything to master, upgrading our python package version.

## What is the new behavior?
Several changes, really! I will try to summarize them below:
* Environment Analysis 2.0: improved, tested, and examplified.
* Fins sweep angle implemented
* Integration with Windy/ECMWF
* New AeroSurfaces class
* postProcess method is died, say hi to cached properties!
* Timezonefinder is now optional

- [ ] Yes
- [x] No

## Other information
Please wait for merge conflicts to be solved before reviewing!


Gui-FernandesBR:
Conflicts were solved and tests are running,

@giovaniceotto and @MateusStano could you guys review this one pls? 
After merging, we can create a new release so the PyPI package can be updated.

I'll wait for review of both of you before merging, this is a large PR.

rocketpy/Flight.py
```diff
@@ -1800,394 +1819,666 @@ def postProcess(self, interpolation="spline", extrapolation="natural"):
...
 
-        # Converts x and y positions to lat and lon
-        # We are currently considering the earth as a sphere.
-        bearing = []
-        distance = [
-            (self.x[i][1] ** 2 + self.y[i][1] ** 2) ** 0.5 for i in range(len(self.x))
+        return np.column_stack((self.time, longitude))
+
+    @cached_property
+    def retrieve_acceleration_arrays(self):
```
MateusStano, line 2277:
```suggestion
    def retrieve_acceleration_arrays(self):
```

Does this need to be a `@cached_property`?

MateusStano (changes requested):

data/dispersion csv files/Valetudo.csv
```diff
@@ -0,0 +1,31 @@
+﻿attribute_class; parameter_name; mean_value; standard_deviation;
```
giovaniceotto, line 1: Is this file crucial for this release? If this is not the case, consider deleting it to prevent accumulating unnecessary files.

docs/user/requirements.rst
```diff
@@ -49,6 +50,7 @@ The packages needed can be installed via ``pip`` by running the following lines
     pip install pytz
     pip install timezonefinder
```
giovaniceotto, line 51:
```suggestion
```

data/dispersion csv files/Valetudo.csv
giovaniceotto: Furthermore, the standard for CSV files is to be comma delimited. So every ";" should be replaced by ",".

tests/fixtures/dispersion/Valetudo_inputs.csv
```diff
@@ -0,0 +1,31 @@
+﻿attribute_class; parameter_name; mean_value; standard_deviation;
```
giovaniceotto, line 1: Consider using "," instead of ";", as it is the standard for CSV files.

rocketpy/utilities.py
```diff
@@ -197,3 +196,46 @@ def du(z, u):
         velocityFunction()
 
     return altitudeFunction, velocityFunction, final_sol
+
+
+def create_dispersion_dictionary(filename):
+    """Creates a dictionary with the rocket data provided by a .csv file.
+    File should be organized in four columns: attribute_class, parameter_name,
+    mean_value, standard_deviation. The first row should be the header.
+    It is advised to use ";" as separator, but "," should work on most of cases.
```
giovaniceotto, line 205: Why is this the case?

rocketpy/utilities.py
```diff
@@ -197,3 +196,46 @@ def du(z, u):
...
+    dictionary
+        Dictionary with all rocket data to be used in dispersion analysis.
+    """
+    try:
+        file = np.genfromtxt(
+            filename, usecols=(1, 2, 3), skip_header=1, delimiter=";", dtype=str
+        )
+    except:
+        print(
+            "Error: The delimiter should be ';'. Using ',' instead, be aware that some resources might not work as expected. Please consider changing the delimiter to ';'."
```
giovaniceotto, line 223: Which resources do not work?

rocketpy/utilities.py
```diff
@@ -197,3 +196,46 @@ def du(z, u):
...
+
+    Parameters
+    ----------
+    filename : string
+        String with the path to the .csv file.
+
+    Returns
+    -------
+    dictionary
+        Dictionary with all rocket data to be used in dispersion analysis.
```
giovaniceotto, line 215: Please describe the "format" of this dictionary. Consider adding examples of when this utility can be useful in the docstring.

giovaniceotto (changes requested):

giovaniceotto:
@Gui-FernandesBR, here is my first review. As this is quite a large PR, I decided to review half of the files today and will review the other half tomorrow, although I have already looked through all the files. Let me know if you have any questions about my comments.

rocketpy/Flight.py
Gui-FernandesBR: The retrieve_acceleration_arrays property is used 6 times at the Flight.py file.
If we don't use it as a cached_property, what you happen is that the functions inside the `def retrieve_acceleration_arrays()` will be called six times, which can significantly impact performance.

I can imagine that better names to this function can be given, if you present some suggestions we can discuss.

rocketpy/Flight.py
```diff
@@ -1800,394 +1819,666 @@ def postProcess(self, interpolation="spline", extrapolation="natural"):
...
+                    ay.append([step[0], ay_value])
+                    az.append([step[0], az_value])
+                    alpha1.append([step[0], alpha1_value])
+                    alpha2.append([step[0], alpha2_value])
+                    alpha3.append([step[0], alpha3_value])
+
+        return ax, ay, az, alpha1, alpha2, alpha3
+
+    @cached_property
+    def retrieve_temporary_values_arrays(self):
```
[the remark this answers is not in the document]
Gui-FernandesBR: Same as above

rocketpy/Flight.py
```diff
@@ -1800,394 +1819,666 @@ def postProcess(self, interpolation="spline", extrapolation="natural"):
...
+        return temporary_values
 
-        lat1 = (
-            3.14159265359 * self.env.lat / 180
-        )  # Launch lat point converted to radians
-        lon1 = (
-            3.14159265359 * self.env.lon / 180
-        )  # Launch long point converted to radians
+    @cached_property
+    def calculate_rail_button_forces(self):
```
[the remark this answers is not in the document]
Gui-FernandesBR: Same as above

data/dispersion csv files/Valetudo.csv
Gui-FernandesBR: We need this file to run the `test_create_dispersion_dictionary` function. It is not a duplicated file, neither a large one, so I would say it is not a big deal having it here.

Regarding the semicolon problem, well..., it was a workaround since there're some items that present commas, as you can see at the ensembleMember, which contains a list.
I think the best way would be having these information as .json files. However, it is not so human-readable as the csv one, so I don't know. I rather take few hours to think what is the best solution here.

rocketpy/Flight.py
MateusStano: Oh okay. This makes sense then, no problems here

data/dispersion csv files/Valetudo.csv
giovaniceotto: YAML files would be the best option in my opinion. I'll raise an issue regarding this, but the implementation works, so I do not believe we should hold back this PR.

However, I do believe the docstring should give a better explanation of why you should use ";".

giovaniceotto:
I'll need one more day to run a couple of checks before we are ready to go. Almost there!

data/dispersion csv files/Valetudo.csv
Gui-FernandesBR: The idea of using csv files was to have an easy interface, allowing the user to write different csv files from excel, for example, and running a batch of simulations with them. YAML does not allow that.

Anyway, I agree we should not spend too many time with this method. It was created mainly to solve a specific problem last year, which was already addressed,

giovaniceotto:
Here is a problem with the documentation, which can be checked here: https://docs.rocketpy.org/en/develop/user/index.html

It seems the Environment Analysis notebook is using H1 headers instead of H2 for Surface Level Analysis, Pressure Level Analysis and Going Further with your Analysis, which causes the main index of the user guide page to get cluttered.

Furthermore, the utilities model notebook is out of date regarding the new function `create_dispersion_dictionary`.

When these issues (and the docstring for `create_dispersion_dictionary` function) are fixed, I will accept the changes and this PR will be able to be merged.

Gui-FernandesBR:
Ok, now tests are passing, and the documentation seems to be good. I also added two notebooks to index.rst file, what I cannot is saying why they weren't included before.

> Furthermore, the utilities model notebook is out of date regarding the new function create_dispersion_dictionary

Hey, I don't think it should be mandatory documenting all the new utilities functions on notebooks. Other classes don't even have a notebook.... Therefore, I vote for skipping this one. Maybe in the future we can provide that. The current utilities module notebook seems good to me.

Everything else was addressed, at least temporarily.

Btw sorry for long commit list, it turns out some tests were passing locally without running so I committed some poor lines thinking that it was running well.

giovaniceotto:
@Gui-FernandesBR, I have fixed the documentation problem related to headings 1 being used mistakenly. I have also defined which exception to catch in the `create_dispersion_dictionary` function and replaced prints with warnings. Please review this changes.

Once done, I believe we are ready to merge. Quick idea: wait until the weekend (just one day) so that we can host a launch party in our discord channel with the new members so that they can see how a release is done (and so that we can promote this new version together)!

giovaniceotto (approved):

Gui-FernandesBR:
I got an idea: use quotes to separate objects in the csv file!!

MateusStano:
We are releasing this today!

PR #326, opened 2023-01-25
MAINT: move last plots from Environment class

Gui-FernandesBR:
## Pull request type

## Pull request checklist

## What is the new behavior?

The very same behavior, the very same results, but now there is no `plt.plot(...` inside the Enrivonment class

- [x] No

## Other information

This PR could be seen as an extension of #296

### DIFF
--- a/rocketpy/AeroSurfaces.py
+++ b/rocketpy/AeroSurfaces.py
@@ -1464,7 +1464,6 @@ class Tail:
         return None
 
     def aerodynamicInfo(self):
-
         print(f"\nTail name: {self.name}")
         print(f"Tail Center of Pressure Position in Local Coordinates: {self.cp} m")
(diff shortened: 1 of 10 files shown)


Gui-FernandesBR:
Ready for review ✅👌

MateusStano (changes requested):
Everything seems good! Just one thing:

image

The lines are above the legend. Maybe this can be fixed using the `zorder` param of matplotlib?

MateusStano:
> What I think is weird is that only the speed of sound have label/lengend. Any reason to that?

Yeah, that is weird. Couldn't figure out how to fix it but found two things:

1. The name of the axes labels are wrong
2. And more importantly, those plots don't need legend because the different axes are already displayed in different colors !!

I think just removing the legend here will work

Gui-FernandesBR:
Ready for a re-review @MateusStano

MateusStano (approved):

PR #330, opened 2023-01-29
Added Piecewise class and other new Function methods

gautamsaiy:
## Pull request type

## Pull request checklist

Enter text here...

## What is the new behavior?

Added a Piecewise Function to to Function.py. (It takes in a dictionary where the keys are the domain and the values are the Functions and returns a Function object). There are also updates to Function.differentiate. The methods derivativeFunction (returns a Function that is the derivative of the original), integralFunction (returns a Function that is the integral of the original), invert (returns a function whose domain is now the range and the range is now the domain), findOptimalValue (finds the best value in the domain corresponding to a value from the range), and compose (returns a function that is f(g(x))) are also added.

- [ ] Yes
- [x] No

## Other information

Enter text here...

### DIFF
--- a/rocketpy/Function.py
+++ b/rocketpy/Function.py
@@ -2129,32 +2129,23 @@ class Function:
         result : Function
             A Function whose domain and range have been inverted.
         """
-        # Check if lower and upper are given
-        if lower is None:
-            if isinstance(self.source, np.ndarray):
-                lower = self.source[0, 0]
-            else:
-                raise ValueError("Lower limit must be given if source is a function.")
-        if upper is None:
-            if isinstance(self.source, np.ndarray):
-                upper = self.source[-1, 0]
-            else:
-                raise ValueError("Upper limit must be given if source is a function.")
-
-        # Create a new Function object
-        xData = np.linspace(lower, upper, datapoints)
-        yData = np.zeros(datapoints)
-        for i in range(datapoints):
-            yData[i] = self.getValue(xData[i])
-
-        try:
+        if isinstance(self.source, np.ndarray):
+            # Swap the columns
+            source = np.concatenate(([self.source[:, 1]], [self.source[:, 0]])).transpose()
+            
             return Function(
-                np.concatenate(([yData], [xData])).transpose(),
+                source,
                 inputs=self.__outputs__,
                 outputs=self.__inputs__,
+                interpolation=self.__interpolation__
             )
-        except ValueError:
-            raise ValueError("Check if given domain of the function is invertible.")
+        else:
+            return Function(
+                lambda x: self.findOptimalInput(x),
+                inputs=self.__outputs__,
+                outputs=self.__inputs__,
+                interpolation=self.__interpolation__
+            )
 
     def findOptimalInput(self, val):
         """
--- a/rocketpy/Function.py
+++ b/rocketpy/Function.py
@@ -2154,8 +2154,6 @@ class Function:
             )
         except ValueError:
             raise ValueError("Check if given domain of the function is invertible.")
-                    inputs=self.__outputs__,
-                    outputs=self.__inputs__)
 
     def findOptimalInput(self, val):
         """
--- a/rocketpy/Function.py
+++ b/rocketpy/Function.py
@@ -2122,7 +2122,7 @@ class Function:
         lower : float
             Lower limit of the new domain. Only required if the Function's source is a callable instead of a list of points.
         upper : float
-            Upper limit of the new domain.
+            Upper limit of the new domain. Only required if the Function's source is a callable instead of a list of points.
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


rocketpy/Function.py
```diff
@@ -2024,9 +2024,262 @@ def integral(self, a, b, numerical=False):
...
+        Finds the optimal input for a given output.
+
+        Parameters
+        ----------
+        val : float
+            The value of the output.
+
+        Returns
+        -------
+        result : float
```
phmbressan, line 2164: The result of [`scipy.optimize.fmin`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.fmin.html) is a `ndarray` (even if it has only one element). This does not cause problems since `ndarray` supports most mathematical operations that a `float` does.

```suggestion
        result : ndarray
```

phmbressan (approved):
Great implementation overall! I had a lot of fun toying around and testing the new features. I have suggested some small changes in the review comments regarding docstrings and one `except` block.

rocketpy/Function.py
```diff
@@ -2024,9 +2024,267 @@ def integral(self, a, b, numerical=False):
...
+        yData = np.zeros(datapoints)
+        for i in range(datapoints):
+            yData[i] = self.integral(lower, xData[i])
+        return Function(np.concatenate(([xData], [yData])).transpose(), 
+                    inputs=self.__inputs__, 
+                    outputs=[o + " Integral" for o in self.__outputs__])
+
+    def invert(self, lower=None, upper=None, datapoints=100):
+        """
+        Inverts the Function. Makes the domain the range and the range the domain.
```
giovaniceotto, line 2117: To follow the pattern stablished by `derivativeFunction` and `integralFunction`, consider renaming this method to `inverseFunction`. This, together with a rewording of the docstring makes it clear that the method generates a new Function, instead of performing the operation in place.

```suggestion
    def inverseFunction(self, lower=None, upper=None, datapoints=100):
        """
        Returns the inverse of the Function. The inverse function of F is a function that undoes the operation of F. The 
        inverse of F exists if and only if F is bijective. Makes the domain the range and the range the domain.
```

rocketpy/Function.py
```diff
@@ -2024,9 +2024,267 @@ def integral(self, a, b, numerical=False):
...
+
+        Parameters
+        ----------
+        lower : float
+            Lower limit of the new domain.
```
giovaniceotto, line 2122:
```suggestion
            Lower limit of the new domain. Only required if the Function's source is a callable instead of a list of points.
```

rocketpy/Function.py
```diff
@@ -2024,9 +2024,267 @@ def integral(self, a, b, numerical=False):
...
+        upper : float
+            Upper limit of the new domain.
```
giovaniceotto, line 2124:
```suggestion
            Upper limit of the new domain. Only required if the Function's source is a callable instead of a list of points.
```

rocketpy/Function.py
```diff
@@ -2024,9 +2024,267 @@ def integral(self, a, b, numerical=False):
...
+        # Create a new Function object
+        xData = np.linspace(lower, upper, datapoints)
+        yData = np.zeros(datapoints)
+        for i in range(datapoints):
+            yData[i] = self.integral(lower, xData[i])
+        return Function(np.concatenate(([xData], [yData])).transpose(), 
+                    inputs=self.__inputs__, 
+                    outputs=[o + " Integral" for o in self.__outputs__])
+
+    def invert(self, lower=None, upper=None, datapoints=100):
```
giovaniceotto, line 2115: I believe this docstring needs to be a bit clearer in terms of what happens with callable defined Functions. From what I understand, it sets it as a discretely defined Function before inverting. What if we wrapped `findOptimalInput` in this case, instead?

I know this is much more costly, but it is more rigorous. If the user wants a faster, less accurate version, he/she can use `Function.setDiscrete` before calling `Function.invert`. 

@gautamsaiy and @phmbressan, what do you think?

rocketpy/Function.py
```diff
@@ -2024,9 +2024,267 @@ def integral(self, a, b, numerical=False):
...
+        yData = np.zeros(datapoints)
+        for i in range(datapoints):
+            yData[i] = self.getValue(xData[i])
+
+        try:
+            return Function(
+                np.concatenate(([yData], [xData])).transpose(),
+                inputs=self.__outputs__,
+                outputs=self.__inputs__,
+            )
```
giovaniceotto, line 2154: Is leaving interpolation and extrapolation arguments empty the best option we have here?

giovaniceotto (changes requested):
@gautamsaiy, these new additions to the Function class are amazing! Congratulations for such a nice idea and an equally well written implementation.

I have raised some small points of discussion for you to think about before we merge this PR.

rocketpy/Function.py
gautamsaiy: I think this a great idea. I will make sure to add it.

giovaniceotto (approved):

Issue #331, opened 2023-02-01
Spurious results when integrating linear Functions


phmbressan:
**Describe the bug**

The method `integral` for `Function` objects produces inconsistent results when interpolation is linear and the numerical parameter is the default False.

**To Reproduce**

1 - Import rocketpy `Function` and create one whose interpolation is linear;
2 - Call the method `integrate` with the parameter `numerical` set to `False`.

The result is incorrect, differing in general from the same method with the numerical parameter set to `True`. The code below illustrates this:

```
>>> from rocketpy import Function
>>> f = Function([(0,0), (1,2), (10,20)], interpolation="linear", extrapolation="natural") # f(x) = 2x
>>> f.integral(0,5) # incorrect result
100.0
>>> f.integral(0,5,True) # correct result
25.0
```

**Expected behavior**

The method ought to return the correct value of the definite integral, taking into account the specified domain. 

```
# expected behavior
>>> f.integral(0,5)
25.0
>>> f.integral(0,5,True)
25.0
```

**Additional context**

Useful links: 
- NumPy `trapz` [documentation](https://numpy.org/doc/stable/reference/generated/numpy.trapz.html) 
- SciPy `quad` [documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.quad.html)

PR #332, opened 2023-02-02
FIX: Implementation of analytical integral of linearly interpolated Functions

giovaniceotto:
## Pull request type

## Pull request checklist

See #331.

## What is the new behavior?

Issue #331 is fixed by implementing a correct analytical integral of linearly interpolated Functions.

- [ ] Yes
- [x] No

## Other information

Refactoring the Flight class is long overdue... I am not comfortable with the size (number of lines of code) of the integral method, but creating "sub" functions goes against the current organization of the Function class.

### DIFF
--- a/rocketpy/Function.py
+++ b/rocketpy/Function.py
@@ -2165,8 +2165,12 @@ class Function:
         ans : float
             Evaluated integral.
         """
+        # Guarantee a < b
+        integrationSign = np.sign(b - a)
+        if integrationSign == -1:
+            a, b = b, a
+        # Different implementations depending on interpolation
         if self.__interpolation__ == "spline" and numerical is False:
-            # Integrate using spline coefficients
             xData = self.source[:, 0]
             yData = self.source[:, 1]
             coeffs = self.__splineCoefficients__
@@ -2241,7 +2245,7 @@ class Function:
                 if b <= yData[-1]:
                     xIntegrationData = np.concatenate((xIntegrationData, [b]))
                     yIntegrationData = np.concatenate((yIntegrationData, [self(b)]))
-                else:
+            else:
                 xIntegrationData = np.concatenate(([a], xIntegrationData))
                 yIntegrationData = np.concatenate(([self(a)], yIntegrationData))
                 xIntegrationData = np.concatenate((xIntegrationData, [b]))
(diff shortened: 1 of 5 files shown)


giovaniceotto:
@phmbressan, please ignore the commit by Lint-Action

rocketpy/Function.py
```diff
@@ -2227,11 +2231,31 @@ def integral(self, a, b, numerical=False):
...
         elif self.__interpolation__ == "linear" and numerical is False:
-            return np.trapz(self.source[:, 1], x=self.source[:, 0])
+            # Integrate from a to b using np.trapz
+            xData = self.source[:, 0]
+            yData = self.source[:, 1]
+            # Get data in interval
+            xIntegrationData = xData[(xData >= a) & (xData <= b)]
+            yIntegrationData = yData[(xData >= a) & (xData <= b)]
+            # Add integration limits to data
+            if self.__extrapolation__ == "zero":
```
phmbressan, line 2241: The idea of adding the integration limits to the data was interesting for consistency, however it introduces new spurious results when integrating with `np.trapz`. See below:

```
>>> from rocketpy import Function
>>> f = Function([(0,0), (1,2), (10,20)], interpolation="linear", extrapolation="zero")
>>> f.integral(0,11) # "exact" integral
110.0  # should be 100.0
>>> f.integral(0,11,True)
100.00083463213139
```

The problem is that the integral limit in the zone of zero extrapolation is causing the trapezoid rule to interpret one more region as a trapezoid whose lesser base is 0 (i.e. a triangle).

I do not have a nice fix for the problem, however, if the integration limit is not added to the dataset, the function gives the correct value.

rocketpy/Function.py
```diff
@@ -2227,11 +2231,31 @@ def integral(self, a, b, numerical=False):
...
+                if a >= xData[0]:
+                    xIntegrationData = np.concatenate(([a], xIntegrationData))
+                    yIntegrationData = np.concatenate(([self(a)], yIntegrationData))
+                if b <= yData[-1]:
```
giovaniceotto, line 2245:
```suggestion
                if b <= xData[-1]:
```

rocketpy/Function.py
giovaniceotto: A typo in line 2245, where `yData` should be `xData` caused the problem. I have fixed it in the latest commit. Therefore, I am requesting your review once again @phmbressan

phmbressan (approved):
Excellent. The method now seems to be correct.

PR #336, opened 2023-03-01
ENH: using pydantic with the dispersion class

MateusStano:
## Pull request type

- [x] Yes


Gui-FernandesBR:
@MateusStano hello!

I believe it is time for a new meeting. Improvements from your side were crucial and I think we are now closer to a better usability of the class.

However, the Dispersion class is still not working, and several TODOs still remain.
Just to summarize my last commits: 
- Moved files to monte_carlo submodule, which I think is more intuitive than ClassData names
- I modified the structure of self.dispersion_dictionary, as well as the yield and the run_dispersion.
- In the notebook you're gonna see what is my proposal of the final version of the class. I hope it doesn't get too hard to understand! 

Btw I've notice that few changes that I've made before were lost here. Idk if it was intentional or not, but just in case, take care because you could have excluded contributions from both me and you.  Let me give u an example:
image


Warm regards,
Guilherme

Gui-FernandesBR:
@MateusStano just wanted to add smth more: we are talking about v1.0.0 here, so please let's use only snake_case from now on.

Gui-FernandesBR:
Well done, @MateusStano 


Is it ready for a re-review?

MateusStano:
Ok... so here we go...

## Changes Overview

Here are the changes explanations of the most relevant commits:

- [d57e871](https://github.com/RocketPy-Team/RocketPy/pull/336/commits/d57e871c8b9320e047bd64c067957e0b53a1003d) and [3087fe5](https://github.com/RocketPy-Team/RocketPy/pull/336/commits/3087fe542613b0d32c8153a65eb9d4302a3e3d57) -> Made `position` of the aerodynamic surfaces be saved both as a tuple in the `rocket.aeroSurfaces` list and as an attribute for the `AeroSurfaces`. Also added it to the DataClasses;
- [3f69073](https://github.com/RocketPy-Team/RocketPy/pull/336/commits/3f69073545d3cdc5c4e180f6a50c43c9503c1e56) -> Created the DispersionModel to be a super class for all the pydantic data classes. This eliminated all the repeated code that existed between the DataClasses;
- [76bb57d](https://github.com/RocketPy-Team/RocketPy/pull/336/commits/76bb57da9e3a1ef20786be920738371b47802317) and [5556a22](https://github.com/RocketPy-Team/RocketPy/pull/336/commits/5556a220e289856558bfc53786ea2eaa3d450581) -> Made ensemble members variable, created a validator for them. Added windXFactor and windYFactor;
- [5787508](https://github.com/RocketPy-Team/RocketPy/pull/336/commits/5787508699ca2f64f44e9543604ff91a5f6631c3) -> This made the updating prints that show the progress of the run_dispersion simpler and now it works both on notebooks and on scripts;
- [ef01444](https://github.com/RocketPy-Team/RocketPy/pull/336/commits/ef0144476f185e502f487b11ebc6a7dee6f7304e) -> This is a big one, I'm going to open a PR just for this one because we will need a review from a certain wizard here. The Flight class had a problem that whenever the parachute lag was equal to zero, a warning would be raised. This was due to some problems in the time nodes and time phases;
- [d442046](https://github.com/RocketPy-Team/RocketPy/pull/336/commits/d442046b4a0b9212e8b5e8de8c77911270f530f9) and [f5d3c59](https://github.com/RocketPy-Team/RocketPy/pull/336/commits/f5d3c596ae31b9a0ec9ea17a6986e7f3ac0d6649) -> Everything about the file saving during run_dispersion and file importing has been changed. Most changes were done to optimize the simulations. Some cached properties were created so that they are only calculated when called, instead of done so during the simulation.

---

## Remaining tasks

I am just going through all the `TODOs` in the code and explaining them.

- [ ] Find a way to deal with Rail Buttons. This needs to be well thought out so we make an implementation that is robust enough to any future changes in the rail buttons.

- [ ] The `get_distribution` function is currently on the tools.py. I like it there but just making sure, this is correct, right?

- [ ] I think it would be valuable to verify if the `distribuition_function` input in the DataClasses (so when a string is inputted) is the name of a valid numpy.random function. This should be done inside of the DataClasses to make the error be raised right when the user is inputting them.

- [ ] Need to check if the other distribution functions work. I guarantee that some of them do not work with just a nominal value and a standard deviation.

- [ ] Create a pretty `__repr__` for all the DataClasses.

- [ ] There is a problem with numpy's distribution function imports. We are currently using `from numpy.random import *` and this causes an issue with `random import choice`. What happens is that `numpy.random` has a `choice` function and this function does not work for us, we need to use `random.choice`. So the `random` import needs to be lower than the numpy's import.

- [ ] Currently, files are all being saved like we always did. Each dictionary is a line of the `.txt`. I wonder if this should be a `.csv`, so the files are easier to read by anyone. If not, we should at least create a method to export the results in a `.csv`.

- [ ] Function objects and the parachute trigger functions are being handled in a not optimal way. Currently, some `if`s are added to handle these special cases. We need to do this because we can not easily save these types of variables to the `.txt`. So in the inputs file, the trigger function is saved as `'trigger': <function drogueTrigger at 0x000001FC26EE9B40>`, and you can not recreate the trigger function from this (which may not necessarily be a problem). Also, the trigger function problem is breaking the `inputs_log` attribute. For the Function objects cases, what is done is that whenever a Function object needs to be saved, the list of the source needs to be gotten, e.g. list(Function.source). The list conversion is necessary to make sure there is no line skips in the string when saved. This way of saving Functions can be bad because it is slow, and is run during the simulation. The only alternative I see for this issue is to use `pickle`.

- [ ] During the simulation a `Dispersion.flight_list` is created. This is a list that saves all the simulated flights and is not used anywhere else in our code, it is just something that is nice to have access to. The only issue here is that this list gets appended inside the `run_dispersion` loop, which might be inefficient. This needs to be tested. The other solution for this would be to make `flight_list` a cached_property and have it recreate the flights from the inputs file. For this to happen, the problem with saving the parachute triggers needs to be fixed.

- [ ] Create evolution plots to analyze the convergence.

- [ ] Maybe when a Dispersion object is created it should try to instantly import a file with the `filename` given. If it cannot import it should just keep going as is.

- [ ] Initialize variables so they can be accessed by MATLAB.

- [ ] Currently, attributes are created and saved in the Dispersion object from the keys in the `dispersion_results` dictionary. I don't think this is necessary anymore. What do you think?

- [ ] There is a `try: except:` to check something about the sweepAngle. This should be removed.

- [ ] Verify what happens if we use Environment instead of McEnvironment.

- [ ] Add hybrid and liquid motor options whenever the Liquid Motors are done.

- [ ] There is an `if else` when passing the `lag` value for the parachute inside of `run_dispersion`. It checks if the lag is bigger than zero since if it is smaller than zero it breaks the Flight simulation. This is not ideal and we should try to find another way to fix this, but there might not be any other option.

- [ ] When an `export_list` is not provided, a message is printed: `"No export list provided, using default list instead."`. This is a little annoying. Is it really necessary?

- [ ] Review and improve everything related to the plots (especially the Ellipses stuff). I didn't do anything in that part of the code.

- [ ] Need to finish all of the documentation. Also need to find a way to document the DataClasses. There is a lot here:
   - [ ] Dispersion.py
   - [ ] DispersionModel.py
   - [ ] mc_aero_surfaces.py
   - [ ] mc_environment.py
   - [ ] mc_flight.py
   - [ ] mc_parachute.py
   - [ ] mc_rocket.py
   - [ ] mc_solid_motor.py

- [ ] Tests


@Gui-FernandesBR a lot of TODOs but not most of them are not that time-consuming. I'll need your help here :). Also, please answer the topics above that I left as questions, they were actually directed at you hehe

Gui-FernandesBR:
OK I'm comenting all the remaining tasks then, let me know if you have any questions.

1 - Find a way to deal with Rail Buttons. This needs to be well thought out so we make an implementation that is robust enough to any future changes in the rail buttons.
- Should we create a simple class to rail buttons? We had discussed that a while ago, and it seems to me that the rail buttons don't follow the standard in the rocket class. If you're ok with this we can open a PR direct to the `beta/v1.0.0` branch and ask for reviews.

2 - The get_distribution function is currently on the tools.py. I like it there but just making sure, this is correct, right?
- It was intentional, and I think we should keep it right there. If you see any problems or a sign of bad practice please describe here, otherwise just let's go.

3 - I think it would be valuable to verify if the distribuition_function input in the DataClasses (so when a string is inputted) is the name of a valid numpy.random function. This should be done inside of the DataClasses to make the error be raised right when the user is inputting them.
- Idk how to do it with the new pydantic implementation. In that case you'd be quicker than me. Do you feel confortable opening another PR with that implementation? but anyway, it is a minor issue, don't need to hold this PR just due to that problem. 

4 - Need to check if the other distribution functions work. I guarantee that some of them do not work with just a nominal value and a standard deviation.
- I believe in you when you say that it may not work for every case. Shall we reduce the number of options (of disstribution functions) we have? What are the most important ones that we should keep in the code? Let me tackle this issue, but feel free to add suggestions.

5 - Create a pretty __repr__ for all the DataClasses.
- That's something that I or Franz can help, just a matter or priorities here, tell me who you think that should solve it earlier.

6 - There is a problem with numpy's distribution function imports. We are currently using from numpy.random import * and this causes an issue with random import choice. What happens is that numpy.random has a choice function and this function does not work for us, we need to use random.choice. So the random import needs to be lower than the numpy's import.
- Ok I can imagine the problem but I still cannot see it. Does it happen at the Dispersion.py file or another? And could you think in a solution? If I understood it correctly, you have forced a import to be in a row lower than other, otherwise you would have a dependency problem, right?

7 - Currently, files are all being saved like we always did. Each dictionary is a line of the .txt. I wonder if this should be a .csv, so the files are easier to read by anyone. If not, we should at least create a method to export the results in a .csv.
- csv are more user friendly but less robust. What would happen if we set a simulation of 200 iterations saving a particular set of variables, and then we run more 400 iterations with a different set of export variables? There will be a difference in the number of columns, right? That would be a real mess. To be honest, I think the outputs would make more sense if saved into .json files. But again, we are discussing smth really not crucial here. We can add a csv feature in the future without hurting anyone, don't worry.

8 - Function objects and the parachute trigger functions are being handled in a not optimal way. Currently, some ifs are added to handle these special cases. We need to do this because we can not easily save these types of variables to the .txt. So in the inputs file, the trigger function is saved as 'trigger': <function drogueTrigger at 0x000001FC26EE9B40>, and you can not recreate the trigger function from this (which may not necessarily be a problem). Also, the trigger function problem is breaking the inputs_log attribute. For the Function objects cases, what is done is that whenever a Function object needs to be saved, the list of the source needs to be gotten, e.g. list(Function.source). The list conversion is necessary to make sure there is no line skips in the string when saved. This way of saving Functions can be bad because it is slow, and is run during the simulation. The only alternative I see for this issue is to use pickle.
- I don't see a case where someone wants to vary the trigger functions inside the Monte Carlo loop, why should we keep it in the flight_setting yield generator? I'm ok with using a pickle, please notice that the EnvironmentAnalysis already introduced a pickle library as a depency, can we start using that? This one I think you should solve faster than me, but let me know if you need any help.

9 - During the simulation a Dispersion.flight_list is created. This is a list that saves all the simulated flights and is not used anywhere else in our code, it is just something that is nice to have access to. The only issue here is that this list gets appended inside the run_dispersion loop, which might be inefficient. This needs to be tested. The other solution for this would be to make flight_list a cached_property and have it recreate the flights from the inputs file. For this to happen, the problem with saving the parachute triggers needs to be fixed.
- Why is that "nice to have access to"? Do you agree that it could become a huge memory problem? Currently in the notebooks we don't save the flights, so I don't see why starting saving it. But if the thing if that you want avoid the append, why don't we create a list of empty values having the length equals to the number of simulations, and by each iteration we replace the corresponding position value for the flight object. What do you think?

10 - Create evolution plots to analyze the convergence.
- minor issue, this should goes into the v1.0.1 milestone, not necessary now. Do you want to create it as an issue? In my opinion leaving the to-do is already good enough.

11 - Maybe when a Dispersion object is created it should try to instantly import a file with the filename given. If it cannot import it should just keep going as is.
- srry but didn't understand the problem here. What is the current behavior and what do you suggest as new? This seems minor to me.

12 - 

13 - Initialize variables so they can be accessed by MATLAB.
- Let's solve this by the end of the PR, since a lot of variables can still be modified.

14 - Currently, attributes are created and saved in the Dispersion object from the keys in the dispersion_results dictionary. I don't think this is necessary anymore. What do you think?
- I didn't understand actually. Can you point out what code lines I should take a look?

15 - There is a try: except: to check something about the sweepAngle. This should be removed.
- Why should it be removed? Is it a relevant problem so we shall open an issue to solve it? 

16 - Verify what happens if we use Environment instead of McEnvironment.
- This seems to be easy, let me try it.

17 - Add hybrid and liquid motor options whenever the Liquid Motors are done.
- Is that a big problem if we postpone this feature to v1.0.1? I would be ok saying that the dispersion class only accepts the solidmotors in the v1.0.0. Please consider how hard would that be.

18 - There is an if else when passing the lag value for the parachute inside of run_dispersion. It checks if the lag is bigger than zero since if it is smaller than zero it breaks the Flight simulation. This is not ideal and we should try to find another way to fix this, but there might not be any other option.
- If somebody create a simulation where it is possible to have a parachute with negative lag, or negative CdS, this is definetely an error with the inputs. We should tell to the user that he/she is making a mistake rather than checking in the simulation. Perhaps it should be checked inside the pydantic validators!! The problem with Mont Carlo analysis is that if you don't pay attention you can actually create some impossible scenarios. The same things is valid to the rocket mass and geometry dimensions like length or areas (all of these must always be positive).

19 - When an export_list is not provided, a message is printed: "No export list provided, using default list instead.". This is a little annoying. Is it really necessary?
- Not necessary, and we can convert the warning to a simple comment if that annoys you.

20 - Review and improve everything related to the plots (especially the Ellipses stuff). I didn't do anything in that part of the code.
- Believe me, this part is already OK since a long time ago. But I'm going to take a look at the results and I tell you if I see any problem.

21 - Change the names of the McXxxxXxxx classes. They are not good.
- Please give me another suggestion, I'm running out of ideas here. If you don't have a suggestion, then I think we should keep the name.

22 - Need to finish all of the documentation. Also need to find a way to document the DataClasses. There is a lot here.
- I can help on the documentation of some classes, but I'm not able to document the DataClasses where it uses the pydantic operations. There are a few things that I'd recommend you to describe using comments. For example, try to comment *why* you have used each validator instead 


Ok, with that being said, I'm going to take a look at the code and try to repreoduce the simulations here. 
Please take a time to anwser my questions too.
Thks for helping btw.

Gui-FernandesBR:
Quick side comment: since you have introduced the windFactors and dragFactors, as long as we save them at every iteration, I don't any reason to save a Function object in the inputs file (and it doesn't matter if it is a pickled object or the source).

The real problem is the parachute trigger function, which btw I also don't see reason to save it after each iteration.

MateusStano:
> 1 - Find a way to deal with Rail Buttons. This needs to be well thought out so we make an implementation that is robust enough to any future changes in the rail buttons.
>   
> * Should we create a simple class to rail buttons? We had discussed that a while ago, and it seems to me that the rail buttons don't follow the standard in the rocket class. If you're ok with this we can open a PR direct to the `beta/v1.0.0` branch and ask for reviews.

Done in #344 

> 2 - The get_distribution function is currently on the tools.py. I like it there but just making sure, this is correct, right?
> 
> * It was intentional, and I think we should keep it right there. If you see any problems or a sign of bad practice please describe here, otherwise just let's go.

Okay, lets leave it there then

> 3 - I think it would be valuable to verify if the distribuition_function input in the DataClasses (so when a string is inputted) is the name of a valid numpy.random function. This should be done inside of the DataClasses to make the error be raised right when the user is inputting them.
> 
> * Idk how to do it with the new pydantic implementation. In that case you'd be quicker than me. Do you feel confortable opening another PR with that implementation? but anyway, it is a minor issue, don't need to hold this PR just due to that problem.

I can work on this eventually

> 4 - Need to check if the other distribution functions work. I guarantee that some of them do not work with just a nominal value and a standard deviation.
> 
> * I believe in you when you say that it may not work for every case. Shall we reduce the number of options (of disstribution functions) we have? What are the most important ones that we should keep in the code? Let me tackle this issue, but feel free to add suggestions.

I think it would be good to just check which distribution functions actually work, and remove the ones that doesn't.

> 5 - Create a pretty **repr** for all the DataClasses.
> 
> * That's something that I or Franz can help, just a matter or priorities here, tell me who you think that should solve it earlier.

Deffinitely low priority but we will need to wait for Liquid Motors to be done to finish this one so it is okay to already start doing this things

> 6 - There is a problem with numpy's distribution function imports. We are currently using from numpy.random import * and this causes an issue with random import choice. What happens is that numpy.random has a choice function and this function does not work for us, we need to use random.choice. So the random import needs to be lower than the numpy's import.
> 
> * Ok I can imagine the problem but I still cannot see it. Does it happen at the Dispersion.py file or another? And could you think in a solution? If I understood it correctly, you have forced a import to be in a row lower than other, otherwise you would have a dependency problem, right?

Yes, it happens in the Dispersion.py file. And yes you understood correctly. I would suggest just removing the `*` from the line `from numpy.random import *` and adding just the names of the functions that we actually use (which is a big list) in the import. This would be good to do together with what we talked about in `4`

> 7 - Currently, files are all being saved like we always did. Each dictionary is a line of the .txt. I wonder if this should be a .csv, so the files are easier to read by anyone. If not, we should at least create a method to export the results in a .csv.
> 
> * csv are more user friendly but less robust. What would happen if we set a simulation of 200 iterations saving a particular set of variables, and then we run more 400 iterations with a different set of export variables? There will be a difference in the number of columns, right? That would be a real mess. To be honest, I think the outputs would make more sense if saved into .json files. But again, we are discussing smth really not crucial here. We can add a csv feature in the future without hurting anyone, don't worry.

Alright, no need to worry for now. But .json files are probably the way to go in the future.

> 8 - Function objects and the parachute trigger functions are being handled in a not optimal way. Currently, some ifs are added to handle these special cases. We need to do this because we can not easily save these types of variables to the .txt. So in the inputs file, the trigger function is saved as 'trigger': <function drogueTrigger at 0x000001FC26EE9B40>, and you can not recreate the trigger function from this (which may not necessarily be a problem). Also, the trigger function problem is breaking the inputs_log attribute. For the Function objects cases, what is done is that whenever a Function object needs to be saved, the list of the source needs to be gotten, e.g. list(Function.source). The list conversion is necessary to make sure there is no line skips in the string when saved. This way of saving Functions can be bad because it is slow, and is run during the simulation. The only alternative I see for this issue is to use pickle.
> 
> * I don't see a case where someone wants to vary the trigger functions inside the Monte Carlo loop, why should we keep it in the flight_setting yield generator? I'm ok with using a pickle, please notice that the EnvironmentAnalysis already introduced a pickle library as a depency, can we start using that? This one I think you should solve faster than me, but let me know if you need any help.



> 9 - During the simulation a Dispersion.flight_list is created. This is a list that saves all the simulated flights and is not used anywhere else in our code, it is just something that is nice to have access to. The only issue here is that this list gets appended inside the run_dispersion loop, which might be inefficient. This needs to be tested. The other solution for this would be to make flight_list a cached_property and have it recreate the flights from the inputs file. For this to happen, the problem with saving the parachute triggers needs to be fixed.
> 
> * Why is that "nice to have access to"? Do you agree that it could become a huge memory problem? Currently in the notebooks we don't save the flights, so I don't see why starting saving it. But if the thing if that you want avoid the append, why don't we create a list of empty values having the length equals to the number of simulations, and by each iteration we replace the corresponding position value for the flight object. What do you think?

I don't know if replacing an item is faster than appending, it might be. But actually you were the one who created the `flight_list` parameter, I just kept it because I thought there could be some useful uses for debugging or something like that. But if you don't think there is much use for it, I am happy to delete it :)

> 10 - Create evolution plots to analyze the convergence.
> 
> * minor issue, this should goes into the v1.0.1 milestone, not necessary now. Do you want to create it as an issue? In my opinion leaving the to-do is already good enough.

No problem lets leave it as a TODO

> 11 - Maybe when a Dispersion object is created it should try to instantly import a file with the filename given. If it cannot import it should just keep going as is.
> 
> * srry but didn't understand the problem here. What is the current behavior and what do you suggest as new? This seems minor to me.

When you create a dispersion object you input the `filename`, which currently is the name (and/or directory) that will be used to create the dispersion files. I think it makes a lot of sense to check if there already are dispersion files created with that filename and, if there is, import them and print out: "file with the same filename was found and was automatically imported /n number of simulations imported : x ". Or something like that

If we don't do this the user would have to create de dispersion object with a filename, then use the `import_results` method indicating the filename of the files to be imported.

> 14 - Currently, attributes are created and saved in the Dispersion object from the keys in the dispersion_results dictionary. I don't think this is necessary anymore. What do you think?
> 
> * I didn't understand actually. Can you point out what code lines I should take a look?

I am referring to the `setattr` in:

https://github.com/RocketPy-Team/RocketPy/blob/d76d0c3d0b95e46dff2f0c56d948d88c7c19b877/rocketpy/Dispersion.py#L281-L304

> 15 - There is a try: except: to check something about the sweepAngle. This should be removed.
> 
> * Why should it be removed? Is it a relevant problem so we shall open an issue to solve it?

It is a try except inside of the yield function. I don't actually know what it is doing, I think you added them there. But I can try to understand it and remove it if necessary.

> 17 - Add hybrid and liquid motor options whenever the Liquid Motors are done.
> 
> * Is that a big problem if we postpone this feature to v1.0.1? I would be ok saying that the dispersion class only accepts the solidmotors in the v1.0.0. Please consider how hard would that be.

I think it might be a really easy change, so we could just do it when Luquids are done. However, if it ends up being too much trouble we can always postpone this.

> 18 - There is an if else when passing the lag value for the parachute inside of run_dispersion. It checks if the lag is bigger than zero since if it is smaller than zero it breaks the Flight simulation. This is not ideal and we should try to find another way to fix this, but there might not be any other option.
> 
> * If somebody create a simulation where it is possible to have a parachute with negative lag, or negative CdS, this is definetely an error with the inputs. We should tell to the user that he/she is making a mistake rather than checking in the simulation. Perhaps it should be checked inside the pydantic validators!! The problem with Mont Carlo analysis is that if you don't pay attention you can actually create some impossible scenarios. The same things is valid to the rocket mass and geometry dimensions like length or areas (all of these must always be positive).

The problem here is due to the std in the dispersion actually. if the lag is 0.3 and its std is 0.2 there is a chance that it ends up as a negative value. I don't think we can check for this in the validators, so the solution I had was to assume the lag is zero if the random value generated is negative (this is done in lines 745~747 inside of the dispersion simulation loop). I just really do not like that if else there because it might slow down things a bit. It would be good if we could figure out a way to avoid this.

Gui-FernandesBR:
Last commit solved number 4, 6, 9 and 18 ✅✅✅✅
Quick explanation:
- since we were already using `import numpy as np`, it was redudant using `from numpy.random import *`
- I checked the numpy's docs and checked what distribution functions could be used in our case. Now it is making more sense I guess.
- I simpled removed the print('bla bla bla') that you pointed out

Now a few updates on the other topics, considering your last comments:
 - 8 - You skipped the number 8 😅
 - 9 - yeah please remove the `Dispersion.flight_list`. But I can garantee that replacing is faster than append().
 - 11 - ok, understood, good point.
 - 14 - ok it is not extremely necessary. But how are the user going to print, for instance, the mean and std of 'apogee'? 
 - 15 - OK, gotcha! The problem is that we cannot set both sweepangle and sweeplength simultansly when creating a fin object, so I tried to catch that exception. Do you see a better solution for this? https://github.com/RocketPy-Team/RocketPy/blob/master/rocketpy/AeroSurfaces.py#L766
 - 16 - I still didn't have time to see this one, but it can be tackled by creating some unit tests maybe.
 - 18 - You mentioned the perfect case of someone trying to set an impossible scenario for a iteration of a monte carlo simulation process. Please notice that simply removing the cases where parachute lag is negative doesn't work properly since it modifies the distribution function behavior. I think we should warn the user that he/she is trying to create something impossible, and it doesn't have to be done inside the dispersion implementation. A simple warning saying to the user that the rocket mass is negative would be a good example of how to approach the problem.

let me know in case u have any questions.

rocketpy/Dispersion.py
```diff
@@ -3,82 +3,96 @@
...
+    from functools import cached_property
+except ImportError:
+    from .tools import cached_property
 
-class Dispersion:
+# TODO: How to save Functions? With pickle? Save just the source?
 
-    """Monte Carlo analysis to predict probability distributions of the rocket's
-    landing point, apogee and other relevant information.
+# TODO: create a method that recreates each flight from inputs_log
```
Gui-FernandesBR, line 30: @MateusStano , can you remind me what would be the benefits of doing so?

rocketpy/Dispersion.py
```diff
@@ -87,841 +101,209 @@ def __init__(
...
+    @dispersion_error_file.setter
+    def dispersion_error_file(self, value):
+        self._dispersion_error_file = value
+        return self._dispersion_error_file
+
+    # cached properties
+    @cached_property
+    def inputs_log(self):
+        """Save inputs_log log from a file into an attribute for easy access"""
+        # TODO: this currently does not work because of parachute trigger function
```
Gui-FernandesBR, line 197: Does it still not work? It seems that one of the last commits solved it.

rocketpy/Dispersion.py
```diff
@@ -87,841 +101,209 @@ def __init__(
...
-            to be varied, and the values can be either tuple or list. If the value
-            is a single value, the corresponding class of the parameter need to
-            be passed on the run_dispersion method.
-        object : Environment, SolidMotor, Rocket, Flight
-            Object that will be used to get attributes from.
-        class_name : string
-            Name of the class used in input_dict.
-        class_name_pretty : string
-            Name of the class used for prints.
+    @cached_property
```
Gui-FernandesBR, line 248: If I run simulations in append='True' mode, the number of loaded_sims should be modified, no? In that case, does it still seem to be a good idea keeping this variable as a cached_property?

rocketpy/Dispersion.py
```diff
@@ -1065,490 +487,503 @@ def run_dispersion(
...
                 )
 
-            rocket_dispersion.setRailButtons(
-                position=[
-                    setting["positionFirstRailButton"],
-                    setting["positionSecondRailButton"],
-                ],
-                angularPosition=setting["railButtonAngularPosition"],
-            )
+            # TODO: Fix rail buttons definition here
```
Gui-FernandesBR, line 680: @MateusStano , we still need to fix rail button in this part of the code, right? I thought you had fixed it already

rocketpy/Environment.py
```diff
@@ -367,6 +367,7 @@ def __init__(
             self.setDate(date, timeZone)
         else:
             self.date = None
+            self.datetime_date = None
```
Gui-FernandesBR, line 370: The `datetime_date` variable could have be introduced directly to the `beta/v1.0.0`, avoiding future dilemmas

rocketpy/Function.py
```diff
@@ -174,7 +174,7 @@ def setSource(self, source):
             def source(x):
                 return temp
 
-        # Handle callable source or number source
+        # Handle callable source or n   umber source
```
Gui-FernandesBR, line 177: this line was clearly modified by mistake

rocketpy/Rocket.py
```diff
@@ -812,20 +810,17 @@ def setRailButtons(self, position, angularPosition=45):
...
-        None
+        rail_buttons : RailButtons
+            RailButtons object created
         """
         # Place top most rail button as the first element of the list
         if self._csys * position[0] < self._csys * position[1]:
             position.reverse()
         # Save important attributes
-        self.railButtons = self.railButtonPair(position, angularPosition)
+        rB = RailButtons(*position, angular_position)
```
Gui-FernandesBR, line 820: don't be lazy, go ahead and use rail_button as the name of the variable

rocketpy/monte_carlo/DispersionModel.py
```diff
@@ -0,0 +1,171 @@
...
+        field_values = "\n\t"
+        field_values += "\t".join(
+            [
+                f"{field_name}={getattr(self, field_name)!r}\n"
+                for field_name in self.__fields__
+            ]
+        )
+        return f"{self.__class__.__name__}({field_values})\n"
+
+    # TODO: find a way to validate if distribution_function string is the name
```
Gui-FernandesBR, line 31: Is this really necessary?

rocketpy/monte_carlo/DispersionModel.py
```diff
@@ -0,0 +1,171 @@
...
+                    values[field] = v
+            elif isinstance(v, (int, float)):
+                # not list or tuple, must be an int or float
+                # get attr and returns (nom_value, std)
+                values[field] = (getattr(values[obj_name], field), v)
+            else:
+                raise ValueError(f"Field '{field}' must be a tuple, list, int or float")
+        return values
+
+    @validator(
```
Gui-FernandesBR, line 130: Good use of factors! I liked it

rocketpy/monte_carlo/mc_environment.py
```diff
@@ -0,0 +1,51 @@
...
+__license__ = "MIT"
+
+from typing import Any, List, Tuple, Union
+
+from pydantic import Field, StrictInt, StrictStr, validator
+
+from ..Environment import Environment
+from .DispersionModel import DispersionModel
+
+# TODO: name suggestions: `DispersionEnvironment`, `DispEnvironment`, EnvironmentDispersion`, EnvironmentDisp`, `MonteCarloEnvironment`, `EnvironmentMonteCarlo`,
```
Gui-FernandesBR, line 12: Tks for all the suggestions. I believe the most intuitive name would be `MonteCarloEnvironment` here. @MateusStano agreed? If not, we may need a third view to help us.

rocketpy/monte_carlo/mc_flight.py
```diff
@@ -0,0 +1,44 @@
...
+    in the Flight class, see its documentation for more information.
+    """
+
+    flight: Flight = Field(..., exclude=True)
+    inclination: Any = 0
+    heading: Any = 0
+    initialSolution: Union[
+        Flight,
+        Tuple[
+            Union[StrictInt, StrictFloat],  # tInitial
```
Gui-FernandesBR, line 28: Does it means that, if we want, we can apply `initialSolution` inside the monte carlo loop?

rocketpy/monte_carlo/mc_rocket.py
```diff
@@ -0,0 +1,264 @@
...
+                "nosecone must be of rocketpy.monte_carlo.McNoseCone or rocketpy.NoseCone type"
+            )
+        if isinstance(nose, NoseCone):
+            # create McNoseCone
+            nose = McNoseCone(nosecone=nose)
+        nose.position = self._validate_position(position, nose.nosecone, "position")
+        return self.nosecones.append(nose)
+
+    def addTrapezoidalFins(self, fins, position=[]):
+        # checks if input is a McTrapezoidalFins type
```
Gui-FernandesBR, line 185: Please help with docs

rocketpy/monte_carlo/mc_rocket.py
```diff
@@ -0,0 +1,264 @@
...
+            raise TypeError("fins must be of McTrapezoidalFins type")
+        if isinstance(fins, TrapezoidalFins):
+            # create McTrapezoidalFins
+            fins = McTrapezoidalFins(trapezoidalFins=fins)
+        fins.position = self._validate_position(
+            position, fins.trapezoidalFins, "position"
+        )
+        return self.fins.append(fins)
+
+    def addEllipticalFins(self, fins, position=[]):
```
Gui-FernandesBR, line 196: Please help with docs

rocketpy/monte_carlo/mc_rocket.py
```diff
@@ -0,0 +1,264 @@
...
+        if isinstance(fins, EllipticalFins):
+            # create McEllipticalFins
+            fins = McEllipticalFins(ellipticalFins=fins)
+        fins.position = self._validate_position(
+            position, fins.ellipticalFins, "position"
+        )
+        return self.fins.append(fins)
+
+    def addTail(self, tail, position=[]):
+        # checks if input is a McTail type
```
Gui-FernandesBR, line 209: Please help with docs

rocketpy/monte_carlo/mc_rocket.py
```diff
@@ -0,0 +1,264 @@
...
+
+        Parameters
+        ----------
+        parachute : McParachute
+            The parachute to be added to the rocket. This must be a McParachute
+            type.
+
+        Returns
+        -------
+        ????
```
Gui-FernandesBR, line 229: help help

tests/conftest.py
```diff
@@ -243,6 +254,194 @@ def func_from_csv():
...
+        span=0.0005,
+    )
+
+    tail = rocket.addTail(
+        topRadius=0.0635, bottomRadius=0.0435, length=0.060, position=-1.194656
+    )
+
+    mc_tail = McTail(tail=tail, bottomRadius=0.0005, length=0.001)
+
+    def mainTrigger(p, y):
```
Gui-FernandesBR, line 363: I lost the count of how many times we define this function in the tests; Maybe we should convert it to a fixture as well!

tests/test_dispersion.py
```diff
@@ -1,267 +1,46 @@
...
+    assert len(dispersion.outputs_log) == 10
+    assert dispersion.plotEllipses(save=True) == None
 
-    os.remove("test_obj_defined.disp_inputs.txt")
-    os.remove("test_obj_defined.disp_outputs.txt")
-    os.remove("test_obj_defined.disp_errors.txt")
-    os.remove("test_obj_defined.svg")
+    # Delete the test file
+    # TODO: can't delete the files because they were not closed, bad practice!
+    # os.remove("test_dispersion_class.disp_errors.txt")
```
Gui-FernandesBR, line 42: Hey, it seems these files have never been closed in the Dispersion class

Gui-FernandesBR: @MateusStano I noticed that you used a lot of `blablabla = open('flename', ...)`, this is actually not the best approach possible ([see](https://realpython.com/why-close-file-python/)) since it doesn't guarantee the files to be closed.
Instead, I think one may use `with open(...)`, and then passing the content of the file to a string, which could be added by simply using the `+`  operand.

Let m know what you think about it

rocketpy/Dispersion.py
MateusStano: This would allow the user to "reconstruct" any of the flights used in the code. It would be really cool to do and easy to implement. Unfortunately, the problem with saving the trigger function and the drag Function makes it impossible. This would require an implementation of pickle

MateusStano: I'd say for now it is not in the scope of v1. We can leave implementing pickle as a future task

rocketpy/Environment.py
MateusStano: True, Ill make a PR for this

tests/conftest.py
MateusStano: Hmmm... maybe this should be a change done in #345. With the changes in that PR we should be able to just remove these functions

rocketpy/monte_carlo/mc_flight.py
MateusStano: Yes! You can pass a initial solution to be used in all flight simulations of the dispersion analysis.

However the way it is currently done you can not add variations to these initial solutions, meaning you can not input a list of initial solutions. You can only input one initial solution that will be used for all flights....

Do you think we should allow the user to input a list of initial solutions?

rocketpy/monte_carlo/mc_environment.py
MateusStano: Thinking back on this I really liked the `Data` names: `EnvironmentData`. I can understand if it is not descriptive enough though.

I would say that I like `DispersionEnvironment` just as much as `MonteCarloEnvironment` heheh

I think we gonna need to open a poll with the team to decide this one

rocketpy/monte_carlo/DispersionModel.py
MateusStano: Well... it is already done hehe

It was done in https://github.com/RocketPy-Team/RocketPy/pull/336/commits/0d5fa7072d0fd15871992bee9366ebedb2a5a828

rocketpy/Dispersion.py
MateusStano: Just forgot to remove the commented lines

rocketpy/Dispersion.py
MateusStano: Yep it is working. Just forgot to remove the Yep it is working. Just forgot to remove the TODO

What I did was I made the Functions be saved just as their `__repr__` ( the `Function R1 -> R1` thingy) and the trigger function just saves as a string too

rocketpy/Dispersion.py
```diff
@@ -677,9 +677,6 @@ def run_dispersion(
                     noise=setting["parachutes"][chute]["noise"],
                 )
 
-            # TODO: Fix rail buttons definition here
-            # rocket_dispersion.setRailButtons()
```
Gui-FernandesBR: Ok, @MateusStano , but now when the rail buttons are being added to the rocket within the monte carlo loop?

tests/conftest.py
Gui-FernandesBR: That's not the most important point actually. Currently a lot of tests repeat code when we need to add aerosurfaces to them, as well as the parachutes. Maybe we should create parachutes and surfaces as fixtures and avoid defining it inside different tests. Example:
- https://github.com/RocketPy-Team/RocketPy/blob/beta/v1.0.0/tests/test_flight.py#L108
- https://github.com/RocketPy-Team/RocketPy/blob/beta/v1.0.0/tests/test_flight.py#L199
- https://github.com/RocketPy-Team/RocketPy/blob/beta/v1.0.0/tests/test_flight.py#L400
- etc.

A new pull request in the future should modify the tests to avoid repeating code. I think I will open an issue with that regards soon.

rocketpy/monte_carlo/mc_flight.py
Gui-FernandesBR: > Do you think we should allow the user to input a list of initial solutions?

I don't think so, that would be really weird to use. What would make more sense would be having an initial solution matrix composed by different parameters like (mean, std. dev, 'distr'). But that is not crucial here, no need to introduce that right now.

What we could do by now is to improve the docs to explain to the user that, if he wants to, he may use initial solution. But if I understood it right, if he sets an iniital solution to the Flight object, it will be automatically used by the McFlight object too.

rocketpy/monte_carlo/mc_environment.py
Gui-FernandesBR: Personally not a fun of polling but I agree we need it. Let's finish it by this week please.

rocketpy/Environment.py
Gui-FernandesBR: > True, Ill make a PR for this

Please don't forget to explain your motivation (problem + solution) in the PR description. Again, avoiding future stress.

rocketpy/Dispersion.py
Gui-FernandesBR: Agreed! As it is already a TODO in the code, no need for opening a new issue

rocketpy/Dispersion.py
MateusStano: They are being defined just a few lines above

rocketpy/Dispersion.py
```diff
@@ -87,841 +101,209 @@ def __init__(
...
+        # TODO: Initialize variables so they can be accessed by MATLAB
 
-        # Aerodynamic Surfaces
-        dictionary = self.__process_aerodynamic_surfaces_from_dict(dictionary)
+        return None
 
-        return dictionary
+    # getters and setters for dispersion input/output/error files
+    @property
+    def dispersion_input_file(self):
```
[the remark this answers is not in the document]
MateusStano: Fixed in https://github.com/RocketPy-Team/RocketPy/pull/336/commits/4e8319e6bbaa1ef7afef23386fe4b5076ba9c78f

rocketpy/Dispersion.py
MateusStano: Fixed in https://github.com/RocketPy-Team/RocketPy/pull/336/commits/4e8319e6bbaa1ef7afef23386fe4b5076ba9c78f

tests/test_dispersion.py
MateusStano: Fixed in https://github.com/RocketPy-Team/RocketPy/pull/336/commits/4e8319e6bbaa1ef7afef23386fe4b5076ba9c78f

rocketpy/monte_carlo/mc_flight.py
MateusStano: > if he sets an iniital solution to the Flight object, it will be automatically used by the McFlight object too.

No it won't. `initialSolution` does not go through the `@root_validator` so it won't check if there already is an initial solution on the flight object. This way the user would need to input the `initialSolution` in `McFlight`. 

Do you think we should change this? I hadn't even thought of this problem

It would be an easy change.

Gui-FernandesBR: > Do you think we should change this? I hadn't even thought of this problem

No, what we need for now is to release v1.0.0 ...

Please just add a simple "initialSolution option will be available in future releases" comment.
If there's demand, we solve the problem in the future, don't worry.

MateusStano:
Hey @Gui-FernandesBR I think everything is done with this one now!! Just need a final look over the in these changes from you:

- The file imports are all working perfectly now. You can review the code if you want, but I'd say that part is good
- Added two methods to the Monte Carlo classes:
    - `generate_dict` --> returns a generator that creates a dict with the randomized values for the inputs arguments of that class, given each arguments distribution function
    - `create_object` --> returns an object created with the randomized value. Each time you call `create_object` it returns a different object

One comment about the `create_object` is the way `Environment` is handled. I made it so instead of recreating an Environment object, it just makes a deep copy and resets the attributes. This is done to avoid having to reinitialized the Atmospheric Model. 

What I wonder is if this is not a faster option for the other classes. Creating a deepcopy and then just changing the attributes of the object might be a way to optimize the code. But this is for later.

Also, the tests that are not passing are unrelated to this PR

Gui-FernandesBR:
Great work, @MateusStano , I loved that!

Please let me review it in the following days =)

Gui-FernandesBR (approved):
Great work, @MateusStano , you were awesome as always.
Keep in mind that we are going to have a hard time to approve this changes to be merged on beta/v1.0.0 since we have done a lot of modifications. However, it is ready to be merged into enh/class_dispersion already

Please let's make an effort to solve the issues with tests.

PR #337, opened 2023-03-10
ENH: burnout time changes. Breaking changes

FranzYuri:
## Pull request type

## Pull request checklist

burnout time is a required variable
## What is the new behavior?

Now the variable is called burn time and can be a tupple with starting and ending times.

The only beaking change is that the name of the variable burnOut changed to burn_time
- [X] Yes
- [ ] No

## Other information

Enter text here...


MateusStano:
Hey, I just commited a few big changes in the code you had made. First I added the use of `Function.setDiscrete` to remove the while loops. Then I improved some of the if else structure to have no repetitions.

There is one big problem still. When using the `burn_time` as tuple the propellant mass does not go to zero.
Here is an example, using the Calisto thrust curve with `burn_time = (1,3)` the thrust curve is:

image


And the propellant mass:

image


@phmbressan can you confirm this is a problem and maybe fix it? I think you will understand what is going on here much faster than me.

giovaniceotto:
- [ ] Document new attributes

phmbressan:
@MateusStano 

The issue of negative propellant mass is fixed now. As you can see in my earlier commits, the main cause were some methods that still considered all burns starting at zero seconds.

On top of that, there were some issues with the type checking and conversion for the parameter `burn_time`, I changed it a bit. It is not the most elegant code since a wide range of input types must checked and converted, but I believe it is working. As always, additional reviews of these changes are important.

Gui-FernandesBR:
What do we need to do in order to get this PR ready for review?

Btw shouldn't it go straight to the `beta/v1.0.0` branch?

MateusStano:
@phmbressan to merge this one we need two more things:

- [ ] Change `motor.thrust` to only take into account the thrust curve during burn time
- [ ] Merge `enh/liquid-motors` branch into this one

phmbressan:
There are still some issues regardig this PR:

- The tests for `SolidMotor` are extremely tight in some cases (_e.g._ tolerances of `1e-16`), thus, due to the new usage of a termination event in grain geometry, there are tests that do not pass.
- There is some revisions of nomenclature and definitions of old methods, such as `reshapeThrustCurve`, I am not sure why the test for this method also does not pass.

MateusStano:
The following code gives incorrect results:

```
Pro75M1670 = SolidMotor(
    thrustSource=[[0,1000],[2,2000],[4,1000]],
    burn_time=3,
    grainsCenterOfMassPosition=-0.85704,
    grainNumber=5,
    grainSeparation=5 / 1000,
    grainDensity=1815,
    grainOuterRadius=33 / 1000,
    grainInitialInnerRadius=15 / 1000,
    grainInitialHeight=120 / 1000,
    nozzleRadius=33 / 1000,
    throatRadius=11 / 1000,
    interpolationMethod="linear",
    nozzlePosition=-1.255,
    coordinateSystemOrientation="nozzleToCombustionChamber",
)
```
Result: 

image


The expected is 

image


The total impulse is also wrong. Result is `3000 N`, and expected is `4750 N`

rocketpy/motors/Motor.py
```diff
@@ -145,88 +155,169 @@ def __init__(
...
+        # Reshape thrustSource if needed
         if reshapeThrustCurve:
+            # Overwrites burn_time and thrust
             self.reshapeThrustCurve(*reshapeThrustCurve)
 
+        # Post process thrust
+        self.thrust = self.clipThrust()
+
+        # Auxiliary quantities
+        self.burnDuration = self.burn_time[1] - self.burn_time[0]
```
MateusStano, line 186: I like this! Really easy to understand

rocketpy/motors/Motor.py
```diff
@@ -145,88 +155,169 @@ def __init__(
...
+            else:
+                raise ValueError("burn_time must be a list or tuple of length 1 or 2.")
+        elif callable(self.thrust.source):
+            raise ValueError(
+                "When using a float or callable as thrust source a burn_time "
+                "range must be specified."
+            )
+        else:
+            self._burn_time = (self.thrust.xArray[0], self.thrust.xArray[-1])
+
```
MateusStano, line 234: Also great!

rocketpy/motors/Motor.py
```diff
@@ -145,88 +155,169 @@ def __init__(
...
+    def clipThrust(self):
```
MateusStano, line 235: Missing docstrings in this function

rocketpy/motors/LiquidMotor.py
```diff
@@ -1,365 +1,365 @@
...
+            also be given as a callable function, whose argument is time in
+            seconds and returns the thrust supplied by the motor in the
+            instant. If a string is given, it must point to a .csv or .eng file.
+            The .csv file shall contain no headers and the first column must
+            specify time in seconds, while the second column specifies thrust.
+            Arrays may also be specified, following rules set by the class
+            Function. See help(Function). Thrust units are Newtons.
+        burnOut : int, float
+            Motor burn out time in seconds.
+        nozzleRadius : int, float
```
MateusStano, line 45: Need to change the docs here, and probably in all motor classes

rocketpy/motors/SolidMotor.py
```diff
@@ -1,754 +1,770 @@
...
+            Maximum thrust value of the given thrust curve, in N.
+        Motor.maxThrustTime : float
+            Time, in seconds, in which the maximum thrust value is achieved.
+        Motor.averageThrust : float
+            Average thrust of the motor, given in N.
+        Motor.burnOutTime : float
+            Total motor burn out time, in seconds. Must include delay time
+            when the motor takes time to ignite. Also seen as time to end thrust
+            curve.
+        Motor.exhaustVelocity : float
```
MateusStano, line 104: Need to add `burn_time` here

phmbressan:
> The following code gives incorrect results:
> 
...
> 
> The total impulse is also wrong. Result is `3000 N`, and expected is `4750 N`

@MateusStano 

Great catch. It is exactly wrong _per se_, but it is not intuitive and will have to be changed. What is happening is that I am clipping the thrust source based on the `burn_range` parameter. Therefore, all data in the `xArray` that is greater than the 3 ends up being removed.

This unintuive result is an unplanned consequence of that. I am thinking on a better way to make it work, an approach could be discretize de domain before clipping, but there is an accuracy loss in doing that.

rocketpy/motors/Motor.py
phmbressan: Solved by [081f0de](https://github.com/RocketPy-Team/RocketPy/pull/337/commits/081f0de83c750b1e936a149b49759e5f97e8ea96).

rocketpy/motors/LiquidMotor.py
phmbressan: Solved by [081f0de](https://github.com/RocketPy-Team/RocketPy/pull/337/commits/081f0de83c750b1e936a149b49759e5f97e8ea96).

rocketpy/motors/SolidMotor.py
phmbressan: Solved by [081f0de](https://github.com/RocketPy-Team/RocketPy/pull/337/commits/081f0de83c750b1e936a149b49759e5f97e8ea96).

phmbressan:
@MateusStano Thank you for your suggestions. I believe the issues you raised were solved, however another review is always important, so I ask you to review these latest changes.

MateusStano (approved):

rocketpy/motors/Motor.py
```diff
@@ -145,88 +160,196 @@ def __init__(
...
         None
         """
         # Retrieve current thrust curve data points
-        timeArray = self.thrust.source[:, 0]
-        thrustArray = self.thrust.source[:, 1]
+        timeArray, thrustArray = self.thrust.xArray, self.thrust.yArray
+
         # Move start to time = 0
         if startAtZero and timeArray[0] != 0:
             timeArray = timeArray - timeArray[0]
```
giovaniceotto, line 333: The behavior described by the docstring suggests that if `startAtZero` is `True`, than we guarantee that `thrust(t) > 0` for `t > 0`. However, this is not enforced by the action performed by the following two lines:

```python
if startAtZero and timeArray[0] != 0:
     timeArray = timeArray - timeArray[0]
```

Option A: enforce documented behavior.

```python
if startAtZero:
    # Get index of first non-zero thrust value
    nonZeroIndex = thrustArray.nonzero()[0][0]
    # Clip timeArray and thrustArray
    nonZeroIndex = max(1, nonZeroIndex) # Avoids problems when subtracting 1 in the next lines
    thrustArray = thrustArray[nonZeroIndex-1, :]
    timeArray = timeArray[nonZeroIndex-1, :]
    timeArray = timeArray  - timeArray[0]
```

Option B: change docstring to reflect current behavior.

```
        startAtZero: bool, optional
            If True, trims the initial thrust curve points which
            are not given, translating the thrust curve so that
            thrust data starts at time equals 0. If False, no translation
            is applied.
```

I have no preference in terms of usefulness of each option. Rationally speaking, option B is simpler and safer, as it is just a subset of option A.

rocketpy/motors/Motor.py
```diff
@@ -149,88 +164,196 @@ def __init__(
...
-        self.burnOutTime = burnTime
-        self.thrust.setInterpolation(self.interpolate)
+        # Reshape time - set burn time to newBurnTime
+        self.burn_time = newBurnTime
 
-        # Reshape thrust - set total impulse
+        # Compute old thrust based on new time discretization
+        newTimeArray = (
+            (self.burn_time[1] - self.burn_time[0]) / (timeArray[-1] - timeArray[0])
+        ) * timeArray + self.burn_time[0]
```
giovaniceotto, line 341: I believe the current implementation only works if `timeArray[0] == 0`.

```suggestion
        # Adjust scale
        newTimeArray = (
            (self.burn_time[1] - self.burn_time[0]) / (timeArray[-1] - timeArray[0])
        ) * timeArray
        # Adjust origin
        newTimeArray = newTimeArray - newTimeArray[0] + self.burn_time[0]
```

giovaniceotto (changes requested):
IMPORTANT: I have reviewed only `rocketpy.motors.Motor.reshapeThrustCurve()`.

Two simple issues need to be addressed. Once they have been made, feel free to automatically consider this approved.

rocketpy/motors/Motor.py
```diff
@@ -247,7 +379,11 @@ def totalImpulse(self):
         self.totalImpulse : float
             Motor total impulse in Ns.
         """
-        return self.thrust.integral(0, self.burnOutTime)
+        # Calculate total impulse
+        self.totalImpulse = self.thrust.integral(*self.burn_time)
```
giovaniceotto, line 383: Just found out that since `self.totalImpulse` is a cached property, this line of code may go by without changing the cached value.

Here is a suggestion based on https://stackoverflow.com/questions/62662564/how-do-i-clear-the-cache-from-cached-property-decorator. Better solutions are welcome.
```suggestion
        # # Clear cache and calculate total impulse
        del self.totalImpulse 
        self.totalImpulse = self.thrust.integral(*self.burn_time)
```

Gui-FernandesBR: I prefer you change the cached property to a simple property then.
I have faced similar issue in the EnvironmentAnalysis class recently, and that was my approach.

Do not try to "clean" the cached property, if you need to update the value, just use property and we will be fine.

The `self.totalImpulse` is not even called a lot of times during the execution, so no worries.

phmbressan: This is a great catch which turned out to be more significant that it seemed since the `reshapeThrustCurve` modified instance attributes that were cached, leading to incorrect values.

There is certainly more than one way to solve this issue, in [e4c75a1](https://github.com/RocketPy-Team/RocketPy/pull/337/commits/e4c75a158985fc72b1c23c73abf7a736a26c1262) I turned the `Motor` methods whose only function is to edit thrust values into `static`.

Therefore, the standard ways to reshape the thrust curve are:

- The same way as before, i.e., through the `__init__` parameter;
- Calling the static method as `Motor.reshapeThrustCurve` and producing a new thrust curve that may be used to instantiate a new motor.

phmbressan:
Sorry for submitting a new review, I had an issue with github pending comments.

The tests are not passing due to the really odd lack of a library in github actions, I have not been able to comprehend why this is happening. Locally they are fine for me.

Gui-FernandesBR:
> Sorry for submitting a new review, I had an issue with github pending comments.
> 
> The tests are not passing due to the really odd lack of a library in github actions, I have not been able to comprehend why this is happening. Locally they are fine for me.

What library? Can you describe it better?

phmbressan:
> > Sorry for submitting a new review, I had an issue with github pending comments.
> > The tests are not passing due to the really odd lack of a library in github actions, I have not been able to comprehend why this is happening. Locally they are fine for me.
> 
> What library? Can you describe it better?

From the testing log in github actions.

```
../../../hostedtoolcache/Python/3.7.17/x64/lib/python3.7/bz2.py:19: in <module>
    from _bz2 import BZ2Compressor, BZ2Decompressor
E   ModuleNotFoundError: No module named '_bz2'
```

I searched a bit and it seems related to the `pandas` import in BelaLui testing. This is odd, since those were not altered and are alright in other branches.

rocketpy/motors/Motor.py
```diff
@@ -5,7 +5,9 @@
 __license__ = "MIT"
 
 import re
+import warnings
 from abc import ABC, abstractmethod
+from rocketpy.utilities import tuple_handler
```
Gui-FernandesBR, line 10: This behavior is not accepted anymore. 

The `tuple_handler` function should be positioned in the tools.py file, otherwise an importError would be raised in the future due to circular reference.

Gui-FernandesBR: Solved by [1a2959e](https://github.com/RocketPy-Team/RocketPy/pull/337/commits/1a2959e0f1695f13ffa9a47b4e28864e8c0b5a17)

Gui-FernandesBR (approved):
Great solving!

PR #338, opened 2023-03-11
Enh/variable gravity

FranzYuri:
## Pull request type

## Pull request checklist

currently gravity is a constant equal to 9,80665 m/s²

## What is the new behavior?

Now, env.g is a callable that returns the gravity given the height.
On the other hand, env.standard_gravity will store the constant value of 9,80665 m/s².

- [ ] Yes
- [X] No

## Other information

Enter text here...


phmbressan:
I believe this PR may be reaching its final steps. Just a few things that I wanted to point out before finishing reviews and merging:

`Flight` class `PotentialEnergy` changes:

- The method was wrong since inputting `self.z` (a `Function`) to the environment gravity `Function` yielded an error at interpolation.
- Trying to multiply `Function` by possible arrays (such as `self.env.g(self.z.source[,:1])`)  does not help.
- Temporary fix was to perform the multiplication "manually".

`Environment` class `gravity` parameter:

- I personally don't think that I took the best decision here, i.e., the default `None` leads to variable gravity, however, since the parameter is a `Function` input, the options were limited. I am open to better suggestions.

Gravity `Function` being discrete:

- The `Environment` discretizes the gravity function height domain for two main reasons: 1) It is likely faster in the case of Somigliana's formula (but there might be a small performance hit compared to returning a constant value); 2) Using lambda sources raises an error at the plotting class when the `.source[]` attributes are used, since it turns many `Functions` of the `Flight` class also lambda based (being a performance hit also).

Nota bene: the fin flutter tests still do not pass for me locally.

MateusStano:
> Nota bene: the fin flutter tests still do not pass for me locally.

Hey, seems that in this case the test actually did its job. The change in the gravity model actually changed the simulation by just enough to make the test break. It seems flutter analysis is very very sensitive

I think I fixed it now, please just try to run it locally to make sure

MateusStano (approved):

giovaniceotto:
@MateusStano, I see you changed the values for the fin flutter test. Shouldn't we keep a test with the old values, but using constant gravity, just to be safe?

rocketpy/Flight.py
```diff
@@ -2121,8 +2121,7 @@ def potentialEnergy(self):
         # Redefine totalMass time grid to allow for efficient Function algebra
         totalMass = deepcopy(self.rocket.totalMass)
         totalMass.setDiscreteBasedOnModel(self.z)
-        # TODO: change calculation method to account for variable gravity
-        potentialEnergy = totalMass * self.env.gravity.compose(self.z)
+        potentialEnergy = totalMass * self.env.gravity.compose(self.z) * self.z
```
giovaniceotto, line 2124: I am afraid this formula, $U_g (z) = mg(z)z$ is no longer valid for variable gravity. Instead, we should use $U_g (R) = \frac{GMm}{R}$ where $R$ is the rocket's center of mass position vector magnitude relative to Earth's center.

MateusStano: So I implemented this in https://github.com/RocketPy-Team/RocketPy/pull/338/commits/90b2bdb1aa59d1128ed4c8955cd64de74297d791 with the following code:

```
        potentialEnergy = GM * totalMass / self.env.earthRadius * (1 / self.z - 1)
```

So what we are calculating here is the potential energy in relation to the sea level. All good now right?

MateusStano: This is the result for Calisto:


Figure 38

PR #343, opened 2023-03-23
BUG: Parachute lag warning when lag is 0

MateusStano:
## Pull request type

- [ x ] Code base additions (bugfix, features)

## What is the current behavior? What is the new behavior?

If there is a parachute with no lag (`lag=0`) in the rocket used in the flight simulation, a warning is raised. This is due to an uDot FlightPhase being added together with a uDotParachute phase when the parachute ejection is detected.

The fix here was just to add a check to see with the parachute lag is not zero. I do not know if this is the best solution though.

The warning in question is also raised when the parachute lag is negative, which I believe should not happen.

### DIFF
--- a/rocketpy/AeroSurfaces.py
+++ b/rocketpy/AeroSurfaces.py
@@ -11,27 +11,45 @@ from abc import ABC, abstractmethod, abstractproperty
 
 
 class AeroSurfaces:
-    """Class used to hold multiple aerodynamic surfaces and their positions."""
+    """Class used to hold one or more aerodynamic surfaces."""
 
     def __init__(self):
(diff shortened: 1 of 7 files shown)


MateusStano:
@giovaniceotto I need your review on this one. It is just a quick fix, just need you to check if this does not mess up with anything in the Flight class

giovaniceotto (changes requested):
My bad, the fin flutter test is failing... is this expected from your end @MateusStano?

It does seem like an unrelated issue, but we need to track down when it started to happen to be safe. It is also a small issue, we are getting values in the order of 1e-4, but the test requires values smaller than 1e-6.

PR #344, opened 2023-03-23
ENH: RailButtons Class

MateusStano:
## Pull request type

## Pull request checklist

## Changes

- Created RailButtons Class In `AeroSurfaces.py`
- Changed `Rocket.railButtons` for `Rocket.rail_buttons`

The only question I have here is, the class I created receives in the init the `upper_button_position` and `lower_button_position` arguments. The `setRailButtons` method receives a `position` argument that is a list. Should we change this to the upper and lower button arguments?

The only advantage there is to using a list is that you don't need to specify an order, which can be more accessible in some cases given the rocket coordinate system.


rocketpy/AeroSurfaces.py
```diff
@@ -1918,3 +1916,55 @@ def allInfo(self):
         self.aerodynamicInfo()
 
         return None
+
+
+class RailButtons:
```
giovaniceotto, line 1921: It seems weird to me to place the RailButton class inside the AeroSurfaces modules. How are these related?

rocketpy/Rocket.py
```diff
@@ -209,14 +209,12 @@ def __init__(
...
 
-        # Rail button data initialization
-        self.railButtons = None
-
         # Aerodynamic data initialization
         self.aerodynamicSurfaces = AeroSurfaces()
         self.nosecone = AeroSurfaces()
         self.fins = AeroSurfaces()
         self.tail = AeroSurfaces()
+        self.rail_buttons = AeroSurfaces()
```
giovaniceotto, line 217: Sorry for the question, but why, again, is everything here AeroSurfaces?

From what I understand, #341 modified the AeroSurfaces class to make the position argument optional... which as far as I know literally makes the AeroSurfaces class degenerate back to the list class when no positional argument is used.  So I am not sure it really make sense to me, in terms of code architecture, to have position argument as optional. When we don't need it, we can just use a list, instead of the AeroSurfaces class.

Furthermore, how are rail_buttons related to aerodynamic surfaces?

rocketpy/AeroSurfaces.py
```diff
@@ -1918,3 +1916,55 @@ def allInfo(self):
...
+
+    def __init__(self, upper_button_position, lower_button_position, angular_position):
+        """Initializes RailButtons Class.
+
+        Parameters
+        ----------
+        upper_button_position : int, float
+            Position of the upper rail button in meters. The upper button is the one
+            closest to the nose cone and furthest from the tail. The coordinate system
+            used is the same as the Rocket that the buttons will be a part of.
```
giovaniceotto, line 1948: Using "the coordinate system used is the same as the Rocket..." is dangerous. As a good practice, we want all components that can be added to a Rocket to be rocket independent, which means it cannot make any assumptions regarding the rocket's coordinate system.

By rocket independent, I mean, we should be able to use any component, in this case, a rail button pair, in any rocket, regardless of the coordinate system, with it working flawlessly.

I believe the current implementation does not satisfy this pillar, which was followed rigorously during the development of the new motor classes.

giovaniceotto (changes requested):
When I started reviewing, I really liked the idea... but then I stumbled upon two issues:

- I don't see the relationship between rail buttons and the `AeroSurfaces` class. Rail buttons have no aerodynamic properties in RocketPy. Furthermore, the changes made to the `AeroSurfaces` class in #341 seem strange to me... it can be two very different things now, and it is not clear when which is which.
- The current implementation violates a key pillar of modularity in terms of rocket components. Just as we did with the new motor classes in #233, we need to guarantee that a component, be it a motor, a nose cone or a rail button, will work the same irrespective of the rocket that we add it to. This is the reason that we had to create a motor coordinate system, to guarantee that a motor can be completely defined without making any assumptions about the rocket that it will be used in. Then, in the `addMotor` method, we specify the missing link that allows for the conversion of all information needed to the rocket system. Notice how the `NoseCone` class has no position argument... it is specified only in the `addNoseCone` method. But in the case of rail buttons, its proposed class pretty much has 3 position arguments. It just doesn't seem like a promising idea to implement it as a separate entity like this, but I would love to be proven wrong.

To conclude, this PR adds 77 lines of code, while only removing 32. It seems it uses more lines of code to accomplish the same thing that was already implemented.

rocketpy/Rocket.py
MateusStano: The AeroSurface class has been changed to act as a list, if there is more then one item in `AeroSurfaces._aerosurfaces`, and act as just the first item of `AeroSurfaces._aerosurfaces` if not.

Example:
If you create a rocket and then add a nose cone to it, you can access the NoseCone object in a few ways:

- With `Rocket.aerodynamicSurface[0]`
- Or `Rocket.nosecone`

If you then add another nosecone and a finset, the  `Rocket.aerodynamicSurface` will be a list with both NoseCones and Fins, while `Rocket.nosecone` will be a list with both NoseCone objects. So to access the first added nosecone now, you can do:

- With `Rocket.aerodynamicSurface[0]`
- Or `Rocket.nosecone[0]`

Essentially what the `AeroSurfaces class` is doing is acting as a list when it needs to, and acting as just an object when a list is not needed.

All of this makes accessing some information in the rocket class much easier and it really makes things easier in the Dispersion Class

> From what I understand, #341 modified the AeroSurfaces class to make the position argument optional

We use the position argument only when appending to the `Rocket.aerodynamicSurfaces`, because it is the only time where it is useful

MateusStano:
> * I don't see the relationship between rail buttons and the `AeroSurfaces` class. Rail buttons have no aerodynamic properties in RocketPy. Furthermore, the changes made to the `AeroSurfaces` class in #341 seem strange to me... it can be two very different things now, and it is not clear when which is which.

Okay, so the reason this change was made was to fit the new Dispersion Class. The way we handle aerodynamic surfaces in the monte carlo simulation is:

- Create the AeroSurface object
- Use this object to create the pydantic DataClass object that also holds the standard deviation
- Add the DataClass object to the Dispersion Rocket

You can check this process in this notebook: https://github.com/RocketPy-Team/RocketPy/blob/6142519d26f8fccb70e0a1cea018395367a69709/docs/notebooks/dispersion_analysis/dispersion_class_usage.ipynb
It is a bit messy still but I think it is good enough to understand the process.

The old RailButton implementation was really hard to deal with in the dispersion class because it completely broke the pattern that worked for all the aerosurfaces. The solution was then to just make a RailButtons class, but to keep the implementation in the Rocket class exactly the same as it was with the named tuple.

Writing the RailButton class inside `AeroSurfaces.py` was just intended as a "future proof" implementation. When we eventually add drag calculations, the buttons will need to be treated as an Aerodynamic surface (or at least that is the idea for now). Currently the RailButton class and the AeroSurface class really have no relation. Also the intended way to add railbutton to the rocket is just the same as before, you are not meant to create a Rail Button object, just to retrieve it from the `setRailButtons` method

Gui-FernandesBR:
I couldn't read all the thread yet, but I already can say that I agree with creating a class to Rail Buttons. 
Currently they are a named tuple, which does not follow any other example in the code.
If parachutes, noses, tails, and fins are saved inside classes, the buttons should also be.
An extra benefits is that it facilitates quite a lot when dealing with the dispersion class.
If someday we decide to compute Cd of each individual part, having these little guys isolated in the could may also help us.

I agree when Giovani complains about being rocket independent. In that case, I think the buttons probably should be treated in the same way as the motors (using 1 cys to the object and other for the rocket).

rocketpy/AeroSurfaces.py
phmbressan: I agree, referencing the Rocket coordinate system will likely lead to the user going back through the Rocket coordinate system documentation only to discover that this position is as simple as the distance to the nozzle (or nosetip) in most cases.

rocketpy/Rocket.py
```diff
@@ -787,7 +785,7 @@ def addParachute(
         # Return self
         return self.parachutes[-1]
 
-    def setRailButtons(self, position, angularPosition=45):
+    def setRailButtons(self, position, angular_position=45):
```
phmbressan, line 788: Due to the fact that the term `position` is used generally in RocketPy as a single value `float`, I would change this parameter's name to something more descriptive, such as `pair_position` or `rb_position`.

rocketpy/Rocket.py
phmbressan: I am having quite a hard timing trying to understand all the different ways that `positions` are being handled:

- Why does all of the aerosurfaces (like `Nosecone`, `Fins` etc) have an attribute `self.position = None` that is not documented? Who sets or uses that value?
- What is the difference between this parameter and the `position` parameter that is appended to the `AeroSurfaces` list? If they are not that same, they must not have the same name.

phmbressan (commented):
I agree with the change of `RailButtons` from a `namedtuple` to its own `class`, since it gives more flexibility. The main issue that may be reconsidered to increase code parallelism is making the coordinate system of the rail buttons rocket independent.

There is a more general issue around the documentation of the `position` parameter in the `AeroSurfaces` class and its components, but this might be discussion for another PR.

rocketpy/AeroSurfaces.py
MateusStano: I don't know if this is a valid issue since rail buttons are a bit different from motors and aerosurfaces in the sense that they are not added to a rocket, but set to one. There is no method that adds a rail button based on a RailButton object, like there is with `addSurfaces` and `addMotor`.

What could be done is add geometrical params to the `RailButtons` class init, but they would not be used in any calculations. They would be eventually useful whenever drag calculations are added, so implementing this would be future-proofing some break changes (but maybe adding drag might cause breaking changes anyway)

MateusStano:
Hey, with the new `AeroSurface` done, I adapted the `RailButtons` class and all tests are passing but two questions remain:

- In the class documentation it is mentioned that the `upper_rail_position` and `lower_rail_position` use the same coordinate system as the Rocket that the buttons will be a part of. This is different from what we did with the `aerosurfaces` and `motors`. However, `RailButtons` are different because they are not "added" to the rocket, but "set". There is no method to set the rail buttons given just a RailButton object. Should this be different?

- The `RailButtons` class does not inherit the `AeroSuface` class like it was planned. This is done because all the `@abstracmethod`'s would have to be redefined to return nothing since they do not generate any lift. When drag is added, the inheritance should be added though. What could be done is implement geometrical parameters to the `RailButtons` class and to the `setRailButtons` method of the `Rocket` class. This would avoid future breaking changes. Is it worth it to do this now?

Gui-FernandesBR:
> Hey, with the new `AeroSurface` done, I adapted the `RailButtons` class and all tests are passing but two questions remain:
> 
> * In the class documentation it is mentioned that the `upper_rail_position` and `lower_rail_position` use the same coordinate system as the Rocket that the buttons will be a part of. This is different from what we did with the `aerosurfaces` and `motors`. However, `RailButtons` are different because they are not "added" to the rocket, but "set". There is no method to set the rail buttons given just a RailButton object. Should this be different?

I think it is worth it to have rail buttons being defined and treated the same way we are already doing with the other components. My suggestions is to just ask the `distance_between_buttons` as an argument of rail buttons class. This would allow the user to add this buttons to any rocket later. Is that functionality something really useful? Well, perhaps not, but at least we keep consistency with the other components.

In the Rocket class, the "position" that is requested when adding a new component would make the same job in the buttons case: you just ask what is the position of one button (for example, the lowest button in the rocket) and then calculate the position of the second one.

> * The `RailButtons` class does not inherit the `AeroSuface` class like it was planned. This is done because all the `@abstracmethod`'s would have to be redefined to return nothing since they do not generate any lift. When drag is added, the inheritance should be added though. What could be done is implement geometrical parameters to the `RailButtons` class `__init__` and change the `setRailButtons` method of the `Rocket` class. This would avoid future breaking changes. Is it worth it to do this now?

What do expect to add as "geometrical parameters"? In my opinion, we would have to require only the Cd and the frontal area of the rail button if we wanted to calculate its drag force. In the very particular case where the user do not know the Cd of the rail button, we could either assume a pre-defined Cd or estimate its values based on geometric parameters.

But I have another question too:
- Should we really proceed with the "rail button" name? Launch lugs are also an option when designing a rocket, and perhaps we could bring a more flexible name here. I'm confortable with rail buttons if we don't have another suggestion tho, but try to think about that please.

MateusStano:
All done! Ready for a quick re-review @Gui-FernandesBR

rocketpy/AeroSurface.py
```diff
@@ -1892,53 +1885,113 @@ def allInfo(self):
...
+
+        Returns
+        -------
+        None
+        """
         return None
 
-    def __repr__(self):
-        rep = f"Rail buttons pair at positions {self.upper_button_position} m and {self.lower_button_position} m"
-        return rep
```
phmbressan: It might be a commentary regarding `__repr__` more generally for the AeroSurfaces PR. However, why to remove the `__repr__` (and `__str__`)? I understand that there is already plenty of info in `allInfo`, however the `__repr__` is quite important to make sense of the object in a compact way, for instance mentioning the `self.name` attribute.

rocketpy/AeroSurface.py
```diff
@@ -1892,53 +1885,113 @@ def allInfo(self):
         return None
 
 
-class RailButtons:
+class RailButtons(AeroSurface):
```
phmbressan, line 1888: Since the class represents a pair of RailButtons, it is fine to have its name pluralized.

rocketpy/AeroSurface.py
```diff
@@ -1892,53 +1885,113 @@ def allInfo(self):
...
+            Name of the rail buttons. Defaul is "Rail Buttons".
 
         Returns
         -------
         None
 
         """
-        self.upper_button_position = upper_button_position
-        self.lower_button_position = lower_button_position
+        self.buttons_distance = buttons_distance
```
phmbressan, line 1920: I remember the distance vs positions PR discussions. Since the class already includes a pair of railbuttons and their distance is working here more like an mathematical value than a core attribute, I will not object too much.

Even thought it would increase code parallelism, it might be a bit "too much" to define another coordinate system here.

phmbressan (approved):

rocketpy/AeroSurface.py
```diff
@@ -1892,3 +1883,115 @@ def allInfo(self):
         self.aerodynamicInfo()
 
         return None
+
+
+class RailButtons(AeroSurface):
+    """Class that defines a generic rail button.
```
Gui-FernandesBR, line 1889: - What do you mean by generic?
- I would complete with the word "pair" here.
- My final suggestion: "Class to define a rail button pair."

rocketpy/AeroSurface.py
```diff
@@ -1892,3 +1883,115 @@ def allInfo(self):
...
+
+    Attributes
+    ----------
+    RailButtons.buttons_distance : int, float
+        Distance between the highest and the lowest rail button in meters.
```
Gui-FernandesBR, line 1894: Highest means "the one with the greatest vertical extent". 
I do not think that you wanted to mean the height of the rail button, but that is what it sounds.

rocketpy/Rocket.py
```diff
@@ -477,6 +482,11 @@ def addSurfaces(self, surfaces, positions):
             regarding the rocket's coordinate system.
             If a list is passed, it will correspond to the position of each item
             in the surfaces list.
+            For NoseCone type, position is relative to the nose cone tip.
+            For Fins type, position is relative to the point belonging to
+            the root chord which is highest in the rocket coordinate system.
```
Gui-FernandesBR, line 487: Does the word "highest" here consider that the coordinate system can be defined as "nozzle-to-ogive" or "ogive-to-nozzle" ?

rocketpy/AeroSurface.py
```diff
@@ -1892,3 +1883,115 @@ def allInfo(self):
...
+        self.clalpha = Function(
+            lambda mach: 0,
+            "Mach",
+            f"Lift coefficient derivative for {self.name}",
+        )
+        self.cl = Function(
+            lambda alpha, mach: 0,
+            ["Alpha (rad)", "Mach"],
+            "Cl",
+        )
```
Gui-FernandesBR, line 1959: Shouldn't we return None here? Just for keeping the consistency...

rocketpy/Rocket.py
```diff
@@ -809,23 +819,25 @@ def addParachute(
...
 
         Parameters
         ----------
-        position : tuple, list, array
-            Two values organized in a tuple, list or array which
-            represent the position of each of the two rail buttons
-            in the rocket coordinate system
-            The order does not matter. All values should be in meters.
+        upper_button_position : int, float
+            Position of the rail button furtherst from the nozzle relative to
```
Gui-FernandesBR, line 833: Well done!

rocketpy/Rocket.py
```diff
@@ -834,15 +846,16 @@ def setRailButtons(self, position, angularPosition=45):
...
         """
-        # Place top most rail button as the first element of the list
-        if self._csys * position[0] < self._csys * position[1]:
-            position.reverse()
-        # Save important attributes
-        self.railButtons = self.railButtonPair(position, angularPosition)
-
-        return None
+        # Create a rail buttons object
+        buttons_distance = abs(upper_button_position - lower_button_position)
```
Gui-FernandesBR, line 853: Great!

Gui-FernandesBR:
I hereby confirm that tests are working on my computer,

@MateusStano , I just left a few comments to you regarding docstrings. That should be very easy to solve tho.
You have my approvement after fixing such small details.

Great one!

rocketpy/Rocket.py
MateusStano: Yes, the csys variation is already being considered

Issue #152, opened 2022-03-06
Keep it Simple, Stupid: Simpler Parachute Triggers


giovaniceotto:
## Is your feature request related to a problem? Please describe.

RocketPy supports parachute trigger functions, which can be quite nice to test ejection activation algorithms. **However, when running simple simulations with parachutes, having to code a trigger function can be an unnecessary overhead.** 

## Describe the solution you'd like

I would like the `trigger` argument of the `Flight.addParachute` method to support, besides a callable, two other types of input:

- a string `'apogee'`, in which case, the parachute would be ejected at apogee (intuitive I believe)
- a numeric value, such as `1400`, in which case the parachute would be ejected at an altitude (above sea level, ASL) corresponding to the numeric value in meters.

Here is a code snippet showcasing both:

```python
Main = Calisto.addParachute(
    "Main",
    CdS=10.0,
    trigger=800,
    samplingRate=105,
    lag=1.5,
    noise=(0, 8.3, 0.5),
)

Drogue = Calisto.addParachute(
    "Drogue",
    CdS=1.0,
    trigger='apogee',
    samplingRate=105,
    lag=1.5,
    noise=(0, 8.3, 0.5),
)
```

## Implementation details

Internally, the `Rocket.addParachute` method may convert both types of new inputs into trigger functions (callables) which will continue to work with the Flight Class without the need of modifying it.

## Additional context

Such a feature would facilitate the creation of parachutes, making RocketPy easier to use. This is particularly important for inexperienced users. Furthermore, the advanced feature of using a trigger function would keep working exactly as it does today. 

It is also important to note that this idea originated when using RocketPy in MATLAB. Currently, a function needs to be defined in Python inside MATLAB to create the triggers, which is awfully bad for user experience, since it can only be done with lambdas. Therefore, such a feature would also greatly enhance the experience for MATLAB users.

Gui-FernandesBR:
Not finished yet

MateusStano:
All done in #345

PR #345, opened 2023-03-30
Enh/simpler parachute triggers

FranzYuri:
## Pull request type

## Pull request checklist

Trigger must be a complex function.

## What is the new behavior?

Now the Trigger may be an altitude or the string "apogee"

- [ ] Yes
- [X] No

## Other information

Enter text here...


giovaniceotto (changes requested):
To be honest, I believe adding the `h` argument to the trigger function does more harm than good.

- It changes a lot of code lines in the code base.
- It breaks compatibility with today's code.
- It is redundant information.
- The only use case I see is to make it easier to create a trigger function based on altitude... but since the altitude can already be inputted directly into the trigger argument in the parachute class, I do not see when this is useful. 

Thus, I recommend removing it. I am open to further discussions with you guys do not agree with my points.

MateusStano:
For some reason git is saying there was 20.000 addition in the deployable payload notebook. I do not know why this is happening. The notebook is the same in beta/v1.0.0 branch

MateusStano:
> * The only use case I see is to make it easier to create a trigger function based on altitude... but since the altitude can already be inputted directly into the trigger argument in the parachute class, I do not see when this is useful.

This is needed because the altitude that comes in the `y` array is ASL. What we were trying to do here was make the `trigger` argument be able to receive a float or int that represents the height AGL that the parachute will open. The only way to convert the altitude from the `y` array into AGL would be by getting the `env.elevation` somehow. Another thing is that this `h` argument in the trigger function also takes into account the `noise` from the parachute, which was only done for `pressure` before.

I thought of a few ways to do this and thought this one was actually the less complex and less redundant one. Do you have any suggestions?

giovaniceotto:
> This is needed because the altitude that comes in the `y` array is ASL. What we were trying to do here was make the `trigger` argument be able to receive a float or int that represents the height AGL that the parachute will open. The only way to convert the altitude from the `y` array into AGL would be by getting the `env.elevation` somehow. Another thing is that this `h` argument in the trigger function also takes into account the `noise` from the parachute, which was only done for `pressure` before.

- Both the user, and the Flight class, have easy access to `env.elevation`, so I do not see that as a major reason for having to use `h`.
- Having noise added to the signal is another story... It is definitely interesting, but generally, altitude is sensed by using a barometric pressure sensor. So, the right path would be: getting the pressure with noise, and then converting it to altitude by using the inverse of `Env.pressure`.

> I thought of a few ways to do this and thought this one was actually the less complex and less redundant one. Do you have any suggestions?

Ok... I have to be honest and say that I couldn't think of anything perfect for now, but here is a quick idea I had:

```python
from rocketpy import trigger_decorators

@trigger_decorators.altitude("z", noise=..., latency=..., type="AGL")
@trigger_decorators.acceleration("a", noise=..., latency=..., reference="body")
@trigger_decorators.pressure("p", noise=..., latency=...)
def main_trigger(_, p, a, z):
    # p is pressure
    # a is an acceleration array (ax, ay, az)
    # z is AGL
    return f(p, a z)
```

For simple trigger methods, only one or two decorators would be used at a time. The way this would work under the hood is: all trigger functions must actually receive the `Flight` object itself, which has access to all information available, which means all that the Flight class needs to do is call the trigger function passing itself as the argument.

What the decorators do is get this Flight object and convert it into what the user wants and then forwards this to the user defined trigger. This is really nice since it makes it possible to only compute what is necessary. For example: computing acceleration is expensive (requires running uDot again), so it would only be calculated when needed. It is also very generic.

I am not saying we need to implement this in this PR, but I would like to know what do you think about the idea, @FranzYuri, @MateusStano?

FranzYuri:
Hey guys! another, simpler idea that allows a trigger that only needs AGL altitude would be to delay the creation of the trigger function itself to the Flight constructor.
image
 So if the user inputs a float/int, the function would not be created, instead self.trigger would return just a number.

image
then, in the Flight constructor the trigger function is created.

Gui-FernandesBR:
I liked your last suggestions @FranzYuri . 
I was reading the description of the issue #152 just to remind the requirements of this implementation.
Here are 3 main aspects that we should keep in mind: 

* The trigger argument of the Flight.addParachute method to support, besides a callable, two other types of input:
    * a string 'apogee', in which case, the parachute would be ejected at apogee (intuitive I believe)
    * a numeric value, such as 1400, in which case the parachute would be ejected at an altitude (*above sea level, ASL*) corresponding to the numeric value in meters.
* Internally, the Rocket.addParachute method may convert both types of new inputs into trigger functions (callables) which will continue to work with the Flight Class without the need of modifying it.

The current implementation violates the last point, since we can see major differences in the Flight.py file.
The idea here is to focus on creating a callable from a shorter information passed by the user.
I agree that we can keep ASL altitude as a standard if needed, it is not a big deal to the user. 
Therefore, what you suggested in the second figure could actually be done in the first figure without any problems.

The idea of decorators presented by @giovaniceotto I honestly didn't understand. But I don't think we need to go too far for that.
 
Let me know in case of any troubles when solving this PR.

MateusStano:
Okay, so the decorator idea is a bit complex but could work well to do some complex things and maybe could be something we put more thought into in the future. However, it goes against the issue that this PR is fixing, which is to make things simpler

I think Franz suggestion is good but I don't like that it defines a parachute-related variable in the Flight class.



> So, the right path would be: getting the pressure with noise, and then converting it to altitude by using the inverse of `Env.pressure`.

This is essentially what the current code is doing. Only thing is that I am not calling `inverseFunction` directly because that requires an approx function that can be a little tricky to use. The code is using `findInput` to calculate the inverse values directly

> * Both the user, and the Flight class, have easy access to `env.elevation`, so I do not see that as a major reason for having to use `h`.

The `h` is different from all inputs in the trigger function:

- `p` is the noisy pressure signal
- `h` is the "noisy" AGL height value 
- `y` is the state vector in which `y[2]` is the rocket altitude ASL

Each of these represents distinct things. 

Could the user calculate the `h` just from the `p` and `env`? Yes.
Would it be simple/intuitive/useful to make the user calculate this by itself? No

I do not think that adding `h` to the trigger function creates any problems to the codes efficiency and just adds to the usability.
The only thing here is that it changes a lot of lines in a lot of files, and to me that is not an issue if the change is justfied

> * a numeric value, such as 1400, in which case the parachute would be ejected at an altitude (_above sea level, ASL_) corresponding to the numeric value in meters.

We could just leave it as ASL. But I just think that it does not make things simpler this way since most rocketeers will design their parachute systems to be based on AGL

MateusStano:
This one is almost done. I still need to add a test for these implementations

PR #351, opened 2023-04-11
BUG: `*` instead of `-` in __sub__

MateusStano:
Yup


giovaniceotto (approved):
My fault, had already fixed in other branches.

PR #353, opened 2023-04-19
Function Comparison and Identity Map

phmbressan:
## Pull request type

## Pull request checklist

The Function class currently lacks basic methods of comparison and identity map.

## What is the new behavior?

The comparison works in a very similar way to comparing arrays in numpy. When this comparison cannot be performed, an error is raised. The identity map is basically an implementation of the function $f(x) = x$ that follows the same discretization as the instance.

- [ ] Yes
- [X] No

### DIFF
--- a/rocketpy/Function.py
+++ b/rocketpy/Function.py
@@ -631,8 +631,8 @@ class Function:
             Zs = np.array(self.getValue(mesh))
             self.setSource(np.concatenate(([Xs], [Ys], [Zs])).transpose())
 
-        self.setInterpolation(modelFunction.__interpolation__)
-        self.setExtrapolation(modelFunction.__extrapolation__)
+        self.setInterpolation(self.__interpolation__)
+        self.setExtrapolation(self.__extrapolation__)
         return self
 
     def reset(
@@ -1892,12 +1892,12 @@ class Function:
         ----------
         other : Function, int, float, callable
             What self will be added to. If other and self are Function
-            objects which are based on interpolation, have the exact same
-            domain (are defined in the same grid points), have the same
-            interpolation method and have the same input name, then a
-            special implementation is used. This implementation is faster,
-            however behavior between grid points is only interpolated,
-            not calculated as it would be.
+            objects which are based on a list of points, have the exact same
+            domain (are defined in the same grid points) and have the same
+            dimension, then a special implementation is used.
+            This implementation is faster, however behavior between grid
+            points is only interpolated, not calculated as it would be;
+            the resultant Function has the same interpolation as self.
 
         Returns
         -------
@@ -1911,8 +1911,7 @@ class Function:
             if (
                 isinstance(other.source, np.ndarray)
                 and isinstance(self.source, np.ndarray)
-                and self.__interpolation__ == other.__interpolation__
-                and self.__inputs__ == other.__inputs__
+                and self.__domDim__ == other.__domDim__
                 and np.array_equal(self.xArray, other.xArray)
             ):
                 # Operate on grid values
@@ -1976,12 +1975,12 @@ class Function:
         ----------
         other : Function, int, float, callable
             What self will be subtracted by. If other and self are Function
-            objects which are based on interpolation, have the exact same
-            domain (are defined in the same grid points), have the same
-            interpolation method and have the same input name, then a
-            special implementation is used. This implementation is faster,
-            however behavior between grid points is only interpolated,
-            not calculated as it would be.
+            objects which are based on a list of points, have the exact same
+            domain (are defined in the same grid points) and have the same
+            dimension, then a special implementation is used.
+            This implementation is faster, however behavior between grid
+            points is only interpolated, not calculated as it would be;
+            the resultant Function has the same interpolation as self.
 
         Returns
         -------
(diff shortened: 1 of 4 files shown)


phmbressan:
It seems that an atmospheric file is not available due to server clutter, it will likely return shortly.

rocketpy/Function.py
```diff
@@ -2118,77 +2195,78 @@ def __sub__(self, other):
                 and isinstance(self.source, np.ndarray)
                 and self.__interpolation__ == other.__interpolation__
```
phmbressan, line 2196: The git diff is a bit weird due to staging, so don't base the changes too much from it. Are these checks really needed in order to perform the operations?

- The `self.__inputs__` I believe is made to ensure unit consistency. However, that only seems reasonable if there were a `unit` parameter for the `Function`, since the `__inputs__` might change or have slightly different names while being the same;

- The interpolation check makes some sense to infer the interpolation of the result, but that seems quite restrictive, wouldn't it be more interesting to not check this and keep the `self` interpolation?

rocketpy/Function.py
```diff
@@ -2116,79 +2227,96 @@ def __sub__(self, other):
             if (
                 isinstance(other.source, np.ndarray)
                 and isinstance(self.source, np.ndarray)
-                and self.__interpolation__ == other.__interpolation__
-                and self.__inputs__ == other.__inputs__
+                and self.__domDim__ == other.__domDim__
+                and np.any(self.xArray - other.xArray) == False
```
phmbressan, line 2231: Is this line really needed? It seems redundant with the line below that checks that the arrays are equivalent.

rocketpy/Function.py
```diff
@@ -600,8 +635,8 @@ def setDiscreteBasedOnModel(self, modelFunction, oneByOne=True):
             Zs = np.array(self.getValue(mesh))
             self.setSource(np.concatenate(([Xs], [Ys], [Zs])).transpose())
 
-        self.setInterpolation(modelFunction.__interpolation__)
-        self.setExtrapolation(modelFunction.__extrapolation__)
+        self.setInterpolation(self.__interpolation__)
+        self.setExtrapolation(self.__extrapolation__)
```
giovaniceotto, line 639: I don't remember discussing this, but I may have forgotten.

From what I see, we have to different behaviors:
 - Keeping interpolation and extrapolation unchanged when discretizing.
 - Using the interpolation and extrapolation of the model Function.

Could you point out the pros and cons of each? How would the first option work in case self originally has a callable source?

Feel free to keep the proposed solution if you find it the best option. However, make sure to document this behavior.

rocketpy/Function.py
```diff
@@ -1676,27 +1713,200 @@ def __interpolateAkima__(self):
...
+        -------
+        numpy.ndarray of bool
+            The result of the comparison one by one.
+        """
+        otherIsFunction = isinstance(other, Function)
+
+        if isinstance(self.source, np.ndarray):
+            if otherIsFunction:
+                try:
+                    return self.yArray >= other.yArray
```
giovaniceotto, line 1768: Since we are not checking if `self.xArray` and `other.xArray` are the same, it seems that if `len(self.xArray) == len(other.yArray)`, then an erroneous comparison will be made without raising any issues.

Is this behavior expected?

This extends to the remaining comparison methods.

rocketpy/Function.py
giovaniceotto: Not needed.

giovaniceotto (changes requested):
Just two final comments:

- Like `min` and `max` are cached properties, perhaps so should `xArray`, `yArray`, `xinitial`, `xfinal`, etc be.
- It worries me that some new methods, such as `min` and `max`, have no handling of 2D Functions. They are rarely used for now tho, so this is not a deal breaker.

rocketpy/Function.py
phmbressan: I admit that this behavior stretches quite a bit the intuitive side of a comparison, but I don't think that raising an error is necessary here. Even if the domains are different, the `yData` is there to be compared.

giovaniceotto:
Here is a to-do list before merging this PR:

- [ ] Resolving conflicts
- [ ] Resolve keeping self vs other interpolation and extrapolation properties
- [ ] Resolve making `xArray`, `yArray`, `xinitial`, `xfinal`, etc as cached properties

The last two are suggestions, thus optional. @phmbressan let me know if you prefer to skip them.

rocketpy/Function.py
phmbressan: Both ways are possible, however, during coding with this method, I found that keeping the `self` properties is usually more useful:

- In general, the Model Function was used only to ensure the same discretization as other in order to perform arithmetic operations. Therefore, the main point was to make the `xArray` equally discretized;
- Due the fact that the arithmetic operations now no longer require the same interpolation for array computations, it seems unnecessary to force `self` into the same interpolation of the model Function. That may cause issues if an "exact" extrapolation (such as linear (i.e. natural)) is substituted by an "inaccurate" extrapolation (such as polynomial)".

As an example of the latter, the IdentityMap is based on accurate linear extrapolation, is one discretizes it based on a model Function of another kind (e.g. something really common) great inaccuracies may appear.

MateusStano (approved):

PR #355, opened 2023-04-27
ENH: AeroSurfaces(ABC) and Improvements to Rocket.aerodynamicSurfaces

MateusStano:
## Pull request checklist

`class AeroSurfaces` used to be a weird tuple-list hybrid. This was inefficient code-wise, especially in the Dispersion class and simulation.

## What is the new behavior?
`class AeroSurfaces` is now an Abstract Class. 

In `class Rocket` the following changes were made:

- `Rocket.aerodynamicSurfaces` is now a list
- `Rocket.nosecone`, `Rocket.fins`, `Rocket.tail` are now lists. Each of them has a getter that checks if the list has only one item, if so returns that item. If the list has more than one item, it returns the complete list. This is just a way to easily access each of the possible aero surfaces attributes. It is not used anywhere in the simulation.
- Added `addSurface` method.
- Added `addSurfaces` method.

- [ ] Yes
- [x] No

## Other information

I am up to debate the implementation of both `addSurface` and `addSurfaces`. Another possibility in this implementation would be to only have one method (`addSurface`) that can also accept lists as inputs for the necessary arguments. 

I prefer having two functions that do two clearly distinct things. Again, I am up for debate on this

Also, this pull request is meant to be merged before #344


rocketpy/AeroSurfaces.py
```diff
@@ -329,7 +352,7 @@ def allInfo(self):
         return None
 
 
-class Fins(ABC):
+class Fins(AeroSurfaces, ABC):
```
phmbressan, line 355: I might be wrong, but, since `AeroSurfaces` is an Abstract Class, its child already inherits `ABC`.

```suggestion
class Fins(AeroSurfaces):
```

rocketpy/Rocket.py
```diff
@@ -449,6 +442,59 @@ def addMotor(self, motor, position):
...
+        if isinstance(surface, NoseCone):
+            self.nosecone.append(surface)
+        elif isinstance(surface, (TrapezoidalFins, EllipticalFins)):
+            self.fins.append(surface)
+        elif isinstance(surface, Tail):
+            self.tail.append(surface)
+        self.evaluateStaticMargin()
+        return None
+
+    def addSurfaces(self, surfaces, positions):
```
phmbressan, line 474: Raising the discussion referenced in the PR description, personally I believe that these two methods have too similar names and functionalities. It does not seem _pythonic_ to slightly differentiate methods naming if the parameter is a single item or a list, the method `addSurface` should take care of both cases, maybe with an `try-except` clause or `if-else`.

rocketpy/Rocket.py
```diff
@@ -449,6 +442,59 @@ def addMotor(self, motor, position):
...
+        None
+        """
+        self.aerodynamicSurfaces.append(surface)
+        surface.position = position
+        if isinstance(surface, NoseCone):
+            self.nosecone.append(surface)
+        elif isinstance(surface, (TrapezoidalFins, EllipticalFins)):
+            self.fins.append(surface)
+        elif isinstance(surface, Tail):
+            self.tail.append(surface)
```
phmbressan, line 470: This is fine, but maybe the objects could sort themselves in the background and not having a two instances in different lists. Let me know what you think of this approach:

1 - Make an attribute of `AeroSurfaces` that is a dictionary

```
{
"Nosecone": [],
"Tail": [],
...
}
```

2 - Use the `__repr__` and `__str__` methods of the objects so as to differentiate the types. That way the `addSurface` method is something along the lines:

```
self.aerodynamicSurfaces[str(surface)].append(surface)
```

The `str(surface)` could be an slice of string or an even an `surface.kind`. I understand that this might be trying to reinvent the wheel to turn the code more polymorphic, so balance the drawbacks on what you think is the best way.

phmbressan (changes requested):
Overall the code is much better and easier to comprehend the relationship between the `AeroSurfaces` and its child classes, nice work. I have just pointed out a few changes that my be considered.

MateusStano:
@giovaniceotto @phmbressan the new `class Components` is done! this is ready for a re-review!

rocketpy/Rocket.py
```diff
@@ -262,6 +270,21 @@ def __init__(
 
         return None
 
+    @property
+    def nosecones(self):
+        self._nosecones = self.aerodynamicSurfaces.get_by_type(NoseCone)
```
Gui-FernandesBR, line 275: I really liked this "get_by_type" thing

Gui-FernandesBR:
> @giovaniceotto @phmbressan the new `class Components` is done! this is ready for a re-review!

I can also give a hand here in the re-review,
just wondering why are the tests not passing

rocketpy/Components.py
```diff
@@ -0,0 +1,59 @@
...
+        component_type_list = [
+            c.component
+            for c in self._components
+            if isinstance(c.component, component_type)
+        ]
+        return (
+            component_type_list[0]
+            if len(component_type_list) == 1
+            else component_type_list
+        )
```
phmbressan, line 42: I remember having already discussed it, this is better than the earlier approach but still haunts me a little the idea that the return type of this method if undefined. Could you summarize the main arguments favoring this implementation?

An idea could be to separate the method into `get_main_by_type` and `get_all_by_type` (with a docstring that the main component is defined as the first added) or `get_component` and `get_components`.

rocketpy/Rocket.py
```diff
@@ -442,56 +466,35 @@ def addMotor(self, motor, position):
...
         -------
         None
         """
-        for surface, position in zip(surfaces):
-            self.addSurface(surface, position)
-            surface.position = position
+        try:
+            for surface, position in zip(surfaces, positions):
+                self.aerodynamicSurfaces.add(surface, position)
+        except:
```
phmbressan, line 495: Try to pinpoint the exception type.

phmbressan (approved):
Great improvements since the first version of this PR. Please take a look at some comments that I made. Nice work.

rocketpy/Components.py
```diff
@@ -0,0 +1,59 @@
+__author__ = "Mateus Stano Junqueira"
+__copyright__ = "Copyright 20XX, RocketPy Team"
+__license__ = "MIT"
+
+from collections import namedtuple
+
+
+class Components:
+    """A group of components, each of which has a position"""
```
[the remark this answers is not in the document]
Gui-FernandesBR: Agreed.

reading fast and without talking to you, I couldn't understand what a component is

rocketpy/Rocket.py
phmbressan: So did I!

rocketpy/Components.py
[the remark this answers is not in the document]
MateusStano: addresed in https://github.com/RocketPy-Team/RocketPy/pull/355/commits/ee9fb3c28e277d3aa1ea24766e524713aa58ae6d

rocketpy/Rocket.py
```diff
@@ -212,10 +220,10 @@ def __init__(
...
 
         # Aerodynamic data initialization
-        self.aerodynamicSurfaces = []
-        self.nosecone = []
-        self.fins = []
-        self.tail = []
+        self.aerodynamicSurfaces = Components()
+        self._nosecones = []
+        self._fins = []
+        self._tails = []
```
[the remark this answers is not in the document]
MateusStano: Addressed in https://github.com/RocketPy-Team/RocketPy/pull/355/commits/378b1ebffe04ed8e4f2ef616f64df5e0c1fb6875

rocketpy/Rocket.py
```diff
@@ -261,6 +270,21 @@ def __init__(
...
 
+    @property
+    def nosecones(self):
+        self._nosecones = self.aerodynamicSurfaces.get_by_type(NoseCone)
+        return self._nosecones
+
+    @property
+    def fins(self):
+        self._fins = self.aerodynamicSurfaces.get_by_type(Fins)
+        return self._fins
```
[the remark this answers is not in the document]
MateusStano: Addressed in https://github.com/RocketPy-Team/RocketPy/pull/355/commits/378b1ebffe04ed8e4f2ef616f64df5e0c1fb6875

rocketpy/Components.py
MateusStano: Addresed in https://github.com/RocketPy-Team/RocketPy/pull/355/commits/49556de1b8118d089ce7071430c84370915de9ca

rocketpy/Rocket.py
MateusStano: Addresed in https://github.com/RocketPy-Team/RocketPy/pull/355/commits/1d8cb9cea1f5937cffff382ffcf92b6b52983120

rocketpy/Components.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def __init__(self):
+        self.component_position = namedtuple("component_tuple", "component position")
```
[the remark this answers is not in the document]
MateusStano: The namedtuple has two parameters: `component` and `position`. I changed the name of the variable to be `self.component_tuple` now

This actually comes from a different implementation of the Components class that I was trying out. Here there is no practical utility besides making it easier to understand and use (at least I think so). 

It could just be a tuple but I find a namedtuple to work great since it essentially works as a tuple (you can get the position of a component with index, e.g. `Calisto.aerodynamicSurfaces[0][0]` or with the attr name, e.g. `Callisto.aerodynamicSurfaces[0].position`)

My initial intention was that this namedtuple would be called just `component`, and the class `Components` would be a holder of `component`'s. I changed the name because this happened:

image

I just didn't like how the word `component` repeats itself.

I am okay with the changing the name or just removing the namedtuple altogether. What is you suggestion?

rocketpy/Rocket.py
```diff
@@ -212,10 +220,10 @@ def __init__(
         self.railButtons = None
 
         # Aerodynamic data initialization
-        self.aerodynamicSurfaces = []
-        self.nosecone = []
-        self.fins = []
-        self.tail = []
+        self.aerodynamicSurfaces = Components()
```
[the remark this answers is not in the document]
MateusStano: Hmmmm..... I don't like it, but it might be easier to understand....

@Gui-FernandesBR what do you think?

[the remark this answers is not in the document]
Gui-FernandesBR: I think `aerodynamicSurfaces` is more intuitive and give us the idea that we are dealing with the outer layer only, regardless mass, density or volume.

In other words...

"Component" reminds me of the nosecone itself, while "Surface" reminds me of its shell, which is what we are working with in this class.

rocketpy/Components.py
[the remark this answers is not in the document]
Gui-FernandesBR: Maybe you forgot to write a word or another, but I couldn't understand it clearly enough.

I will give you a suggestion slightly different, what do you think? (the ** indicates my changes)

>     """A class to hold components of the Rocket class. **Each component is an object
>     that is added to the rocket at a specific position relative to the rocket's
>     coordinate system origin**. Components can be added to the rocket using the
>     'add' methods. This class is currently used specifically for holding
>     aerodynamic surfaces.
> 
>     Attributes
>     ----------
>     _components : list
>         A list of named tuples representing **all the components and their positions relative to the rocket**.
>     component_tuple : namedtuple
>         A named tuple representing a component and its position within the rocket.
>     """

rocketpy/Components.py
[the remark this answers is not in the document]
Gui-FernandesBR: I think with the new "component_tuple" name we are already in a better situation than in the figure above.

Naming the tuples argument seems more convenient to me when reading.

rocketpy/Components.py
```diff
@@ -0,0 +1,125 @@
...
+        """Initialize an empty components list instance."""
+        self.component_tuple = namedtuple("component_tuple", "component position")
+        self._components = []
+
+    def __repr__(self):
+        """Return a string representation of the Components instance."""
+        return repr(self._components)
+
+    def __len__(self):
+        """Return the number of components in the collection."""
```
Gui-FernandesBR, line 33: In the class docstring you cite this as a list, but from now on you mention as a collection.

What is the point? Is it a list or a collection afterwards?

rocketpy/Components.py
```diff
@@ -0,0 +1,125 @@
...
+    def __iter__(self):
+        """Return an iterator over the components."""
+        return iter(self._components)
+
+    def add(self, component, position):
+        """Add a component to the collection.
+
+        Parameters
+        ----------
+            component: AeroSurface, Motor
```
Gui-FernandesBR, line 49: You defined in the class' docstring that "this is currently used only for aerodynamic surfaces".
1 - Well, should we modify the class docstring or the add method?
2 - What other objects could possible be components in the future?
    - Rail buttons for sure
    - Maybe parachutes should be too?
    - Any other?

Gui-FernandesBR: Btw these lines are out of identation. This will break the docs if we merge like this.

rocketpy/Components.py
```diff
@@ -0,0 +1,125 @@
...
+                coordinate system origin.
+
+        Returns
+        -------
+        None
+        """
+        self._components.append(self.component_tuple(component, position))
+
+    def get_by_type(self, component_type):
+        """Get the components of a specified given type.
```
Gui-FernandesBR, line 62: I would recommed """Search the components list (or colletions?) and return a list with all the components of a given type

rocketpy/Components.py
```diff
@@ -0,0 +1,125 @@
...
+            The component to be removed from the rocket.
+
+        Returns
+        --------
+        None
+        """
+        for index, comp in enumerate(self._components):
+            if comp.component == component:
+                self._components.pop(index)
+                break
```
Gui-FernandesBR, line 97: What would happen if the same component is added twice in the rocket's component list?

For instance, I can create a single fin and add it 3 times to the rocket. And if I want I can even set them at the very same position.

Shall we be worried about that?

rocketpy/Components.py
```diff
@@ -0,0 +1,125 @@
...
+    def __repr__(self):
+        """Return a string representation of the Components instance."""
+        return repr(self._components)
+
+    def __len__(self):
+        """Return the number of components in the collection."""
+        return len(self._components)
+
+    def __getitem__(self, index):
+        """Return the number of components in the collection."""
```
Gui-FernandesBR, line 37: This docstring doesn't seem right.

rocketpy/Components.py
Gui-FernandesBR: Suggestion in case you're tired and wants to move forward: just remove the break and let it search in the entire list XD

rocketpy/AeroSurfaces.py
```diff
@@ -2,60 +2,89 @@
...
+class AeroSurfaces(ABC):
+    """Abstract class used to define aerodynamic surfaces."""
 
-    def __init__(self):
-        self._aeroSurfaces = []
+    def __init__(self, name):
+        self.cpx = 0
+        self.cpy = 0
+        self.cpz = 0
+        self.position = None  # in relation to rocket
```
Gui-FernandesBR, line 21: Suggestion: "relative to the rocket"

rocketpy/AeroSurfaces.py
```diff
@@ -2,60 +2,89 @@
...
+        self.name = name
```
Gui-FernandesBR, line 22: really like the new place for the self.name definition. Great!

rocketpy/AeroSurfaces.py
```diff
@@ -234,20 +253,20 @@ def evaluateLiftCoefficient(self):
...
-        self.cl : Function
-            Function of the angle of attack (Alpha) and the mach number
-            (Mach) expressing the lift coefficient of the nose cone. The inputs
-            are the angle of attack (in radians) and the mach number.
-            The output is the lift coefficient of the nose cone.
+        None
         """
         # Calculate clalpha
-        self.clalpha = 2 * self.radiusRatio**2
+        self.clalpha = Function(
```
Gui-FernandesBR, line 259: I can understand the reason why the mach is being ignored in the first self.clalpha definition, but please keep in mind that a regular user/reader won't understand it so easily. 

Matbe adding a comment or TODO could make future lives better

rocketpy/AeroSurfaces.py
```diff
@@ -1901,7 +1867,7 @@ def aerodynamicInfo(self):
         print(f"\nAerodynamic Information of {self.name}")
         print("-------------------------------")
         print(f"Tail Center of Pressure Position in Local Coordinates: {self.cp} m")
-        print(f"Tail Lift Coefficient Slope: {self.clalpha:.3f} 1/rad")
+        print(f"Tail Lift Coefficient Slope: {self.clalpha(0):.3f} 1/rad")
```
Gui-FernandesBR, line 1870: Shall we specify that the coefficient slope is being taken in alpha==0 ? (I understand that the alpha is the angle of attack)

rocketpy/Rocket.py
```diff
@@ -13,8 +13,15 @@
 
 from .Function import Function
 from .Parachute import Parachute
-from .AeroSurfaces import AeroSurfaces
-from .AeroSurfaces import NoseCone, TrapezoidalFins, EllipticalFins, Tail
+from .AeroSurfaces import (
+    AeroSurfaces,
```
Gui-FernandesBR, line 17: AeroSurfaces currently not being used in the rest of the code.

Error in code or should we just remove this import?

rocketpy/Rocket.py
```diff
@@ -449,6 +460,38 @@ def addMotor(self, motor, position):
         self.evaluateStaticMargin()
         return None
 
+    def addSurfaces(self, surfaces, positions):
+        """Adds one or more aerodynamic surfaces to the rocket. The aerodynamic
+        surface must be an instance of a class that inherits from the
+        AerodynamicSurface
```
Gui-FernandesBR, line 466:
```suggestion
        AerodynamicSurface (e.g. NoseCone, Tail, etc.)
```

rocketpy/Rocket.py
```diff
@@ -449,6 +460,38 @@ def addMotor(self, motor, position):
...
+
+        Parameters
+        ----------
+        surfaces : list, AeroSurfaces, NoseCone, TrapezoidalFins, EllipticalFins, Tail
+            Aerodynamic surface to be added to the rocket. Can be a list of
+            AeroSurfaces if more than one surface is to be added.
+            See AeroSurface class for more information.
+        positions : int, float, list
```
Gui-FernandesBR, line 474: See "AeroSurfaces" class instead of "AeroSurface".

Btw, since we have a Motor class instead of Motors class, sholdn't we also name AeroSurface instead of Aerosurfaces?

Sorry for this annoying comment, but it can improve code's consistency

rocketpy/Rocket.py
```diff
@@ -449,6 +460,38 @@ def addMotor(self, motor, position):
...
+            Position, in m, of the aerodynamic surface's center of pressure
+            relative to the user defined rocket coordinate system.
+            See `Rocket.coordinateSystemOrientation` for more information
+            regarding the rocket's coordinate system.
+            If a list is passed, will correspond to the position of each item
```
Gui-FernandesBR, line 479: it will *

Gui-FernandesBR (changes requested):
I made several comments above, please ensure you consider them before we merge this one.

Overall it was an incredible work and I'm surprised with the new component class as it is short and straightforward.

I can also confirm that tests are passing in my machine (windows 11, python 3.10.9)
The only exception is the ``test_utilities.py::test_fin_flutter_analysis``, same old same old.

rocketpy/Rocket.py
phmbressan: Generally, it is good practice keeping class names as a noun in the singular form, so `AeroSurface` would be correct here.

The same should apply to the other classes as well. Nice catch.

rocketpy/Components.py
MateusStano: > You defined in the class' docstring that "this is currently used only for aerodynamic surfaces".
> 1 - Well, should we modify the class docstring or the add method?
> 2 - What other objects could possible be components in the future?
> - Rail buttons for sure
> - Maybe parachutes should be too?
> - Any other?

It could some day be used for other stuff. I guess I could change the documentation to be `any` instead of specifying a type. Because it really could be anything

rocketpy/Components.py
MateusStano: I actually thought of that when writing the code. What I thought was that if the method is implemented to remove all the same components even if it is in the list more than once, the user would not be able to easily remove only one of those instances.

The way it is currently done, the user could just call `remove` as many times as needed and it is in their control.

By the way, this is how a standard Python list deals with `.remove`

rocketpy/Components.py
Gui-FernandesBR: Keep in mind that every time we change the documentation, we are actually creating a breaking change,
So good idea to be flexible by now ;)

rocketpy/Components.py
Gui-FernandesBR: OK then, it makes a lot of sense to me now.
But I still think that we could mention, in the documentation, that it will remove only the first appearance of the object.

rocketpy/AeroSurfaces.py
MateusStano: `clalpha` is a function of mach, so `clalpha(0)` is the Tail Lift Coefficient Slope at Mach 0 (or at speed 0).
I think it is a good idea to specify that it is at mach 0

rocketpy/Rocket.py
MateusStano: Should the file `AeroSurfaces.py` be changed to `AeroSurface.py` too? The file current holds several classes

MateusStano:
@Gui-FernandesBR I resolved the conversation of all the review comments that were added in the last commit (https://github.com/RocketPy-Team/RocketPy/pull/355/commits/f2a84a0fea8f699b127d4388a0ea4a4aba89d3fe)

rocketpy/Rocket.py
Gui-FernandesBR: @MateusStano , yes the file should be renamed too.
In PEP convention, it should be:
- File: aero_surface.py
- Class: AeroSurface

As we are still not adopting snake_case, it is ok to use AeroSurface.py for now.

But please do not simply change the name of the file, instead use `git rename` so it do not lose the track of commits in the file.

rocketpy/AeroSurfaces.py
Gui-FernandesBR: Ok, sorry for my mistake and thanks for solving this one in the last commit :)

rocketpy/Components.py
Gui-FernandesBR: I'm re-opening this conversation as I'm not sure if we are allowed to simply put "any" as the argument type requirement.

@MateusStano have you seen this approach in any other place?

I understand and agree that we should be flexible, but this seems new and tricky to me, I'm not sure if everybody would understand it.

Gui-FernandesBR:
> @Gui-FernandesBR I resolved the conversation of all the review comments that were addressed in the last commit ([f2a84a0](https://github.com/RocketPy-Team/RocketPy/commit/f2a84a0fea8f699b127d4388a0ea4a4aba89d3fe))

Great!! Many thanks @MateusStano ,
Just 3 more conversations to be solved and I will be able to approve this one

rocketpy/Components.py
MateusStano: Yes I have seen this in pydantic and actually good catch because the `a` should be capitalized (e.g. `Any`)

MateusStano: Example from https://docs.python.org/3/library/typing.html

image

rocketpy/Rocket.py
MateusStano: Done in https://github.com/RocketPy-Team/RocketPy/pull/355/commits/5b25113544a2e787d9cda07ff113e6fd9c1f8241

rocketpy/Components.py
MateusStano: Done in https://github.com/RocketPy-Team/RocketPy/pull/355/commits/cec534a57655859b4c7aa47e372a2f051bd4312b

rocketpy/Components.py
MateusStano: Done in https://github.com/RocketPy-Team/RocketPy/pull/355/commits/e53404f81bf89862181c1b8732331d01db6b10f0

Gui-FernandesBR (approved):
One of the best PRs I ever saw, it really address the problem with a clever Components class.

1 year ago we didn't even had a separate file for the aero surfaces definition, everything was in Rocket.py.

Now they have a parent (ABC) and a component class to improve the level of abstraction.

I'm honestly happy here :)

PR #357, opened 2023-04-27
ENH: Environment changes for Dispersion

MateusStano:
## Pull request checklist

## Discussion

Some of the arguments in the `Environment` class are being saved on `self` with a different name, e.g. the argument `railLength` is saved with `self.rL = railLength`.

If `getattr` is run with the name of the arguments it will not be able to get the `railLength` value inputted. This is a huge problem for the `Dispersion` class, since it is heavilydependent in `getattr`.

The following attributes were changed:

- `self.rL` --> `self.railLenght`
- `self.lat` --> `self.latitude`
- `self.lon` --> `self.longitude`
- created `self.datetime_date`, which is what `self.date` was before these changes. Meaning, it is the argument `date` transformed into a datetime object with the inputted timezone
- made `self.date` save the exact argument inputted in `date`

- [x] Yes
- [ ] No

## Other information

`self.g` is also a problem here but it is already being fixed in #338

### DIFF
--- a/docs/notebooks/topography_usage.ipynb
+++ b/docs/notebooks/topography_usage.ipynb
@@ -139,7 +139,7 @@
             "metadata": {},
             "outputs": [],
             "source": [
-                "elevation = Env.getElevationFromTopographicProfile(Env.lat, Env.lon)\n"
+                "elevation = Env.getElevationFromTopographicProfile(Env.latitude, Env.longitude)\n"
             ]
         },
         {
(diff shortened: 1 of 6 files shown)


phmbressan (approved):
I approve the changes since they improve code consistency regarding the parameters and attributes of classes and are required for the current `Dispersion` implementation. However, just as a side-note, this might be a bit restrictive regarding the namespace for `attributes` and `properties`.

PR #358, opened 2023-04-27
ENH: add " ' " to Function __repr__

MateusStano:
## Pull request checklist

## Discussion

There is a problem in the Dispersion class when it comes to saving Function objects in the `.txt` files. This is a special problem in the `.disp_inputs.txt` files, since the values of `powerOffDrag`, `powerOnDrag` and others are received as Functions.

What happens in these cases is, when saving in the `.txt` file we simply pass a dictionary, which contains the info of the simulation inputs, into the `write` function. The interaction between the `write` function and the `Function` class is so that what gets saved in the file is the `Function` `__repr__`. For example:

```python
 { ... "powerOffDrag": Function from R1 to R1 : (Scalar) → (Scalar), ... }
```

Further on, in the dispersion class, there is an attribute called `inputs_log` that generates, from the input file, a list in which each item is a line of the file. A list of dictionaries with all the inputs necessary to recreate any iteration of the simulation.

Now, two problems arise from this:

- The way `Function`s are saved is not "redefinable", so we can't recreate that `Function` object just with its `__repr__` information. I believe the only way to do this (from what I know) would be to start using `pickle`. Unfortunately, that is out of the scope of v1.0 since it would take some extra effort to implement
- When the `inputs_log` is created, it can not interpret each line of the file because `Function from R1 to R1 : (Scalar) → (Scalar)` is not a recognizable Python object. It is not a string.

The temporary solution I found for this was to just change add a `'` as the first and last character of the `__repr__` string. This way, it gets saved in the `.txt` as string and can be read by as a python object:

```python
 { ... "powerOffDrag": 'Function from R1 to R1 : (Scalar) → (Scalar)', ... }
```

I am creating this PR to both ask if there is any better way to do this, or if this is an okay solution to the problem

- [ ] Yes
- [X] No

### DIFF
--- a/rocketpy/Function.py
+++ b/rocketpy/Function.py
@@ -564,7 +564,7 @@ class Function:
         interpolations.
 
         >>> g.setDiscreteBasedOnModel(f)
-        Function from R1 to R1 : (Scalar) → (Scalar)
+        'Function from R1 to R1 : (Scalar) → (Scalar)'
         >>> h = f * g
         >>> h.source
         array([[ 0.,  0.],
--- a/rocketpy/Function.py
+++ b/rocketpy/Function.py
@@ -1223,8 +1223,8 @@ class Function:
 
     def __repr__(self):
         "Return a string representation of the Function"
-        return (
-            "'Function from R"
+        return repr(
+            "Function from R"
             + str(self.__domDim__)
             + " to R"
             + str(self.__imgDim__)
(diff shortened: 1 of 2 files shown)


phmbressan (approved):
I don't see problem in making the `inputs` and `outputs` of the `Function` a "string of a string" _per se_, however that is a bit of a weird change since it alters the behavior of all `Function` instances so as meet the requirements of `Dispersion`. Perhaps it could also be achieved with an `adapter` method at the `Dispersion` class.

giovaniceotto (approved):
As suggested, the repr() of a string was used to keep the Python pattern.

PR #364, opened 2023-05-22
Enh/liquid-motors-linear-algebra

giovaniceotto:
## Pull request type

## Pull request checklist

Enter text here...

## What is the new behavior?

Enter text here...

- [ ] Yes
- [ ] No

## Other information

Enter text here...


rocketpy/tools.py
```diff
@@ -23,3 +26,978 @@ def __get__(self, instance, owner=None):
...
+        precedence of the operators.
+        """
+        x = self[1] * other[2] - self[2] * other[1]
+        y = -self[0] * other[2] + self[2] * other[0]
+        z = self[0] * other[1] - self[1] * other[0]
+        return Vector([x, y, z])
+
+    def __matmul__(self, other):
+        """Dot product between two R3 vectors."""
+        return sum([self[i] * other[i] for i in range(3)])
```
phmbressan, line 269: There is also a roughly 2x speed increase in performing:

```py
return self[0]*other[0] + self[1]*other[1] + self[2]*other[2]
```

The same comment as above about the usage of `self.x` applies.

rocketpy/tools.py
```diff
@@ -23,3 +26,978 @@ def __get__(self, instance, owner=None):
...
+            msg = "One or more elements of this vector is not callable."
+            raise TypeError(msg) from exc
+
+    def __len__(self):
+        return 3
+
+    @cached_property
+    def x(self):
+        """First component of the vector."""
+        return self.components[0]
```
phmbressan, line 158: If all relevant operations with Vectors involves this value (that will be eventually cached as an attribute) I would remove the small delay of the first caching and input `self.x` as a plain attribute on `__init__`.

Also, I would deconstruct the list so as not to call `__getitem__` thrice:

```py
self.x, self.y, self.z = components
```

rocketpy/tools.py
```diff
@@ -23,3 +26,978 @@ def __get__(self, instance, owner=None):
...
+        >>> M = Matrix([[1, 0, 0], [0, 2, 0], [0, 1e-5, 3]])
+        >>> M.is_diagonal
+        False
+        """
+        for i, j in product(range(3), range(3)):
+            if i == j:
+                continue
+            if abs(self[i, j]) > tol:
+                return False
+        return True
```
phmbressan, line 691: Just as a remider (to myself or anyone who wants to toy with timeit) to compare the speed of this implementation and an huge explicit `if` with `ands` (since Python evaluates them lazily). I really hope its not much.

rocketpy/tools.py
```diff
@@ -23,3 +26,978 @@ def __get__(self, instance, owner=None):
...
+        """
+        try:
+            result = [[0, 0, 0], [0, 0, 0], [0, 0, 0]]
+            for i, j in product(range(3), range(3)):
+                result[i][j] = Vector(self[i, :]) @ Vector(
+                    [other[0][j], other[1][j], other[2][j]]
+                )
+            return Matrix(result)
+        except (TypeError, IndexError):
+            return Vector([Vector(self[i, :]) @ other for i in range(3)])
```
phmbressan, line 849: I will test the difference in speed later. However, the only way I can think of speeding up this method is an explicit computation in a 3x3 format. I think it is better to handle the `other` type with an `len` comparison.

phmbressan (commented):
Overall great enhancements, really straight to the point code to speed up simple linear algebra operations vs numpy. I still have some testing to do and left some comments, I will approve it after it.

rocketpy/Flight.py
```diff
@@ -1723,66 +2051,86 @@ def op(a):
...
+                    (1 / 2 * rho * compStreamSpeed)
                     * referenceArea
                     * (2 * surfaceRadius) ** 2
-                    * Cldomega(freestreamMach)
+                    * Cldomega(compStreamMach)
                     * omega3
                     / 2
                 )
                 M3 += M3f - M3d
             except AttributeError:
```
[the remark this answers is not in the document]
giovaniceotto: Every iteration, depending on which rocket components exist. Some do not have all the attributes needed in this branch. With the new abstract classes, this should be solved.

rocketpy/tools.py
```diff
@@ -23,3 +26,978 @@ def __get__(self, instance, owner=None):
...
+        return (self.x**2 + self.y**2 + self.z**2) ** 0.5
+
+    def __neg__(self):
+        """-1 times R3 vector self."""
+        return Vector([-self.x, -self.y, -self.z])
+
+    def __add__(self, other):
+        """Sum two R3 vectors."""
+        result = [self[i] + other[i] for i in range(3)]
+        return Vector(result)
```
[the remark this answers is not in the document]
giovaniceotto: Good suggestion! I have updated this in the new implementation.

rocketpy/tools.py
giovaniceotto: Great suggestion! I have implemented it in the new commits.

rocketpy/tools.py
giovaniceotto: `len` of both Matrices and Vectors is `3` unfortunately. But I did implement the explicit 3x3 computation to speed up execution significantly. Furthermore, I tried the Laderman algorithm, but it is much slower in Python.

rocketpy/tools.py
```diff
@@ -23,3 +26,978 @@ def __get__(self, instance, owner=None):
...
+
+    def __rsub__(self, other):
+        """Subtract two R3 vectors."""
+        result = [other[i] - self[i] for i in range(3)]
+        return Vector(result)
+
+    def __mul__(self, other):
+        """Component wise multiplication between R3 vector and scalar other."""
+        return self.__rmul__(other)
```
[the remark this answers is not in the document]
phmbressan: Adding to the discussion, I performed some `timeit` tests for an analogous `__mul__` case and have the following relative performance:

```
Mul:     0.5847252989999561
Short RMul:      0.6325959810001223
Mul:     0.5638121239999236
Explicit RMul:   0.5642033120000178
Ratio 1.1212198998932894
```

All in seconds after `1_000_000` runs. The simple operations should be the worst case scenario, since they are the "fastest" operations relative to the extra function call time. Therefore, the worst case is about `10 %` slower.

In my opinion, it is worth making explicit this `__mul__`. I will leave the suggestion, but feel free to decide what you think is best here.

```suggestion
        return Vector([other * self.x, other * self.y, other * self.z])
```

rocketpy/tools.py
phmbressan: Just checked and the results are remarkebly close and within run to run variance. Therefore this implementation is much better.

MateusStano (commented):
Hey, so tried to debug the new EOM to find where the the difference was coming from. 

What I did was I ran two debugs in the `getting_started.ipynb` (without wind) in two different VScodes simultaneously. I compared the results of "standard" `uDot` and the "NumpyFree" `uDot` vector in the first step of the uDot phase. The results of `uDotRail1` were the same in both cases (as expected).

Here is what I found:

- `I_11_dot` (from `uDotVariableMassNumpyFree`) is different from `TiDot` (from standard `uDot`). From what I understand, these two should be the same, or am I wrong?

- The resulting `uDot` from "standard" is:
    - [0.0, 2.248419533601745, 25.699552867618234, 0.0, 8.656902256438004, 89.14219557097906, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
- The resulting `uDot` from "NumpyFree" is:
    - [0.0, 2.248419533601745, 25.699552867618234, 0.0, 8.59772619988504, 89.14312454809829, 0.0, 0.0, 0.0, 0.0, 0.48555730861286195, 0.0, 0.0]

- The differences are in `ay`, `az`, and specially in `w1` 
   - `w1`'s differences come from: `I_CM`, `T20` and `T21`
       - It seems that the biggest factors for difference (on both `T20` and `T21`) come from `weightB` and `r_CM` involving factors.
       
       
Overall I'd recommend to check `weightB`, `r_CM` and all the `T`'s. Seems the difference is deriving from there one way or another

giovaniceotto:
Great debugging work @MateusStano. I'll attempt to use your observations to fix the trajectory difference and get back to you ASAP.

I'll first review the motor mass PR and then get back to this.

phmbressan (approved):

rocketpy/Flight.py
```diff
@@ -2092,7 +2092,7 @@ def uDotVariableMassNumpyFree(self, t, u, postProcessing=False):
 
         T20 = ((w ^ T00) ^ w) + (w ^ T03) + T04 + weightB + Vector([R1, R2, R3])
 
-        T21 = ((I @ w) ^ w) + T05 @ w + (weightB ^ r_CM) + Vector([M1, M2, M3])
+        T21 = ((I @ w) ^ w) + T05 @ w - (weightB ^ r_CM) + Vector([M1, M2, M3])
```
phmbressan, line 2095: Superb work catching this mistake @MateusStano. Great to see that the results agree now. This linear algebra pull request allow for much more readable formulas in code, on top of that, with this last fix, the results appear to be consistent.

giovaniceotto:
- [ ] Review equations of motion based Stano's feedback.
- [ ] Remove unnecessary implementations.
- [ ] Create a toggle to use older equations of motion.

giovaniceotto:
> * `I_11_dot` (from `uDotVariableMassNumpyFree`) is different from `TiDot` (from standard `uDot`). From what I understand, these two should be the same, or am I wrong?

For a moment, I agreed with your hypothesis. But it is not true because `I_11_dot` and `TiDot` are relative to different origins. Since there is the `m*d²` term when translating inertia components, and both the mass and the distance are time varying functions, the relationship between `I_11_dot` and `TiDot` is non-trivial.

>Is the multiplication and division of * self.rocket.motor.mass / total_mass correct here? It seems that this was done to simplify the expression of v_dot, but I am not sure if the other expressions that use r_CM_z were also changed.

>Also r_CM_z is meant to be the CM position relative to CDM in the body frame. This multiplication and division makes it so r_CM_z represents something else, or am I wrong?

I have verified that this is in fact correct. The logic is:

$r_{CM} \cdot m_{CM} = r_{CDM} \cdot m_{CDM} + r_{CPM} \cdot m_{CPM}$

where $CM$ is for Center of Mass, $CDM$ is for Center of Dry Mass and $CPM$ is for Center of Propellant Mass. Keep in mind that $m_{CDM}$ is the total dry mass (constant) and $m_{CPM}$ is the propellant mass (time-varying).

Since, for the equations of motion, the origin is fixed back to the $CDM$, then $r_{CDM} = 0$. This simplifies the equation to:

$r_{CM} = r_{CPM} \cdot m_{CPM} / m_{CM} $

I believe this settle the debate about the weird mass multiplication and division in the code.

With all EOMs doubts addressed, @MateusStano, feel free to review my last commits and merge.

MateusStano (approved):

giovaniceotto:
Seems like I had an issue before with the beta v1 merge (user error). Just merged it again, and the liquid motor branch as well. Everything working perfectly now, with all tests passing. No code changes, in the scope of this PR, were needed.

@MateusStano, please skim through and merge when you feel ready. Let me know if you have any questions. For reviewing porpuses, I would recommend reviewing each commit individually.

MateusStano (approved):

Issue #265, opened 2022-11-02
BUG: Unable to setup RocketPy using python 3.11


Gui-FernandesBR:
**Describe the bug**

When used the most recent [python 3.11 version](https://docs.python.org/3/whatsnew/3.11.html), rocketpy cannot be properly installed, it seems that the interpreter is currently having issues when installing the netCDF4 library. More specifically, a `error: metadata-generation-failed` is raised.

**To Reproduce**

Initialize a fresh environment of python and ensure you are using python 3.11.0 version.

**Expected behavior**

I want to see something like this on my terminal after setting up rocketpy package:
`Successfully installed certifi-2022.9.24 cffi-1.15.1 cftime-1.6.2 charset-normalizer-2.1.1 contourpy-1.0.6 cycler-0.11.0 fonttools-4.38.0 h3-3.7.4 idna-3.4 jsonpickle-2.2.0 kiwisolver-1.4.4 matplotlib-3.6.1 netCDF4-1.6.1 numpy-1.23.4 packaging-21.3 pillow-9.3.0 pycparser-2.21 pyparsing-3.0.9 python-dateutil-2.8.2 pytz-2022.6 requests-2.28.1 rocketpy-0.12.0 scipy-1.9.3 setuptools-65.5.0 simplekml-1.3.6 six-1.16.0 timezonefinder-6.1.6 urllib3-1.26.12 windrose-1.8.0`

**Screenshots**

**Trying to install rocketpy with python 3.11:**
![image](https://user-images.githubusercontent.com/63590233/199374720-3101e52a-ec79-47f1-9f07-2faf3ae91283.png)
![image](https://user-images.githubusercontent.com/63590233/199374750-2bd446ee-bc22-4925-8f5d-326e2568d62c.png)

**Now, when trying on python 3.10:**
![image](https://user-images.githubusercontent.com/63590233/199374869-80efad26-b144-409d-9ca6-d3ecb3c1fd5c.png)
![image](https://user-images.githubusercontent.com/63590233/199374886-1c0b8a8e-7a3d-4e71-b5d8-a10bbf329a59.png)
![image](https://user-images.githubusercontent.com/63590233/199374903-7d3f474a-7cbe-4c79-9dd1-d7ff63e51d47.png)

Gui-FernandesBR:
Does anyone have any idea on how to solve it?

@giovaniceotto @MateusStano @phmbressan @MrGribel

phmbressan:
[...]

PR #365, opened 2023-05-26
MAINT: Dropping support for python 3.7

Gui-FernandesBR:
## Pull request type

- [x] GitHub maintenance

## Pull request checklist

RocketPy supports python 3.7, which will stop being maintained/updated in the very-short-near future. 
Considering that the v1.0.0 is probably going to be release later than that, I'm already proposing to drop python 3.7 support.

## What is the new behavior?

- The lowest version accepted to run rocketpy package is now the 3.8
- The automations were adjusted to be run considering py 3.8
- When modifying requirements, I used the same logic as before, but changing 3.7 to 3.8

- [x] No

## Other information

* Only 1 month and 3 days left for the official support of python 3.7 (https://endoflife.date/python)
* Sorry for the blank spaces added in some automation files, it was done automatically by the prettier.
* **The official drop will happen, automatically, only after making the release**


setup.py
```diff
@@ -54,5 +46,5 @@
         "License :: OSI Approved :: MIT License",
         "Operating System :: OS Independent",
     ],
-    python_requires=">=3.7",
+    python_requires=">=3.8",
```
Gui-FernandesBR, line 49: This is the most important line.
The PR were basically named after that.

setup.py
```diff
@@ -7,21 +7,13 @@
...
-    if sys.version_info[1] < 11:
-        # Support for Linux with Python < 3.11
-        # Avoid problems due to netcdf4 issue linked below
-        netCDF4_requirement = "netCDF4>=1.4,<1.6"
-    else:
-        # Support for Linux with Python >= 3.11
-        # Might have problems with openDAP in the Environment class due to
-        # https://github.com/Unidata/netcdf4-python/issues/1179
-        netCDF4_requirement = "netCDF4>=1.6.2"
+if sys.platform == "linux" and sys.version_info[1] < 11:
```
Gui-FernandesBR, line 12: Try to understand that I'm achieving the same result using only one `if` statement

requirements.txt
```diff
@@ -1,8 +1,8 @@
 numpy>=1.0
 scipy>=1.0
 matplotlib>=3.0
-netCDF4>=1.4,<1.6; python_version == "3.7"
-netCDF4>=1.6.2; python_version != "3.7"
+netCDF4>=1.4,<1.6; python_version < "3.11"
```
Gui-FernandesBR, line 4: The old script didn't make sense to me, this one is more aligned with the requirements definition made in setup.py

MateusStano: Yeah... so it seems python 3.7 requires `netCDF4>=1.4,<1.6`. Since we are dropping 3.7 we can just remove the requirement

requirements.txt
```diff
@@ -1,8 +1,8 @@
...
+netCDF4>=1.6.2; python_version >= "3.11"
 windrose>=1.6.8
```
MateusStano, line 6: The correct fix would be:


```suggestion
netCDF4>=1.4; python_version < "3.11"
netCDF4>=1.6.2; python_version >= "3.11"
windrose>=1.6.8
```

MateusStano (changes requested):
Please keep in mind that this will cause conflicts with #368

docs/user/requirements.rst
```diff
@@ -20,7 +20,7 @@ The following packages are needed in order to run RocketPy:
 - Numpy >= 1.0
 - Scipy >= 1.0
 - Matplotlib >= 3.0
-- netCDF4 >= 1.4, < 1.6 for Python 3.7+, netCDF4 >= 1.6.2 for Python 3.11
+- netCDF4 >= 1.4, < 1.6 for Python 3.8+, netCDF4 >= 1.6.2 for Python 3.11
```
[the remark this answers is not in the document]
Gui-FernandesBR: @MateusStano can you confirm that python 3.8, 3.9 and 3.10 work well with netCDF==1.62 and above? 
Because in the above suggestion it is the expected behavior.

Gui-FernandesBR:
In the next days I am going to create some codespaces (https://docs.github.com/en/codespaces/getting-started/deep-dive) to test the package build under different operational systems, dependency and python versions.

I'm also curious with this last comment: https://github.com/RocketPy-Team/RocketPy/issues/265#issuecomment-1555409091

Gui-FernandesBR:
@MateusStano , this one is ready for a re-review.

RocketPy will always be built considering the requirements described in the setup.py file;
Other files are used only for documentation purposes.
With that been said, if you agree with the changes made in the setup.py file, please notice that all the other files are simply following what is written in the setup.py.

I may open another PR to bump up netcdf requirement to 1.6.4 version. Until that, the code should be working properly regardless the OS or python version, as you can see by the successful github workflow tests.

Gui-FernandesBR:
This one is ready for review and merge!

docs/user/requirements.rst
```diff
@@ -7,7 +7,7 @@ But read further if this is not your case!
 Python Version
 --------------
 
-RocketPy supports Python 3.7 and above.
+RocketPy supports Python 3.8 and above.
 Sorry, there are currently no plans to support earlier versions.
 If you really need to run RocketPy on Python 3.6 or earlier, feel free to submit an issue and we will see what we can do!
```
phmbressan, line 12: I believe this comment should be updated to consider `Python 3.7` as legacy.

```suggestion
If you really need to run RocketPy on Python 3.7 or earlier, feel free to submit an issue and we will see what we can do!
```

docs/user/requirements.rst
```diff
@@ -20,16 +20,15 @@ The following packages are needed in order to run RocketPy:
 - Numpy >= 1.0
 - Scipy >= 1.0
 - Matplotlib >= 3.0
-- netCDF4 >= 1.4, < 1.6 for Python 3.7+, netCDF4 >= 1.6.2 for Python 3.11
+- netCDF4 >= 1.6.4
```
phmbressan, line 23: I recall hearing that this version of `netCDF4` solves the issues that we were having, however I did not verify this. Hence, may I ask if you were able to check whether the `collab` and `3.11` install work on `1.6.4`?

phmbressan (approved):

docs/user/requirements.rst
Gui-FernandesBR: Yes, ot is running perfectly.

Gui-FernandesBR:
Moving forward given that all the commentaries from both reviews were addressed already.

PR #366, opened 2023-05-26
TST: Remove Flutter Tests

MateusStano:
## Pull request type

The flutter analysis yields inconsistent results across different computers, with the most significant variations occurring when changing operating systems. Surprisingly, even when the operating system remains the same, discrepancies still arise.

Although we are uncertain about the root cause, there may be a possible correlation with the issue described in this Stack Overflow post: https://stackoverflow.com/questions/38228088/same-python-code-same-data-different-results-on-different-machines.


Gui-FernandesBR:
Modifying tests is a critical decision always, excluding them is just unacceptable.
However, since (1) it is not a core functional of rocketpy, and (2) it is slowing down our development, I agree with deleting the corresponding code.

It is worth mentioning that **this is a temporary fix only**. We should recreate tests for the flutter feature right after the v1.0.0 release. 

Unfortunately, a new issue should be opened in our repo. This way we avoid "forgetting that it was a problem in the past".

Gui-FernandesBR (approved):

Gui-FernandesBR:
PR already approved and new related issue raised.
Feel free to merge this one whenever you want @MateusStano , thanks!

PR #368, opened 2023-05-26
MAINT: convert windrose ipywidgets and jsonpickle to optional dependencies

Gui-FernandesBR:
## Pull request type

## Pull request checklist

These little guys are always imported, even though they are barely used during rocketpy experience.

## What is the new behavior?

Following the timezonefinder example, I converted the libraries to optional dependencies.
Now the rocketpy library is ligther, faster and more consistent.
I also took care to update the corresponding documentation files. 

- [x] No


requirements_test.txt
```diff
@@ -3,4 +3,5 @@ pytest-coverage
 black
 flake8-black
 pandas
-numericalunits==1.25
\ No newline at end of file
+numericalunits==1.25
+ipywidgets>=7.6.3
```
Gui-FernandesBR, line 7: I did this to ensure all the tests would pass without catching importError

rocketpy/EnvironmentAnalysis.py
```diff
@@ -2064,6 +2065,13 @@ def animate_average_wind_rose(self, figsize=(8, 8), filename="wind_rose.gif"):
         -------
         Image : ipywidgets.widgets.widget_media.Image
         """
+        try:
+            import ipywidgets as widgets
```
Gui-FernandesBR, line 2069: I had to repeat this code block twice.
If you have an idea on how to avoid that, I would appreciate.
but it works.

docs/user/requirements.rst
```diff
@@ -67,7 +61,7 @@ To update Scipy and install netCDF4 using Conda, the following code is used:
 Optional Packages
 -----------------
 
-Optionally, you can install timezonefinder to allow for automatic timezone detection when performing Enviornment Analysis.
+Optionally, you can install timezonefinder to allow for automatic timezone detection when performing Environment Analysis.
```
MateusStano, line 64: I think this should include the other optional packages too

docs/user/requirements.rst
```diff
@@ -77,6 +71,21 @@ This can be done by running the following line of code in your preferred termina
...
+
+You can install these packages by running the following lines in your preferred terminal:
+
+.. code-block:: shell
+
+    pip install "windrose>=1.6.8"
+    pip install "ipywidgets>=7.6.3"
+    pip install jsonpickle
+
+
```
MateusStano, line 88: And this should be included in the part above. It makes more sense to me

MateusStano (changes requested):

rocketpy/EnvironmentAnalysis.py
MateusStano: It is not that much repetition in a part of the code that is not crucial. I think it is fine to leave it like this

requirements_test.txt
MateusStano, line 1: Shouldn't this file be called `requirements_tests.txt`? The missing `s` annoys me a bit heheh

MateusStano:
Another thing, the optional packages are still in the `requirements.txt`. They should be removed from there

rocketpy/EnvironmentAnalysis.py
Gui-FernandesBR: agreed

Gui-FernandesBR:
> One last thing, the optional packages are still in the `requirements.txt`. They should be removed from there I think

Paraphrasing Giovani, I have to say: "Yes. The requirements.txt file is used for development only, it is not included in the packaged served through PyPI."

requirements_test.txt
Gui-FernandesBR: What really annoys me is that VSCode doesn't recognize it as a requirements file.
The best name would be `requirements-tests.txt`. 

I'm fixing that then, hope you don't mind to have that modification in this PR.

docs/user/requirements.rst
Gui-FernandesBR: Solved here: https://github.com/RocketPy-Team/RocketPy/pull/368/commits/8a45c7b896e5ef869ef2072b22e18e1e86f5e18a

Yes, indeed it improved clarity, thanks!

docs/user/requirements.rst
Gui-FernandesBR: Solved here: https://github.com/RocketPy-Team/RocketPy/pull/368/commits/8a45c7b896e5ef869ef2072b22e18e1e86f5e18a

requirements_test.txt
Gui-FernandesBR: Solved with: https://github.com/RocketPy-Team/RocketPy/pull/368/commits/636b3b7632ec8c475c9549b44f92b438832f1b45

Gui-FernandesBR:
@MateusStano ready for a re-review!

@phmbressan @giovaniceotto , the old `requirement_test.txt` file was renamed to `requirements-tests.txt`, I would appreciate having an approvement from you too.

VSCode now understands it is a requirements file:
image

docs/user/requirements.rst
```diff
@@ -67,15 +61,22 @@ To update Scipy and install netCDF4 using Conda, the following code is used:
 Optional Packages
 -----------------
 
-Optionally, you can install timezonefinder to allow for automatic timezone detection when performing Enviornment Analysis.
-This can be done by running the following line of code in your preferred terminal:
+Optionally, you can install the following packages to allow for additional functionality:
+
```
MateusStano, line 65:
```suggestion
Optionally, you can install the following packages to allow for additional functionalities when using the EnvironmentAnalysis class:

```

rocketpy/EnvironmentAnalysis.py
phmbressan: Could you update be on the reasoning behind moving the auxiliary imports to their own methods? My understanding is that one should only be able to create an `EnvironmentAnalisys` instance if all of its methods are readily available. Then, the imports would be on the top of the file, like:

```py
try:
    import ipywidgets, windrose, timezonefinder
except ImportError import_error:
    raise ImportError(f"Please install module {import_error}.")

### Please note that this is just a sketch code block as an example, I don't even know if it works.
```

I understand that the some of these imports are only needed sometimes in only one method. However, this is not the way the other imports are handled, for instance, there is no check for `matplotlib` at every `allInfo`, even though its the only place the plots module is needed.

phmbressan (changes requested):
I have no problem with the content of this pull request and I will approve it if first some things are taken into account:

1. It might be troublesome requiring from the user to install individually the extras package as needed, since it not only requires the python and pip knowledge to do so, but also the recognition of a possible `ImportError` in code.
2. I believe it would be better if features from [PEP 508](https://peps.python.org/pep-0508/) are employed to separate a bit better the project dependencies:
    - For standard use `pip install rocketpy`;
    - For analysis use `pip install rocketpy[envanalysis]`.
    - [Usage description](https://stackoverflow.com/questions/64685527/pip-install-with-all-extras).

   For future use this can be expanded further in the future, e.g. `pip install rocketpy[orbitsim]`. That way we have more control that the necessary imports for a feature are indeed installed.

This might be out of scope for this pull request. If you thick this is a good idea, we might create another one.

requirements_test.txt
Gui-FernandesBR: No longer needed!

rocketpy/EnvironmentAnalysis.py
Gui-FernandesBR: > Could you update be on the reasoning behind moving the auxiliary imports to their own methods? My understanding is that one should only be able to create an `EnvironmentAnalisys` instance if all of its methods are readily available. Then, the imports would be on the top of the file, like:
> 
> ```python
> try:
>     import ipywidgets, windrose, timezonefinder
> except ImportError import_error:
>     raise ImportError(f"Please install module {import_error}.")
> 
> ### Please note that this is just a sketch code block as an example, I don't even know if it works.
> ```
> 
> I understand that the some of these imports are only needed sometimes in only one method. However, this is not the way the other imports are handled, for instance, there is no check for `matplotlib` at every `allInfo`, even though its the only place the plots module is needed.

Great question,
I agree with your argument,
I guess I only tried to copy what had been done before with the timezonefinder packages, 
without actually thinking about what the code was doing.

Thanks for pointing out this kind of things.

rocketpy/EnvironmentAnalysis.py
```diff
@@ -20,6 +20,17 @@
...
+    import ipywidgets as widgets
+    import jsonpickle
+    from timezonefinder import TimezoneFinder
+    from windrose import WindroseAxes
+except ImportError:
+    raise ImportError(
+        "At least one required module could not be imported."
+        + "The EnvironmentAnalysis requires additional dependencies."
+        + "Please install them by running 'pip install rocketpy[env_analysis]'."
+    )
```
phmbressan, line 33: Here, I think it would be better to point which module had an import error. I understand that the most common case would be the user not having any of them (since an "common" install was performed), however it certainly helps in debugging if the problematic module is indicated:

```suggestion
try:
    import ipywidgets as widgets
    import jsonpickle
    from timezonefinder import TimezoneFinder
    from windrose import WindroseAxes
except ImportError as error:
    raise ImportError(
        f"The following error was encountered while importing dependencies: '{error}'. "
        "Please note that the EnvironmentAnalysis requires additional dependencies, "
        "which can be installed by running 'pip install rocketpy[env_analysis]'."
    )
```

This is just a suggestion to make it clearer, feel free to change the error message.

setup.py
```diff
@@ -23,23 +23,30 @@
...
+    "simplekml",
+]
+
+env_analysis_require = [
+    "timezonefinder",
+    "windrose>=1.6.8",
+    "ipywidgets>=7.6.3",
+    "jsonpickle",
+]
+
```
phmbressan, line 42: Fantastic variable usage here, also it is much easier to locate the configuration for the netCFD4 package. Well done.

phmbressan (approved):

rocketpy/EnvironmentAnalysis.py
Gui-FernandesBR: Great suggestion, many thanks @phmbressan !!

Gui-FernandesBR:
@MateusStano , the merging is blocked because I need your approve before, since you have requested changes in the past.

Please take a look to review/approve this one ;)

MateusStano (approved):
all good

PR #361, opened 2023-05-04
ENH: Snake Case

MateusStano:
Pull request checklist

  - [x] Tests for the changes have been added
  - [ ] Docs have been reviewed and added / updated if needed
  - [x] Lint (`black rocketpy`) has passed locally and any fixes were made
  - [x] All tests (`pytest --runslow`) have passed locally

## What is the new?

snake_case!

## Does this introduce a breaking change?

- [x] Yes
- [ ] No

## Other information

Enter text here...

PR #369, opened 2023-05-26
BUG: fix standard atmosphere

Gui-FernandesBR:
## Pull request type

## Pull request checklist

@MateusStano recently caught an inconsistency with the geopotential <=> geometric altitudes in the Environment Class.
I hereby confirm that is currently a BUG and the code has not been working properly so far.

## What is the new behavior?

I verified every equation in the Standard Atmosphere model, comparing each of them with the ISO 2533 available online.
All the inconsistencies were removed.
A few comments were provided to warn users of model limitations.

- [x] Yes (it modifies a basic calculation regarding environment pressure profile and temperature in ISA. As the difference between geopotential and geometric altitudes are usually lower and 1%, the impact will be very small. It doesn't even affects the current acceptance and unit tests).

## Other information

* It is not hard to find the ISO-2533 online, just google it including "pdf". However, it seems that it was update in 2021 but only a paid version can be found so far: https://www.iso.org/standard/7472.html 
* For future reference: something that **really** annoys me is that the code consider height as a synonym of altitude. Technically they are different things, and I would agree with a quick name refactoring in the future. As a rule of thumb, whenever we wrote "height", just replace to "altitude" and we will be more accurate.

### DIFF
--- a/rocketpy/Environment.py
+++ b/rocketpy/Environment.py
@@ -2903,7 +2903,6 @@ class Environment:
         # Convert geopotential height to geometric height
         ER = self.earthRadius
         height = [ER * H / (ER - H) for H in geopotential_height]
-        height = geopotential_height
 
         # Save international standard atmosphere temperature profile
         self.temperatureISA = Function(
@@ -2917,12 +2916,12 @@ class Environment:
         g = self.standard_g
         R = self.airGasConstant
 
-        # Create function to compute pressure profile
+        # Create function to compute pressure at a given geometric height
         def pressure_function(h):
             # Convert geometric to geopotential height
             H = ER * h / (ER + h)
-            H = h
 
+            # Check if height is within bounds, return extrapolated value if not
             if H < -2000:
                 return pressure[0]
             elif H > 80000:
@@ -3019,6 +3018,8 @@ class Environment:
         """Compute the dynamic viscosity of the atmosphere as a function of
         height by using the formula given in ISO 2533 u = B*T^(1.5)/(T+S).
         This function is automatically called whenever a new atmospheric model is set.
+        Warning: This equation is invalid for very high or very low temperatures
+        and under conditions occurring at altitudes above 90 km.
 
         Parameters
         ----------


phmbressan:
I compared the results with the ISA implementation of the `Ambiance` module and the pressure results are really off due to the lack of the Molar Mass term in the pressure formula. I will rectfy this shortly.

phmbressan (approved):
Good that weird code block was corrected. Just to be sure I checked the results with the module `Ambience` and they are good.

MateusStano:
> * For future reference: something that **really** annoys me is that the code consider height as a synonym of altitude. Technically they are different things, and I would agree with a quick name refactoring in the future. As a rule of thumb, whenever we wrote "height", just replace to "altitude" and we will be more accurate.

Interesting. We can do this right before or right after (or during I guess...) the snake case changes

MateusStano (approved):

Gui-FernandesBR:
> Good that weird code block was corrected. Just to be sure I checked the results with the module `Ambience` and they are good.

Great idea, @phmbressan ! I never used this `Ambience` module, could you provide me a link to its project? 
I found this one, is the same that you used? https://pypi.org/project/ambiance/

Gui-FernandesBR:
> > Good that weird code block was corrected. Just to be sure I checked the results with the module `Ambience` and they are good.
> 
> Great idea, @phmbressan ! I never used this `Ambience` module, could you provide me a link to its project? I found this one, is the same that you used? https://pypi.org/project/ambiance/

Agreed! Therefore I'm mentioning #361 here too

phmbressan:
> > Good that weird code block was corrected. Just to be sure I checked the results with the module `Ambience` and they are good.
> 
> Great idea, @phmbressan ! I never used this `Ambience` module, could you provide me a link to its project? 
> I found this one, is the same that you used? https://pypi.org/project/ambiance/

Yes, that one. I made a typo in the module name it is `Ambiance`.

Gui-FernandesBR:
> > > Good that weird code block was corrected. Just to be sure I checked the results with the module `Ambience` and they are good.
> > 
> > 
> > Great idea, @phmbressan ! I never used this `Ambience` module, could you provide me a link to its project?
> > I found this one, is the same that you used? https://pypi.org/project/ambiance/
> 
> Yes, that one. I made a typo in the module name it is `Ambiance`.

Great, many thank for slid reviewing it.

Based on the two approved reviews, I'm proceeding with the merge operation

Issue #222, opened 2022-09-10
Let's Rocket.draw it so we can double check our parameters


Gui-FernandesBR:
**Is your feature request related to a problem? Please describe.**

Based on recent comment we got on our discord server, I believe that it would be really good if we have a function to draw the rocket geometry after we design the 

**Describe the solution you'd like**

Simple: After inputting every parameter, I'd like to use Rocket.draw() to see what is the final geometry of the rocket that will be used on the simulation, specially when talking about the fins, this is quite important!

**Additional context**
![image](https://user-images.githubusercontent.com/63590233/189438937-9119c0df-4b6c-422e-aa2e-426f34d0b97b.png)

![image](https://user-images.githubusercontent.com/63590233/189480367-cdc0d816-5040-4e79-891c-3beabd76dfec.png)

Gui-FernandesBR:
It turns out that we've already see something like this working on a hackathon a while ago... It's possible to check it by here: https://github.com/RocketPy-Team/Hackathon_2020/pull/96

Gui-FernandesBR:
Already started on #263

Gui-FernandesBR:
Rocket draws are ready, just waiting for the motors draws (#436 ) before closing this one.

Gui-FernandesBR:
This is coming in the next release. Great one.

PR #372, opened 2023-06-01
ENH: nosecone draw and bluffness 

CabGT:
## Pull request type

## Pull request checklist

## What is the new behavior?

This update introduces the function to create different nosecone type curves. 
It considers the bluffiness chosen by the users. 
Also adds a drawing function for the nosecone, which allows visualization of it.

- [ ] Yes
- [x] No


tests/test_rocket.py
```diff
@@ -370,13 +370,7 @@ def test_evaluate_static_margin_assert_cp_equals_cm(kg, m, dimensionless_rocket)
...
 @pytest.mark.parametrize(
     "k, type",
-    (
-        [2 / 3, "conical"],
-        [0.466, "ogive"],
-        [0.563, "lvhaack"],
-        [0.5, "default"],
-        [0.5, "not a mapped string, to show default case"],
-    ),
+    ([2 / 3, "conical"], [0.46469957130675876, "ogive"], [0.563, "lvhaack"]),
```
Gui-FernandesBR, line 373: Regarding the `0.46469957130675876` value: 
I did the bureaucracy so the tests could pass, but I'd love to have more reasons on why did the "ogive" format had its k value changed.

Regarding the others that were removed, it is expected since the now would be caught in exception error by the kind setter in the NoseCone class.

rocketpy/AeroSurface.py
```diff
@@ -202,15 +211,27 @@ def kind(self):
 
     @kind.setter
     def kind(self, value):
```
Gui-FernandesBR, line 213: Are we allowed to do so many things inside the `kind` setter? 
What other possibilites could we set instead?

rocketpy/AeroSurface.py
```diff
@@ -219,10 +240,85 @@ def kind(self, value):
...
+                + '\n\t"conical"'
+                + '\n\t"ogive"'
+                + '\n\t"lvhaack"'
+                + '\n\t"tangent"'
+                + '\n\t"vonkarman"'
+                + '\n\t"elliptical"'
+                + '\n\t"parabolic"\n'
+            )
+
+        n = 127  # Points on the final curve.
```
Gui-FernandesBR, line 283: Should this really be hardcoded?

rocketpy/AeroSurface.py
```diff
@@ -219,10 +240,85 @@ def kind(self, value):
...
+            xtol=1e-7,
+        )[0]
+
+        # Corrects circle radius if it's too small.
+        if xInit > 0:
+            r = self.bluffiness * self.baseRadius
+        else:
+            r = 0
+            print(
+                "ATTENTION: The chosen bluffiness ratio is insufficient for the selected nosecone category, thereby the effective bluffiness will be 0."
```
Gui-FernandesBR, line 302: I suggest changing to "WARNING" instead os "ATTENTION".

However, this message is currently being raised everytime that the user do not define the bluffiness value, which may be annoying sometimes.

Can we think a better approach to raise this warning only when the user specify a bluffness value without it being valid enough?

rocketpy/AeroSurface.py
```diff
@@ -159,6 +167,7 @@ def __init__(
         self._rocketRadius = rocketRadius
         self._baseRadius = baseRadius
         self._length = length
+        self.bluffiness = bluffiness
```
Gui-FernandesBR, line 170: I'm not an expert in English skills, but my text suggester is telling me that "bluffness" (without the "i") would sound better

rocketpy/AeroSurface.py
```diff
@@ -202,15 +211,27 @@ def kind(self):
...
+        # Sets the k for Cp calculation
+        # Sets the function which creates the respective curve
         self._kind = value
+        value = (value.replace(" ", "")).lower()
+
         if value == "conical":
             self.k = 2 / 3
-        elif value == "ogive":
-            self.k = 0.466
+            self.y_nosecone = Function(lambda x: x * self.baseRadius / self.length)
```
Gui-FernandesBR, line 222: The `y_nosecone` variable name is not clear enough, since it is becoming an attribute of the class.

I would like if we could think in a better name together.

rocketpy/AeroSurface.py
```diff
@@ -202,15 +211,27 @@ def kind(self):
...
+        # Analyzes nosecone type
+        # Sets the k for Cp calculation
+        # Sets the function which creates the respective curve
         self._kind = value
+        value = (value.replace(" ", "")).lower()
+
         if value == "conical":
             self.k = 2 / 3
-        elif value == "ogive":
-            self.k = 0.466
```
Gui-FernandesBR, line 210: Hey, this is an important breaking change!

It seems that you are modifying the expected k value of the "ogive" nose kind. Could you elaborate on the reasons for such change please?

rocketpy/AeroSurface.py
Gui-FernandesBR, line 1: Overall, it was a vey interesting PR, implementing something that seems to be easy but actually is quite tricky.

The addition of some new attributes caught my attention here, we should be careful to be clear enough when selecting these. And btw they all should be added to the docs.

I still have some questions tho, let's discuss on top of this revision please.

Gui-FernandesBR (changes requested):
Well done, @CabGT , I see a good future for this PR, meaning that small adjusts would allow us to merge.

@MateusStano , please give us a little help in this one as it is needed to be merged **before** we start to move plots/prints from the AeroSurface.py file

rocketpy/AeroSurface.py
CabGT: You are right. 
Already corrected.

rocketpy/AeroSurface.py
CabGT: Maybe we could just set up the type and nosecone equation. 
And then create an function to do the rest?

rocketpy/AeroSurface.py
CabGT: This was done following the literature found in "THE DESCRIPTIVE GEOMETRY OF NOSECONES" writen by Gary A. Crowell Sr. in 1996.
In the document, several types of nosecone are mentioned alongside their respective equations and k which stand for the coefficient that can be multplied by the length of the nosecone to find it's center of pressure as the following formual implies:

C_p = k * L
Being, 
C_p  ->  Nosecone's center of pressure 
k  ->  coefficient
L  ->  Nosecone's length

The numbers used were extracted straight from the document, which states these values can be calculated through the formula:

X_n  = V / A,  in which
X_n  ->  center of pressure measured from the base to the tip on a nosecone of length = 1m
V  ->  Nosecone's volume
A  ->  Nosecone's base area 

The k coefficient in RocketPy is complementary to X_n on the literature as it's measured from the tip to the base of the nosecone. 

The formulas and defnitions said here can be found on pages 5 and 6 of the beforementioned document and the marked value (k of ogive nosecone = 0.466) is the result of 
k = 1 - X_n 
with the respective C_n for ogive nosecone defined on page 6 of the same document.

rocketpy/AeroSurface.py
CabGT: It could be set up as an user input.

rocketpy/AeroSurface.py
CabGT: Done.

tests/test_rocket.py
CabGT: I didn't understand the problem here.

rocketpy/AeroSurface.py
Gui-FernandesBR: I think the "evaluate_geometrical_parameters" should be the best place for what we have now in the kind.setter method.
Seems like an item for a to-do list in future PRs.

tests/test_rocket.py
Gui-FernandesBR: > I didn't understand the problem here.
> 
> Could you be more specific please?

I think you already solved the question by answering another comment below. What I was wondering was regarding the value change of the variable k. But now it is clear that it comes from the new reference that you pointed out.

Gui-FernandesBR (approved):
Nice work here @CabGT , congratulations.

The implementation allows for sophisticated plotting the nosecone shape. 
This is an important step in the roadmap of the rocket drawing function #222 .

I can confirm that the code is working in my machine. 
I've tested it for different nose kinds and different bluffness ratio values.
I did some minor adjustments (basically some typos) in the last 3 PRs. Those were so tiny details that I preferred committing instead of requesting change.

A future PR containing some refactoring operations would be welcome. Here as some tasks:
- [ ] create unittests to check different values of bluffness
- [ ] clean the `kind.setter` method, moving some of its new operations to the "evaluate_geometry_parameters" method
- [ ] move the `draw()` method to the aero_surface_plots.py file, to accomplish
- [ ] I think it's time for us creating a `coods` parameter for the AeroSurface class. Again, this would help us to generate the #222 solution.
- [ ] When the bluffness==0, it sometimes throw an warning. But I didn't see the impact of that in the rest of the code

Please check my commits and merge it when you feel confortable.

PR #373, opened 2023-06-04
MAINT: EnvironmentAnalysis major refactor

Gui-FernandesBR:
## Pull request type

## Pull request checklist

- EnvironmentAnalysis is a very slow class...:
    - with everything running inside its `__init__` method
    - several methods are duplicated inside the code, resulting in doubled loops that really slow down the running speed.
    - the code is messy and a new user would have a hard time to understand the logic behind it
- There are several class attributes being defined inside the plots methods.
- There's an issue/to-do regarding the  `y_lim` in the wind speed distribution plots, where the bounds were not parametrized.

## What is the new behavior?
- `@cached_property` decorators are now being applied inside the EnvironmentAnalysis class:
    - Positive outcomes: initializing the class is much faster, and the most of the old attributes are now calculated only when needed (and calculated only once, if cached). 
    - Negative outcomes: after initializing the class, the user can no longer modify the unit system to perform more unit conversions. So the choice of unit_system in the class construction should be wise. This was necessary to adopt the cache functionality. In general, I prefer to have to create a new object with a new unit_system than changing the current object to another unit system (this last option costs a lot due to several loops needed).
- Solved the issue regarding the `y_lim` mentioned above.
- Docstring were updated or completed, wherever needed. The class is ready to not suffer breaking changes 
- Tests were double checked to match with the new architecture. And new tests were created when needed.
- The 2 Notebooks of env_analysis were updated (and I already checked that everything is running as expected).

- [x] Yes - Different methods were renamed in order to become more readable and intuitive. Docstrings also suffered from some modifications.

## Other information

This PR is not as short as I would like it to be, but I can ensure that the overall readability of the code is now better. Also, I ran all the tests in my machine perfectly, obtaining a minimum of 94% of code coverage within the env_analysis files (plots, prints, etc.)

I hope we can include these breaking changes in the next major release, a hard work of code architecture :)
Please hit me up in case of problems in the review.

### DIFF
--- a/rocketpy/Environment.py
+++ b/rocketpy/Environment.py
@@ -2903,7 +2903,6 @@ class Environment:
         # Convert geopotential height to geometric height
         ER = self.earthRadius
         height = [ER * H / (ER - H) for H in geopotential_height]
-        height = geopotential_height
 
         # Save international standard atmosphere temperature profile
         self.temperatureISA = Function(
@@ -2917,12 +2916,12 @@ class Environment:
         g = self.standard_g
         R = self.airGasConstant
 
-        # Create function to compute pressure profile
+        # Create function to compute pressure at a given geometric height
         def pressure_function(h):
             # Convert geometric to geopotential height
             H = ER * h / (ER + h)
-            H = h
 
+            # Check if height is within bounds, return extrapolated value if not
             if H < -2000:
                 return pressure[0]
             elif H > 80000:
@@ -3019,6 +3018,8 @@ class Environment:
         """Compute the dynamic viscosity of the atmosphere as a function of
         height by using the formula given in ISO 2533 u = B*T^(1.5)/(T+S).
         This function is automatically called whenever a new atmospheric model is set.
+        Warning: This equation is invalid for very high or very low temperatures
+        and under conditions occurring at altitudes above 90 km.
 
         Parameters
         ----------
--- a/rocketpy/EnvironmentAnalysis.py
+++ b/rocketpy/EnvironmentAnalysis.py
@@ -256,20 +256,6 @@ class EnvironmentAnalysis:
         -------
         index : int
             Index of the nearest value in the array.
-
-        Examples
-        --------
-        >>> array = np.array([-180, -90, 0, 90, 180])
-        >>> value = 0
-        >>> index = self.__get_nearest_index(array, value)
-        >>> index
-        2
-
-        >>> array = np.array([0, 90, 180, 270, 360])
-        >>> value = 0
-        >>> index = self.__get_nearest_index(array, value)
-        >>> index
-        0
         """
         # Create value convention
         if np.min(array) < 0:
--- a/rocketpy/EnvironmentAnalysis.py
+++ b/rocketpy/EnvironmentAnalysis.py
@@ -14,7 +14,6 @@ import jsonpickle
 import netCDF4
 import numpy as np
 import pytz
-from cftime import num2pydate
 
 from rocketpy.Environment import Environment
 from rocketpy.Function import Function
@@ -22,6 +21,7 @@ from rocketpy.units import convert_units
 
 from .plots.environment_analysis_plots import _EnvironmentAnalysisPlots
 from .prints.environment_analysis_prints import _EnvironmentAnalysisPrints
+from .tools import bilinear_interpolation, time_num_to_date_string
 
 
 class EnvironmentAnalysis:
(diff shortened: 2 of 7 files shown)


Gui-FernandesBR:
Before updating the notebooks, just for future reference...
At this point, this PR adds +2,738 code lines and remove −1,328.

During the review, I would recommend the reviewer to consider some of the following steps:
- [ ] Ensure all the tests are passing locally, specially the --runslow tests.
- [ ] Run both EnvAnalysis notebooks and check if there's nothing weird in the results. 
- [ ] Take a deep breath and try reading the EnvironmentAnalysis file. See how the methods are organized, does that make sense to you?
- [ ] Check the prints and plots files, ensuring everything do not look weird
- [ ] Point out suggestions of improvements in this PR.

Gui-FernandesBR:
Hey, I had a hard time discovering how to run the [doctest](https://docs.python.org/3/library/doctest.html#module-doctest) tests after adding "Examples >>>" in some docstrings. Tests weren't passing on GitHub even though I had performed tests flawlessly in my computer.

For future reference:

```bash
cd rocketpy
pytest --doctest-modules
```

However, finally, we are ready for review here @MateusStano

Gui-FernandesBR:
> Hey! Seems everything is good with this one, just two questions:
> 
> * Why do the tests open a window with each plot when they are ran? I thought this was not supposed to happen

That's weird and it was not happening with me. Can you confirm the steps you're following to generate such issue? And what tests exactly are creating this? 

> * Seems there is only one remaining TODO in then EnvAnalysis file `# TODO: the average_wind_speed_profile_by_hour and similar methods could be more abstract than currently are`. Anything to say about this one? Or should we just leave it there

Yes, let's leave it for the future. I don't think there's an easy way to solve that for now. Basically, there're some methods following the same steps to calculate properties in the class. This violates "don't repeat yourself" principle, but was already done before and it is kinda useful now since each cached property holds a single, different property.
In the future we can re-discuss it, therefore the TODO is remaining, but right now the cost-benefit is not so attractive.

MateusStano:
> That's weird and it was not happening with me. Can you confirm the steps you're following to generate such issue? And what tests exactly are creating this?

I simply ran `pytest --runslow`, but if it is not happening to you it might be something on my end

All in all, this is good to merge!

MateusStano (approved):

Gui-FernandesBR:
> > > That's weird and it was not happening with me. Can you confirm the steps you're following to generate such issue? And what tests exactly are creating this?
> > 
> > 
> > I simply ran `pytest --runslow`, but if it is not happening to you it might be something on my end
> > All in all, this is good to merge!
> 
> Yes, I confirm that it is happening with me too. However, the problem is also present in the Environment.py related tests
> 
> Maybe the `--runslow` option is confusing the patch option. **I need to investigate this in the following days.** (for reference: https://stackoverflow.com/questions/51429465/how-to-mock-pyplot-show-in-python-to-prevent-showing-plots)

I tried to solve the problem for hours today :/
At least I can confirm that the problem is started when we use `--runslow` and patch.mock options together, in other words:
```python
@pytest.mark.slow
@patch("matplotlib.pyplot.show")
```

Some of my current thoughts:
* I wonder why this issue have been caught only now. Was nobody using `--runslow` option? Probably not. Was the pytest version modified? Neither. So when did the problem started and why?

Gui-FernandesBR:
> > > > That's weird and it was not happening with me. Can you confirm the steps you're following to generate such issue? And what tests exactly are creating this?
> > > 
> > > 
> > > I simply ran `pytest --runslow`, but if it is not happening to you it might be something on my end
> > > All in all, this is good to merge!
> > 
> > 
> > Yes, I confirm that it is happening with me too. However, the problem is also present in the Environment.py related tests
> > Maybe the `--runslow` option is confusing the patch option. **I need to investigate this in the following days.** (for reference: https://stackoverflow.com/questions/51429465/how-to-mock-pyplot-show-in-python-to-prevent-showing-plots)
> 
> I tried to solve the problem for hours today :/ At least I can confirm that the problem is started when we use `--runslow` and patch.mock options together, in other words:
> 
> ```python
> @pytest.mark.slow
> @patch("matplotlib.pyplot.show")
> ```
> 
> Some of my current thoughts:
> 
> * I wonder why this issue have been caught only now. Was nobody using `--runslow` option? Probably not. Was the pytest version modified? Neither. So when did the problem started and why?

As discussed in our last meeting, the problem with the --runslow option was not introduced by this PR, it can be seen also in the beta/v1.0.0 branch.

A new issue should be opened to address the problem. Ans this PR is ready to be merged

Issue #205, opened 2022-07-28
RuntimeError: Unable to load latest weather data for GFS


git-akihakune:
**Describe the bug**

Running the code in [Colab Notebook preview](https://colab.research.google.com/github/giovaniceotto/rocketpy/blob/master/docs/notebooks/getting_started_colab.ipynb) led to an error.

```python
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
[<ipython-input-17-71fdef7bdc29>](https://localhost:8080/#) in <module>()
----> 1 Env.setAtmosphericModel(type="Forecast", file="GFS")

[/usr/local/lib/python3.7/dist-packages/rocketpy/Environment.py](https://localhost:8080/#) in setAtmosphericModel(self, type, file, dictionary, pressure, temperature, wind_u, wind_v)
    935                 if not success:
    936                     raise RuntimeError(
--> 937                         "Unable to load latest weather data for GFS through " + file
    938                     )
    939             elif file == "FV3":

RuntimeError: Unable to load latest weather data for GFS through https://nomads.ncep.noaa.gov/dods/gfs_0p25/gfs20220717/gfs_0p25_06z
```

**To Reproduce**

Just run the default code in the preview.

**Expected behavior**

It's expected to run the demo smoothly.

**Screenshots**

![image](https://user-images.githubusercontent.com/87116762/181552481-a9be47e3-62f6-4da9-9c72-5c1f694a6c67.png)

**Additional context**

Checking the [GrADS Data Server](https://nomads.ncep.noaa.gov/dods/gfs_0p25), indeed, I'm currently not seeing any "gfs20220717" dir. Maybe fixing the link can solve this issue.

giovaniceotto:
@git-akihakune Thanks for letting us know that you are having an issue. We recently became aware of this and it seems to be a problem relating the netCDF4 library and Colab.

We have not been able to reproduce the error locally so far. I will perform further investigations and let you know.

giovaniceotto:
I have just confirmed that running on a linux server (GitHub Codespaces) works properly. It really seems to be a problem specific to Google Colab.

Furthermore, I have tested a simple netCDF4 example using Colab and  also got an error. You can see the example here: https://colab.research.google.com/drive/17juZulxoz8sahoDJVuU4uOMSnBC_xXDA?usp=sharing

On other systems, the example runs fine.

This maybe a related issue: https://github.com/Unidata/netcdf4-python/issues/753

[...]

PR #374, opened 2023-06-09
MAINT: upgrade netcdf4 requirement

Gui-FernandesBR:
## Pull request type

- [x] Other (please describe): requirements update

## Pull request checklist

The requirements for netcdf4 dependency is a bit more complex than what would be ideal. This has been a problem since the BUG #205 was reported.

## What is the new behavior?
Everything is easier with the newest version of netcdf4, which may work without error all environments, including google colab.

- [x] No

## Other information
- It is sad that this PR isn't going to the master branch directly, but let's be patient and wait until the v1 gets finally merged into master
- This PR should wait for the #365 to be merged into beta/v1.0.0
- This should finally fix the BUG mentioned above (#205 )

### DIFF
--- a/docs/user/requirements.rst
+++ b/docs/user/requirements.rst
@@ -20,7 +20,7 @@ The following packages are needed in order to run RocketPy:
 - Numpy >= 1.0
 - Scipy >= 1.0
 - Matplotlib >= 3.0
-- netCDF4 >= 1.6.2 for Windows, MacOS or Python < '3.11'; netCDF4 >=1.4,<1.6 for Python 3.11+ in Linux
+- netCDF4 >= 1.6.4
 - windrose >= 1.6.8
 - requests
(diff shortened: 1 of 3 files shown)


docs/user/requirements.rst
MateusStano, line 1: Check out line 34 on this file. I think it also needs to be changed

MateusStano (changes requested):
Just one change here, everything else seems good

docs/user/requirements.rst
Gui-FernandesBR: Nice catch, 
I just fixed it in the last commit.

Thanks for reviewing it =)

Gui-FernandesBR:
Going to merge this one into `maint/drop-py37`, ad from there we merge into beta/v1. 
I think that's a faster way

Issue #350, opened 2023-04-10
Stability Margin with Compressibility Factor


giovaniceotto:
**Is your feature request related to a problem? Please describe.**

Currently, the `Flight` class has a property named `Flight.staticMargin`, which returns the static margin as a function of time. However, since it is the static margin, it always assumes `Mach=0.0`, and, thus, the [Prandtl-Glauert compressibility correction factor](https://en.wikipedia.org/wiki/Prandtl%E2%80%93Glauert_transformation) $beta$ is always 1. 

**Describe the solution you'd like**

For the `Rocket` class:
- `Rocket.initialStabilityMargin(mach)`: a method which computes the initial stability margin given a Mach number.

For the `Flight` class:
- `Flight.stabilityMargin()`: a property which returns the stability margin, as a Function of time, for the entire flight, accounting for Mach numbers different from 0, and changing center of mass position.

**Additional context**

This is a common request from OpenRocket users.

Gui-FernandesBR:
Hey, this is a common request from RocketPy users, let's increase this priority right after release v1.0.0

Please @MateusStano  👨‍💻

FYI,
![image](https://github.com/RocketPy-Team/RocketPy/assets/63590233/e3db00e9-008d-41a4-bf76-592b736816fa)

Gui-FernandesBR:
Seems like the issue statement was solved by the #377 
We still have some enhancements to do regarding the stability margin, tho.

PR #377, opened 2023-06-11
ENH: Adding Stability Margin with Mach dependency

Gui-FernandesBR:
## Pull request type

## Pull request checklist

- The `flight.staticMargin` doesn't seem good enough nowadays. It do not consider the variation of the center of pressure position as the mach number varies during the flight.
- Users usually have troubles to compare stability across different rocket simulators.

## What is the new behavior?
- Implementation followed the instructions/requirements established here: https://github.com/RocketPy-Team/RocketPy/issues/350
- The Rocket and Flight classes now communicates very well to compute more robust `stabilityMargin` attribute, while the trajectory simulation is not being affected.
- Everything thanks to the most magical class of rocketpy: Function

- [x] Yes (static margin is no longer a synonym of stability margin in rocketpy).

## Other information
- I tried to do a chart like [this](https://github.com/RocketPy-Team/RocketPy-Hackathon-2022/pull/99#issuecomment-1229254045), but I got no success at all. This can be done in the future.
- Please, review the changes first, and then we update the notebooks.


rocketpy/Rocket.py
```diff
@@ -225,11 +224,6 @@ def __init__(
         # Rail buttons data initialization
         self.rail_buttons = Components()
 
-        self.cpPosition = 0
```
Gui-FernandesBR, line 228: Notice this was duplicated with line 248 (same file)

rocketpy/Rocket.py
```diff
@@ -388,48 +384,83 @@ def evaluateThrustToWeight(self):
...
+        their center of pressure position and derivative of the coefficient of
+        lift as a function of Mach number.
 
         Parameters
         ----------
         None
 
         Returns
         -------
-        self.staticMargin : float
```
Gui-FernandesBR, line 403: as `None` was being returned, this documentation is currently wrong at the default branch 

(sadly)

rocketpy/Rocket.py
```diff
@@ -245,7 +239,9 @@ def __init__(
             "linear",
             "constant",
         )
-        self.cpPosition = 0  # Set by self.evaluateStaticMargin()
+        self.centerOfPressure = Function(
+            lambda mach: 0
+        )  # This will be set by the self.evaluateCenterOfPressure()
```
phmbressan, line 244: The side comment length is causing the linter to divide this simple line into multiple.

Just a aesthetical change, an equivalent approach would be using `Function(0)`, I believe it makes it clearer that is only a temporary value.

rocketpy/plots/rocket_plots.py
```diff
@@ -78,7 +78,18 @@ def staticMargin(self):
         None
         """
 
-        self.rocket.staticMargin()
+        # TODO: it would be interesting to make a 3D plot of stability margin
+        # (https://matplotlib.org/stable/gallery/mplot3d/surface3d.html)
```
phmbressan, line 82: Great suggestion!

phmbressan (commented):
Overall, the Mach dependency is a great addition.

This PR still needs to be updated to follow `snake_case` conventions before approval.

Gui-FernandesBR:
I'm waiting for the #395 PR to be merged before solving the conflicts here.

MateusStano:
Hey! I changed a few things, mostly structure wise.

- I re-added static margin because it is still useful (it is just a special case of the stability margin)
- I made stability margin a 2D Function, so now it makes a 3D plot in `Rocket.all_info()`
- Flight calculates the stability margin through the flight and plots it

Think this is ready for a review

Gui-FernandesBR:
Good work, @MateusStano , thank you so much for improving the stability margin calculation in RocketPy, this was a huge gap since the beginning.

I made some small adjusts to improve plots and prints. You can check them by my last 2 commits. 
Some tests that I think you would like to see:
image

image


I'm also sending you a copy of a notebook running the simulation in case you don't wanna run the notebook in your computer.

I want to approve this PR by I was the one who created.
Well, please consider this as resolved/approved, and feel free to merge into develop, there're no breaking changes.

rocketpy/rocket/rocket.py
```diff
@@ -419,47 +445,79 @@ def evaluate_thrust_to_weight(self):
...
+            Static margin is defined as the distance between the center of
+            pressure and the center of mass, divided by the rocket's diameter.
+        """
         # Calculate static margin
-        self.static_margin = (self.center_of_mass - self.cp_position) / (
-            2 * self.radius
+        self.static_margin.set_source(
+            lambda time: (self.center_of_mass(time) - self.cp_position(0))
+            / (2 * self.radius)
+            * self._csys
```
phmbressan, line 515: I have no problems with this, but out of curiosity why define a `lambda` source and discretize right after?

Wouldn't it work using `Function` operations directly? (same comment to the `stability_margin` evaluation)

phmbressan (approved):

rocketpy/rocket/rocket.py
```diff
@@ -419,47 +445,79 @@ def evaluate_thrust_to_weight(self):
...
         """
-        # Initialize total lift coefficient derivative and center of pressure position
-        self.total_lift_coeff_der = 0
-        self.cp_position = 0
+        # Re-Initialize total lift coefficient derivative and center of pressure position
+        self.total_lift_coeff_der.set_source(lambda mach: 0)
+        self.cp_position.set_source(lambda mach: 0)
 
         # Calculate total lift coefficient derivative and center of pressure
         if len(self.aerodynamic_surfaces) > 0:
```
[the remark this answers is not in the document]
MateusStano: It shouldn't be necessary, but something goes very wrong when adding the `EmptyMotor` in the rocket initialization. We should probably look into it some other time

rocketpy/rocket/rocket.py
MateusStano: For some reason, static_margin used to break pretty badly with some liquid and hybrid motors. We added this discretization when developing the liquid and hybrid motors, but it might have been done before we started discretizing things in those classes

I just tested it and it seems it can be remover without issues

PR #378, opened 2023-06-21
MAINT: Move the rail length definition to the Flight Class

Gui-FernandesBR:
## Pull request type

## Pull request checklist

- Weirdly, the user needs to give the rail length as an argument of the Environment class, even given it is something that only the Flight Class will need.

## What is the new behavior?
- Now you can define an environment without thinking about the rail length
- The Flight class needs the rail length to be initialized now.

- [x] Yes, says the title of this PR

## Other information

- I know that I already started this discussion earlier, but... Should the railLength really be a mandatory argument?
    - Not every rocket flies from a launch rail !!
    - If we use an initial solution in the flight (deployable payloads, multi-stage rockets, etc.), we would still have to pass an useless railLength value, 
    - Maybe the standard value of this argument should be `None`, meaning that the Flight class will skip the uDotRail if a float/int is not provided.
- I will only modify the notebooks and docs after your review.


Gui-FernandesBR:
All tests passing and PR description updated ✅

This one is ready for a review!

Please consider the last comments in the PR description.

rocketpy/Flight.py
```diff
@@ -512,6 +514,7 @@ def __init__(
         self,
         rocket,
         environment,
+        railLength,
         inclination=80,
         heading=90,
```
MateusStano, line 519: I think it is totally valid to change rail_length to be optional and to its default value be `None`. Is the Flight class already structured for this change? Or would not having `railLength` break it?

Even better question, what happens when `railLength = 0`?

Also should inclination and heading have `None` as default too?

MateusStano (changes requested):
All good here. I am requesting a review because of the comment above and of what you mentioned in the PR comment:

-  [x] Discuss if the `railLength` should be optional
- [ ] Change notebooks

rocketpy/Flight.py
Gui-FernandesBR: > I think it is totally valid to change rail_length to be optional and to its default value be `None`. Is the Flight class already structured for this change? Or would not having `railLength` break it?

I tried but the flight class was raising problem with some weird ZeroDivisionError.
I think it is just easier if we move with this PR without having a standard value for the railLength, and maybe in the future we change it in another PR, this shouldn't create any break changes.

> Even better question, what happens when `railLength = 0`?

A ValueError will be raised if a rail length <= 0 is passed, just see the last commits.

> Also should inclination and heading have `None` as default too?
Not in the scope of this PR. Great suggestion though, let's address this in another PR in the future, for sure.

Gui-FernandesBR:
Quick update here:
- [ ] I'm changing the notebooks here and will update them very soon.
- [ ] I need to update MATLAB documentation, but have no idea on how to do it. A search and replace will not work in this case.
- [ ] I created tests to the railLength parameter, but it is not working due to another bug that I already solved locally but didn't update the solution yet.

MateusStano (approved):

tests/test_flight.py
```diff
@@ -849,17 +868,62 @@ def test_latlon_conversions2(mock_show):
...
+    """
+    test_flight = Flight(
+        rocket=rocket,
+        environment=example_env,
+        railLength=rail_length,
+        inclination=85,
+        heading=0,
+        terminateOnApogee=True,
+    )
+    assert (test_flight.z(test_flight.outOfRailTime) - out_of_rail_time) < 1e-6
```
Gui-FernandesBR, line 929: Am I just extremely tired here, or this testing is asserting that the altitude at lift off is lower than the out of rail time?
This is non-sense.

rocketpy/Flight.py
```diff
@@ -2546,14 +2512,19 @@ def calculate_rail_button_forces(self):
         F22: Function
             Rail Button 2 force in the 2 direction
         """
+        # First check for no rail phase or rail buttons
+        nullForce = []
```
Gui-FernandesBR, line 2516: Actually, it is important that even the nullForce to be represented as a Function object with value 0 and length higher than 1. Otherwise I'm afraid the calculations of forces (in the funcified cached properties) will not work properly.
See for example line 2295.

tests/test_flight.py
Gui-FernandesBR: we should either use abs(a - b) < 1e-6 or other approx. function available in pytest or numpy.

Gui-FernandesBR:
@MateusStano , unfortunately, I cannot re-open the PR because it was already merged. 
But the 2 comments above need to be addressed asap since it is already running in `beta/v1.0.0`

PR #383, opened 2023-06-25
Bug/flight without rail buttons

Gui-FernandesBR:
## Pull request type

## Pull request checklist

- Try to create a flight using a rocket without rail buttons, and then try to access the rail button forces, you're going to have a bug since the class cannot access rail buttons positions (given that there's no rail button defined).

## What is the new behavior?
- I improved the conditions to capture errors when a "buttonless" rocket is being provided to the Flight Class.
- Rail buttons forces are only calculated when there's a rail button phase and a rail button pair present.
- In case you try to access rail buttons forces without properly defining rail buttons, you are going to know exactly the reasons of any error, and the forces will be equals to zero in some cases.

- [x] No

## Other information
- I need this to be reviewed and merged before proceeding with the #378

### DIFF
--- a/rocketpy/Flight.py
+++ b/rocketpy/Flight.py
@@ -2522,7 +2522,9 @@ class Flight:
             rail_buttons_tuple.component.buttons_distance + rail_buttons_tuple.position
         )
         lower_button_position = rail_buttons_tuple.position
-        angular_position_rad = rail_buttons_tuple.component.angular_position * np.pi / 180
+        angular_position_rad = (
+            rail_buttons_tuple.component.angular_position * np.pi / 180
+        )
         D1 = (
             upper_button_position - self.rocket.centerOfDryMassPosition
         ) * self.rocket._csys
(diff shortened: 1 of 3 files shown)


Gui-FernandesBR:
Just noticed something, in some cases the forces on the rail buttons could be exclusively negatives, thus the max force is ambiguos.
I think we should return the max of abs() value.

image

rocketpy/Flight.py
```diff
@@ -2237,96 +2238,72 @@ def staticMargin(self):
...
+        button, the function returns 0. See the rail buttons documentation to
+        learn more about the alpha angle.
+        """
+        try:
+            alpha = self.rocket.rail_buttons[0].component.angular_position * (
+                np.pi / 180
+            )
+        except IndexError:
+            alpha = 0  # there's no rail button defined, use any alpha value
+        return alpha
```
MateusStano, line 2253: Alpha angle? What is this supposed to be?

Also this is not in the rail buttons documentation (I think....)

rocketpy/Flight.py
```diff
@@ -2531,7 +2508,7 @@ def retrieve_temporary_values_arrays(self):
         return temporary_values
 
     @cached_property
-    def calculate_rail_button_forces(self):
+    def __calculate_rail_button_forces(self):
```
MateusStano, line 2511: Any reason to make this private?

rocketpy/Flight.py
Gui-FernandesBR: Alpha is the angular position of the rail buttons.

I agree this could be better described

rocketpy/Flight.py
Gui-FernandesBR: 1. I just did not want to allow user to get the whole tuple of 4 Function objects.
2. The return is a tuple of 4 elements describing the forces in the rocket coordinates system. This needs to be converted to "rail button" reference system. If we don't convert, we can't call it shear or normal force. I don't think it would be good if we set this "intermediate" result as public.

rocketpy/Flight.py
MateusStano: Okay! I prefer if we just use `angular_position` as the name of this

MateusStano (approved):

PR #384, opened 2023-06-26
FIX: funcify titles - pt1

phmbressan:
## Pull request type

## Pull request checklist

The `funcify_method` decorator from `Function` did not update the titles for methods that returned `Function` instances.

## What is the new behavior?

This issue was fixed by updating the `reset` method from `Function` with a call to `setTitle`.

- [ ] Yes
- [X] No

### DIFF
--- a/rocketpy/Function.py
+++ b/rocketpy/Function.py
@@ -83,21 +83,8 @@ class Function:
         # Set source
         self.setSource(source)
         #  Set function title
-        if title:
-            self.setTitle(title)
-        else:
-            if self.__domDim__ == 1:
-                self.setTitle(
-                    self.__outputs__[0].title() + " x " + self.__inputs__[0].title()
-                )
-            elif self.__domDim__ == 2:
-                self.setTitle(
-                    self.__outputs__[0].title()
-                    + " x "
-                    + self.__inputs__[0].title()
-                    + " x "
-                    + self.__inputs__[1].title()
-                )
+        self.setTitle(title)
+
         # Return
         return None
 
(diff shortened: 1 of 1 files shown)


MateusStano:
I do not think this is fixing the issue.

Running `getting_started.ipynb` I still get the same results as before:

image


image

phmbressan:
> I do not think this is fixing the issue.
> 
> Running `getting_started.ipynb` I still get the same results as before:
> 

Thank you for pointing it out. This is odd since I have run some motor notebooks to test it and they seemed normal. Perhaps there is an specific method in getting started that is triggering it. I will check it ASAP.

phmbressan:
I understand it now. The issue here is that the `beta V.1` branch does not implement `funcify` on the `Motors` methods. Therefore the Function titles will be propagated through many calculations leading to the result above. 

My understanding of the problem was too shallow, I am sorry about that, I tought that it only happened due to `funcify`. I will fix it.

Gui-FernandesBR:
> I understand it now. The issue here is that the `beta V.1` branch does not implement `funcify` on the `Motors` methods. Therefore the Function titles will be propagated through many calculations leading to the result above.
> 
> My understanding of the problem was too shallow, I am sorry about that, I tought that it only happened due to `funcify`. I will fix it.

Don't worry. At least you tried.
Please ask for review again when it is ready, I converted the PR to draft.

There's no problem if we leave this one out of our pre-release

Gui-FernandesBR (approved):
Works on my machine 🧑‍💻

Partial solution only, tho.

Hope to see you solving the bigger issue regarding Function operations titles later :)

PR #385, opened 2023-06-27
ENH: Resetable Funcified Methods

giovaniceotto:
## Pull request type

## Pull request checklist

All methods that use the `funcify_method` decorator from the Function class are cached. Once an underlying quantity that effects the value of these funcified methods is changed, they cannot be easily updated to incorporate the changes made to the class instance.

## What is the new behavior?

A simple function called `reset_funcified_methods` has been created in the Function module. It can be imported by any multable class which makes use of the `funcify_method` decorator.

Both `HybridMotors` and `LiquidMotors` have been updated to incorporate the function `reset_funcified_methods` in end of their ´addTank´ method. 

- [ ] Yes
- [x] No

## Other information

This can be further incorporated in other parts of the code if judged as a good implementation.

This PR does not consider resetting cached properties. In a simple understanding, such methods can be decorated simply by `@property` instead of `@cached_property`. If caching is still useful, a similar approach to what was implemented in the `reset_funcified_methods`  can be done to provide the desired functionality.


rocketpy/Function.py
```diff
@@ -3025,6 +3025,7 @@ def __get__(self, instance, owner=None):
                     )
 
                 val.__doc__ = self.__doc__
+                val.__cached__ = True
```
phmbressan, line 3028: Great solution to mark cached attributes, I would only check if there are not any conflicts with this [standard usage](https://docs.python.org/3/reference/import.html#__cached__).

rocketpy/Function.py
```diff
@@ -3034,6 +3035,27 @@ def __get__(self, instance, owner=None):
...
+    """Recalculates all the funcified methods of the instance. It does so by
+    deleting the current Functions, which will make the interperter redefine
+    them when they are called. This is useful when the instance has changed
+    and the methods need to be recalculated.
+
+    Parameters
+    ----------
+    instance : object
+        The instance of the class whose funcified methods will be recalculated.
+        The class must have a multable __dict__ attribute.
```
phmbressan, line 3048:
```suggestion
        The class must have a mutable __dict__ attribute.
```

rocketpy/motors/HybridMotor.py
```diff
@@ -493,6 +493,7 @@ def addTank(self, tank, position):
         """
         self.liquid.addTank(tank, position)
         self.solid.massFlowRate = self.totalMassFlowRate - self.liquid.massFlowRate
+        reset_funcified_methods(self)
```
phmbressan, line 496: Really clean. Perhaps in the future we can deal with this caching in another way, but this reset really helps to reset the caching when needed and to not miss any funcify by mistake.

phmbressan (approved):

rocketpy/Function.py
```diff
@@ -3034,6 +3035,27 @@ def __get__(self, instance, owner=None):
         return funcify_method_decorator
 
 
+def reset_funcified_methods(instance):
+    """Recalculates all the funcified methods of the instance. It does so by
```
MateusStano, line 3039:
```suggestion
    """Resets all the funcified methods of the instance. It does so by
```

rocketpy/Function.py
```diff
@@ -3034,6 +3035,27 @@ def __get__(self, instance, owner=None):
...
+    deleting the current Functions, which will make the interperter redefine
+    them when they are called. This is useful when the instance has changed
+    and the methods need to be recalculated.
```
MateusStano, line 3042: This actually resets every cached_property, no? Not only funcified ones

rocketpy/motors/HybridMotor.py
MateusStano, line 496: So every time a Tank is added to a motor every cached_property of that motor object gets reset?

This probably works for the Motor classes but it can be dangerous in other places

phmbressan: Could you elaborate more what do you mean by "dagerous"? My first though was exactly the opposite, all the cached parameters are being reset, even if some don't need it, therefore the code is playing safe here to ensure correctness.

rocketpy/Function.py
phmbressan: I don't think so. I believe the `__cached__` tag that was added only apply to `funcify`. They have no parallel in `cached_properties`.

rocketpy/motors/HybridMotor.py
MateusStano: We might reset a property that does not need recalculation and is computationally heavy to calculate. Eventually, that property could need to be recalculated just because of the unnecessary reset. 

I can't think of an example where this would happen though

rocketpy/Function.py
MateusStano: Okay, got it!

rocketpy/motors/HybridMotor.py
phmbressan: I agree with you, although this is not likely to happen in methods that change fundamentally the instance, such as addTank, since most of the cache would need to be updated anyway.

I don't have a good solution for the scenario that you described. In the short term, the `del` can be used to select specific attributes to be deleted.

Gui-FernandesBR: I think it is not dangerous, instead, it is inefficient.  

We don't need to be fast now, let's wait to see if any part of the code starts to increase its runtime due to this function.
We can also address it in the future.

rocketpy/Function.py
Gui-FernandesBR: This link that you pointed out is related to python modules.
It is useful to show to the user where the module is located in the user's machine, specially when working with virtual environments.

Here is an example:

```bash
C:\Users\guiga>python
Python 3.10.9 (tags/v3.10.9:1dd9be6, Dec  6 2022, 20:01:21) [MSC v.1934 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy
>>> numpy.__cached__
'C:\\Users\\guiga\\AppData\\Roaming\\Python\\Python310\\site-packages\\numpy\\__pycache__\\__init__.cpython-310.pyc'
>>> numpy.__file__
'C:\\Users\\guiga\\AppData\\Roaming\\Python\\Python310\\site-packages\\numpy\\__init__.py'
>>>
```

On the other hand, what Giovani has created here is a `__cached__` attribute to the funcifyed methods.
This will allow us to know if the method was cached in the class by the funcify decorator.

The `__cached__` attribute is not being assigned to any module, therefore there are not any conflicts here.

Actually, I think it is pretty intuitive using the name `__cached__` here, since it keeps a degree of consistency with the Python conventions. 

See an example:
image

Gui-FernandesBR:
For future reference, here are the classes that could start using this resetable function: 
- SolidMotor (evaluateGeometry method)
- Rocket (evaluate..... several methods)

Gui-FernandesBR (approved):
Great fix, many thanks @giovaniceotto !

I hereby confirm that I've tested in the notebooks and it was running perfectly.
Before the v1.0.0 we should definitely apply this reset feature to other classes too.

giovaniceotto:
Thanks for all the feedbacks.

@Gui-FernandesBR, I have added the new method to the `evaluateGeometry` method in the `SolidMotor` class. However, I have not added it to the `Rocket` class since it does not have any funcified methods.

I believe all the concerns have already been thoroughly discussed and addressed. Feel free to merge.

rocketpy/motors/HybridMotor.py
giovaniceotto: Not everytime.... Only when the methods have already been cached. Methods are only cached once called. In the general RocketPy workflow, you define everything before actually calling funcified methods.

PR #394, opened 2023-07-30
BUG: Wrong Effective Rail

MateusStano:
## Pull request type

## Pull request checklist

- [ ] Yes
- [x] No

## Information

The calculation for the effective rail length was wrong ever since the Rail Buttons and Distance x Positions PRs were merged.

This causes serious calculations errors in the out-of-rail speed values

### DIFF
--- a/rocketpy/Flight.py
+++ b/rocketpy/Flight.py
@@ -1207,13 +1207,12 @@ class Flight:
 
     @cached_property
     def effective_1rl(self):
-        nozzle = (
-            self.rocket.motor_position - self.rocket.center_of_dry_mass_position
-        ) * self.rocket._csys  # Kinda works for single nozzle
+        nozzle = self.rocket.motor_position
         try:
             rail_buttons = self.rocket.rail_buttons[0]
             upper_r_button = (
-                rail_buttons.component.buttons_distance + rail_buttons.position
+                rail_buttons.component.buttons_distance * self.rocket._csys
+                + rail_buttons.position
             )
         except IndexError:  # No rail buttons defined
             upper_r_button = nozzle
@@ -1222,9 +1221,7 @@ class Flight:
 
     @cached_property
     def effective_2rl(self):
-        nozzle = (
-            self.rocket.motor_position - self.rocket.center_of_dry_mass_position
-        ) * self.rocket._csys
+        nozzle = self.rocket.motor_position
         try:
             rail_buttons = self.rocket.rail_buttons[0]
             lower_r_button = rail_buttons.position


rocketpy/Flight.py
```diff
@@ -1207,13 +1207,12 @@ def __init_equations_of_motion(self):
 
     @cached_property
     def effective_1rl(self):
-        nozzle = (
-            self.rocket.motor_position - self.rocket.center_of_dry_mass_position
-        ) * self.rocket._csys  # Kinda works for single nozzle
+        nozzle = self.rocket.motor_position
```
phmbressan, line 1210: I don't think this is correct. The new behavior of the motor position parameter is not anymore bound to the nozzle. The nozzle position is specified by the motor parameter `nozzle_position`, therefore I believe it should be:

```py
# Example only, not checked throughout
nozzle = self.rocket.motor_position + self.rocket.motor.nozzle_position
```

Furthermore, I am not sure how the coordinate system orientation is handled in your additions, remember that the motor also has an orientation.

phmbressan (changes requested):
Good job in noticing the mistake. I had a few comments regarding the fix, please review them so we clarify the implementation.

rocketpy/Flight.py
MateusStano: Ah right! I forgot about the motor csys

Anyway, I think what you suggested is not right, it should be:

```
nozzle = self.rocket.motor_position + self.rocket._csys*self.rocket.motor.nozzle_position
```

I will add it to the code right away

Also, this makes me think that we need some tests with the Motor and Rocket classes that asserts these position related attributes

MateusStano: Actually, there might be a bigger misunderstanding here:

image

The documentation says that the `motor_position` is the position of the nozzle in relation to the rocket's coordinate system.

Was this meant to be different?

If so, why? To me it makes a lot more sense to input the position of the nozzle outlet in relation to the rocket csys

phmbressan: I believe that the parameter `position` should be position of the `motor` system "zero" (that might be the nozzle) in the Rocket reference frame.

As a related sidenote that may be corrected in another PR, see the `add_tank` method in `LiquidMotor`, the documentation there seems to be also out of date: it says position from the nozzle tip, however this is not what is assumed in code regarding the tanks.

phmbressan (approved):

MateusStano:
Merging this PR now. Fixes to motor position should be coming in a future PR

PR #396, opened 2023-08-12
ENH: Add Maximum Acceleration During Burn Print

MateusStano:
## Pull request type

## Pull request checklist

If a flight simulation happens to have a parachute whose trigger generates an acceleration bigger than the acceleration generated by the motor, only the acceleration due to the parachute trigger will be printed out in the `info` and `all_info` methods

## What is the new behavior?

Now, a print is added in cases where the motor acceleration is not what caused the highest acceleration:

image

Note that when the maximum acceleration **is** caused by the motor, the extra printed lines are not added

- [ ] Yes
- [x] No

## Other information

I also tried adding a print for the acceleration at the `open_time` of the parachute, but the acceleration at that time is not the max acceleration caused by the parachute opening. The max only happens a bit of time after the event and can not easily be determined.

### DIFF
--- a/rocketpy/Flight.py
+++ b/rocketpy/Flight.py
@@ -20,7 +20,7 @@ from scipy import integrate
 from .Function import Function, funcify_method
 from .plots.flight_plots import _FlightPlots
 from .prints.flight_prints import _FlightPrints
-from .tools import Matrix, Vector
+from .tools import Matrix, Vector, find_closest
 
 
 class Flight:
@@ -2164,17 +2164,11 @@ class Flight:
     @cached_property
     def max_acceleration_power_on(self):
         """Maximum acceleration reached by the rocket during motor burn."""
-        closest_value = float("inf")
-        index_to_slice = None
-
-        for index, item in enumerate(self.acceleration):
-            if abs(item[0] - self.rocket.motor.burn_out_time) < abs(
-                closest_value - self.rocket.motor.burn_out_time
-            ):
-                closest_value = item[0]
-                index_to_slice = index
+        burn_out_time_index = find_closest(
+            self.acceleration.source[:, 0], self.rocket.motor.burn_out_time
+        )
         max_acceleration_time_index = np.argmax(
-            self.acceleration[: index_to_slice + 1, 1]
+            self.acceleration[:burn_out_time_index, 1]
         )
         return self.acceleration[max_acceleration_time_index, 1]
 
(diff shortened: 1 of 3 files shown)


rocketpy/Flight.py
```diff
@@ -2161,6 +2161,41 @@ def acceleration(self):
...
+
+        for index, item in enumerate(self.acceleration):
+            if abs(item[0] - self.rocket.motor.burn_out_time) < abs(
+                closest_value - self.rocket.motor.burn_out_time
+            ):
+                closest_value = item[0]
+                index_to_slice = index
+        max_acceleration_time_index = np.argmax(
+            self.acceleration[: index_to_slice + 1, 1]
+        )
```
phmbressan, line 2178: The `for` loop seems to be searching the index of the acceleration time array that is the closest to the burn out time. Since the time array is ordered, wouldn't it be much faster to perform a binary search as follows:

```py
from bisect import bisect_left

def find_closest(ordered_sequence, value):
    pivot_index = bisect_left(ordered_sequence, value)
    if pivot_index == 0:
        return pivot_index
    if pivot_index == len(ordered_sequence):
        return pivot_index - 1
    
    smaller, greater = ordered_sequence[pivot_index-1], ordered_sequence[pivot_index]
    
    return pivot_index - 1 if value - smaller <= greater - value else pivot_index
```

If I understood the problem correctly, this function could be placed in `tools.py` and has performance $\mathcal{O}\left(\log(n)\right)$.

rocketpy/prints/flight_prints.py
```diff
@@ -369,6 +369,28 @@ def maximum_values(self):
...
+        if (
+            self.flight.max_acceleration_time
+            != self.flight.max_acceleration_power_on_time
+        ):
+            print(
+                "Maximum Acceleration During Motor Burn: {:.3f} m/s² at {:.2f} s".format(
+                    self.flight.max_acceleration_power_on,
+                    self.flight.max_acceleration_power_on_time,
+                )
+            )
```
phmbressan, line 383: Why is the maximum acceleration on motor burn only logged when it differs from the overall maximum acceleration? In my opinion, this is quite an useful information that may always be shown.

phmbressan (approved):

Gui-FernandesBR:
I think I would simply split the vector to go from 0s up to the burn time,
and then I used np.argmax. performing loops and searches with dedicated
functions may be faster sometimes.

rocketpy/prints/flight_prints.py
```diff
@@ -356,40 +356,35 @@ def maximum_values(self):
...
+        )
+        print(
+            "Maximum Gs After Motor Burn: {:.3f} g at {:.2f} s".format(
+                self.flight.max_acceleration_power_off
+                / self.flight.env.gravity(
+                    self.flight.z(self.flight.max_acceleration_power_off_time)
+                ),
+                self.flight.max_acceleration_power_off_time,
             )
+        )
```
phmbressan, line 387: Sorry for not noticing this before. I might be mistaken, however I believe accelerations expressed in terms of the unit `g` are computed based of the standard gravity `9.80665 m/s²`.

As a reference guide for what I mean, see: this [Wikipedia page](https://en.wikipedia.org/wiki/G-force#Unit_and_measurement).

rocketpy/Flight.py
```diff
@@ -2164,35 +2164,46 @@ def acceleration(self):
...
+        """Time at which the rocket reaches its maximum acceleration after
+        motor burn."""
+        burn_out_time_index = find_closest(
+            self.acceleration.source[:, 0], self.rocket.motor.burn_out_time
+        )
         max_acceleration_time_index = np.argmax(
-            self.acceleration[: index_to_slice + 1, 1]
+            self.acceleration[burn_out_time_index:, 1]
         )
         return self.acceleration[max_acceleration_time_index, 0]
```
phmbressan, line 2208: These two methods are essentially the same. Personally, I would keep the `max_acceleration_power_off_time` as it is, and use its value in the `max_acceleration_power_off` as

```py
return self.acceleration(self.max_acceleration_power_off_time)
```

Which is similar to the way that the methods `max_speed` and `max_speed_time` are implemented.

Issue #390, opened 2023-07-09
BUG: elliptical fins cannot have a span smaller than the rocket radius


brunosorban:
**Describe the bug**

If a fin from class `EllipticalFins` is added with the `span` value lower than `rocketRadius`, a imaginary number is obtained, thus, it's not a valid value.

**To Reproduce**
Try the following line in a `Rocket` object with `radius=0.55`.

`FinSet = Rocket.addEllipticalFins(3, span=0.13, rootChord=0.13, distanceToCM=-0.48)`

**Expected behavior**

That a positive real value is obtained without throwing any error.

**Screenshots**

![image](https://github.com/RocketPy-Team/RocketPy/assets/68908065/6cf7e4e3-c427-41a1-81f3-7ea547dad0c1)

**Additional context**

PR #398, opened 2023-08-14
BUG: EllipticalFins Roll Damping Factor Calculation

MateusStano:
## Pull request type

## Pull request checklist

An `EllipticalFins` object could be defined if it had the span smaller than the rocket radius, as stated in Issue #390. This was due to the expression used to evaluate the `roll_damping_interference_factor` that wasn't defined at that domain.

## What is the new behavior?

With a whole bunch of calculus, I got the correct expressions for when the span is bigger to the rocket radius and for when the span is equal to the radius. 

The new expressions have been added to the documentation and added to the code accordingly

- [ ] Yes
- [x] No

### DIFF
--- a/docs/technical/aerodynamics/elliptical_fins.rst
+++ b/docs/technical/aerodynamics/elliptical_fins.rst
@@ -159,7 +159,7 @@ When :math:`S < r`:
 
 .. math::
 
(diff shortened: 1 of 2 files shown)


Gui-FernandesBR (approved):
Well done!!

To confirm the correct behavior, I've tested different sets of fins, just to confirm that the roll damping coefficient has a continuous behavior when varied against the span. 

To illustrate:

```python3
from rocketpy import EllipticalFins


def test_elliptical_fins(rc, span, rr):
    fins = EllipticalFins(
        n=4,
        root_chord=rc,
        span=span,
        rocket_radius=rr,
        cant_angle=0,
        airfoil=None,
        name="Fins",
    )
    return fins.roll_damping_interference_factor


span = range(1, 500, 10)
span = [s / 1000 for s in span]
damping = [] * len(span)

for s in span:
    damping.append(test_elliptical_fins(rc=0.2, span=s, rr=0.1))
plt.scatter(span, damping)
plt.plot(span, damping)
plt.xlabel("Span [m]")
plt.ylabel("Roll Damping Interference Factor")
plt.title("Roll Damping Interference Factor vs Span")
plt.xlim(0, 0.5)
plt.show()
```

image

PR #401, opened 2023-09-03
ENH: airfoil outputs (resolve #144)

### LINKED ISSUE #144: Airfoil Rocket.allInfo() method
Not so long ago, we had #140 solving bugs regarding airfoil implementation.
It's true that now rocketpy is dealing very well with these non-flat fins, however I feel we could implement some plots in rocket.py file to show to the user.

Some of them could be:
- Simple plot of airfoil data the user inputed by addFins() method
- Derivative of Lift coefficient by angle of attack

But of course feel free to add any other information you consider important!

Gui-FernandesBR:
## Pull request type

## Pull request checklist

You give an airfoil curve as input, but RocketPy do not let you to check if the curve is correct.

## What is the new behavior?
A few more plots or prints were added to improve the visualization of airfoil data when it is used.
This closes #144 

- [x] No

## Other information
Easy to review.

### DIFF
--- a/rocketpy/AeroSurface.py
+++ b/rocketpy/AeroSurface.py
@@ -566,24 +566,24 @@ class Fins(AeroSurface):
         """
         if not self.airfoil:
             # Defines clalpha2D as 2*pi for planar fins
-            clalpha2D = Function(lambda mach: 2 * np.pi / self.__beta(mach))
+            clalpha2D_incompressible = 2 * np.pi
         else:
-            # Defines clalpha2D as the derivative of the
-            # lift coefficient curve for a specific airfoil
+            # Defines clalpha2D as the derivative of the lift coefficient curve
+            # for the specific airfoil
             self.airfoil_cl = Function(
                 self.airfoil[0],
                 interpolation="linear",
             )
 
             # Differentiating at alpha = 0 to get cl_alpha
-            clalpha2D = self.airfoil_cl.differentiate(x=1e-3, dx=1e-3)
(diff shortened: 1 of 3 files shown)


phmbressan (approved):
Good PR, solves an old issue.

rocketpy/AeroSurface.py
```diff
@@ -576,14 +576,14 @@ def evaluate_lift_coefficient(self):
             )
 
             # Differentiating at x = 0 to get cl_alpha
-            clalpha2D_mach0 = self.airfoil_cl.differentiate(x=1e-3, dx=1e-3)
+            clalpha2D_alpha0 = self.airfoil_cl.differentiate(x=1e-3, dx=1e-3)
```
MateusStano, line 579: This change is wrong here. The `clalpha2D_mach0` is a float, and is independent of the mach number. We then apply a correction to make it a function of `mach` and call it `clalpha2D`

Also, since we are using Barrowman the `clalpha` is constant at all angle of attack

MateusStano: The previous naming is not that good, but its there just to differentiate `clalpha2D_mach0` form `clalpha2D`

Gui-FernandesBR:
> The previous naming is not that good, but its there just to differentiate `clalpha2D_mach0` form `clalpha2D`

Please check the last version of the code (screenshot below).

I believe the `clalpha2D_incompressible` is a clearer name for such variable.
Also, I refactor the code bit so the beta correction is called just once.

image

MateusStano (approved):

PR #399, opened 2023-08-17
FIX: optional imports for EnvironmentAnalysis prevent default install.

phmbressan:
## Pull request type

Please check the type of change your PR introduces:

- [X] Code base additions (bugfix, features)
- [ ] Code maintenance (refactoring, formatting, renaming, tests)
- [ ] ReadMe, Docs and GitHub maintenance
- [ ] Other (please describe):

## Pull request checklist

Please check if your PR fulfills the following requirements, depending on the type of PR:

- Code base additions (for bug fixes / features):

  - [ ] Tests for the changes have been added
  - [ ] Docs have been reviewed and added / updated if needed
  - [X] Lint (`black rocketpy`) has passed locally and any fixes were made
  - [X] All tests (`pytest --runslow`) have passed locally

## What is the current behavior?

The default `rocketpy` install through `pip` does not install the optional dependencies as intended. However, importing the package loaded the `EnvironmentAnalysis` module that required them, raising an error.

## What is the new behavior?

The imports of the optional modules are made lazily, i.e. on instantiation of an `EnvironmentAnalysis` object. Then, the expected install behavior is assured.

## Does this introduce a breaking change?

- [ ] Yes
- [X] No

## Other information

The way the imports were done goes against PEP8 convention of only handling imports in top level statements. However, the implemented way is simpler for just these few dependencies. In the long term, if more complex install conditions are needed, an file to process this requirements (as in `pandas` repository) may be considered.

PR #403, opened 2023-09-06
FIX: env analysis dependency imports

Gui-FernandesBR:
## Pull request type

- [ x] Other (please describe): dependency management

## Pull request checklist

You cannot run "import rocketpy" without getting an error

## What is the new behavior?
Inspired in pandas management of optional dependency, I introduced a short function inside tools.py that will deal with the lazy imports of Environment Analysis class. I hope you all like it!

Please notice that all the optional decencies related to Environment Analysis are actually not crucial to run the class.

Plus: I created the `requirements-optional.txt` to avoid confusions.

- [x] No

## Other information
- This is supposed to be the correct version of the old PR #399 .
- The bug was introduced with the PR #368


Gui-FernandesBR:
This bug was reported in our discord server!

rocketpy/tools.py
```diff
@@ -1231,6 +1231,42 @@ def find_closest(ordered_sequence, value):
...
+    """
+    package_name = name.split(".")[0]
+    try:
+        module = importlib.import_module(name)
+    except ImportError as exc:
+        raise ImportError(
+            f"{package_name} is an optional dependency and is not installed.\n"
+            + f"\t\tUse 'pip install {package_name}' to install it or "
+            + "'pip install rocketpy[all]' to install all optional dependencies."
+        ) from exc
```
phmbressan, line 1266:
```suggestion
    try:
        module = importlib.import_module(name)
    except ImportError as exc:
        package_name = name.split(".")[0]
        raise ImportError(
            f"{package_name} is an optional dependency and is not installed.\n"
            + f"\t\tUse 'pip install {package_name}' to install it or "
            + "'pip install rocketpy[all]' to install all optional dependencies."
        ) from exc
```

rocketpy/tools.py
```diff
@@ -1231,6 +1231,42 @@ def find_closest(ordered_sequence, value):
     return pivot_index - 1 if value - smaller <= greater - value else pivot_index
 
 
+def import_optional_dependency(
+    name,
+):
+    """Import an optional dependency.
+
+    github.com/pandas-dev/pandas/blob/main/pandas/compat/_optional.py
+    If the dependency is not installed, an ImportError is raised.
```
phmbressan, line 1240: There seems to be an issue with this docstring, perhaps part of the first phrase that leads to the pandas link.

I would write something along the lines "based on the implementation found in ...".

rocketpy/tools.py
```diff
@@ -1231,6 +1231,42 @@ def find_closest(ordered_sequence, value):
     return pivot_index - 1 if value - smaller <= greater - value else pivot_index
 
 
+def import_optional_dependency(
+    name,
+):
```
phmbressan, line 1236: For some reason, `black` does not flag this when linting. I don't think there is any need to break the line here.

```suggestion
def import_optional_dependency(name):
```

rocketpy/EnvironmentAnalysis.py
```diff
@@ -382,10 +399,18 @@ def __localize_input_dates(self):
...
+                timezonefinder = import_optional_dependency("timezonefinder")
+                tf = timezonefinder.TimezoneFinder()
+                self.preferred_timezone = pytz.timezone(
+                    tf.timezone_at(lng=self.longitude, lat=self.latitude)
+                )
+            except ImportError:
+                print(
+                    "'timezonefinder' not installed, defaulting to UTC."
+                    + " Install timezonefinder to get local time zone."
+                )
```
phmbressan, line 412: Perhaps, to make the error messages more consistent, suggest the command to install `timezonefinder`.

phmbressan (commented):
Really good implementation. Concise, organized and easily replicable/scalable.

Still a partial review only since I will test it further later, but things are awesome so far.

rocketpy/EnvironmentAnalysis.py
Gui-FernandesBR: Solved: [7cf2435](https://github.com/RocketPy-Team/RocketPy/pull/403/commits/7cf2435ed069f331608bafd471081888d3e40ad4)

rocketpy/tools.py
Gui-FernandesBR: Solved: [7cf2435](https://github.com/RocketPy-Team/RocketPy/pull/403/commits/7cf2435ed069f331608bafd471081888d3e40ad4)

rocketpy/tools.py
Gui-FernandesBR: Solved: [7cf2435](https://github.com/RocketPy-Team/RocketPy/pull/403/commits/7cf2435ed069f331608bafd471081888d3e40ad4)

Gui-FernandesBR:
@phmbressan I was wondering if there's any easy way to test the `import rocketpy` import after building the package with pip install without using the `[all]` option.

Basically the bug passed through our git actions because we always set it to download the package with [all] option.

Idk if this is a matter of solving it with better github actions commands or new creating unit test with pytest. But I think it may be important for us in the future

MateusStano:
> @phmbressan I was wondering if there's any easy way to test the `import rocketpy` import after building the package with pip install without using the `[all]` option.

You can use in Colab:

```
!pip install git+https://github.com/RocketPy-Team/RocketPy.git@fix/env-analysis-dependency-imports#egg=rocketpy[env_analysis]
```

I tested it here and it seems some warnings are raised when even when the packages are installed:

image

Gui-FernandesBR:
> > @phmbressan I was wondering if there's any easy way to test the `import rocketpy` import after building the package with pip install without using the `[all]` option.
> 
> You can use in Colab:
> 
> ```
> !pip install git+https://github.com/RocketPy-Team/RocketPy.git@fix/env-analysis-dependency-imports#egg=rocketpy[env_analysis]
> ```
> 
> I tested it here and it seems some warnings are raised when even when the packages are installed:
> 
> image

Nice catch @MateusStano !

I solved the issue within the last commit. The code was trying to import "windrose>=1.6.8" instead of "windrose", of course it should not work properly.

I think it is running correctly now.

Gui-FernandesBR:
@MateusStano , @phmbressan ,

This PR is ready for a re-review now.
I've to added IPython as an optional requirement for the environment analysis class.
Running the class always through jupyter gave me the false idea that IPython was built-in python module.

I hope the code will be working on your machines as well.

rocketpy/EnvironmentAnalysis.py
```diff
@@ -236,15 +237,19 @@ def __check_extra_requirements(self):
         env_analysis_require = [
             "timezonefinder",
             "windrose>=1.6.8",
+            "IPython>=8.8.0",
             "ipywidgets>=7.6.3",
             "jsonpickle",
         ]
```
phmbressan, line 243: If the package name is splitted and only the first part of the name matters for checking if it is present, why put the version numbers in this list?

I might have overlooked something, but it seems to me that 

```py
        env_analysis_require = [
            "timezonefinder",
            "windrose",
            "IPython",
            "ipywidgets",
            "jsonpickle",
        ]
```

would have the same effect for this function without the need of the versioning. Perhaps it would be useful to keep the versioning and also check if the installed version is compliant, it is likely that the `importlib` has ways to check that.

MateusStano:
Still getting a warning when running on Colab:

image

I think maybe we should lower the IPython version required

rocketpy/EnvironmentAnalysis.py
```diff
@@ -236,15 +237,19 @@ def __check_extra_requirements(self):
...
+            pckg_name = requirement
+            for op in operators:
+                pckg_name = pckg_name.split(op)[0]
+            is_present = importlib.util.find_spec(pckg_name)
+            if is_present is None:
                 print(
-                    f"{requirement:20} is not installed. Some methods may not work."
+                    f"{pckg_name:20} is not installed. Some methods may not work."
                     + " You can install it by running 'pip install rocketpy[env_analysis]'"
                 )
```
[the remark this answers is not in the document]
Gui-FernandesBR: Thanks for solid review.

1. I didn't know that.  See in the last commit how i implemented your suggestion
2. gotcha!
3.  Alright, I created a new ``tools.check_requirement_version()` function, let's see if you like it. Still prefer to keep other function in the env analysis class so we can do dedicated exception management

rocketpy/EnvironmentAnalysis.py
Gui-FernandesBR: Yes I had left the numbers bc I thought in the future it would be nice to have version numbering tests.
I just "hadn't realized by my own" that this future were so close, so I implemented version checks in my last commit. 
This feature can be used in other classes in the future.

Hope you like it now!

Gui-FernandesBR:
> Still getting a warning when running on Colab:
> 
> image
> 
> I think maybe we should lower the IPython version required

Hey @MateusStano , thks for noticing that. I  was wondering if there's any automatic test that we could implement to check if the simulation is running in colab environment. If weren't by you, I would never know the code isn't working there.

Anyway, I have simply dropped the version requirement of ipython. I believe there's no reason for us to demand specific version of this package. 

Ready for a re-review then.

Gui-FernandesBR:
Ready for re-review @phmbressan

Gui-FernandesBR:
I just improved the version comparison of python packages. It seems that using eval() may not work in 100% of the cases, besides being a security risk. 

https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python

phmbressan:
While testing the latest changes, I ran into this import error:

```
Python 3.11.5 (main, Sep  7 2023, 19:25:29) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import importlib
>>> importlib.metadata
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'importlib' has no attribute 'metadata'
>>> 
```

It seems weird that the metadata is not in the standard `__init__` of `importlib`. The tests also were passing, so to make sure the error was not due to a corrupted venv I tested with a new venv and got the same error. I fixed it with my newest commit, now it seems to be working.

phmbressan (approved):
As I mentioned I my partial review, I really enjoyed this approach to organize optional dependencies imports. Well done!

The main change this PR introduces while adding new packages is to ensure the package is also added to the optional dependency dictionary of the corresponding class exactly as it is in the `setup.py`.

Perhaps, in the future, it may be interesting to verify these dependencies automatically with `importlib.metadata.requires('rocketpy')`.

Gui-FernandesBR:
Thank you all for solid reviewing this PR.

I hereby confirm that the code is running locally an in Goggle Collab.

Let's move forward.

PR #413, opened 2023-09-20
ENH: Minor Improvements and Replacement of NoseCone.draw()

MateusStano:
## Pull request type

## Pull request checklist

Warnings are raised when they should not be and the `draw()` method is not in the `aero_surfaces_plot.py` file

## What is the new behavior?

The warnings have been fixed by adding a special check to see if the nosecone is of `elliptical` shape, and the method has been moved to the correct place.

I've also improved the documentation a bit.

- [ ] Yes
- [x] No

## Other information

Special thanks to @CabGT for doing all the hard work here.

### DIFF
--- a/rocketpy/AeroSurface.py
+++ b/rocketpy/AeroSurface.py
@@ -4,7 +4,6 @@ import warnings
 import numpy as np
 
 from .Function import Function
-import matplotlib.pyplot as plt
 import numpy as np
 from scipy.optimize import fsolve
 from .plots.aero_surface_plots import (
--- a/rocketpy/AeroSurface.py
+++ b/rocketpy/AeroSurface.py
@@ -194,10 +194,16 @@ class NoseCone(AeroSurface):
         self._rocket_radius = rocket_radius
         self._base_radius = base_radius
         self._length = length
-        self.bluffness = bluffness
+        if bluffness is not None:
+            if bluffness > 1 or bluffness < 0:
+                raise ValueError(
+                    f"Bluffness ratio of {bluffness} is out of range. It must be between 0 and 1."
+                )
+        self._bluffness = bluffness
         self.kind = kind
 
         self.evaluate_lift_coefficient()
+        self.evaluate_center_of_pressure()
 
         self.plots = _NoseConePlots(self)
         self.prints = _NoseConePrints(self)
@@ -213,6 +219,7 @@ class NoseCone(AeroSurface):
         self._rocket_radius = value
         self.evaluate_geometrical_parameters()
         self.evaluate_lift_coefficient()
+        self.evaluate_nose_shape()
 
     @property
     def base_radius(self):
@@ -223,6 +230,7 @@ class NoseCone(AeroSurface):
         self._base_radius = value
         self.evaluate_geometrical_parameters()
         self.evaluate_lift_coefficient()
+        self.evaluate_nose_shape()
 
     @property
     def length(self):
@@ -232,6 +240,7 @@ class NoseCone(AeroSurface):
     def length(self, value):
         self._length = value
         self.evaluate_center_of_pressure()
+        self.evaluate_nose_shape()
 
     @property
     def kind(self):
@@ -310,6 +319,7 @@ class NoseCone(AeroSurface):
 
         self.evaluate_center_of_pressure()
         self.evaluate_geometrical_parameters()
+        self.evaluate_nose_shape()
 
     @property
     def bluffness(self):
@@ -323,6 +333,7 @@ class NoseCone(AeroSurface):
                     f"Bluffness ratio of {value} is out of range. It must be between 0 and 1."
                 )
         self._bluffness = value
+        self.evaluate_nose_shape()
 
     def evaluate_geometrical_parameters(self):
         """Calculates and saves nose cone's radius ratio.
@@ -409,13 +420,14 @@ class NoseCone(AeroSurface):
         final_shape_vec = np.vectorize(final_shape)
 
         # Create the vectors X and Y with the points of the curve
-        self.nosecone_x = (self.length - (circle_center - r_circle)) * (
+        nosecone_x = (self.length - (circle_center - r_circle)) * (
             np.linspace(0, 1, n) ** p
         )
-        self.nosecone_y = final_shape_vec(self.nosecone_x + (circle_center - r_circle))
+        nosecone_y = final_shape_vec(nosecone_x + (circle_center - r_circle))
 
         # Evaluate final geometry parameters
-        self.length = self.nosecone_x[-1]
+        self.shape_vec = (nosecone_x, nosecone_y)
+        self._length = nosecone_x[-1]
         self.fineness_ratio = self.length / (2 * self.base_radius)
 
         return None
--- a/rocketpy/AeroSurface.py
+++ b/rocketpy/AeroSurface.py
@@ -97,12 +97,27 @@ class NoseCone(AeroSurface):
     ----------
     NoseCone.length : float
         Nose cone length. Has units of length and must be given in meters.
-    NoseCone.rocket_radius : float
-        The reference rocket radius used for lift coefficient normalization,
-        in meters.
     NoseCone.kind : string
         Nose cone kind. Can be "conical", "ogive", "elliptical", "tangent",
         "von karman", "parabolic" or "lvhaack".
+    NoseCone.bluffness : float
+        Ratio between the radius of the circle on the tip of the ogive and the
+        radius of the base of the ogive. Currently only used for the nose cone's
+        drawing. Must be between 0 and 1. Default is None, which means that the
+        nose cone will not have a sphere on the tip. If a value is given, the
+        nose cone's length will be slightly reduced because of the addition of
+        the sphere.
+    NoseCone.rocket_radius : float
+        The reference rocket radius used for lift coefficient normalization,
+        in meters.
+    NoseCone.base_radius : float
+        Nose cone base radius. Has units of length and must be given in meters.
+    NoseCone.radius_ratio : float
+        Ratio between the nose cone base radius and the rocket radius. Has no
+        units. If base radius is not given, the ratio between base radius and
+        rocket radius is assumed as 1, meaning that the nose cone has the same
+        radius as the rocket. If base radius is given, the ratio between base
+        radius and rocket radius is calculated and used for lift calculation.
     NoseCone.name : string
         Nose cone name. Has no impact in simulation, as it is only used to
         display data in a more organized matter.
@@ -158,7 +173,10 @@ class NoseCone(AeroSurface):
         bluffness : float, optional
             Ratio between the radius of the circle on the tip of the ogive and
             the radius of the base of the ogive. Currently only used for the
-            nose cone's drawing. Must be between 0 and 1.
+            nose cone's drawing. Must be between 0 and 1. Default is None, which
+            means that the nose cone will not have a sphere on the tip. If a
+            value is given, the nose cone's length will be slightly reduced
+            because of the addition of the sphere.
         rocket_radius : int, float, optional
             The reference rocket radius used for lift coefficient normalization.
             If not given, the ratio between ``base_radius`` and
@@ -328,8 +346,9 @@ class NoseCone(AeroSurface):
         return None
 
     def evaluate_nose_shape(self):
-        """Calculates and saves nose cone's shape. The shape is saved as two
-        vectors, one for the x coordinates and one for the y coordinates.
+        """Calculates and saves nose cone's shape as lists and reavulates the
+        nose cone's length for a given bluffness ratio. The shape is saved as
+        two vectors, one for the x coordinates and one for the y coordinates.
 
         Returns
         -------
--- a/rocketpy/AeroSurface.py
+++ b/rocketpy/AeroSurface.py
@@ -1,11 +1,12 @@
 from abc import ABC, abstractmethod
+import warnings
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


rocketpy/AeroSurface.py
```diff
@@ -368,11 +313,94 @@ def evaluate_geometrical_parameters(self):
...
+        final_shape_vec = np.vectorize(final_shape)
+
+        # Create the vectors X and Y with the points of the curve
+        self.nosecone_x = (self.length - (circle_center - r_circle)) * (
+            np.linspace(0, 1, n) ** p
+        )
+        self.nosecone_y = final_shape_vec(self.nosecone_x + (circle_center - r_circle))
+
+        # Evaluate final geometry parameters
+        self.length = self.nosecone_x[-1]
```
Gui-FernandesBR, line 399: This method is modifying the length of the NoseCone, right?

Should this affect the position of the C.P.? 

Imagine you create a rocket with a quite long nosecone,
Then you run the draw method, which calls the `evaluate_nose_shape` method, the length will be modified and therefore the flight simulation would have a different behavior.

I'm keeping this short, but try to clearly define if the method actually modifies the CP or not

rocketpy/AeroSurface.py
```diff
@@ -368,11 +313,94 @@ def evaluate_geometrical_parameters(self):
...
+            return self.y_nosecone(x) if x >= x_init else create_circle(x)
+
+        # Vectorize the final_shape function
+        final_shape_vec = np.vectorize(final_shape)
+
+        # Create the vectors X and Y with the points of the curve
+        self.nosecone_x = (self.length - (circle_center - r_circle)) * (
+            np.linspace(0, 1, n) ** p
+        )
+        self.nosecone_y = final_shape_vec(self.nosecone_x + (circle_center - r_circle))
```
Gui-FernandesBR, line 396: You name this variable as `nosecone_y` and `y_nosecone` in the same code, which doesn't feel right.
Also, there's no documentation for these attributes.

I'd recommend creating a single atribute called `coords`, this can be a list of: 
- two items: the list of x coords and the list of y coords.
- n items, with each item being a (xi, yi) pair.

If we manage to create this kind of robust system to store the coordinates, the rocket drawing problem will be tackled really quicker.

rocketpy/AeroSurface.py
```diff
@@ -368,11 +313,94 @@ def evaluate_geometrical_parameters(self):
...
+        # Calculate a function to find the tangential intersection point between the circle and nosecone curve.
+        def find_x_intercept(x):
+            return x + self.y_nosecone(x) * self.y_nosecone.differentiate(x)
+
+        # Calculate a function to find the radius of the nosecone curve
+        def find_radius(x):
+            return (self.y_nosecone(x) ** 2 + (x - find_x_intercept(x)) ** 2) ** 0.5
+
+        # Check bluffness circle and choose whether to use it or not
+        if self.bluffness is None or self.bluffness == 0:
```
Gui-FernandesBR, line 351: Keep in mind that matching the `self.bluffness == 0`  condition is really hard. 
Easier solution is to set a tolerance (let's say, 1e-8) and check if the value is inside the interval.

numpy.isclose

rocketpy/AeroSurface.py
```diff
@@ -873,7 +855,15 @@ def __fin_num_correction(_, n):
             return n / 2
 
     def draw(self):
+        """Draw the fin shape along with some important information, including
```
Gui-FernandesBR, line 858: nice catch!

rocketpy/plots/aero_surface_plots.py
```diff
@@ -71,7 +71,75 @@ def __init__(self, nosecone):
...
     def draw(self):
-        # This will de done in the future
+        """Draw the nosecone shape along with some important information,
+        including the center line and the center of pressure position.
+
+        Returns
+        -------
+        None
+        """
+        self.aero_surface.evaluate_nose_shape()
```
Gui-FernandesBR, line 81: If you call the draw() method N times, this will run the method N times as well, 

Maybe some try/except or even properties and cached_properties would be beneficial here, what do you think?

To be more specific, I think we should retrieve only the coords, not invoke the evaluate_nose_shape method everytime.

Gui-FernandesBR (changes requested):
Overall a good PR, 
I made some comments because I think it is worth to make small adjusts before following through

rocketpy/AeroSurface.py
```diff
@@ -1,11 +1,12 @@
 from abc import ABC, abstractmethod
+import warnings
 
 import numpy as np
 
 from .Function import Function
 import matplotlib.pyplot as plt
```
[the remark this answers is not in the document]
MateusStano: Why is it useless? It is still used in the file

rocketpy/AeroSurface.py
```diff
@@ -156,7 +157,8 @@ def __init__(
             ``rocket_radius`` will be assumed as 1.
         bluffness : float, optional
             Ratio between the radius of the circle on the tip of the ogive and
-            the radius of the base of the ogive.
+            the radius of the base of the ogive. Currently only used for the
+            nose cone's drawing. Must be between 0 and 1.
```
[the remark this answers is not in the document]
MateusStano: Solved in https://github.com/RocketPy-Team/RocketPy/pull/413/commits/49df84bc39fcb8c748d5e2e8cb51f1fd7f367d1a

rocketpy/AeroSurface.py
```diff
@@ -368,11 +313,94 @@ def evaluate_geometrical_parameters(self):
         -------
         None
         """
+
+        # If base radius is not given, the ratio between base radius and
```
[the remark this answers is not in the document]
MateusStano: Solved in https://github.com/RocketPy-Team/RocketPy/pull/413/commits/49df84bc39fcb8c748d5e2e8cb51f1fd7f367d1a

rocketpy/AeroSurface.py
MateusStano: Yeah, this was an overlook. Had to change a bit of code to take into account this length change when there is bluffness

MateusStano: Solved in https://github.com/RocketPy-Team/RocketPy/pull/413/commits/e847406d9df8eb36befc3f1d33fb9e46c25f00e4

rocketpy/AeroSurface.py
MateusStano: Solved in https://github.com/RocketPy-Team/RocketPy/pull/413/commits/e847406d9df8eb36befc3f1d33fb9e46c25f00e4

rocketpy/plots/aero_surface_plots.py
MateusStano: This was done on purpose for a few reasons, but they are no longer valid because of the `length` change that the blufness generates.

Removed this line in https://github.com/RocketPy-Team/RocketPy/pull/413/commits/e847406d9df8eb36befc3f1d33fb9e46c25f00e4

rocketpy/AeroSurface.py
MateusStano: This 0 is just to check when the use inputs the blufness as 0 (which is the same as None). Having a value of 1e-8 to the bluffness does not really make sense but should still work

rocketpy/AeroSurface.py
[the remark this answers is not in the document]
Gui-FernandesBR: Well, not apparently.

image

Gui-FernandesBR:
So, if the length of the NoseCone is being modified when we set a bluffness ratio, do you agree that the length argument is being kinda redundant at the moment?

Should  we convert the length to an optional argument? Just like the sweepLenght and sweepAngle arguments cannot be used at the same time when defining a trapezoidal fin set.
 

Imagine you as an user, you have set a nose_lenght at the beginning, but the simulation have used another. What are the implications for that?

rocketpy/AeroSurface.py
Gui-FernandesBR: ok

rocketpy/AeroSurface.py
```diff
@@ -174,10 +194,14 @@ def __init__(
         self._rocket_radius = rocket_radius
         self._base_radius = base_radius
         self._length = length
+        if bluffness is not None:
```
Gui-FernandesBR, line 197: This is duplicated with lines 330

rocketpy/plots/aero_surface_plots.py
```diff
@@ -71,7 +71,78 @@ def __init__(self, nosecone):
...
-        # This will de done in the future
+        """Draw the nosecone shape along with some important information,
+        including the center line and the center of pressure position.
+
+        Returns
+        -------
+        None
+        """
+        # Create the vectors X and Y with the points of the curve
+        nosecone_x = self.aero_surface.shape_vec[0]
```
Gui-FernandesBR, line 82: This way is faster:

```python
nosecone_x, nosecone_y = self.aero_surface.shape_vec
```

rocketpy/AeroSurface.py
```diff
@@ -368,11 +342,96 @@ def evaluate_geometrical_parameters(self):
...
+        final_shape_vec = np.vectorize(final_shape)
+
+        # Create the vectors X and Y with the points of the curve
+        nosecone_x = (self.length - (circle_center - r_circle)) * (
+            np.linspace(0, 1, n) ** p
+        )
+        nosecone_y = final_shape_vec(nosecone_x + (circle_center - r_circle))
+
+        # Evaluate final geometry parameters
+        self.shape_vec = (nosecone_x, nosecone_y)
```
Gui-FernandesBR, line 429: tuples are immutable in python, but np.array are mutable.

Here we are defining a tuple of 2 np.arrays.

I believe it could be better to define as a list (or an array) instead of a tuple, 

One may also suggest to go ahead and define it as a rocketpy.Function object, but this should not work so well in the case of TrapezoidalFins, so I'd recommed not using Function for now.

rocketpy/AeroSurface.py
MateusStano: It is intended. It is done this way so that the bluffness is checked both when initializing the object and when changing the bluffness parameter (through the setter). We can put these lines of code into a function and call it instead, but I don't think that saves code lines

Also, I think the setters and getters structure in this class is a weird and makes the class more complex than it should be. Something to think about later

rocketpy/AeroSurface.py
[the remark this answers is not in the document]
MateusStano: Ah, I thought you were talking about the warnings. Okay, will be removed soon

rocketpy/AeroSurface.py
MateusStano: Are we ever going to mutate the shape_vec?

I'll change into a list just for safety

Also, I think we could use `PiecewiseFunction` for the fins drawings too. The problem is that making the plots with Matplotlib requires the vectors to be in a specific format, which is different from the format accepted in the Function class.

MateusStano:
> So, if the length of the NoseCone is being modified when we set a bluffness ratio, do you agree that the length argument is being kinda redundant at the moment?

The length parameter is needed to calculate the "modified" length after the bluffness calculations. You could instead receive the fineness ratio from the user, but the same problem would apply because the fineness ratio also changes with bluffness. 

Essentially, we need the length of the nosecone to define the nosecone and, if there is bluffness, calculate the new length.

Adding bluffness to a nosecone causes the length to change, its a geometrical problem. We can make it clearer in the documentation that this happens though.

rocketpy/AeroSurface.py
Gui-FernandesBR: > It is intended. It is done this way so that the bluffness is checked both when initializing the object and when changing the bluffness parameter (through the setter). We can put these lines of code into a function and call it instead, but I don't think that saves code lines

... it does not save code lines but is does avoid repeating code lines. But anyway I don't think this should be something to worry right now. In the future we could create some validators to check if a general input is being used within its valid range (mass, cds, radius, etc.). Right now your code is working and it is not killing anyone.

> Also, I think the setters and getters structure in this class is a weird and makes the class more complex than it should be. Something to think about later

Absolutely, you have no idea about how lost I'm with these setters in AeroSurface.py. 
This is definitely something to discuss later.

rocketpy/AeroSurface.py
Gui-FernandesBR: > Are we ever going to mutate the shape_vec?

I'm not sure, but I think we could need to do some operations with this vector when trying to draw the entire rocket, but let's see.

> I'll change into a list just for safety
> 
> Also, I think we could use `PiecewiseFunction` for the fins drawings too. The problem is that making the plots with Matplotlib requires the vectors to be in a specific format, which is different from the format accepted in the Function class.

Alright, this is something to add to the #222 discussion

Gui-FernandesBR:
> > So, if the length of the NoseCone is being modified when we set a bluffness ratio, do you agree that the length argument is being kinda redundant at the moment?
> 
> The length parameter is needed to calculate the "modified" length after the bluffness calculations. You could instead receive the fineness ratio from the user, but the same problem would apply because the fineness ratio also changes with bluffness.
> 
> Essentially, we need the length of the nosecone to define the nosecone and, if there is bluffness, calculate the new length.
> 
> Adding bluffness to a nosecone causes the length to change, its a geometrical problem. We can make it clearer in the documentation that this happens though.

Hopefully there will be a method in the future to define the nosecone without modifying the length that was given by the user.
When inputting the data, the user should be able to measure what he is providing.

Gui-FernandesBR (approved):

PR #419, opened 2023-09-25
ENH: rocket drawing

Gui-FernandesBR:
## Pull request type

## Checklist

## Current behavior
We know that sometimes it is hard to check if we have correctly defined the rocket before simulating it. 
Also, sometimes we just need to ensure that our rocket looks good on RocketPy.

## New behavior
Just added a method to draw the entire rocket. You can already try the `Rocket.draw()` method.

Here the first example:
image

## Breaking change

- [ ] Yes
- [x] No

## Additional information
Some comments on the reviewing process:
- We still need to test the code under different scenarios. 
- I currently don't know if by any chance we could create an unit test that will check if the plot was created correct. Instead, we keep just checking if the method is called without raising any errors.
- Waiting the review to elaborate some more specific tests, accepting suggestions meanwhile.


rocketpy/plots/__init__.py
Gui-FernandesBR, line 1: I removed these code lines because they were actually not crucial for using the package.
Also, it was contributing to a circular import when I introduced the following code line to the rocket_plots.py file:

```python
from rocketpy.rocket.aero_surface import Fins, NoseCone, Tail
```

rocketpy/plots/rocket_plots.py
```diff
@@ -104,6 +105,181 @@ def thrust_to_weight(self):
...
+                "body": "dimgrey",
+                "fins": "black",
+                "motor": "black",
+                "buttons": "black",
+                "line_width": 2.0,
+            }
+            A full list of color names can be found at:
+            https://matplotlib.org/stable/gallery/color/named_colors
+        """
+        if vis_args is None:
```
Gui-FernandesBR, line 129: I'd like an opinion on this trade-off:

Is it better to let this standard dictionary as the default value of the method? Or just let it as it is here, checking if vis_args was passed?

I believe the Flight.export_data() method does a similar approach to this.

rocketpy/plots/rocket_plots.py
```diff
@@ -104,6 +105,181 @@ def thrust_to_weight(self):
...
+        plt.scatter(cm, 0, color="black", label="Center of Mass", s=30)
+        plt.scatter(cm, 0, facecolors="none", edgecolors="black", s=100)
+
+        cp = self.rocket.cp_position
+        plt.scatter(cp, 0, label="Center Of Pressure", color="red", s=30, zorder=10)
+        plt.scatter(cp, 0, facecolors="none", edgecolors="red", s=100, zorder=10)
+
+        # Set plot attributes
+        plt.title(f"Rocket Geometry")
+        plt.ylim([-self.rocket.radius * 4, self.rocket.radius * 6])
```
Gui-FernandesBR, line 275: This code line may be improved in the future to take the "maximum semi span" of the rocket, usually the sum of the fin span and  the rocket radius. But at least it is working for now.

rocketpy/plots/rocket_plots.py
giovaniceotto: Adding it to the default argument value may simplify documentation. But since you are already describing it in the docstring, I believe the current implementation is fine.

rocketpy/plots/rocket_plots.py
```diff
@@ -104,6 +105,181 @@ def thrust_to_weight(self):
...
+        plt.plot(
+            [x_nozzle, x_last],
+            [0, -y_last],
+            color=vis_args["body"],
+            linewidth=vis_args["line_width"],
+        )
+
+        # Draw nosecone
+        nosecones = self.rocket.aerodynamic_surfaces.get_tuple_by_type(NoseCone)
+        for nose, position in nosecones:
```
giovaniceotto, line 198: Is color the only reason why you need to create different for loops for each aerodynamic surface type?

Gui-FernandesBR: For NoseCone and Tails, yes, 
but Fins got something different due to the rocket.radius

I got suggestion of creating a "draw_surface" function, but didn't stop too much to think about it because it wasn't working

Here is the suggestion anyway: 

```python
def plot_surface(surface, position, y_multiplier=1):
        x_surface = -csys * surface.shape_vec[0] + position
        y_surface = surface.shape_vec[1] * y_multiplier
        surface_type = surface.__class__.__name__
        plt.plot(x_surface, y_surface, color=color_args[surface_type], linewidth=color_args["line_width"])
```

rocketpy/plots/rocket_plots.py
```diff
@@ -104,6 +105,181 @@ def thrust_to_weight(self):
...
+        reverse = csys == 1
+        self.rocket.aerodynamic_surfaces.sort_by_position(reverse=reverse)
+        y_tube, x_tube = [], []
+        for surface, position in self.rocket.aerodynamic_surfaces:
+            if isinstance(surface, (NoseCone, Tail)):
+                # Append the x and y coordinates of the surface shape_vec to the respective lists
+                x_tube.extend((-csys) * surface.shape_vec[0] + position)
+                y_tube.extend(surface.shape_vec[1])
+            if isinstance(surface, Fins):
+                pass
```
phmbressan, line 159: What are these two lines doing? Couldn't you simply not perform the the `Fins` check by simply removing the lines?

rocketpy/plots/rocket_plots.py
```diff
@@ -104,6 +105,181 @@ def thrust_to_weight(self):
...
+            y_tube_negated,
+            color=vis_args["body"],
+            linewidth=vis_args["line_width"],
+        )
+
+        # Get nozzle position (kinda not working for EmptyMotor class)
+        x_nozzle = self.rocket.motor_position + self.rocket.motor.nozzle_position
+
+        # Find the last point of the rocket
+        idx = -1 if csys == 1 else 0
```
phmbressan, line 177: This variable `idx` could have a better name. Perhaps something like `end_index`. The comment description is good, but does not supersedes intention revealing names.

rocketpy/plots/rocket_plots.py
```diff
@@ -104,6 +105,181 @@ def thrust_to_weight(self):
...
+            vis_args = {
+                "background": "#EEEEEE",
+                "tail": "black",
+                "nose": "black",
+                "body": "dimgrey",
+                "fins": "black",
+                "motor": "black",
+                "buttons": "black",
+                "line_width": 2.0,
+            }
```
phmbressan, line 139: Optional comment fix. Notice that you have created a dictionary, but in all of its uses, the keys are called hard-coded. This is indicative that, perhaps, you didn't have a dictionary structure in mind, but something more alike to a `namedtuple`.

rocketpy/rocket/rocket.py
```diff
@@ -2,20 +2,20 @@
...
-from ..mathutils.function import Function
-from ..motors.motor import EmptyMotor
-from ..plots.rocket_plots import _RocketPlots
-from ..prints.rocket_prints import _RocketPrints
-from .aero_surface import (
+from rocketpy.mathutils.function import Function
+from rocketpy.motors.motor import EmptyMotor
+from rocketpy.plots.rocket_plots import _RocketPlots
+from rocketpy.prints.rocket_prints import _RocketPrints
+from rocketpy.rocket.aero_surface import (
```
phmbressan, line 9: Any particular reason for the conversion of the relative imports to absolute?

Gui-FernandesBR: > Any particular reason for the conversion of the relative imports to absolute?

Just to avoid circular import error, it was happening with some modules here, unfortunately. 
I honestly couldn't find where exactly the problem were, but using absolute import solved everything quickly.

Gui-FernandesBR:
Have you ever wondered if there is something better than "spectacular"?
Well, @MateusStano have found it.

Great Work!! I've been able to debug some rockets thanks to the new implementations.

The only flaw I could find so far is the position of the legend, this is a problem with matplotlib in general.
Avoiding legend overlap is sort of annoying task.

image

Gui-FernandesBR:
+ we get the following error if the user do not set the rail buttons:

image

Gui-FernandesBR (commented):
I would approve if I wasn't the PR's author

MateusStano (approved):

rocketpy/plots/rocket_plots.py
Gui-FernandesBR: no longer used

PR #423, opened 2023-10-03
BUG: motor coordinates

phmbressan:
## Pull request type

## Checklist

## Current behavior

A bug was identified by @Gui-FernandesBR regarding `add_motor` coordinate conversion. A legacy assumption of `nozzle_position=0` was being considered in code.

## New behavior

The coordinate conversion was fixed and now is consistent with the documentation: positions are defined with respect to the coordinate system's origin of the objects involved.

A parametrized test was added to check this behavior.

## Breaking change

- [ ] Yes
- [X] No


tests/test_rocket.py
```diff
@@ -379,6 +379,56 @@ def test_add_fins_assert_cp_cm_plus_fins(calisto, dimensionless_calisto, m):
...
+def test_add_motor_coordinates(
+    calisto_motorless,
+    cdm_position,
+    grain_cm_position,
+    nozzle_position,
+    coord_direction,
+    motor_position,
+    expected_motor_cdm,
+    expected_motor_cpp,
+):
```
Gui-FernandesBR, line 402: This test lacks a docstring. Here is a suggestion:

"""
Test the method add_motor and related position properties in a Rocket instance.

This test checks the correctness of the `add_motor` method and the computed 
motor_center_of_dry_mass_position and center_of_propellant_position properties in the 
`Rocket` class using various parameters related to the motor's position, nozzle's position,
and other related coordinates. Different scenarios are tested using parameterization, 
checking scenarios moving from the nozzle to the combustion chamber and vice versa, 
and with various specific physical and geometrical characteristics of the motor.

Parameters
----------
calisto_motorless : Rocket instance
    A predefined instance of a Rocket without a motor, used as a base for testing.
cdm_position : float
    Position of the center of dry mass of the motor.
grain_cm_position : float
    Position of the grains' center of mass.
nozzle_position : float
    Position of the nozzle.
coord_direction : str
    Direction for coordinate system orientation; 
    it can be "nozzle_to_combustion_chamber" or "combustion_chamber_to_nozzle".
motor_position : float
    Position where the motor should be added to the rocket.
expected_motor_cdm : float
    Expected position of the motor's center of dry mass after addition.
expected_motor_cpp : float
    Expected position of the center of propellant after addition.

"""

tests/conftest.py
```diff
@@ -111,16 +111,10 @@ def cesaroni_m1670():  # old name: solid_motor
 
 
 @pytest.fixture
-def calisto(cesaroni_m1670):  # old name: rocket
+def calisto_motorless():
```
Gui-FernandesBR, line 114: Well done with this calisto_motorless fixture, loved.

rocketpy/rocket/rocket.py
```diff
@@ -642,14 +642,14 @@ def add_motor(self, motor, position):
...
         self.motor_center_of_mass_position = (
-            self.motor.center_of_mass - self.motor.nozzle_position
-        ) * _ + self.motor_position
+            self.motor.center_of_mass * _ + self.motor_position
+        )
         self.motor_center_of_dry_mass_position = (
-            self.motor.center_of_dry_mass_position - self.motor.nozzle_position
-        ) * _ + self.motor_position
+            self.motor.center_of_dry_mass_position * _ + self.motor_position
+        )
```
Gui-FernandesBR, line 652: I know this PR is meant to just solve the BUG, so I agree this is already complete.
However, I must comment in advance that in the future we might need add a `self.nozzle_position` attribute to the Rocket class, just for convenience.

Gui-FernandesBR (approved):
Overall a simple and effective solution. I really liked it.

As this is a bug solving, I believe we should merge it directly to the master branch so we don't spend much time until actually getting these changes to the default branch.

tests/test_rocket.py
phmbressan: Well noted. Thanks for the suggestion, added on [c437c9e](https://github.com/RocketPy-Team/RocketPy/pull/423/commits/c437c9ebcb4d279c0f55f46ddc6aed3be3ae5b65).

rocketpy/rocket/rocket.py
phmbressan: Thank you for the reminder. I still think that we should not add this attribute preemptively in this PR, when the need comes the feature is rather simple and the test may be updated accordingly.

MateusStano (changes requested):
I think there are still some things wrong thorought the code. Here is what I found:

- `rocket_prints.py`: Line 97,
- `rocket.py`: Line 102, documentation is wrong
- `flight.py`: `effective_1rl` and effective_2rl` seem wrong
- `flight.py`: `b` and `c` parameters in `u_dot`
- `flight.py`: `r_NOZ` in `u_dot_generalized`
- `flight.py`: in `b` parameter of `rotational_energy`

Btw. The effective rail cached properties were duplicated, I just deleted them here because they are still related to this bug

rocketpy/rocket/rocket.py
```diff
@@ -650,6 +650,7 @@ def add_motor(self, motor, position):
         self.motor_center_of_dry_mass_position = (
             self.motor.center_of_dry_mass_position * _ + self.motor_position
         )
+        self.nozzle_position = self.motor.nozzle_position * _ + self.motor_position
```
Gui-FernandesBR, line 653: We need to add `nozzle_position` to the documentation as well.

rocketpy/simulation/flight.py
```diff
@@ -1171,31 +1171,6 @@ def __init_equations_of_motion(self):
         if self.equations_of_motion == "solid_propulsion":
             self.u_dot_generalized = self.u_dot
 
-    @cached_property
-    def effective_1rl(self):
```
Gui-FernandesBR, line 1175: this was super weird, my god

rocketpy/simulation/flight.py
```diff
@@ -2413,7 +2388,7 @@ def aerodynamic_spin_moment(self):
     def rotational_energy(self):
         # b = -self.rocket.distanceRocketPropellant
         b = (
-            -(self.rocket.motor_position - self.rocket.center_of_dry_mass_position)
+            -(self.rocket.nozzle_position - self.rocket.center_of_dry_mass_position)
```
Gui-FernandesBR, line 2391: I think this is wrong... See the line 2389 right above... the parameter `b` should be the distance from the rocket's CDM to the propellant center of mass.

The way it is implemented right now, it is always computing a constant distance, instead of a floating one.

Gui-FernandesBR: What is written here is actually the r_nozzle

rocketpy/rocket/rocket.py
MateusStano: Done in https://github.com/RocketPy-Team/RocketPy/pull/423/commits/9b44fbaf0204154e5aed4fac1ad7f3a2245f56b4

rocketpy/simulation/flight.py
MateusStano: Fixed in https://github.com/RocketPy-Team/RocketPy/pull/423/commits/81e4d1a44a988675b30e89074283da6e932b69dd

MateusStano (approved):

Gui-FernandesBR (approved):
Approved again, good work here!

Issue #273, opened 2022-11-05
Sensors


giovaniceotto:
To aid in exporting different flight information, such as pressure at a specific point along the fuselage during flight or an acceleration as if measured by an accelerometer at a particular orientation mounted at a particular location in the rocket, this issue proposes the implementation of a `Sensor` abstract class, along if modifications to the `Rocket` and `Flight` class to interface with it.

## Proposed Usage Example

```python
from rocketpy import Rocket, Flight

from rocketpy.sensors import Accelerometer

test_rocket = Rocket(...)

nose_cone_accel = Accelerometer(
    orientation=(1, 0, 0),
    sample_rate=100,
    saturation_values=(-160, 160),
    accuracy=1e-3,
    noise=...,
)

test_rocket.add_sensor(
    sensor=nose_cone_accel,
    position=(x, y, z)
)

def throttle_control():
    if nose_cone_accel.current_value[0] > 100:
        # To fast, throttle_down
        ...      

test_rocket.add_discrete_controller(
    controller=throttle_control
    refresh_rate=100
)  

test_flight = Flight(...)

nose_cone_accel.export_data('accel.csv')
```

## Requirements

1. Support for the following types of sensors:
    - Clock
    - Accelerometer
    - Gyroscope
    - Magnetometer
    - Barometer
    - Thermometer (optional but desired)
    - PitotTube (optional but desired)
    - Luminosity (optional but desired)

2. Support for the following methods/properties:
    - `Sensor.current_value`: to be used during flight by controllers
    - `Sensor.export_data`: to be used after flight

3. Each sensor should be able to emulate:
    - Saturation limits
    - Noise
    - Accuracy
    - Malfunction (random defective values if specified)

## Proposed Architecture

Create a submodule `rocketpy.sensors`. The submodule should have at least the following classes:

- Sensor(abc)
- Accelerometer(Sensor)
- Gyroscope(Sensor)
- Magnetometer(Sensor)
- Barometer(Sensor)

## Proposed Milestone

This feature should be implemented in a major release with a better support for continuous and discrete controllers, along with a refactoring of the `Flight` class.

Therefore, it can be released with **v2.0.0**.

## Context

This idea was first proposed by @Lucas-KB. It was a great idea and we are excited to implement it!

PR #426, opened 2023-10-04
ENH: Air Brakes

MateusStano:
## Pull request type

## Checklist

## New behavior

A simple implementation of an airbrakes simulation has been added. The added code:

- `Class Airbrakes`: Inherits from `AeroSurface`. Defines an airbrakes system with a cd curve that is a function of the deployed level and mach number.
- `Class Controller`: Defines a controller that has a `controller_function`. This function is defined by the user and can alter the state of any number of `observed_objects`. The controller will call the `controller_function` every `sampling_rate` (in simulation time).
- `add_controllers` method in `Rocket`: Similar to `add_surfaces`
- `add_airbrakes` method to `Rocket`: Similar to the aero surfaces `add` methods. Simplifies the addition of the air brakes for the user

## Breaking change

- [ ] Yes
- [x] No


rocketpy/simulation/flight.py
```diff
@@ -1422,6 +1429,30 @@ def u_dot(self, t, u, post_processing=False):
...
+                    ]
+                )
+            else:
+                airbrakes.state_history.append(
+                    [
+                        t,
+                        airbrakes.deployed_level,
+                        0,
+                    ]
+                )
```
MateusStano, line 1455: This is a bit of a work around the interpolation/extrapolation of a 2D `Function` object.

Physically speaking, when the deployed_level is zero, there must be no drag from the airbrakes. However, when passing the cd_curve the `Function` class will usually interpolate the data in a way that there will be drag even when deployed level is 0. 

The `Function` class is interpolating correctly, we just have to add this extra check to guarantee physical coherence

rocketpy/rocket/rocket.py
```diff
@@ -1119,6 +1148,90 @@ def add_parachute(
...
+        reference_area = reference_area if reference_area is not None else self.area
+        airbrakes = Airbrakes(
+            cd_curve=cd_curve,
+            reference_area=reference_area,
+            deployed_level=0,
+            name=name,
+        )
+        # save sampling rate on airbrakes so that it is available to inside
+        # controller function if needed
+        airbrakes.sampling_rate = sampling_rate
```
MateusStano, line 1227: This is a little weird, but I found no better way to make the user have access to the `sampling_rate` when defining its `controller_function`

rocketpy/plots/aero_surface_plots.py
```diff
@@ -448,3 +448,97 @@ def __init__(self, tail):
...
+        # Add labels and a title
+        plt.xlabel("Time (s)")
+        plt.ylabel("Drag Coefficient")
+        plt.title("Drag Coefficient X Time (s)")
+
+        # Show the plot
+        plt.grid(True)  # Add grid lines if needed
+        plt.show()
+
+        return None
```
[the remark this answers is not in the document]
phmbressan: Why not use the `force_data` in the `Function` plot. It should be similar to a scatter plot.

giovaniceotto (changes requested):
This PR looks great. Simple yet powerful implementation of controllers. I was amazed by how such a simple controller could perfectly make the rocket reach the desired apogee.

@phmbressan @brunosorban and I helped in this PR at its infancy, but the example created by @MateusStano is fantastic. Hats off! 🚀

I do have one issue regarding the `airbrakes.state_history` attribute. I am not sure storing the state history in the airbrake instance is perfectly appropriate. I see a couple of problems:

- `airbrakes.state_history` is empty before a Flight instance is created.
- `airbrakes.state_history` may contain the history of multiple flights if various Flight instances are created using it, unless it is manually reset.
- Plots related to `airbrakes.state_history` are totally empty before a flight.

That is why I wonder if this should be stored in the Flight class somehow, or if it is ok to have the airbrake instance mutate during flight. I would like a third opinion on this. @phmbressan @Gui-FernandesBR, any quick thoughts? 

And a quick note regarding sources. I believe that the airbrake drag curve comes from a European Team. If confirmed, we should ask their permission to use the data and cite them.

*If fast approval is needed, we can go ahead and merge this PR and I'll create an issue to handle what is left afterwards. I hereby state that this PR is fully functional as far as I was able to review and test it. What is still necessary is:

- [ ] Tests
- [x] Detailed docs
- [ ] `airbrakes.state_history` discussion
- [x] Proper citation of data sources

rocketpy/rocket/aero_surface.py
```diff
@@ -1889,3 +1892,180 @@ def all_info(self):
...
+        """
+        self.clalpha = Function(
+            lambda mach: 0,
+            "Mach",
+            f"Lift coefficient derivative for {self.name}",
+        )
+        self.cl = Function(
+            lambda alpha, mach: 0,
+            ["Alpha (rad)", "Mach"],
+            "Cl",
```
Gui-FernandesBR, line 2012:
```suggestion
            "Lif Coefficient",
```

This ensures no one ever have any question about CL

rocketpy/rocket/aero_surface.py
```diff
@@ -1889,3 +1892,180 @@ def all_info(self):
         """
         self.prints.all()
         return None
+
+
+class Airbrakes(AeroSurface):
```
Gui-FernandesBR, line 1897: I'm getting an annoying spelling warning everytime it is written "airbrake" (without separation). 

Apparently, both "airbrake" and "air brake" is accpeted, but the second option is more common. Indeed, searching on wikipedia, it is way more common to find it with separation (https://en.wikipedia.org/wiki/Air_brake_%28aeronautics%29).

Also, I wonder if it should be in plural or not. 
- Fins and RailButtons are plural
- Tail and NoseCone are not
  
Any suggestions here?

Therefore, I suggest changing the class name to either `AirBrake` or `AirBrakes`

Gui-FernandesBR: Also, I would be happy if we change every instance of "airbrakes" to "air_brakes" :)

rocketpy/rocket/rocket.py
Gui-FernandesBR: Maybe we could define the sampling rate as an argumnet of the `AirBrake` class (already using the new name sorry hahahaha).

One may suggest that the same aerosurface could be added to different rockets with different sampling rates...
But well, it turns out that the Parachutes also have their sampling rate as an argument!! So I think it would make more sense adding sampling_rate directly to the __init__ method of `AirBrake`

rocketpy/simulation/flight.py
```diff
@@ -1422,6 +1429,30 @@ def u_dot(self, t, u, post_processing=False):
             drag_coeff = self.rocket.power_off_drag.get_value_opt(free_stream_mach)
         rho = self.env.density.get_value_opt(z)
         R3 = -0.5 * rho * (free_stream_speed**2) * self.rocket.area * (drag_coeff)
+        for airbrakes in self.rocket.airbrakes:
```
Gui-FernandesBR, line 1432: Just out of curiosity...
Do you see a case where someone would need to add multiple air brakes to a rocket?

No changes requested, just for personal knowledge.

rocketpy/simulation/flight.py
```diff
@@ -1422,6 +1429,30 @@ def u_dot(self, t, u, post_processing=False):
...
+                )
+                airbrakes.state_history.append(
+                    [
+                        t,
+                        airbrakes.deployed_level,
+                        airbrakes.cd(airbrakes.deployed_level, free_stream_mach),
+                    ]
+                )
+            else:
+                airbrakes.state_history.append(
```
Gui-FernandesBR, line 1449: I wonder what would happen if you simply forget the if statement and use only the next code lines: 

```python
R3 += (
                    -0.5
                    * rho
                    * (free_stream_speed**2)
                    * airbrakes.reference_area
                    * airbrakes.cd(airbrakes.deployed_level, free_stream_mach)
                )
                airbrakes.state_history.append(
                    [
                        t,
                        airbrakes.deployed_level,
                        airbrakes.cd(airbrakes.deployed_level, free_stream_mach),
                    ]
                )
```

Also, the content inside the append is slightly different from one condition to another. Any particular reasons for that?

rocketpy/simulation/flight.py
```diff
@@ -1707,7 +1738,30 @@ def u_dot_generalized(self, t, u, post_processing=False):
         else:
             drag_coeff = self.rocket.power_off_drag.get_value_opt(free_stream_mach)
         R3 += -0.5 * rho * (free_stream_speed**2) * self.rocket.area * (drag_coeff)
-
+        for airbrakes in self.rocket.airbrakes:
```
Gui-FernandesBR, line 1741: duplicating code is bad... 

Should we encapsulate this into another function?

Or maaybe deprecate one of the `u_dot` functions...

rocketpy/simulation/flight.py
```diff
@@ -2001,6 +2055,11 @@ def w3(self):
         """Rocket angular velocity ω3 as a Function of time."""
         return self.solution_array[:, [0, 13]]
 
+    @funcify_method("Time (s)", "Altitude (m/s²)", "spline", "zero")
+    def altitude(self):
+        """Rocket altitude as a Function of time."""
```
Gui-FernandesBR, line 2060:
```suggestion
        """Rocket altitude above ground level (AGL) as a Function of time."""
```

rocketpy/simulation/flight.py
```diff
@@ -3476,6 +3535,20 @@ def add_parachutes(self, parachutes, t_init, t_end):
                 ]
                 self.list += parachute_node_list
 
+        def add_controllers(self, controllers, t_init, t_end):
```
Gui-FernandesBR, line 3538: The `add_parachutes` and `add_controllers` methods look the same, matbe we should have a single function to do the loop that is being done inside the two functions.

What do you think?

rocketpy/plots/aero_surface_plots.py
```diff
@@ -448,3 +448,84 @@ def __init__(self, tail):
...
+        plt.scatter(x, y, marker="o", linestyle="-", color="b", s=0.5)
+        plt.xlabel("Time (s)")
+        plt.ylabel("Deployed Level")
+        plt.title("Deployed Level X Time (s)")
+        plt.grid(True)
+        plt.show()
+
+        return None
+
+    def cd(self):
```
Gui-FernandesBR, line 496: isn't `cd` a too short name for the method? 
Did you consider using `drag_coefficient` instead?

rocketpy/plots/aero_surface_plots.py
```diff
@@ -448,3 +448,84 @@ def __init__(self, tail):
...
+        plt.xlabel("Time (s)")
+        plt.ylabel("Drag Coefficient")
+        plt.title("Drag Coefficient X Time (s)")
+        plt.grid(True)
+        plt.show()
+
+        return None
+
+    def draw(self):
+        raise NotImplementedError
```
Gui-FernandesBR, line 519: nice!

Gui-FernandesBR (changes requested):

Gui-FernandesBR (commented):
Amazing work here! The plots of the rocket reaching the exact target apogee was simply amazing.

Code is working, that's goos. I've made some small comments to ensure the best performance possible. Open to discuss.

Also, I have noticed some small flaws when reviews, so I will commit them directly to save time.

rocketpy/simulation/flight.py
Gui-FernandesBR: > Physically speaking, when the deployed_level is zero, there must be no drag from the airbrakes. However, when passing the cd_curve the Function class will usually interpolate the data in a way that there will be drag even when deployed level is 0.

What if the air brakes are set to not have null drag in the default configuration ("openness level"=0%) ?

Is that possible?

Gui-FernandesBR:
> This PR looks great. Simple yet powerful implementation of controllers. I was amazed by how such a simple controller could perfectly make the rocket reach the desired apogee.
> 
> @phmbressan @brunosorban and I helped in this PR at its infancy, but the example created by @MateusStano is fantastic. Hats off! 🚀
> 
> I do have one issue regarding the `airbrakes.state_history` attribute. I am not sure storing the state history in the airbrake instance is perfectly appropriate. I see a couple of problems:
> 
> * `airbrakes.state_history` is empty before a Flight instance is created.
> * `airbrakes.state_history` may contain the history of multiple flights if various Flight instances are created using it, unless it is manually reset.
> * Plots related to `airbrakes.state_history` are totally empty before a flight.
> 
> That is why I wonder if this should be stored in the Flight class somehow, or if it is ok to have the airbrake instance mutate during flight. I would like a third opinion on this. @phmbressan @Gui-FernandesBR, any quick thoughts?
> 
> And a quick note regarding sources. I believe that the airbrake drag curve comes from a European Team. If confirmed, we should ask their permission to use the data and cite them.
> 
> *If fast approval is needed, we can go ahead and merge this PR and I'll create an issue to handle what is left afterwards. I hereby state that this PR is fully functional as far as I was able to review and test it. What is still necessary is:
> 
> * [ ]  Tests
> * [ ]  Detailed docs
> * [ ]  `airbrakes.state_history` discussion
> * [ ]  Proper citation of data sources

Hey @giovaniceotto , good comments.

- I think about the `AirBrake` as a more sophisticated version of the `Parachute` class. 
- We currently do not save anything about the flight into the Parachute class, 
- However, we need to start thinking about saving flight properties into AeroSurface instances (which is the case of the AirBrake as well). For instance, what was the total angle of attack for each aerodynamic surface during a flight? Instead of just calculating the rocket's total angle of attack.
- The way this PR is proposing is good enough to me. But you're right noticing that the state_history may be empty sometimes, and this could generate some errors when plotting the 
- An alternative is to save an object inside the Flight class with the historical state of every aero_surface/callback associated with that Flight. 

Regarding data:
- Any random air brake curve would fit well here, it is just an example. I don't think he used inappropriate data as you mention. But the easiest ""solution"" is to generate random columns for the example file. It is just an example, don't need to be real data.

rocketpy/rocket/aero_surface.py
MateusStano: Good catch! I will change it to AirBrakes, since this seems to be the most common way to refer to this type of system

giovaniceotto: > Also, I would be happy if we change every instance of "airbrakes" to "air_brakes" :)

Interesting note: during initial development, this was discussed quite extensively, and we even changes from AirBrakes to Airbrakes.

We got this from some EuRoC technical reports... But I have to agree that your references seem more popular!

rocketpy/simulation/flight.py
MateusStano: No, but I guess it is possible .... ?

rocketpy/simulation/flight.py
MateusStano: It is possible, but with our current implementation we assume that a 0% deployed level is the same as the air brakes having no effect in the rocket

I think this is a really reasonable hypothesis, but I will make this clearer in the documentation

rocketpy/rocket/rocket.py
MateusStano: It is a solution, but the thing is that `sampling_rate` is an argument for the `Controller`, which is a separate thing than the air brakes. It makes more sense for the controller to have the sampling rate than the air brakes

If we add `sampling_rate` to the `AirBrakes` init, then there will be two places where we have to pass the same `sampling_rate`, the `Controller` and the `AirBrakes`

The other way to get the `sampling_rate` inside the Controller function would be to have it as a input, meaning the controller function would have 4 inputs now (time, sampling_rate, state, list of objects).

I need opnions on this @Gui-FernandesBR @giovaniceotto @phmbressan

rocketpy/simulation/flight.py
MateusStano: solved in https://github.com/RocketPy-Team/RocketPy/pull/426/commits/42093674c7432469c0b30c9f786c2c44a1974a5b

rocketpy/rocket/aero_surface.py
```diff
@@ -1889,3 +1892,180 @@ def all_info(self):
...
+        )
+        self.reference_area = reference_area
+        self.deployed_level = deployed_level
+        self.prints = _AirbrakesPrints(self)
+        self.plots = _AirbrakesPlots(self)
+
+        # Initialize tracker lists
+        self.state_history = []
+
+    def set_deployed_level(self, deployed_level):
```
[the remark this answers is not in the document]
MateusStano: It already is a property. The default python setter is already doing what we need in this case.

The `set_deployed_level` method was made to make the usage of the air brakes inside the controller function easier. Calling this method is the same as doing `air_brakes.deployed_level = x`. The method just make it more intuitive, and also it is simpler for people coming from other programming languages

rocketpy/rocket/aero_surface.py
MateusStano: Question here, should we change `add_airbrakes` to `add_air_brakes`?

rocketpy/simulation/flight.py
MateusStano: They are not the same, there is a small difference when initializing the TimeNode object

rocketpy/rocket/rocket.py
Gui-FernandesBR: @MateusStano did you solve this one already or not? Please ignore the next lines if you did.

I still don't understand the downsides of this option:

> The other way to get the sampling_rate inside the Controller function would be to have it as a input, meaning the controller function would have 4 inputs now (time, sampling_rate, state, list of objects).

rocketpy/rocket/aero_surface.py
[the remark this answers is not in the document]
Gui-FernandesBR: Trusting you on this.
Good to learn.

rocketpy/rocket/rocket.py
MateusStano: Yes, the sampling rate has been added as an argument to the controller function

MateusStano:
Hey! Changed and improved a whole bunch here:

- https://github.com/RocketPy-Team/RocketPy/pull/426/commits/e0a7aca12c07abdc9f09ab6ecfaae177cd4a91ba -> Changed how the `state_history` works, I am sure this is still going to raise a lot of discussion. Just keep in mind that the problem here is that the air brakes also needs a state vector itself, because it has information that is updated at every step of the simulation. Since we are not changing the `Flight` class core behavior, I think it is fair to save that information in the `AirBrakes` object

- https://github.com/RocketPy-Team/RocketPy/pull/426/commits/e123897b7aa09b146dd0ba3df2e187fc8e4b40a7 -> Added more arguments to the controller function. This should be enough information for any user to define anything. It is a shame that we can not pass the Forces acted upon the rocket here though

- https://github.com/RocketPy-Team/RocketPy/pull/426/commits/e123897b7aa09b146dd0ba3df2e187fc8e4b40a7 and https://github.com/RocketPy-Team/RocketPy/pull/426/commits/7906efae619300a64a6060e972d8319708c0cea3 -> Created `clamp` argument

--- 

I will be adding the tests later, but this is enough for a re-review @phmbressan @Gui-FernandesBR @giovaniceotto

docs/user/airbrakes.rst
Gui-FernandesBR, line 1: Building this page on documentation is taking forever.
Options: 
1. leave it as it is and refactor the documentation in the future to use some kind of fixture that will be loaded only once during the docs building.
2. don't execute the cells during the simulation building process. This will speedup the docs, but will not render the results of each cell.

Gui-FernandesBR (changes requested):
Good work!

rocketpy/rocket/rocket.py
```diff
@@ -1124,6 +1154,101 @@ def add_parachute(
...
+            objects as needed. The function should return None.
+
+            .. note:: The function will be called according to the sampling rate
+            specified.
+
+        sampling_rate : float
+            The sampling rate of the controller function in Hertz (Hz). This
+            means that the controller function will be called every
+            `1/sampling_rate` seconds.
+        clamp : bool, optional
```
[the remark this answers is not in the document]
MateusStano: It is just a term commonly used for limiting a observed variable in controllers. MATLAB uses I think and Pytorch uses it for sure

giovaniceotto (changes requested):
This is quite an extensive PR, so I made a lot of specific comments throughout each file.

In a more general note, I must repeat that this is a groundbreaking contribution to RocketPy. And for this reason (and because I want to avoid future breaking changes as much as possible), I am being extra careful here.

I have a couple of architecture related comments:

- [Discussion topic] Parachute triggers and controller functions seem awfully similar to be separated into two different things inside the Flight class. How about we migrate parachute triggers into controllers? (Can be pursued outside of this PR).
- [Change requested] I believe each controller should have an "overshootable" property. It may be the case that a high refresh rate controller is running but rarely makes any changes to the rocket, and thus can be overshot, while other controllers with low refresh rates may be making consistent changes and cannot be overshot. It does not seem reasonable to run the simulation at the high refresh rate just for this. (This is important both if parachute triggers become controllers and if not).
- [Change requested] I think that all `interactable_objects` should be checked to see if they are really interactable during simulation. We could have special attributes saying if an object is interactable or not. If it is not, a warning or error should be raised. This avoids the user from trying to modify a component that was not meant to be modified.
- [Discussion topic] It seems to me that our controller class is doing some heavy lifting that may not always be necessary: passing the entire state history and storing the observable variables. Don't get me wrong, the solution seems simple and quite useful. However, the user may not make use of either of these things. And those who may need it can implement the same functionality by defining a class and using a method as a controller function. This makes me wonder if an abstract "controller function" class would be the most appropriate solution here. Users could build on top of it by inheriting it. This would bring more flexibility, but it would also require more Python knowledge. (Let me know if this statement is confusing. I can provide a working example if needed.)
- [Painful discussion topic] How would the arguments of the controller function change when sensors are implemented (see #273)? Would we have braking changes?

MateusStano:
> * [Discussion topic] Parachute triggers and controller functions seem awfully similar to be separated into two different things inside the Flight class. How about we migrate parachute triggers into controllers? (Can be pursued outside of this PR).

I think they are two different things, it's just that inside the flight class they are treated almost the same. I think having an improved structure for callbacks would be beneficial. Definitely something to think about in the future


> * [Discussion topic] It seems to me that our controller class is doing some heavy lifting that may not always be necessary: passing the entire state history and storing the observable variables. Don't get me wrong, the solution seems simple and quite useful. However, the user may not make use of either of these things. And those who may need it can implement the same functionality by defining a class and using a method as a controller function. This makes me wonder if an abstract "controller function" class would be the most appropriate solution here. Users could build on top of it by inheriting it. This would bring more flexibility, but it would also require more Python knowledge. (Let me know if this statement is confusing. I can provide a working example if needed.)

I do not understand the issue with passing the `state history` to the `controller_function`. It is just an argument passed to the function, that the user can use if desired. Does passing a big list to a function affect performance somehow? 

I don't think I see how this is a problem, to be honest

> * [Painful discussion topic] How would the arguments of the controller function change when sensors are implemented (see

Well, we always expected that adding Sensors and making a truly robust controller implementation would require breaking changes somehow. I ask if we could even implement Sensors without doing a complete refactor in the `Flight` class, and if we could make a complete refactor in `Flight` without breaking changes.

---

To sum up and answer the remaining questions:

I think we have been looking at the Air Brakes implementation wrong.

This PR adds a controller class, but was never meant to add a solution to controllers. The `Controller` class was initially created as a support for the Air Brakes. Obviously, we would like to make it as robust as possible, as to be able to make the most out of it without the introduction of breaking changes. However, I think it is impossible to make a truly robust controller implementation without a much more structured approach.

As we have discussed a few times before, a good controller implementation would inevitably lead to breaking-changes, or at least be limited if we do not create it with breaking-changes involved in its construction.

The current controller implementation is very limited, and these limitations go way beyond what is necessary for an `AirBrakes` class. Other types of "controllable things" will not suffice with whatever is added here, and would require further changes in many different places in the code. As an example, if we were to add TVC or controlled Parafoil, we would need to change quite a bit in the `Flight` class and even create some new classes. The controller class can pretty much only work with air brakes right now, and maybe a rare few of other aero surfaces.

I see three different ways to implement air brakes, without having to many problems with whatever we make of the `Controller` class:

1. Make the `Controller` class private (`_Controller`) and remove (or private) the `add_controller` from the `Rocket` class;
2. Remove the `Controller` class altogether and include everything it does inside the `AirBrakes` class. This way the `AirBrakes` implementation is contained in itself;
3. Leave it as is, but change the documentation of the `Controller` class to clearly explain its limitations and capabilities.

docs/user/airbrakes.rst
```diff
@@ -0,0 +1,402 @@
...
+
+        # Get previous state from state_history
+        previous_state = state_history[-1]
+        previous_vz = previous_state[5]
+
+        # If we wanted to we could get the returned values from observed_variables:
+        # returned_time, deployed_level, drag_coefficient = observed_variables[-1]
+
+        # Check if the rocket has reached burnout
+        if time > Pro75M1670.burn_out_time:
```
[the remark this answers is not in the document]
MateusStano: You have a good point here. This is actually a pretty big problem that we have overlooked.

Is there any way to nicely solve this, without making significant changes to the `Flight` class?

We could just add some extra calculations to get the accelerations, in case there are air brakes are in the simulation. That is the only way I can think of right now

MateusStano:
Getting back to this one:

- https://github.com/RocketPy-Team/RocketPy/pull/426/commits/c0408cec859b612ae13546dbe99fc0ddfc17268d --> Added option to have the air brakes' cd curve substitute the rocket's power of drag curve. This is important for use cases with wind tunnel data, and maybe even certain CFD simulations

- https://github.com/RocketPy-Team/RocketPy/pull/426/commits/b11b21d0c5140cd310f86f70c4cc6e6229965d14 --> Make the Controller class and all related attributes private

- https://github.com/RocketPy-Team/RocketPy/pull/426/commits/11df71f9c503cf8a8ad1ccbdaae117aa5783bcfb --> This one requires some discussion. Because we are now saving all *variables of interest* in the controller `observed_attributes`, interacting with the controller class is pretty important. However, since we are making this class private, I tried finding a way around the need to interact with it. This was done with the `get_controller_observed_variables` cached property. Please pay special attention to this in your review

Final comment here is that we are definitely going to be missing the #273 Sensors implementation. I believe improving RocketPy's  control capabilities should be our priority from here on out

rocketpy/control/controller.py
```diff
@@ -0,0 +1,132 @@
...
+class _Controller:
+    """A class for storing and running controllers on a rocket. Controllers
+    have a controller function that is called at a specified sampling rate
+    during the simulation. The controller function can access and modify
+    the objects that are passed to it. The controller function also stores the
+    variables of interest in the objects that are passed to it."""
+
+    def __init__(
+        self,
+        interactable_objects,
```
Gui-FernandesBR, line 13: Altough the word 'interactable' makes a lot of sense, it doesn't exist in the most commom English dictionaries. 

https://www.oxfordlearnersdictionaries.com/us/spellcheck/english/?q=interactable

I believe the word you're looking for is "interactive".

I could simply search and replace it but in this case I prefer open this comment to see if you have any different opinion.

rocketpy/control/controller.py
```diff
@@ -0,0 +1,132 @@
...
+            time,
+            self.sampling_rate,
+            state_vector,
+            state_history,
+            self.observed_variables,
+            self.interactable_objects,
+        )
+        if observed_variables is not None:
+            self.observed_variables.append(observed_variables)
+
```
Gui-FernandesBR, line 122: Just to double check here...
It is really the intention to not include a "else" statement?

Something like:

else:
            self.observed_variables = []

rocketpy/control/controller.py
```diff
@@ -0,0 +1,132 @@
...
+    """A class for storing and running controllers on a rocket. Controllers
+    have a controller function that is called at a specified sampling rate
+    during the simulation. The controller function can access and modify
+    the objects that are passed to it. The controller function also stores the
+    variables of interest in the objects that are passed to it."""
+
+    def __init__(
+        self,
+        interactable_objects,
+        controller_function,
```
Gui-FernandesBR, line 14: This is already the `Controller` class, do we really need to point that this is the `controler` function?

Of course, letting an argument to be named as `function` is also not a good idea since this is one special variable name in python.

My suggestion is to change this argument to `func`, which is gives the same understandment to the user while being shorter.

rocketpy/rocket/aero_surface.py
```diff
@@ -1889,3 +1892,210 @@ def all_info(self):
...
+    AirBrakes.name : str
+        Name of the air brakes.
+    """
+
+    def __init__(
+        self,
+        drag_coefficient_curve,
+        reference_area,
+        clamp=True,
+        substitute_rocket_drag_coefficient=False,
```
Gui-FernandesBR, line 1923: This argument name is tooooo long, it starts to get annoying somehow.

Can we change to `override_cd`?

rocketpy/simulation/flight.py
```diff
@@ -2881,6 +2926,21 @@ def retrieve_temporary_values_arrays(self):
...
 
+    @cached_property
+    def get_controller_observed_variables(self):
+        """Retrieve the observed variables related to air brakes from the
+        controllers. If there is only one set of observed variables, it is
+        returned as a list. If there are multiple sets, the list containing
+        all sets is returned."""
+        observed_variables = [
+            controller.observed_variables for controller in self._controllers
+        ]
```
Gui-FernandesBR, line 2923: I'm affraid that multiple controllers can generate duplicates in this list when added at to the same rocket. Would that be a problem?

docs/user/airbrakes.rst
```diff
@@ -0,0 +1,431 @@
...
+    def controller_function(
+        time, sampling_rate, state, state_history, observed_variables, air_brakes
+    ):
+        # state = [x, y, z, vx, vy, vz, e0, e1, e2, e3, wx, wy, wz]
+        altitude_ASL = state[2]
+        altitude_AGL = altitude_ASL - env.elevation
+        vx, vy, vz = state[3], state[4], state[5]
+        mach_number = (vx**2 + vy**2 + vz**2) ** 0.5 / env.speed_of_sound(
+            altitude_AGL - env.elevation
+        )
```
Gui-FernandesBR, line 179: mmmm the mach number should be calculated considering the freestream speed instead of the rocket velocities, right?

In a more sophisticated fashion, we would do `mach = ((vx - wind_u)**2 + (vy - wind_v)**2 + ...)**0.5`.

Te current function works when the wind is zero only, which is the case of the simulation in this page.

docs/user/airbrakes.rst
```diff
@@ -0,0 +1,431 @@
...
+                    new_deployment_level = air_brakes.deployment_level - max_change
+
+                air_brakes.set_deployment_level(new_deployment_level)
+
+            # Return variables of interest to be saved in the observed_variables list
+            return (
+                time,
+                air_brakes.deployment_level,
+                air_brakes.drag_coefficient(air_brakes.deployment_level, mach_number),
+            )
```
Gui-FernandesBR, line 217: I've made some "simplifications" to these code blocks when I updated the air brakes notebook. Do you mind checking the notebook and evaluating if it is worth to implement that in this page as well?

I'm sorry for keeping this short.

Gui-FernandesBR (commented):
Nice changes since the previous reviews. I think we are quite close to conclude this one. I've made some minor comments that I think should contribute to the PR.

Tests could be better (i.e. faster) but I understand it doesn't need to be perfect right now.

The PR already allows users to simulate they airbrakes in RocketPy, which is remarkable!

rocketpy/control/controller.py
MateusStano: Wow, good catch

Ill change it asap

rocketpy/control/controller.py
MateusStano: Yes, that is the intended behavior

rocketpy/control/controller.py
phmbressan: Personally I believe `func` is a bit of a too vague name, of course this is just a nomenclature debate (things are clarified in the description), but my suggestion would be `control_function`.

MateusStano: I get that repeating the name controller can be a little messy, but I think the clarity we gain from it is pretty useful.

I could see this being named  `callback_function`, but honestly I just prefer `controller_function`

rocketpy/simulation/flight.py
MateusStano: What do you mean by duplicates? This is iterating through all controllers, so it is meant to have multiple lists

phmbressan: I find this method quite odd for these reason:

- Is it only useful as a workaround since the `Controller` is private? If yes I believe we should discuss a bit if it is useful creating another method that we will need to maintain for the sake of preventing interaction with the `Controller` class. 
    - My understanding (correct me if I am forgetting something important) is that if the name of the class is already private `_Controller`, then its attributes are "privated". Otherwise we would have to create methods in the `Flight` class for each of the `_Controller` attributes. 
- Properties are almost never named `get...`, they are the getters themselves;
- What is the reasoning behind this method being `cached`? It is not called frequently in the simulation nor it is a particularly slow method (generally there are few controllers).

docs/user/airbrakes.rst
MateusStano: Oof, you are right. I'll add the right calculation for the sake of correctness, but this is a problem with the current controller implementation, and maybe actually the flight class.

Currently our state vector has `(vx vy vz)`, and I believe it would be better to instead use `(free_stream_speed attack_angle sideslip_angle)`. This seems to be the standard when dealing with aerodynamics (aviation uses this as well)

rocketpy/rocket/aero_surface.py
MateusStano: I'll set with `override_rocket_drag` if thats good enough for you

rocketpy/simulation/flight.py
phmbressan: Ok, I understand the reasoning behind the method. Just be sure to ponder over what Gui said about the duplicates. I don't think here the duplicates would be a problem, but a simulation in which two different controllers change the same variable would be quite messy.

Last thing, I am not suggesting to simply let the user engage with the Controller, it is more that I would be ok with it. Your solution is more strict on not allowing breaking change, with is good.

rocketpy/control/controller.py
Gui-FernandesBR: Ok, guess I've lost this one kids,
but no problem

rocketpy/rocket/aero_surface.py
Gui-FernandesBR: Well done!

rocketpy/control/controller.py
Gui-FernandesBR: Solved here: [bf07e2e](https://github.com/RocketPy-Team/RocketPy/pull/426/commits/bf07e2e260a5b8352fceda328a7ae32528e76f5a)

Thanks!

docs/user/airbrakes.rst
Gui-FernandesBR: The state vector is constraining a lot of our developments in the Flight class lately.
I'm sure soon enough we will be modifying it somehow.

rocketpy/rocket/aero_surface.py
```diff
@@ -1889,3 +1888,210 @@ def all_info(self):
...
+            not clamp the deployment level and will instead raise a warning if
+            the deployment level is out of bounds. Default is True.
+        override_rocket_drag : bool, optional
+            If False, the air brakes drag coefficient will be added to the
+            rocket's power off drag coefficient curve. If True, during the
+            simulation, the rocket's power off drag will be ignored and the air
+            brakes drag coefficient will be used for the entire rocket instead.
+            Default is False.
+        deployment_level : float, optional
+            Current deployment level, ranging from 0 to 1. Deployment level is the
```
phmbressan, line 1973: With `Current` here is it meant `initial`?

rocketpy/rocket/aero_surface.py
```diff
@@ -1889,3 +1888,210 @@ def all_info(self):
...
+        self.drag_coefficient_curve = drag_coefficient_curve
+        self.drag_coefficient = Function(
+            drag_coefficient_curve,
+            inputs=["Deployment Level", "Mach"],
+            outputs="Drag Coefficient",
+        )
+        self.reference_area = reference_area
+        self.clamp = clamp
+        self.override_rocket_drag = override_rocket_drag
+        self._deployment_level = deployment_level
```
phmbressan, line 1992: If this `deployment_level` is an initial value, I would run it through the standard property setter, as `self.deployment_level = deployment_level`.

This way out of bound initial inputs are clamped if this behavior was selected.

rocketpy/rocket/aero_surface.py
```diff
@@ -1889,3 +1888,210 @@ def all_info(self):
...
+            else:
+                # Raise warning if clamp is False
+                warnings.warn(
+                    f"Deployment level of {self.name} is smaller than 0 or "
+                    + "larger than 1. Extrapolation for the drag coefficient "
+                    + "curve will be used."
+                )
+        self._deployment_level = value
+
+    def set_deployment_level(self, deployment_level):
```
phmbressan, line 2018: Given the existance of the `deployment_level.setter`, is this method still necessary?

rocketpy/rocket/rocket.py
```diff
@@ -1124,6 +1159,145 @@ def add_parachute(
...
+        sampling_rate,
+        clamp=True,
+        reference_area=None,
+        initial_observed_variables=None,
+        override_rocket_drag=False,
+        return_controller=False,
+        name="AirBrakes",
+        controller_name="AirBrakes Controller",
+    ):
+        """Creates a new air brakes, storing its parameters such as
```
phmbressan, line 1175: I believe this description should be updated to match the input names more closely, since they don't have a `trigger function`.

rocketpy/rocket/aero_surface.py
MateusStano: No, this changes throughout the simulation

rocketpy/rocket/aero_surface.py
MateusStano: Good catch

rocketpy/rocket/aero_surface.py
MateusStano: It is there just for usability purposes. Makes it easier to make it cleared what is happening in the controller function for examples. It is technically not needed though.

Do you think we should remove it?

rocketpy/rocket/rocket.py
MateusStano: Good catch!

rocketpy/rocket/aero_surface.py
phmbressan: I find it odd to have the property and a setter, so I would remove it.

If you really prefer keeping it, just remember that we should only change the property.setter and keep the current method only calling the property.setter.

Gui-FernandesBR: I am with Bressan on this one. Please refer to this past comment in this same PR. Sorry not getting into too much details.

image

rocketpy/simulation/flight.py
Gui-FernandesBR: Thanks for solving this one.

phmbressan (approved):
Amazing pull request, these are truly some of the most interesting graphs I have seen in RocketPy.

Great to see more than 100 comments on this PR request with inputs and suggestion.

Superb work @MateusStano in putting this all together.

Gui-FernandesBR (approved):
LGTM, @MateusStano whenever you feel confortable, please merge it.

rocketpy/prints/aero_surface_prints.py
```diff
@@ -291,3 +291,16 @@ def geometry(self):
             f"Angular position of the buttons: {self.aero_surface.angular_position:.3f} deg\n"
         )
         return None
+
+
+class _AirBrakesPrints(_AeroSurfacePrints):
```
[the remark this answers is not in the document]
Gui-FernandesBR: Well, although I agree we could improve the `_AirBrakesPrints`, I don't think it's necessary to be done in this PR.

Three reasons why:
- The PR is already too long, hard to maintain
- This feature can be easily implemented in a different PR in a short-term future.
- The lacking of prints in this class does not prevent the user to use simulate air brakes in their rockets.

giovaniceotto (approved):
A masterpiece.

It is hard to make it 100% future proof, but this is definitely the best that could have been done for now. I am eager to see how control will evolve in RocketPy. Couldn't ask for a better start.

PR #441, opened 2023-10-14
ENH: new Flight.get_solution_at_time() method

Gui-FernandesBR:
## Pull request type

## Checklist

## Current behavior
If by any chance you want to know the state vector that corresponds to a specific time of the flight, there's no easy way to do so.

## New behavior
There's this new method that returns the numpy array of the solution state at a specific time. If the user choose a time that do not exists in the timesteps, a warning is raised and the code automatically changes to the closest time step.

## Breaking change

- [x] No

## Additional information
- This feature is particularly important for us to work on multi-stage and deployable flights.
- I was not sure if it was better to return the np.array object or a list. I chose array because I thought it could be more useful and faster, but honestly this was arbitrary and I don't think there is a big difference between the two options.
- as I am creating this PR on windows, I wonder if the integrators of the Flight class will give the exact same behavior on linux when I upload this to github actions. Let's see!

### DIFF
--- a/rocketpy/simulation/flight.py
+++ b/rocketpy/simulation/flight.py
@@ -1934,7 +1934,7 @@ class Flight:
         """Returns time array from solution."""
         return self.solution_array[:, 0]
 
-    def get_solution_at_time(self, t):
+    def get_solution_at_time(self, t, atol=1e-3):
         """Returns the solution state vector at a given time. If the time is
         not found in the solution, the closest time is used and a warning is
         raised.
@@ -1943,6 +1943,10 @@ class Flight:
         ----------
         t : float
             Time in seconds.
+        atol : float, optional
+            Absolute tolerance for time comparison. Default is 1e-3. If the
+            difference between the time and the closest time in the solution
+            is greater than this value, a warning is raised.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


rocketpy/simulation/flight.py
```diff
@@ -1934,6 +1934,33 @@ def time(self):
...
+        Returns
+        -------
+        solution : np.array
+            Solution state at time t. The array follows the format of the
+            solution array, with the first column being time like this:
+            [t, x, y, z, vx, vy, vz, e0, e1, e2, e3, omega1, omega2, omega3].
+
+        """
+        time_index = find_closest(self.time, t)
+        if abs(self.time[time_index] - t) > 1e-5:
```
phmbressan, line 1956: The way this is implemented here is good, but I would like to raise discussion about the following points:

- It might be convenient to have a `tol=1e-5` parameter;
- The counter argument would be that this parameter has no effect in code, only toggles a warning.

I mentioned this, because this value `1e-5` seems quite arbitrary. Is this warning really important?

phmbressan (approved):

rocketpy/simulation/flight.py
Gui-FernandesBR: I just thought it would be wrong if the user sets a time but we take the state of a time close but not equal to the time he/she set.

I think adding this tolerance parameter could help, I will make it optional.

Gui-FernandesBR: Solved by [422bb51](https://github.com/RocketPy-Team/RocketPy/pull/441/commits/422bb51cf1ffcb1a586f79f7c063addaa103eb0b)

PR #446, opened 2023-10-29
DOCs: linter corrections to Function class

Gui-FernandesBR:
## Pull request type

## Checklist

## Current behavior
- The `Function` class is hard to read: code is mostly dated from 4 years ago, some list operations takes several code lines
- The black linter doesn't cover all the problems reported by pylint, which makes my IDE throw several errors and fix suggestions. 
- Some docstrings are missing
- Some variables still use camelCase

## New behavior
- Small changes, no break changes, less errors, more readability!
- Overall pylint score in the function.py file:
    - before: 7.25/10
    - after: 9.55/10 
- Also, added `.flake8` and `.pylintrc` files to set the configuration for our project. Ideally, this should be 
- For the future: introduce `isort` and `mypy` integration.
- **All of these is improving the general quality of our code source, putting us in a better position as a python library**

## Breaking change
- [x] No

## Additional information
- There is still a major problem in this file which is the too generic exception inside the funcify decorator, as recently pointed out by @phmbressan  I left a TODO for the future.
- Please notice the diff shows 800 added lines, but more 600 are just the new .pylintrc file, which is generated automatically and doesn't need to be reviewed. So this PR essencially introduces 200 new code lines.
- Waiting for review if the action get checked


Gui-FernandesBR:
Hey, I'd like to discuss this suggestion:

image

Gui-FernandesBR:
Should we add a page in the documentation describing how to use the linters during the development phases?

.flake8
MateusStano, line 1: Are the `.flake8` and `.pylintrc` files needed? If so, why?

MateusStano (changes requested):

MateusStano:
> Hey, I'd like to discuss this suggestion:

I prefer `plot1D` over `plot1d`, in fact I think that's why I did not change it in the snake case PR. But I get that it technically is not snake case

If `plot1d` is the best way forward then I think the suggestion is the best way to go

rocketpy/mathutils/function.py
```diff
@@ -954,7 +954,9 @@ def to_frequency_domain(self, lower, upper, sampling_frequency, remove_dc=True):
         >>> time = np.linspace(0, 10, 1000)
         >>> signal = np.sin(2 * np.pi * main_frequency * time)
         >>> time_domain = Function(np.array([time, signal]).T)
-        >>> frequency_domain = time_domain.to_frequency_domain(lower=0, upper=10, sampling_frequency=100)
+        >>> frequency_domain = time_domain.to_frequency_domain(
+        ...     lower=0, upper=10, sampling_frequency=100
+        ... )
```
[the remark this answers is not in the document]
Gui-FernandesBR: Works fine for me.

image

.flake8
Gui-FernandesBR: They are important to set the configurations for flake8 and pylint in the whole project. This is useful when one is working with other developers in the same project and wants to check linters results. 

From now on, we can edit these files to set custom linter rules for the RocketPy repo.
The size of these files are also irrelevant.

rocketpy/mathutils/function.py
```diff
@@ -1420,47 +1420,42 @@ def __interpolate_polynomial__(self):
...
         # Store coefficients
-        self.__spline_coefficients__ = np.array([y[0:-1], b, c[0:-1], d])
+        self.__spline_coefficients__ = np.vstack([y[:-1], b, c[:-1], d])
 
     def __interpolate_akima__(self):
         """Calculate akima spline coefficients that fit the data exactly"""
         # Get x and y values for all supplied points
-        x = self.x_array
-        y = self.y_array
+        x, y = self.x_array, self.y_array
```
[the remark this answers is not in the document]
Gui-FernandesBR: If you do not feel confortable with the changes after reading the code, please consider that there's a `test_integral_spline_interpolation` test function to cover the spline interpolations, so we should be free of any possible errors here.

I understand your concern as the diff changes of this function are not exactly easy to read, but I think that the final version is better in both speed and readability.

Gui-FernandesBR:
> > Hey, I'd like to discuss this suggestion:
> 
> I prefer `plot1D` over `plot1d`, in fact I think that's why I did not change it in the snake case PR. But I get that it technically is not snake case
> 
> If `plot1d` is the best way forward then I think the suggestion is the best way to go

I understand that you prefer one option above other. But I think we should try to be as intuitive as possible with the methods names. If rocketpy follows snake_case, then it is easier to imagine that te function is called `plot1d` than `plot1D`. Imagine the case where you have to spell the function name to someone else, that would be confusing I guess. Also, I cannot see any advantage for using the `plot1D` name besides of your preference, can you?

Alternatively, we could also write `plot1d = plot1D` in any part of this file so both options are accepted, but this is a sort of a workaround.

rocketpy/mathutils/function.py
```diff
@@ -2286,6 +2273,7 @@ def differentiate(self, x, dx=1e-6, order=1):
             return (
                 self.get_value(x + dx) - 2 * self.get_value(x) + self.get_value(x - dx)
             ) / dx**2
+        return None
```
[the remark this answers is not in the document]
Gui-FernandesBR: Solved. Thanks for noticing this.

rocketpy/mathutils/function.py
```diff
@@ -144,8 +154,8 @@ def set_source(self, source):
         # Import CSV if source is a string or Path and convert values to ndarray
         if isinstance(source, (str, Path)):
             # Read file and check for headers
-            f = open(source, "r")
-            first_line = f.readline()
+            with open(source, mode="r", encoding="utf-8") as f:
+                first_line = f.readline()
```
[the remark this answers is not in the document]
Gui-FernandesBR: Same comment as in #442

.pylintrc
```diff
@@ -0,0 +1,636 @@
...
+# List of plugins (as comma separated values of python module names) to load,
+# usually to register additional checkers.
+load-plugins=
+
+# Pickle collected data for later comparisons.
+persistent=yes
+
+# Minimum Python version to use for version dependent checks. Will default to
+# the version used to run pylint.
+py-version=3.11
```
phmbressan, line 91: Should we change this line to RocketPy minimum supported version (`Python 3.8`)?

rocketpy/mathutils/function.py
```diff
@@ -1420,47 +1419,42 @@ def __interpolate_polynomial__(self):
...
             return self.set_interpolation("spline")
         # Create coefficient matrix1
-        A = np.zeros((degree + 1, degree + 1))
+        matrix = np.zeros((degree + 1, degree + 1))
         for i in range(degree + 1):
-            A[:, i] = x**i
+            matrix[:, i] = x**i
         # Solve the system and store the resultant coefficients
-        self.__polynomial_coefficients__ = np.linalg.solve(A, y)
+        self.__polynomial_coefficients__ = np.linalg.solve(matrix, y)
```
phmbressan, line 1426: I know PyLint flags this, but between `matrix` and `A` I almost prefer `A` since it is a common "standard" in mathematical interpolation equation. However, perhaps an even better name would be: `sys_coefficient` or `vandermond_matrix`.

See: https://en.wikipedia.org/wiki/Polynomial_interpolation#Second_proof

rocketpy/mathutils/function.py
```diff
@@ -2080,7 +2066,7 @@ def __rpow__(self, other):
         Parameters
         ----------
         other : int, float, callable
-            What self will exponentiate.
+            The object that will be exponentiate by function.
```
phmbressan, line 2069:
```suggestion
            The object that will be exponentiated by function.
```

phmbressan (approved):

rocketpy/mathutils/function.py
```diff
@@ -68,22 +73,26 @@ def __init__(
         -------
         None
         """
+        # initialize variables to avoid pylint W0201
```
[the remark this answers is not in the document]
Gui-FernandesBR: You're 100% right.

I updated the commentary tho, bc I think I would be someone asking why is the initialization of those variables important.

See [3fa58a1](https://github.com/RocketPy-Team/RocketPy/pull/446/commits/3fa58a1831229c8084b99581a0c1099017aa5cf4)

.pylintrc
Gui-FernandesBR: Nice catch! I didn't see this line.

Solved here: [3fa58a1](https://github.com/RocketPy-Team/RocketPy/pull/446/commits/3fa58a1831229c8084b99581a0c1099017aa5cf4)

rocketpy/mathutils/function.py
Gui-FernandesBR: Honestly, I'm not a fan of the idea of violating the [style guide](https://docs.rocketpy.org/en/latest/development/style_guide.html#code-style) using "mathematical excuses" as this is not a final attribute.
I can quickly pass my eyes through this function and reading `A` is harder than the complete form `matrix_A`.

Given your link, it seems like vandermond_matrix is another thing (the X), so I think we could stay with `sys_coeffs`.

This variable occurs only 3 times in the code, not the end of the world.

I solved the problem here: [3fa58a1](https://github.com/RocketPy-Team/RocketPy/pull/446/commits/3fa58a1831229c8084b99581a0c1099017aa5cf4)
Open for future discussions.

rocketpy/mathutils/function.py
Gui-FernandesBR: solved here: [3fa58a1](https://github.com/RocketPy-Team/RocketPy/pull/446/commits/3fa58a1831229c8084b99581a0c1099017aa5cf4)

Gui-FernandesBR:
All good now. Just waiting for @MateusStano 's feedback regarding my recent changes and further discussions on the `plot2D` snake case situation.

MateusStano (approved):

Gui-FernandesBR:
As discussed, we are going adopt the approach of raising deprecation and renaming the function to snake case.

Gui-FernandesBR:
The `--runslow` option is running perfectly here too.

Issue #443, opened 2023-10-27
BUG: Function not working properly if interpolation method is set for n-d arrays


brunosorban:
**Describe the bug**
If a `Function` object is created using, for example, a 3D array and the user sets the interpolation method different from `shepard`, it is accepting only 1 input when it should accept 2. There should be a protection so the user cannot define the interpolation method if a n-d array is given as input.

**To Reproduce**

![image](https://github.com/RocketPy-Team/RocketPy/assets/68908065/e63e10f0-6b8a-4b8f-8b72-44a3321db2e9)

**Expected behavior**

![image](https://github.com/RocketPy-Team/RocketPy/assets/68908065/50795c0d-5017-4837-aaa8-fab257112c2f)

Also, the extrapolation method must be manually set for n-D arrays, meaning it is not optional.

![image](https://github.com/RocketPy-Team/RocketPy/assets/68908065/663f0f8b-7369-4254-84d4-6b8afa7743ac)

Gui-FernandesBR:
@brunosorban did you consider this commentary?

https://github.com/RocketPy-Team/RocketPy/blob/0d0079718753fe7b3f9262b63271ddef4ac50e7e/rocketpy/mathutils/function.py#L54

Gui-FernandesBR:
Just for reference:

```python
>>> from rocketpy import function
>>> a = [(0,0,1), (1,1,2)]
>>> f = Function(a, ['x', 'y'], ['z'])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "RocketPy\rocketpy\mathutils\function.py", line 82, in __init__
    self.set_source(source)
  File "RocketPy\rocketpy\mathutils\function.py", line 237, in set_source
    self.set_interpolation("shepard")
  File "RocketPy\rocketpy\mathutils\function.py", line 290, in set_interpolation
    self.set_get_value_opt()
  File "RocketPy\rocketpy\mathutils\function.py", line 331, in set_get_value_opt
    raise ValueError(f"Invalid extrapolation type {self.__extrapolation__}")
ValueError: Invalid extrapolation type None
>>>
>>> f = Function(a, ['x', 'y'], ['z'], extrapolation="natural") 
>>> f(0,1)
1.5
```

- Ok, it's true that it throwing an error when you do not set the extrapolation type, this should be documented or fixed somehow.
[...]

Issue #444, opened 2023-10-28
ENH: Support for type hints


Gui-FernandesBR:
**Is your feature request related to a problem? Please describe.**

Type hints were a great addition of Python 3.5 (https://docs.python.org/3/library/typing.html), and it has been improved in the latest python 3.12 version (https://docs.python.org/3/whatsnew/3.12.html). Here are some points to consider:

* Type hints improve code readability as it makes clear what are the expected values. 
* Enhances static type checking (mypy) to prevent potential errors before execution.
* IDEs can use these type hints and this can extremely help you when coding.
* It's easier to review a code when you know exactly the expected types of each argument (we add docstrings for this same reason).
* All of these provide a better development experience

**Describe the solution you'd like**

* There are some solutions on the market that would allow us to automatically generate the type annotations for all the functions, for example, https://github.com/Instagram/MonkeyType 
* Also, [mypy](https://mypy.readthedocs.io/en/stable/#) could be used to check the annotations as a linter.
* We should implement type hints in every single function or method. This would be a long process, so I'd say we should introduce this gradually, maybe opening PRs class by class to allow better .
* Ideally, every new function added to the code should have type hints, ensuring we achieve a good support for type hints.

**Additional context**

- Discussion needed here.
- Type hints do not ensure you're code is free of errors, so it is not a "pill of safety" for users. That's why I think this could benefit devs more than users.
- For a simplified introduction to type hints, see [PEP 483](https://peps.python.org/pep-0483/)

Gui-FernandesBR:
We decided to **not** adopt type hints and annotations in this repository as they would create a need for maintenance that is higher than the benefit.

We prefer to exhaustively document our code with doc string, which allows for a good documentation page as well.

Can be re-discussed in the future.

PR #451, opened 2023-11-01
BUG: User input checks added for Function class

brunosorban:
Enhanced Validation and Default Settings via `_check_user_input` Method for the `Function` class. The features added comprise:

## Pull request type

## Checklist

## Current behavior

See issue https://github.com/RocketPy-Team/RocketPy/issues/443.

## New behavior
**Type Checking and Conversion for Source Data:**
The method now checks for list and np.ndarray the the data is consistent. It uses NumPy conversion to an array of type np.float64 for input validation. This ensures consistent data type handling and error triggering for non-homogeneous or non-numeric arrays.

**Output Parameter Validation:**
Enhanced checks for the `outputs` parameter have been added. If `outputs` is a string, it's converted into a list. Additionally, the method now raises a `ValueError` if outputs has a dimension greater than 1, enforcing strict dimensionality constraints.

**Input, Interpolation, and Extrapolation Handling for Multidimensional Functions:**
For `sources` with dimensions greater than 2:
The inputs are automatically set to ["Input 1", "Input 2", ...] if they were initially ["Scalar"], and a warning is issued to inform the user of this default setting.
The interpolation method defaults to `"shepard"` for multidimensional `Functions`, with a warning if any other method is specified, indicating limited support for interpolation methods.
The `extrapolation` is set to "natural" if not specified, with a corresponding warning for clarity and default behavior.

**Source Dimensionality Consistency Check:**
The method now verifies the consistency between the dimensions of `source` and the combined length of `inputs` and `outputs`. It raises a `ValueError` if there's a mismatch, ensuring congruence between source data and expected `input`/`output` dimensions.

**Function Source Handling:**
When the source is a function, the method checks and updates the `inputs` and `outputs` to match those of the source function, issuing warnings if there are discrepancies.

These updates aim to make the method more intuitive, user-friendly, and error-resistant, providing more precise feedback and automatic user adjustments. The added validations and default settings should minimize common configuration errors and guide users toward correct usage patterns, particularly in handling multidimensional data and function sources.

**Added Argument Count Check:**
The code now includes a conditional check to compare the length of args (arguments passed to the function) against the expected domain dimensionality (self.__dom_dim__).
If the number of arguments (len(args)) does not match the expected domain dimension (self.__dom_dim__), the function raises a ValueError.
The error message clearly states the discrepancy, indicating both the expected number of arguments and the number provided. For example: "This Function takes 3 arguments, 2 given."

## Breaking change

- [ ] Yes
- [x] No

## Additional information

These checks should avoid common user errors, but more can always be added. Please feel free to give feedback so we can improve it even more.


rocketpy/mathutils/function.py
```diff
@@ -1,9 +1,11 @@
 from inspect import signature
```
Gui-FernandesBR, line 1: Please use [isort](https://pycqa.github.io/isort/) to sort the imports in a .py file.

rocketpy/mathutils/function.py
```diff
@@ -19,12 +21,12 @@ class Function:
 
     def __init__(
         self,
-        source,
```
Gui-FernandesBR, line 22: Type hints are not supported yet (although I like your approach), so I'm afraid we need to remove the hints before merging.

rocketpy/mathutils/function.py
```diff
@@ -2612,6 +2622,77 @@ def compose(self, func, extrapolate=False):
                 extrapolation=self.__extrapolation__,
             )
 
+    def _check_user_input(
```
Gui-FernandesBR, line 2625: The function doesn't access or set any attribute of the class. So I think we should convert it to a [staticmethod](https://www.geeksforgeeks.org/class-method-vs-static-method-python/). By doing this, we would allow for using the method without initializing the Function class, which may be useful if someone wants to validate the arguments

rocketpy/mathutils/function.py
```diff
@@ -2612,6 +2622,77 @@ def compose(self, func, extrapolate=False):
...
+            in_out_dim = len(inputs) + len(outputs)
+            if source_dim != in_out_dim:
+                raise ValueError(
+                    f"Source dimension ({source_dim}) does not match input and output dimension ({in_out_dim})."
+                )
+
+        # if function, check for inputs and outputs
+        if isinstance(source, Function):
+            # check inputs
+            if inputs != None and inputs != source.get_inputs():
```
Gui-FernandesBR, line 2683: Try to use "is not None" instead of "!= None", this improves readability in Python.

rocketpy/mathutils/function.py
```diff
@@ -2612,6 +2622,77 @@ def compose(self, func, extrapolate=False):
...
+        # if list or ndarray, check for dimensions, interpolation and extrapolation
+        if isinstance(source, (list, np.ndarray)):
+            # this will also trigger an error if the source is not a list of numbers or if the array is not homogeneous
+            source = np.array(source, dtype=np.float64)
+
+            # check dimensions
+            source_dim = source.shape[1]
+
+            # check output type and dimensions
+            if isinstance(outputs, str):
```
Gui-FernandesBR, line 2644: I think these two lines should be at a higher level of this method. I suggest moving this to line 2634.

Also, you can do the same thing for the inputs. You're gonna see that this will fix some tests ;)

rocketpy/mathutils/function.py
```diff
@@ -2612,6 +2622,77 @@ def compose(self, func, extrapolate=False):
...
+                if inputs == ["Scalar"]:
+                    inputs = [f"Input {i+1}" for i in range(source_dim - 1)]
+                    warnings.warn(
+                        f"Inputs not set, defaulting to {inputs} for multidimensional functions.",
+                    )
+
+                if interpolation != None and interpolation != "shepard":
+                    interpolation = "shepard"
+                    warnings.warn(
+                        "Interpolation method for multidimensional functions is set to 'shepard', currently other methods are not supported.",
```
Gui-FernandesBR, line 2664: Small comment but still valuable...
Some comment and warning message lines are larger than 88 characters. We'd appreciate if you break it into more lines please.

rocketpy/mathutils/function.py
```diff
@@ -2612,6 +2622,77 @@ def compose(self, func, extrapolate=False):
...
+                    )
+
+                if extrapolation == None:
```
Gui-FernandesBR, line 2667:
```suggestion
                if extrapolation is None:
```

rocketpy/mathutils/function.py
```diff
@@ -2612,6 +2622,77 @@ def compose(self, func, extrapolate=False):
...
+            # check interpolation and extrapolation
+            if source_dim > 2:
+                # check for inputs and outputs
+                if inputs == ["Scalar"]:
+                    inputs = [f"Input {i+1}" for i in range(source_dim - 1)]
+                    warnings.warn(
+                        f"Inputs not set, defaulting to {inputs} for multidimensional functions.",
+                    )
+
+                if interpolation != None and interpolation != "shepard":
```
Gui-FernandesBR, line 2661:
```suggestion
                if interpolation not in [None , "shepard"]:
```

This is more direct and readable

Gui-FernandesBR (changes requested):
Hey @brunosorban very nice job starting this one. I think we are going to the correct target.
I have some comments on your additions, please feel free to ask for further details if needed.

a task list before merging:

- [ ] Different tests are not passing. Maybe you were under the false impression since only the black linter. We need to fix that before proceeding (54 failed, 777 passed, 22 skipped, 35 warnings, 69 errors in 49.32s)
- [ ] Add unit tests to cover the new `_check_user_input` method
- [ ] Document the `_check_user_input` (docstrings)
- [ ] Ensure that all input validations are being done inside the new method and there's no redundant comparisons in the rest of the class.

rocketpy/mathutils/function.py
brunosorban: Done

rocketpy/mathutils/function.py
brunosorban: Updated

rocketpy/mathutils/function.py
brunosorban: Done

rocketpy/mathutils/function.py
brunosorban: You cannot make `source_dim = source.shape[1]` if the input is a scalar, which currently is a supported inputs from function class

rocketpy/mathutils/function.py
brunosorban: Updated

rocketpy/mathutils/function.py
brunosorban: Updatex

rocketpy/mathutils/function.py
brunosorban: Updated

brunosorban: I applied the requested changes but I don't have the information about the testes you said it didn't pass and the unit tests

rocketpy/mathutils/function.py
brunosorban: I didn't understand which errors you want to change, but feel free remove the ones that are not supported

Gui-FernandesBR: Type hinting (https://realpython.com/lessons/type-hinting/) is a Python feature that we didn't adopt in RocketPy yet. We had some discussions on top of #444 recently.

Example:
image

I'm removing the hints and annotations from this PR.  Again, I appreciate the intention, but we decided for not accepting hinting in this library yet, this will make maintainers life easier.

rocketpy/mathutils/function.py
Gui-FernandesBR: RocketPy adopts pytest suit for maintaining unit tests that cover small parts of the code. The GitHub actions are used to run the pytest module on every Pull Request. It's possible to check that the tests didn't pass: 

https://github.com/RocketPy-Team/RocketPy/pull/451/checks
image

I'm going to fix it as soon as I understand where the bug is.

Gui-FernandesBR (approved):
Approved by now. More unit tests could be added to improve the code, but I manages to cover at least the very basic functionality of the new function.

There was another problem in the plot2D function, already solved in my last commit.

To be used in the future:
https://pytest-with-eric.com/introduction/pytest-assert-exception/#:~:text=Exceptions%20can%20be%20tested%20in,doesn%27t%20raise%20an%20exception.

Gui-FernandesBR:
As I was the last person to edit the PR and my changes were kinda delicate, I'd like to wait for another review here before merging it into the master branch.

@MateusStano @phmbressan , could one of you review this one? Please don't get biased by my approved review, this one is important for the future releases...

rocketpy/mathutils/function.py
```diff
@@ -2646,6 +2719,152 @@ def compose(self, func, extrapolate=False):
...
+        if isinstance(outputs, str):
+            outputs = [outputs]
+        if isinstance(inputs, str):
+            inputs = [inputs]
+
+        elif len(outputs) > 1:
+            raise ValueError(
+                "Output must either be a string or have dimension 1, "
+                + f"it currently has dimension ({len(outputs)})."
+            )
```
phmbressan, line 2806: I don't know if this is considered a higher dimensional output, but this code runs fine:

```py
import numpy as np
from rocketpy import Function

def source(x):
    return np.array([x, x+1, x+2])

f = Function(source)

f.plot()
```

image

I can even differentiate it:

```bash
>>> print(f.differentiate(5))
array([1., 1., 1.])
```

rocketpy/mathutils/function.py
```diff
@@ -1225,13 +1297,14 @@ def plot2D(
         x = np.linspace(lower[0], upper[0], samples[0])
         y = np.linspace(lower[1], upper[1], samples[1])
         mesh_x, mesh_y = np.meshgrid(x, y)
-        mesh_x_flat, mesh_y_flat = mesh_x.flatten(), mesh_y.flatten()
-        mesh = [[mesh_x_flat[i], mesh_y_flat[i]] for i in range(len(mesh_x_flat))]
+        mesh = np.column_stack((mesh_x.flatten(), mesh_y.flatten()))
+
         # Evaluate function at all mesh nodes and convert it to matrix
-        z = np.array(self.get_value(mesh)).reshape(mesh_x.shape)
+        z = np.array(self.get_value(mesh[:, 0], mesh[:, 1])).reshape(mesh_x.shape)
```
phmbressan, line 1303: This line is causing a mistake due the inner workings of shepard interpolation do not expect the data in this format. What I noticed:

1. Shepard iterator check needs to be updated to handle ndarrays. This is easy I may be done with `collections.abc.Iterable` or adding `np.ndarray` to the list of `isinstance`. However this alone does not fix the mistake.
2. The `x` data from shepard expects a stacked format instead of the first argument list. Still working on a solve.

Just so you can compare:

Before

image


After

image

phmbressan (changes requested):
Great checks for input consistency here, but I caught an error in plotting 2D functions due to shepard interpolation that must be fixed before merging.

Requesting changes for now so nobody merges, I will try to fix the Plot2D error meanwhile.

Gui-FernandesBR:
@phmbressan please improve tests so we prevent from getting the same error in the future.

phmbressan:
I have fixed the plotting error that I mentioned on my earlier review and implemented some tests for the values of `shepard` interpolation to prevent this from happening again.

I would rather if someone double checked the changes before merging. Could you review these @Gui-FernandesBR ?

rocketpy/mathutils/function.py
phmbressan: Solved by [0f76b3e](https://github.com/RocketPy-Team/RocketPy/pull/451/commits/0f76b3e4c4a9c35155a6543e6d59afae6e7ce7b3).

rocketpy/mathutils/function.py
```diff
@@ -749,18 +756,83 @@ def get_value(self, *args):
...
-                    return [self.source(arg) for arg in args[0]]
-            elif len(args) == 1 and isinstance(args[0], np.ndarray):
-                return self.source(args[0])
+            # if the function is 1-D:
+            if self.__dom_dim__ == 1:
+                # if the args is a simple number (int or float)
+                if isinstance(args[0], (int, float)):
+                    return self.source(args[0])
+                # if the arguments are iterable, we map and return a list
+                if isinstance(args[0], (list, tuple, np.ndarray)):
```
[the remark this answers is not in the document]
phmbressan: Applied in [322e332](https://github.com/RocketPy-Team/RocketPy/pull/451/commits/322e33268c8784183c11f66bcf1f35734c10eadb).

rocketpy/mathutils/function.py
Gui-FernandesBR: image

phmbressan: Indeed it seems to be a half supported feature, since it cannot be made with array based Functions and there is this odd R1->R1.

The plots work since `matplotlib` supports multiple values on the "y".

I agree moving this PR foward since implementing fully this feature would be out of scope of this PR.

Gui-FernandesBR: Yes, agreed.

Just realized something here... So the "outputs" argument should always be a list of one single string. Right?

And should we flag a TODO do understand (and maybe remove) this ValueError raise in this part of the code?

Gui-FernandesBR (approved):
Great job. Overall a good correction

Gui-FernandesBR:
@phmbressan please solve your request changes and merge it into develop. We have other branches coming up tomorrow,

rocketpy/mathutils/function.py
phmbressan: Indeed, if the Function only supported output dimension is one, then the `outputs` argument should only be a list of a single string.

I do think that we can support multidimensional outputs due the fact `numpy.ndarray` bahaves quite well in this cases, with a behavior basically undistinguished from a number (e.g. `array([1,1]) + array([1,1]) => array([2,2])`).

phmbressan (approved):

rocketpy/mathutils/function.py
Gui-FernandesBR: Does that kind of Function make sense to you? A mathematical operator that takes 1 value in the domain and associate it to 3 different values (3 different images?).

phmbressan: I believe mathematically it is a well defined operation, but cannot see right now an application for it in the code.

PR #455, opened 2023-11-07
ENH: Spherical Caps Included in Total Length

MateusStano:
## Pull request type

## Checklist

## Current behavior

When creating a cylindrical tank with spherical caps, the `height` argument only refers to the height of the tank, _excluding_ the tank. This is counterintuitive and just forces the user to measure/calculate impractical stuff.

Also, this was not explicit in the documentation

## New behavior

The `height` argument now refers to the entire tank, with or without spherical caps

## Breaking change

- [ ] Yes
- [x] No

## Additional information

Seems there was also a bug with some geometries due to numerical errors in the calculations of the spherical caps, so really long tanks could not be defined. This was fixed by adding an `abs` to the calculation of the spherical caps geometry


Gui-FernandesBR:
I understand your motivation and it is fair enough to me.

However, this PR technically introduces a breaking change since the `add_spherical_caps` is not a private method and you're adding a new mandatory argument to it. I recommend removing the new `total_height` argument. but it would also be possible to set this argument as optional.

Also, the code is considering that the caps will have a perfect sphere at the its both tips. If you give a blind search, it's almost impossible to find a perfect sphere format (https://www.google.com/search?q=cylindrical+tank&sca_esv=581032301&tbm=isch&source=lnms&sa=X&ved=2ahUKEwiw-9vHl7iCAxUpJrkGHYFKBBIQ_AUoAXoECAIQAw&biw=1358&bih=642&dpr=1). Usually it has a height different than the tank radius. Shall we start considering this phenome as well?

rocketpy/motors/tank_geometry.py
```diff
@@ -369,28 +369,37 @@ def __init__(self, radius, height, spherical_caps=False, geometry_dict=dict()):
         height : float
             Height of the cylindrical tank, in meters.
         spherical_caps : bool, optional
-            If True, the tank will have spherical caps. The default is False.
+            If True, the tank will have spherical caps. The height of the
+            tank is maintained. The default is False.
```
phmbressan, line 373: I would expand this documentation (and of the height parameter) to be clearer on what is being considered as height when the spherical caps is either True of False.

Of course this PR already helps on this parameter being more consistent and intuitive, but clear docs will make sure that no confusions are made.

rocketpy/motors/tank_geometry.py
```diff
@@ -369,28 +369,37 @@ def __init__(self, radius, height, spherical_caps=False, geometry_dict=dict()):
...
-        self.add_geometry((-height / 2, height / 2), radius)
-        self.add_spherical_caps() if spherical_caps else None
+        if spherical_caps:
+            self.add_geometry((-height / 2 + radius, height / 2 - radius), radius)
+            self.add_spherical_caps(height)
+        else:
+            self.add_geometry((-height / 2, height / 2), radius)
 
-    def add_spherical_caps(self):
+    def add_spherical_caps(self, total_height):
```
phmbressan, line 385: I would remove this parameter since this value of `total_height` might have any value so that the Tank geometry may be completely reshaped. Therefore I don't think this should be the behavior of a method that simply add spherical caps.

I implemented this method so that someone that made an instance of a CylindricalTank without caps can add standard caps to it should he want. Therefore, in this reasoning, it made sense to change the total height of the tank since its geometry was changed (without any need for the `total_height` parameter).

MateusStano:
@Gui-FernandesBR @phmbressan added the warning as we talked

Please give a re-review here

rocketpy/motors/tank_geometry.py
```diff
@@ -369,28 +369,46 @@ def __init__(self, radius, height, spherical_caps=False, geometry_dict=dict()):
...
-        """
+        and at the top of the tank with the same radius as the cylindrical
+        part. The height is not modified, meaning that the total volume of
+        the tank will decrease.
+        """
+        # raise warning
+        print(
+            "Warning: Adding spherical caps to the tank will not modify the "
+            + "height of the tank."
+        )
```
phmbressan, line 398: A would rather the warning was a bit more clear on what is the method outcome, for example:

```suggestion
        print(
            "Warning: Adding spherical caps to the tank will not modify the "
            + f"total height of the tank {self.height}."
            + f"Its cylindrical portion height will be reduced to {self.height - 2*self.radius}."
        )
```

phmbressan (approved):
Good, this way the object creation seems much more intuitive. The docstrings are also much clearer.

rocketpy/motors/tank_geometry.py
```diff
@@ -369,28 +369,46 @@ def __init__(self, radius, height, spherical_caps=False, geometry_dict=dict()):
...
         """
         Adds spherical caps to the tank. The caps are added at the bottom
-        and at the top of the tank. If the tank already has caps, it raises a
-        ValueError.
-        """
+        and at the top of the tank with the same radius as the cylindrical
+        part. The height is not modified, meaning that the total volume of
+        the tank will decrease.
+        """
+        # raise warning
```
Gui-FernandesBR, line 394: this comment is not necessary

```suggestion
```

Gui-FernandesBR (approved):
Nice problem solving, the code is clearer now, I'm comfortable with its final state.

PR #457, opened 2023-11-09
ENH: Clean Plots and Prints Subpackages init files

MateusStano:
## Pull request type

## Checklist

## Current behavior

Because of some circular imports errors, the `__init__.py` file of the `plots` subpackage was changed to be empty. This is not an issue because the plots classes are only meant to be imported in other files of the library. However, the `CompareFlights` class is in this subpackage, and it is meant to be easily imported.

## New behavior

The `prints` subpackage `__init.py__`  file is also empty and the imports for the Compare classes have been added

## Breaking change

- [ ] Yes
- [x] No

## Additional information

Just to make things clear. Currently, in `master` we can do:

```
from rocketpy.plots import _EnvironmentPlots
```

or

```
from rocketpy.plots.environment_plots import _EnvironmentPlots
```

and for compare we had to do

```
from rocketpy.plots.compare import CompareFlights
```

With the changes of this PR, we can do only

```
from rocketpy.plots.environment_plots import _EnvironmentPlots
```

and for compare we now can do

```
from rocketpy import CompareFlights
```

or 

```
from rocketpy.plots import CompareFlights
```


rocketpy/__init__.py
```diff
@@ -35,3 +35,4 @@
     TrapezoidalFins,
 )
 from .simulation import Flight
+from .plots.compare import Compare, CompareFlights
```
Gui-FernandesBR, line 38: I'm not a big fan of this code line. I think restricting the user to access the Compare class from only the `plots` module would allow for a cleaner `rocketpy.__init__.py` file.

However, it is a matter of preference since we don't have specific rules for that. So go ahead and merge it. 

Just keep in mind that *it is not mandatory* to expose all the classes to the main __init__.py file

Gui-FernandesBR (approved):
Well done!
The plots and prints modules are more consistent now.

rocketpy/__init__.py
phmbressan: I also prefer accessing it only through the plots module, but the way it is done is also fine and not a deal breaker for this PR. I believe it is more a matter of preference/style than correctness.

phmbressan (approved):

PR #463, opened 2023-11-15
DOC: refactor dispersion analysis notebook

giovaniceotto:
## Pull request type

## Checklist

## Current behavior

Dispersion Analysis Notebook does not work in v1 since it still uses camelCase.

## New behavior

Minor changes were made to the notebook to support v1.

## Breaking change

- [ ] Yes
- [x] No

## Additional information

Due to time constrains, the number of simulations executed were just 100.
Fake data was used for new parameters such as motor casing inertia tensor.


Gui-FernandesBR (approved):
Works on my machine. Great solve, @giovaniceotto !!

A couple of commentaries tho:
- Remember to apply black before pushing. I wonder if our linter should also be blaking .ipynb files. To use black on notebooks: `pip install black[jupyter]` 
-  The `parachute_drop_from_helicopter.ipynb`example is still not updated, but no worries.

Issue #284, opened 2022-11-12
ENH: Built-in compare flights


Gui-FernandesBR:
**Is your feature request related to a problem? Please describe.**

More and more teams keep asking us about RocketPy performance. More specifically, they usually want to compare rocketpy simulations against other software or even real flight data. 

**Describe the solution you'd like**

Simple but flexible, create methods that receive data from other simulators or from avionics logs. And then it should 
- [x] Create a "Flight" object from recorded flight data (avionics or other software)
- [x] Compare rocketpy Flight against the user data
- [x] Plot and Print comparisons (should be easy with the comparison subplot)

**Additional context**

See [RocketPaper](https://github.com/RocketPy-Team/RocketPaper) repository to find some (perhaps) useful codes with the same purpose.

Gui-FernandesBR:
Altough not exactly the same problem, the #252 is kind related to this one

Monta120:
Hi! I’d like to contribute to Issue #284. Since FlightDataImporter already handles importing external flight data, I can implement the next layer: a FlightComparator class that takes a RocketPy Flight and imported flight data and automatically generates comparison plots (altitude, velocity, acceleration, attitude) and summary error metrics (RMS error, max deviation, etc.).

This would provide the built-in “compare flights” functionality requested in the issue. Let me know if this aligns with what you expect.

PR #474, opened 2023-11-20
ENH: create FlightDataImporter class

Gui-FernandesBR:
## Pull request type

## Checklist

## Current behavior
See issue #284 . 

## New behavior
I created this `DataFrameToFlight` class that I'd like to discuss before going to develop. I'm still struggling to find the best architecture for this, but at least it is a first step.

## Breaking change

- [x] No


Gui-FernandesBR:
To see an example of the new class being used, see this branch: https://github.com/RocketPy-Team/RocketPy/tree/enh/add-juno3-example

rocketpy/simulation/df2flight.py
```diff
@@ -0,0 +1,208 @@
...
+        The interpolation method to be used in the rocketpy.Function objects.
+        The default is "linear".
+    extrapolation : str, optional
+        The extrapolation method to be used in the rocketpy.Function objects.
+        The default is "zero".
+    """
+
+    name: str
+    time_col: str
+    df: pd.DataFrame
```
giovaniceotto, line 120: Pandas is quite a heavy library (> 10 MB while RocketPy is < 300 kB). So, adding it as import (which will significantly slow down importing this file) just to define the type of a data class attribute seems a bit exaggerated to me.

In fact, rumor says that even if you don't import pandas, this line still runs fine since python completely ignores type annotations, as long as what is written is correct syntax wise (see https://death.andgravity.com/dataclasses).

So, I suggest two options:

1. Don't import pandas and keep `df: pd.DataFrame`. Works for python, may cause problems with linters.
2. Don't import pandas and change `df: pd.DataFrame` to `'df: pd.DataFrame'`. Works for python, has less chance of causing problems with linters.

```suggestion
    df: 'pd.DataFrame'
```

rocketpy/simulation/df2flight.py
```diff
@@ -0,0 +1,208 @@
...
+from dataclasses import dataclass
+from functools import cached_property, wraps
+
+import numpy as np
+
+from rocketpy.mathutils import Function
+from rocketpy.tools import import_optional_dependency
+
+pandas = import_optional_dependency("pandas")
+pd = pandas
```
giovaniceotto, line 14:
```suggestion
```

rocketpy/simulation/df2flight.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+def df2property(
```
giovaniceotto, line 17: If this decorator only works inside the `DataFrameToFlight` class, wouldn't it be better to move it there?

See https://www.geeksforgeeks.org/creating-decorator-inside-a-class-in-python/

Also, consider other names such as `df_2_function_property` or similar. I feel we are missing the Function class in the name, since it is core part of this decorator.

rocketpy/simulation/df2flight.py
```diff
@@ -0,0 +1,208 @@
...
+    decorator
+        The decorator to be used in the DataFrameToFlight class. This only works
+        in the DataFrameToFlight class.
+    """
+
+    class Decorator:
+        """A decorator class that wraps a function and provides create a
+        property from a data frame. This is made via class because it's the only
+        way to use the __set_name__ method. This method is used to get the name
+        of the property in the class.
```
giovaniceotto, line 48:
```suggestion
        """A decorator class that wraps a method to create a property from a data
        frame, as a Function instance.
        This is made via class because it's the only way to use the __set_name__
        method. This method is used to get the name of the property in the class.
```

rocketpy/simulation/df2flight.py
```diff
@@ -0,0 +1,208 @@
...
+                source, "Time (s)", output_name, interpolation, extrapolation
+            )
+
+    return Decorator
+
+
+@dataclass
+class DataFrameToFlight:
+    """A class to create a rocketpy.Flight object from a pandas data frame. Each
+    property of a Flight object will be created using values from the data frame
```
giovaniceotto, line 84:
```suggestion
    property of a Flight object will be created using values from the data frame.
```

rocketpy/simulation/df2flight.py
```diff
@@ -0,0 +1,208 @@
...
+        sensitive. This column is crucial because it is used to create the
+        other rocketpy.Function objects.
+    df : pd.DataFrame
+        The data frame containing the flight data. The data frame must contain
+        the time column.
+    columns_map : dict, optional
+        A dictionary containing the mapping between the column names in the
+        data frame and the names to be used in the Flight object. This is useful
+        when the column names in the data frame are not the same as the ones
+        used in the Flight object. The default is None.
```
giovaniceotto, line 109: Does the default option currently work? What happens if you leave it as default?

giovaniceotto (changes requested):
Great addition @Gui-FernandesBR! Can't wait to start using this more!

In fact, I hope we get a similar version that works directly with CSV files and other file formats common for rocket data loggers. I'll reach out to the community to see what kind of files we should support in a more long term scope.

I was pretty close to approving this PR as is, however, I believe I can help improve the implementation just a bit with two comments:

1. Prefer a more explicit file name: `data_frame_to_flight.py`
2. Avoid importing pandas, since it is a big module and will slow down importing this module. When you analyze the code, you really don't need pandas at all, even though you are working with pandas directly (Python is weird...)

rocketpy/simulation/df2flight.py
Gui-FernandesBR: Good catch, I fixed this based on your comment.

rocketpy/simulation/df2flight.py
Gui-FernandesBR: Yes, it would be a better idea.

But the decorator no longer exists.

Gui-FernandesBR:
For future reference... 

Something that is getting into my nerves is that sometimes the flight data is separated into different files. This may cause differences in the time steps and others.

I guess that an easy solution would be converting the current private method `__create_attributes` to a public one, so the user could call it multiple times if needed. I think that could be a brilliant solution, but I'm not able to test it immediately so saving this for later.

Example:

image

rocketpy/simulation/flight_data_importer.py
```diff
@@ -0,0 +1,196 @@
...
+                warnings.warn(
+                    f"Unable to find column '{col}' in the header of the file. "
+                    + f"The attribute '{name}' won't be set. The available "
+                    + "columns are:"
+                    + str(self.original_columns),
+                    UserWarning,
+                )
+                continue
+
+            # Extract values for the current column
```
MateusStano, line 171: The way you are iterating through the `columns_map` seem to inverted to the Juno3 example and to the `FLIGHT_LABEL_MAP`. The way its done here, the key is the name of the column in the file and the value is the name of the Flight attribute

Seems like it was intended to be the other way around (which I think is better also)

rocketpy/simulation/flight_data_importer.py
```diff
@@ -0,0 +1,196 @@
...
+    Initially, this class is designed to work with the option '1.'
+
+    Parameters
+    ----------
+    name : str
+        The name of the flight. This is useful to identify the flight in the
+        simulation. Example: "Rocket Calisto, Spaceport America, 2019"
+    filepath : str, Path
+        The path
+    columns_map : dict, optional
```
MateusStano, line 47: I wonder this could be a list of filepaths and, as long as they have headers and a "time" column, it should be easy to make it work

MateusStano (commented):
So overall it seems the user needs two things to use this class:

1 - A file with data in a valid format
2 - A dictionary mapping his data to flight values

I would say that makes a lot of sense, we just need a lot of documentation to support these requirements

rocketpy/simulation/flight_data_importer.py
Gui-FernandesBR: You are right, I inverted that without noticing. 
The point is that the idea of "mapping" the columns is to get the corresponding Flight attribute for each column name of the dataset. I understand if may not be so beautiful, but I think if makes more sense to have something like `{"column_name": "Flight attribute name", ...}` than the opposite.

rocketpy/simulation/flight_data_importer.py
```diff
@@ -0,0 +1,196 @@
...
+        if self.columns_map:
+            for col, atr in self.columns_map.items():
+                if atr == "time":
+                    self.time_col = col
+                    return None
+        if "time" in self.original_columns:
+            self.time_col = "time"
+            return None
+        raise ValueError("Unable to determine the time column...")
+
```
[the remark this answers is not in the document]
Gui-FernandesBR: That's right.

rocketpy/simulation/flight_data_importer.py
Gui-FernandesBR: Good suggestions, done!

rocketpy/simulation/flight_data_importer.py
```diff
@@ -0,0 +1,196 @@
...
+    def __handle_data(self):
+        raw_data = np.genfromtxt(
+            self.filepath,
+            delimiter=self.separator,
+            encoding=self.encoding,
+            dtype=np.float64,
+            names=True,
+            deletechars="",
+        )
```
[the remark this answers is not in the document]
Gui-FernandesBR: I still need to update the juno3 example

rocketpy/simulation/df2flight.py
Gui-FernandesBR: pandas is no longer necessary

Gui-FernandesBR:
@RocketPy-Team/code-owners I believe we are ready for a review now.
I'm still working on supplying an extended example on how to use the class and create good comparisons, but the `FlighDataImporter` itself is ready, I believe so.

Gui-FernandesBR:
Please take a look at #513 to see an example of this class being used.

rocketpy/simulation/flight_data_importer.py
```diff
@@ -0,0 +1,399 @@
...
+        interpolation : str, optional
+            The interpolation method to use for missing data.
+            Defaults to "linear", see rocketpy.mathutils.Function for more
+            information.
+        extrapolation : str, optional
+            The extrapolation method to use for data outside the range.
+            Defaults to "zero", see rocketpy.mathutils.Function for more
+            information.
+        separator : str, optional
+            The separator used in the data file. Defaults to ",".
```
MateusStano, line 96: `delimiter` is a more common name for this

rocketpy/simulation/flight_data_importer.py
```diff
@@ -0,0 +1,399 @@
...
+        Returns
+        -------
+        str
+            A string representation of the FlightDataImporter class.
+        """
+        return f"FlightDataImporter(name='{self.name}', paths='{self.paths}')"
+
+    def __reveal_files(self, path):
+        """Get a list of all the .csv or .txt files in the given path, or simply
+        return the path of the file if it is a file.
```
MateusStano, line 143: Do we follow any conventions when naming methods meant to be private? Sometimes we do it with one underscore, sometimes we do it with two.

From what I've researched it is more common to use only one underscore for the use case of making a method "private" or "protected"

rocketpy/simulation/flight_data_importer.py
```diff
@@ -0,0 +1,399 @@
...
+        """Initializes the FlightDataImporter class.
+
+        Parameters
+        ----------
+        name : str
+            The name of the FlightDataImporter object.
+        paths : list
+            A list of paths or strings representing the file paths or directories
+            containing the flight data files. Only .csv and .txt files are supported.
+        columns_map : dict, optional
```
MateusStano, line 81: This can also be a string, not just list

MateusStano (approved):
I think most comments I made are regarding documentation, so feel free to merge if you want to fix those in another PR.

Everything else seems good

Gui-FernandesBR:
> I think most comments I made are regarding documentation, so I am approving this one.
> 
> But unless you are planning to add proper documentation soon, I do not feel comfortable merging this to develop just yet

What do you mean by "proper documentation"?

rocketpy/simulation/flight_data_importer.py
Gui-FernandesBR: Writing "Class._method" and "Class.__method" are not the same thing.

Private methods in python actually doesn't exist.
Using double underscore will make python "to mangle" the attribute, which is basically adding a "__ClassName__" at the beginning of the attribute's name, which happens to be a uncomfortable way of "privatizing" attributes and methods within a class.

Using a single underscore is the informal way python developers found to define "private methods" without mangling its name.

I left these methods with double underscores because I don't think users (or developers) should invoke these methods out of this class. I mean, I wanted them private, not "weakly private".

rocketpy/simulation/flight_data_importer.py
```diff
@@ -0,0 +1,399 @@
...
+    ):
+        """Initializes the FlightDataImporter class.
+
+        Parameters
+        ----------
+        name : str
+            The name of the FlightDataImporter object.
+        paths : list
+            A list of paths or strings representing the file paths or directories
+            containing the flight data files. Only .csv and .txt files are supported.
```
MateusStano, line 80: Seems, that this can be a path to a folder, and thin will read all the .txt and .csv files in there, right?

How does this behave with the `columns_map` argument?

rocketpy/simulation/flight_data_importer.py
Gui-FernandesBR: The juno3 notebook was updated in another PR

rocketpy/simulation/flight_data_importer.py
```diff
@@ -0,0 +1,399 @@
...
+    "drift": "Drift (deg)",
+    "e0": "Quaternion e0",
+    "e1": "Quaternion e1",
+    "e2": "Quaternion e2",
+    "e3": "Quaternion e3",
+    "latitude": "Latitude (deg)",
+    "longitude": "Longitude (deg)",
+    "mach_number": "Mach number",
+    "name": "Time (s)",
+    "speed": "Speed (m/s)",
```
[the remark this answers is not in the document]
Gui-FernandesBR: nice catch. Solved  here: [4a473d6](https://github.com/RocketPy-Team/RocketPy/pull/474/commits/4a473d6a7a5bb48057917b308e37b39244e8e889)

rocketpy/simulation/flight_data_importer.py
```diff
@@ -0,0 +1,399 @@
...
+    frame must contain a time column in seconds.
+
+    There are different ways of calculating a property from the data frame.
+        1. The property is already in data frame and you want to use it as is
+        2. The property is not in the data frame but you can calculate it from
+            other properties in the data frame.
+
+    Initially, this class is designed to work with the option '1.'
+    """
+
```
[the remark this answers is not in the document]
Gui-FernandesBR: I would say "outdated" instead of "wrong".

Solved here: [4a473d6](https://github.com/RocketPy-Team/RocketPy/pull/474/commits/4a473d6a7a5bb48057917b308e37b39244e8e889)

MateusStano:
> What do you mean by "proper documentation"?

My bad, I meant to delete that srry

rocketpy/simulation/flight_data_importer.py
```diff
@@ -0,0 +1,399 @@
...
+    def __init__(
+        self,
+        name,
+        paths,
```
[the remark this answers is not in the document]
Gui-FernandesBR: My initial answer was gonna be:
- Yes, because for a single rocket flight I'm expecting to have different FlightDataImport and and want to ensure the user is naming the objects correctly.

But tbh, there's no hard reason to have this as a mandatory argument. The code should still run without it. 
So I changed it here: [c7ec3ce](https://github.com/RocketPy-Team/RocketPy/pull/474/commits/c7ec3ceb55ce741ff1ef0bba183577b56c4928b2)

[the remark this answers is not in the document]
Gui-FernandesBR: Thank you for noticing it.

rocketpy/simulation/flight_data_importer.py
```diff
@@ -0,0 +1,399 @@
...
+        ----------
+        name : str
+            The name of the FlightDataImporter object.
+        paths : list
+            A list of paths or strings representing the file paths or directories
+            containing the flight data files. Only .csv and .txt files are supported.
+        columns_map : dict, optional
+            A dictionary mapping column names to desired column names.
+            Defaults to None, which will keep the original column names.
+        units : dict, optional
```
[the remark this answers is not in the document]
Gui-FernandesBR: I gave a better description with this commit: [4a473d6](https://github.com/RocketPy-Team/RocketPy/pull/474/commits/4a473d6a7a5bb48057917b308e37b39244e8e889)

Also, the #513 is going to add a good example that will help understanding how to use this class.

rocketpy/simulation/flight_data_importer.py
Gui-FernandesBR: Good catch. The numpy module also uses delimiter, so I've changed in my latest commit.

rocketpy/simulation/flight_data_importer.py
Gui-FernandesBR: Solved.

rocketpy/simulation/flight_data_importer.py
Gui-FernandesBR: Yes it can indeed read all the files within a folder, at once.
In this case the same columns_map will be used for all the files.
Again this will become clearer after we merge example notebooks, but basically it is way easier to define the same mapping instead of one particular map for each file.

I ensured the code is working even if you mess up the map... So no worries.

PR #477, opened 2023-11-21
ENH: Get Instance Attributes

MateusStano:
## Pull request type

## Checklist

## Description

This PR introduces a function to `utilities.py` that returns all relevant attributes of all main classes in RocketPy. This is a special request for the development of the Infinity-API.

I have also added a deprecation warning to the old `info_returned` methods, since they do not need to be used anymore

## Breaking change

- [ ] Yes
- [x] No

## Additional information

There a couple of fixes here to the `Environment.time_array` attribute. In some atmospheric models, this was not being saved as an array, but instead as a NetCDF4 object. This attribute is not used anywhere in the code, but it was breaking the `get_instance_attributes` function


MateusStano:
@GabrielBarberini @luimot A review from you guys is extra important!

Please tell me if this works well with the API

GabrielBarberini (approved):
LGTM

rocketpy/utilities.py
```diff
@@ -654,3 +655,25 @@ def liftoff_speed(mass):
...
+    Returns
+    -------
+    dictionary
+        Dictionary with all attributes of the given instance.
+    """
+    attributes_dict = dict()
+    members = inspect.getmembers(instance)
+    for member in members:
+        # Filter out methods and protected attributes
+        if not inspect.ismethod(member[1]) and not member[0].startswith("__"):
```
phmbressan, line 677: Just to make sure this is the intended behavior, "weakly private" attributes that start with only one `_` are being added to the dict (the `member[0].startswith("__")` has two underscores).

phmbressan (approved):

rocketpy/utilities.py
MateusStano: Yes, this is intended. Things like `rocket._csys` is pretty important to have

rocketpy/environment/environment.py
```diff
@@ -2365,7 +2365,7 @@ def process_forecast_reanalysis(self, file, dictionary):
         self.wind_vs = wind_vs
         self.levels = levels
         self.temperatures = temperatures
-        self.time_array = time_array
+        self.time_array = time_array[:].tolist()
```
Gui-FernandesBR, line 2368: I think we should keep consistency here. Please notice that all the variables in this block are being saved as as numpy masked arrays (not netcdf4 objects), changing the type of just one is not a good idea imo. 

I think that what we should do is to modify the `get_instance_attributes`method to automatically convert numpy array objects to lists before saving it. Some classes even have a "to_serializable" object. If you find this complicated, you should definitely us try/except block inside the `get_instance_attributes` function.

```suggestion
        self.time_array = time_array
```

rocketpy/environment/environment.py
```diff
@@ -3126,7 +3126,18 @@ def all_plot_info_returned(self):
...
         plot_info : Dict
             Dict of data relevant to plot externally
+
+        Warning
+        -------
+        Deprecated in favor of `utilities.get_instance_attributes`.
+
         """
+        warnings.warn(
+            "The method 'all_plot_info_returned' is deprecated. "
```
Gui-FernandesBR, line 3136: Deprecation message should contain the specific version in which the method was deprecated, and also the version in which that will be completely removed. Also, it should indicates new methods to be used in the place of the deprecated method.

My suggestion is to deprecate it in v1.2.0 and fully remove it in v.1.4.0

warnings.warn(
        "old_method() is deprecated as of version X.Y and will be removed in version X.(Y+2). "
        "Please use new_method() instead.", 
        DeprecationWarning, 
        )

rocketpy/utilities.py
```diff
@@ -654,3 +655,25 @@ def liftoff_speed(mass):
     if plot:
         retfunc.plot(min_mass, max_mass, points)
     return retfunc
+
+
+def get_instance_attributes(instance):
```
Gui-FernandesBR, line 660: Did you consider using the built-in `vars()` from Python? See [link](https://docs.python.org/3/library/functions.html#vars).
This would allow you to get the __dict__ of the instance and the code would become much easier to understand.

rocketpy/utilities.py
```diff
@@ -654,3 +655,25 @@ def liftoff_speed(mass):
...
+    """Returns a dictionary with all attributes of a given instance.
+
+    Parameters
+    ----------
+    instance : object
+        Instance of a class.
+
+    Returns
+    -------
```
Gui-FernandesBR, line 669: This function needs tests before merging

Gui-FernandesBR (changes requested):

rocketpy/environment/environment.py
MateusStano: I am a little confused here.

All the variables in this block are saved as either lists or masked arrays, the only exception was `time_array` which was incorrectly saved as netcdf4 object

Other parts of the code that also define this attribute already did this exact conversion (using `.tolist()`). I just made these two cases consistent with the rest

rocketpy/utilities.py
MateusStano: Unfortunately this does not work with `@funcify_method` and `@cached_properties`

rocketpy/environment/environment.py
Gui-FernandesBR: As discussed, you're right here, as usual

Gui-FernandesBR (approved):

PR #478, opened 2023-11-21
HOTFIX: 2D .CSV Function and missing set_get_value_opt call

MateusStano:
## Pull request type

## Checklist

## Current behavior

If you create a Function based on a 2D .csv file, it will not work properly due to the interpolation being set to None. It also has no `get_value_opt` defined

## New behavior

2D .csv based Functions now work as expected.

Also, tests checking the issues here have beend added

## Breaking change

- [ ] Yes
- [x] No


rocketpy/mathutils/function.py
```diff
@@ -2825,7 +2825,12 @@ def _check_user_input(
 
         # check source for data type
         # if list or ndarray, check for dimensions, interpolation and extrapolation
-        if isinstance(source, (list, np.ndarray)):
+        if isinstance(source, (list, np.ndarray, str, Path)):
+            # Deal with csv or txt
+            if isinstance(source, (str, Path)):
+                # Convert to numpy array
+                source = np.loadtxt(source, delimiter=",", dtype=float)
+
```
phmbressan, line 2833: I understand that this is necessary in order to check the validity of the inputs/outputs, but it is quite strange to read the csv here, throw away the result, and read it again in `set_source`.

rocketpy/mathutils/function.py
```diff
@@ -2825,7 +2825,12 @@ def _check_user_input(
 
         # check source for data type
         # if list or ndarray, check for dimensions, interpolation and extrapolation
-        if isinstance(source, (list, np.ndarray)):
+        if isinstance(source, (list, np.ndarray, str, Path)):
+            # Deal with csv or txt
+            if isinstance(source, (str, Path)):
```
phmbressan, line 2830: I believe a `try-except` with `FileNotFound` would be beneficial here, just because it is a good practice. But again, this is also done in `set_source`.

phmbressan (approved):

rocketpy/mathutils/function.py
giovaniceotto: I would say that loading it twice is really not a good idea. It can be a huge problem if the CSV file is large and takes a while to read.

Furthermore, I would not recommend changing this in this hotfix PR. It doesn't exactly fix a bug.

tests/conftest.py
```diff
@@ -1112,9 +1112,21 @@ def func_from_csv():
...
-        outputs=["Scalar"],
-        interpolation="linear",
-        extrapolation="natural",
+    )
+    return func
+
+
+@pytest.fixture
+def func_2d_from_csv():
+    """Create a 2d function based on a csv file. The csv file contains the
```
giovaniceotto, line 1121: I think something is missing here.

giovaniceotto (changes requested):

rocketpy/mathutils/function.py
phmbressan: Indeed, in retrospect, it is a bit odd to perform the same operations multiple times in check inputs and while setting the source.

Not in the scope of this hotfix PR, but the check inputs could be made as a way of pre-processing the source so as to only return a `np.array` or a `callable` to the `set_source`.

This would decouple responsabilities a fair bit and could be implemented in various ways, maybe with `functools.singledispatch`.

PR #479, opened 2023-11-21
HOTFIX: Tanks Overfill not Being Detected 

phmbressan:
## Pull request type

## Checklist

## Current behavior

The `MassBasedTank` did not compute correctly overfills or underfills due to some code checks that were not working properly, including:

- The `Function.compose` extrapolation check was not being applied correctly;
- The overfill check was only being done regarding the computed `gas_height`, however it is more accurate to check before computations in `fluid_volume`.

## New behavior

The main changes this hotfix introduces:

- The `Function.compose` extrapolation was corrected;
- An overfill validation was added in `MassBasedTank.fluid_volume`;
- An assertion in `test_mass_based_tank` regarding the `fluid_volume` so as to prevent this kind of error in the near future;

## Breaking change

- [ ] Yes
- [X] No


tests/test_tank.py
```diff
@@ -150,8 +150,8 @@ def test_mass_based_tank():
     real_geometry = TankGeometry(
         {
             (0, 0.0559): bottom_endcap,
-            (0.0559, 0.7139): lambda y: 0.0744,
-            (0.7139, 0.7698): top_endcap,
+            (0.0559, 0.8039): lambda y: 0.0744,
+            (0.8039, 0.8698): top_endcap,
```
phmbressan, line 154: Comment on these changes: after making the assertion from lines 192-195, the tests failed due to the fact that the defined geometry volume is too low. Therefore the tank ended up being overfilled and had to be increased in size.

rocketpy/motors/tank.py
```diff
@@ -1292,7 +1292,20 @@ def fluid_volume(self):
...
+
+        # Check if within bounds
+        diff = fluid_volume - self.geometry.total_volume
+
+        if (diff > 1e-6).any():
+            raise ValueError(
+                "The `fluid_volume`, defined as the sum of `gas_volume` and "
+                + "`liquid_volume`, is not equal to the total volume of the tank."
+                + "\n\t\tThe difference is more than 1e-6 m^3 at "
+                + f"{diff.x_array[np.argmin(diff.y_array)]} s."
```
giovaniceotto, line 1305: Just checking: we need to guarantee, for now, that `fluid_volume` does not exceed `self.geometry.total_volume`, but we don't really care yet if `fluid_volume` is much less then `self.geometry.total_volume`?

If this is the case, shouldn't we us `np.argmax` instead of `np.argmin`?

phmbressan: I do think you are right regarding `np.argmax`, but, just to make sure we are on the same page, this would need to be changed in all error messages of this class, right?

MateusStano, line 1305: I think this error message could be improved a bit

This is a MassBasedTank, meaning the user will have given mass curves to the tank. If this error pops up warning about volume issues, the user might get confused and the correction to this error will not be clear.

I would suggest telling the user that the masses, the densities or the tank geometry might be wrong

phmbressan: Actually, I correct myself: the computation should be `np.argmin` since we are loking for the smallest of the differences between the fluid volume and the tank volume to determine exactly where it overfilled.

MateusStano (approved):

rocketpy/motors/tank.py
giovaniceotto: @phmbressan with `np.argmin`, don't you have the risk of getting a negative value, in which the tank is overfilled?

phmbressan: The check here is made following these assumptions:

1. The `diff` pointed out that the `Tank` is, at some time, overfilled;
2. There is a value of time that the fluid volume surpassed the total volume of the Tank;
3. In general, this specific point in time is not part of the time samples so the minimum value of the diff is used to select it. I do think it is possible that this value may be a bit before the overfill crossing or a bit after, so there is a certain margin of error.

Perhaps I should mention that the time is approximate or, likely a better solution, use `Function.find_input`. I favored the implemented way due the fact this is a hotfix and the `np.argmin` was already being used in the other exceptions of this class.

Let me know if I made it clearer now or if there is any mistake in my reasoning.

Issue #482, opened 2023-11-23
BUG: Overfilled Tanks produce absurd calculations without an error thrown


theDiego21:
I encountered a Bug, in which, when the Volume of a Fluid is greater than the Tank Volume, the engine gets calculated just fine, but when you add the engine to a rocket, the centre of gravity of the rocket is set to thousands of meters negative.

The Rocket therefore does not get drawn properly and the calculation of the Flight goes (i think) into an endless loop. 
(I tried letting it run for 15 mins, then i cancelled the calculation)

Sometimes, an error (Overfill or underfill) is thrown, but most of the times, the calculation of the engine does not throw an error
Environment used: Google Colab

phmbressan:
Thank you for reporting this Bug and creating an issue. 

This was already detected and is being fixed by the hotfix PR #479 . A release containing this hotfix will be released in a few days.

Additionally, could you provide more information about which `Tank` class produced this issue (or even a code sample) so we can check if the planned hotfix covers this case?

theDiego21:
The Tank used was a MassFlowRateBasedTank.
Here is the code sample, which caused the bug:

# Define fluids
oxidizer_liq = Fluid(name="LOX", density=1141)
fuel_liq = Fluid(name="ethanol", density=789)
pressurant_gas = Fluid(name="N2", density=1.034)

# Define tanks geometry
eth_tank_shape = CylindricalTank(radius = 13 / 100, height = 0.345409407550068, spherical_caps = True)
lox_tank_shape = CylindricalTank(radius = 13 / 100, height = 0.28827610851596, spherical_caps = True)
pressurant_tank_shape = CylindricalTank(radius = 6 / 100, height = 0.22, spherical_caps = True)

# Define tanks
oxidizer_tank = MassFlowRateBasedTank(
    name="oxidizer tank",
    geometry=eth_tank_shape,
    flux_time=5,
    initial_liquid_mass=4.695,
    initial_gas_mass=0,#.1880196263187668,
    liquid_mass_flow_rate_in=0,
    liquid_mass_flow_rate_out=lambda t: 0.4815384615384616, #/ 3 * exp(-0.25 * t),
    gas_mass_flow_rate_in=0,
    gas_mass_flow_rate_out=0,
    liquid=oxidizer_liq,
    gas=pressurant_gas,
)

fuel_tank = MassFlowRateBasedTank(
    name="fuel tank",
    geometry=lox_tank_shape,
    flux_time=5,
[...]

PR #483, opened 2023-11-23
HOTFIX: v1.1.1

MateusStano:
## Fixed

https://github.com/RocketPy-Team/RocketPy/pull/479:
- Opposite static margin bug, which affected all rockets defined using a nose_to_tail coordinate system, causing positive static margins to become negative and negative ones to become positive. Note: this bug did not lead to any issues in terms of the trajectory simulation itself.

https://github.com/RocketPy-Team/RocketPy/pull/478:
- Fixed Bug in 2D .csv-Based Functions for Proper Interpolation Handling and Added Relevant Test Coverage

https://github.com/RocketPy-Team/RocketPy/pull/476:
- Corrected Function.compose extrapolation.
- Added overfill validation in MassBasedTank.fluid_volume.
- Introduced assertion in test_mass_based_tank to prevent future errors in fluid_volume.


Gui-FernandesBR (changes requested):
Thank you @MateusStano !!

2 changes that we still need to do before merging:
- Update the CHANGELOG file.
- Re-run the SEBLM notebook to update the static margin plot. See:

image

Gui-FernandesBR:
We should have more tests for the tanks, not all the lines are being covered right now, as reported by the codecov

Gui-FernandesBR:
This one solves #482

Gui-FernandesBR (approved):
Great one!

tests/test_tank.py
```diff
@@ -190,6 +189,10 @@ def test_mass_based_tank():
         discretize=None,
     )
 
+    # Assert volume bounds
+    assert (real_tank_lox.gas_height <= real_tank_lox.geometry.total_volume).all
```
phmbressan, line 193: This line is wrong, it should be 


```suggestion
    assert (real_tank_lox.gas_height <= real_tank_lox.geometry.top).all
```

phmbressan (approved):
Great fixes. I just noticed a small assertion mistake and will commit a fix.

PR #484, opened 2023-11-23
ENH: Prevent out of bounds Tanks from Instantiation

phmbressan:
## Pull request type

## Checklist

## Current behavior

Due to the fact that the `Tank` getters are evaluated lazily (post instance), it is possible that a `Tank` with invalid inputs is only detected by specific methods that may not be called.

## New behavior

This PR introduces methods to check if the `Tank` fluids are inside the appropriate volume limits and calls them on instantiation. 

Therefore, a invalid `Tank` shall not be constructed.

## Breaking change

- [ ] Yes
- [X] No


rocketpy/motors/tank.py
```diff
@@ -410,6 +409,73 @@ def inertia(self):
         """
         return self.liquid_inertia + self.gas_inertia
 
+    def _check_volume_bounds(self):
+        """Checks if the tank is overfilled or underfilled. Raises a ValueError
+        if the tank is overfilled or underfilled.
+        """
+        if (self.fluid_volume > self.geometry.total_volume + 1e-6).any():
+            raise ValueError(
+                f"The tank '{self.name}' is overfilled. The fluid volume is "
```
Gui-FernandesBR, line 418: maybe you could create a single string to be used in both error messages, thus reducing the number of lines here.

Something like this.

error_msg = "The tank '{self.name}' is {fill_condition}'. " + "Consider increasing input fluid quantities or checking the fluid density values.\n\t\t" + "The fluid volume is {volume:.3f} m³ at {time:.3f} s."

raise ValueError(error_msg.format(volume=self.fluid_volume.y_array[max_volume_idx],
            time=self.fluid_volume.x_array[max_volume_idx]))

Sorry for keeping this short.

rocketpy/motors/tank.py
```diff
@@ -410,6 +409,73 @@ def inertia(self):
...
+                + "greater than the total volume of the tank.\n\t\t"
+                + "Try increasing the tank height and check out the fluid density"
+                + " values.\n\t\t"
+                + f"The fluid volume is {np.max(self.fluid_volume.y_array):.3f} m³ at "
+                + f"{self.fluid_volume.x_array[np.argmax(self.fluid_volume.y_array)]:.3f} s."
+            )
```
MateusStano, line 424:
```suggestion
            raise ValueError(
                f"The tank '{self.name}' is overfilled. The fluid volume is "
                + "greater than the total volume of the tank.\n\t\t"
                + "Try increasing the tank height and check out the fluid density"
                + " values.\n\t\t"
                + f"The fluid volume is {np.max(self.fluid_volume.y_array):.3f} m³ at "
                + f"{self.fluid_volume.x_array[np.argmax(self.fluid_volume.y_array)]:.3f} s.\n\t\t"
                + f"The tank total volume is {self.geometry.total_volume:.3f} m³."
            )
```

MateusStano: I suggest that we also print the tank total volume all the over/underfilled error messages

rocketpy/motors/tank.py
```diff
@@ -410,6 +409,73 @@ def inertia(self):
...
+                + f"{self.fluid_volume.x_array[np.argmin(self.fluid_volume.y_array)]:.3f} s."
+            )
+
+    def _check_height_bounds(self):
+        """Checks if the tank is overfilled or underfilled. Raises a ValueError
+        if the tank is overfilled or underfilled.
+        """
+        top_tolerance = self.geometry.top + 1e-4
+        bottom_tolerance = self.geometry.bottom - 1e-4
+
```
MateusStano, line 441: Can the `geometry.top` or the `geometry.bottom` somehow be equal to zero?

If so, would the bottom_tolerance allow negative values of liquid_height? Could this break something further on in a simulation?

rocketpy/motors/tank.py
```diff
@@ -410,6 +409,73 @@ def inertia(self):
...
+                + f"the tank bottom at {self.liquid_height.x_array[np.argmin(self.liquid_height.y_array)]:.3f} s."
+            )
+        elif (self.gas_height > top_tolerance).any():
+            raise ValueError(
+                f"The tank '{self.name}' is overfilled. "
+                + "The gas height is above the tank top.\n\t\t"
+                + "Try increasing the tank height and check out the fluid density"
+                + " values.\n\t\t"
+                + f"The gas height is {np.max(self.gas_height.y_array):.3f} m above "
+                + f"the tank top at {self.gas_height.x_array[np.argmax(self.gas_height.y_array)]:.3f} s."
```
MateusStano, line 467: `gas_height` is usually a negative value right? Will this check work as intended?

rocketpy/motors/tank.py
```diff
@@ -1230,6 +1292,10 @@ def __init__(
         # Discretize input if needed
         self.discretize_masses() if discretize else None
 
+        # Check if the tank is overfilled or underfilled
+        self._check_volume_bounds()
+        self._check_height_bounds()
```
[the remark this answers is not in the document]
phmbressan: I did not understand your comment, my intention is to run these checks methods only in the class init.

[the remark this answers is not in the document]
Gui-FernandesBR: Sorry I was under a false impression here.

Gui-FernandesBR:
Hey, last comment, should we add tests to ensure the exceptions are being raised correctly?

rocketpy/motors/tank.py
phmbressan: The `liquid_height` value is computed relative to the `Tank` geometry reference so it could be negative without any input error.

If you meant that `geometry.top` and `geometry.bottom` could both be zero at the same time, then the tank geometry would degenerate and I don't know precisely its consequences.

rocketpy/motors/tank.py
phmbressan: I do think this check works: both `geometry.top` and `gas_height` are calculated with respect to the `Tank` geometry reference, so they could be both negative or positive.

If the `gas_height` surpasses the top tolerance (e.g. -0.9 > -1), the check would point the overfill.

MateusStano (approved):

PR #485, opened 2023-11-25
BUG: Hotfix v1.1.1 Docs Build - Function CSV

phmbressan:
## Pull request type

## Checklist

## Current behavior

The `Function.__check_user_inputs` did not support headered csv, which caused an error on docs build.

## New behavior

The support for headered csv was added and there was a `LiquidMotor` example in the docs that needed refactor to the new cap height definition.

## Breaking change

- [X] No

### DIFF
--- a/rocketpy/mathutils/function.py
+++ b/rocketpy/mathutils/function.py
@@ -2830,7 +2830,13 @@ class Function:
             # Deal with csv or txt
             if isinstance(source, (str, Path)):
                 # Convert to numpy array
-                source = np.loadtxt(source, delimiter=",", dtype=float)
+                try:
+                    source = np.loadtxt(source, delimiter=",", dtype=float)
+                except ValueError:
+                    # Skip header
+                    source = np.loadtxt(source, delimiter=",", dtype=float, skiprows=1)
+                except Exception:
+                    raise ValueError("The source file is not a valid csv or txt file.")
 
             else:
                 # this will also trigger an error if the source is not a list of
--- a/docs/user/motors/liquidmotor.rst
+++ b/docs/user/motors/liquidmotor.rst
@@ -40,7 +40,7 @@ Then we must first define the tanks:
   fuel_gas = Fluid(name="ethanol_g", density=1.59)
 
   # Define tanks geometry
-  tanks_shape = CylindricalTank(radius = 0.1, height = 1, spherical_caps = True)
+  tanks_shape = CylindricalTank(radius = 0.1, height = 1.2, spherical_caps = True)
 
   # Define tanks
   oxidizer_tank = MassFlowRateBasedTank(
@@ -48,7 +48,7 @@ Then we must first define the tanks:
       geometry=tanks_shape,
       flux_time=5,
       initial_liquid_mass=32,
-      initial_gas_mass=0.1,
+      initial_gas_mass=0.01,
       liquid_mass_flow_rate_in=0,
       liquid_mass_flow_rate_out=lambda t: 32 / 3 * exp(-0.25 * t),
       gas_mass_flow_rate_in=0,


Gui-FernandesBR (approved):
LGTM!

All tests are passing locally, including with the `--run-slow` option

The `check_user_inputs` method is raising more problems than what I thought in the beginning. Improvements for this method are welcome, and I believe there's already an issue for that.

Good news: docs are passing! See https://docs.rocketpy.org/en/bug-v1.1.1-docs-build/

rocketpy/mathutils/function.py
```diff
@@ -165,20 +174,20 @@ def set_source(self, source):
...
-            else:
-                source = np.loadtxt(source, delimiter=",", dtype=float)
+            with open(source, "r") as file:
+                first_line = file.readline()
+                try:
+                    # Try to convert the first line to floats
+                    np.array([float(item) for item in first_line.split(",")])
+                    # If successful, no headers are present
+                    file.seek(0)  # Reset file read position
+                    source = np.loadtxt(file, delimiter=",", dtype=float)
```
phmbressan, line 184: I might have misunderstood something, but these checks before `np.loadtxt` seem pointless:

- Why to cast every item in the first line to a `float`, return to the beginning, and then apply `np.loadtxt` that basically will try to cast all the items to a `float`?
- Isn't it the same as using `np.loadtxt` in the first place and, if there is an error, that means there is a header.

Furthermore, I do think that should be another exception layer in case the `np.gentxt` fails. That means the input is malformed and the user should be informed.

Of course, I am not trying to bias you to the same logic I did in this PR at `__check_user_inputs`, but it seems that there may be some unnecessary operations here.

rocketpy/mathutils/function.py
```diff
@@ -39,14 +39,9 @@ def __init__(
...
-            The actual function. If type is function, it will be called for
-            evaluation. If type is int or float, it will be treated as a
-            constant function. If ndarray, its points will be used for
-            interpolation. An ndarray should be as [(x0, y0, z0), (x1, y1, z1),
-            (x2, y2, z2), ...] where x0 and y0 are inputs and z0 is output. If
-            string, imports file named by the string and treats it as csv.
-            The file is converted into ndarray and should not have headers.
+        source : function, scalar, ndarray, string, Function
+            The data source to be used for the function. See the ``set_source``
+            method's documentation for more details.
```
giovaniceotto, line 44: @Gui-FernandesBR, I wonder if it is fair to forward users to another documentation in a basic description of constructor arguments.

I understand your idea here, to remove duplicate docs, but it can be quite frustrating when you are trying to understand how to initialize a class and you must navigate through its methods to do so.

rocketpy/mathutils/function.py
```diff
@@ -133,25 +128,39 @@ def set_outputs(self, outputs):
...
-            string, imports file named by the string and treats it as csv.
-            The file is converted into ndarray and should not have headers.
-            If the source is a Function, its source will be copied and another
-            Function will be created following the new inputs and outputs.
+        source : function, scalar, ndarray, string, or Function
+            The data source to be used for the function:
+
+            - Python function: Called for evaluation with input values.
+
+            - int or float: Treated as a constant value.
```
giovaniceotto, line 141:
```suggestion
            - int or float: Treated as a constant value function.
```

rocketpy/mathutils/function.py
MateusStano, line 44: Having a description of what source is here is much more useful than it just referencing another method. I know it makes for repeated documentation, but I would argument that sometimes repetition in documentation can be good, and this is one of those cases.

Honestly, from time to time I have to check the description of `source` and having to look at a method's argument to understand what I need to do to initialize the object instead of just looking at `__init__` would be annoying and could be confusing for people not used to python library documentation styles

The current description in `set_source` is really good, I'd just copy and paste it here

rocketpy/mathutils/function.py
```diff
@@ -133,25 +128,39 @@ def set_outputs(self, outputs):
...
-            interpolation. An ndarray should be as [(x0, y0, z0), (x1, y1, z1),
-            (x2, y2, z2), ...] where x0 and y0 are inputs and z0 is output. If
-            string, imports file named by the string and treats it as csv.
-            The file is converted into ndarray and should not have headers.
-            If the source is a Function, its source will be copied and another
-            Function will be created following the new inputs and outputs.
+        source : function, scalar, ndarray, string, or Function
+            The data source to be used for the function:
+
+            - Python function: Called for evaluation with input values.
```
giovaniceotto, line 139: I believe the correct term here would be callable. It does not need to be a Python function. Any callable will do, such as classes with the `__call__` magic method, or any class method. See https://realpython.com/python-callable-instances/#:~:text=In%20Python%2C%20a%20callable%20is,examples%20of%20callables%20in%20Python. for details.

```suggestion
            - Callable: Called for evaluation with input values. Must have the desired inputs as arguments and return a single output value. Input order is important. Example: Python functions, classes, and methods.
```

MateusStano (changes requested):

rocketpy/mathutils/function.py
```diff
@@ -165,20 +174,20 @@ def set_source(self, source):
...
+                    # Try to convert the first line to floats
+                    np.array([float(item) for item in first_line.split(",")])
+                    # If successful, no headers are present
+                    file.seek(0)  # Reset file read position
+                    source = np.loadtxt(file, delimiter=",", dtype=float)
+                except ValueError:
+                    # If an error occurs, headers are present
+                    source = np.genfromtxt(
+                        source, delimiter=",", dtype=float, skip_header=1
+                    )
```
giovaniceotto, line 189: This opens the same file twice, which seems like a code smell to me. Can we refactor it avoiding at all costs reading the same file twice?

giovaniceotto (changes requested):

rocketpy/mathutils/function.py
Gui-FernandesBR: I think your approach is better.

rocketpy/mathutils/function.py
Gui-FernandesBR: I don't think repeating docstrings is a good idea. I could agree that having it in the init method is better than in the set_source method.

But I won't say anything if you duplicate this docstring to accomplish with your comment.

Gui-FernandesBR: Same as above

rocketpy/mathutils/function.py
Gui-FernandesBR: PLease keep in mind thath "callable" also includes the rocketpy' Function class.

rocketpy/mathutils/function.py
Gui-FernandesBR: Same as Bressan's comment.

Just out of curiosity, how long do you imagine that could take to read a really really long file using genfromtxt? 
I tested with a 130k lines file these days, and the time is not significant. That worth a try at least.

MateusStano (approved):

Gui-FernandesBR (approved):
This Looks Good To Me

rocketpy/mathutils/function.py
```diff
@@ -39,14 +39,33 @@ def __init__(
...
+            (x1, y1, z1), ..., (xn, yn, zn)], where 'x' and 'y' are inputs,
+            and 'z' is the output.
+
+            - string: Path to a CSV file. The file is read and converted into an
+            ndarray. The file can optionally contain a single header line.
+
+            - Function: Copies the source of the provided Function object,
+            creating a new Function with adjusted inputs and outputs.
+
+        Notes
```
giovaniceotto, line 62: This section should be moved to after line 91.

giovaniceotto (approved):

rocketpy/mathutils/function.py
phmbressan: Done in [cd9d10c](https://github.com/RocketPy-Team/RocketPy/pull/485/commits/cd9d10cd4c485582e23ed3629e915e3771a51dd6)

PR #488, opened 2023-11-28
ENH: Function Reverse Arithmetic Priority

phmbressan:
## Pull request type

## Checklist

## Current behavior

Operations that involve reverse `Function` operations with `numpy.ndarrays` result, usually unexpectedly, in a `numpy.ndarray` as a result. This happens since `numpy.ndarray` implement an `__radd__` that overrides `Function.__radd__` operation. 

## New behavior

The variable `__array_ufunc__` is a way to enforce `Function` priority regarding reverse operations.

## Breaking change

- [X] No

## Additional information

This same issue was found during `mathutils.vector` and `mathutils.matrix` development and the same fix was applied (using `__array_ufunc__`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ straightforward as possible.
 
 ### Changed
 
--
+- ENH: Function Reverse Arithmetic Priority [#488](https://github.com/RocketPy-Team/RocketPy/pull/488)
 
 ### Fixed
 
--- a/rocketpy/mathutils/function.py
+++ b/rocketpy/mathutils/function.py
@@ -1840,7 +1840,9 @@ class Function:
                 return Function(lambda x: (self.get_value(x) + other(x)))
         # If other is Float except...
         except AttributeError:
-            if isinstance(other, (float, int, complex, np.ndarray)):
+            if isinstance(
+                other, (float, int, complex, np.ndarray, np.integer, np.floating)
+            ):
                 # Check if Function object source is array or callable
                 if isinstance(self.source, np.ndarray):
                     # Operate on grid values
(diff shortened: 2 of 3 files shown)


giovaniceotto (approved):
This will prevent me from dying a couple of years earlier due to stress. Thank you!

Gui-FernandesBR:
Why exactly is this important? I couldn't figure that out by simply reading the PR description.
Seeing the tests that were created gave me an idea, but I'm not sure.

Does it also works for lists instead of np.ndarrays ?

So are you saying this magic line __array_ufunc__ = None is extremely important for solving the issue, is that right?

Gui-FernandesBR:
The CHANGELOG.md needs to be updated before merging.

Gui-FernandesBR:
Some of the modified lines were not covered by tests. Could you check/fix that please?

giovaniceotto:
> Why exactly is this important? I couldn't figure that out by simply reading the PR description. Seeing the tests that were created gave me an idea, but I'm not sure.

To speed things up, let me try to help here.

As the PR description says:

> Operations that involve reverse Function operations with numpy.ndarrays result, usually unexpectedly, in a numpy.ndarray as a result.

Which means, for example, `np.array([1, 2, 3]) * Function([(0, 0), (1, 1), (2, 2)])` returns a `np.array` instead of returning a `Function`. Even worse:  `a * Function([(0, 0), (1, 1), (2, 2)])` returns a `np.array` if `a` is a `np.float64` or another numpy dtype (https://numpy.org/doc/stable/user/basics.types.html).

This causes a lot of problems, since we expect `a * Function([(0, 0), (1, 1), (2, 2)])` to return a Function. The solution for this is presented in this PR, and probably inspired by references such as: https://numpy.org/neps/nep-0018-array-function-protocol.html

phmbressan:
@Gui-FernandesBR just made some commits expanding the tests for these operations and updated the CHANGELOG file.

Gui-FernandesBR (approved):
LGTM,

Good use of tests

PR #490, opened 2023-11-28
MNT: Add __repr__ method to Parachute class

Gui-FernandesBR:
## Pull request type

## Checklist

## Current behavior

Enter text here...

## New behavior
This is useful for debugging or when you simply call `Rocket.parachutes` and want to see something more readable.

## Breaking change

- [ ] Yes
- [x] No

## Additional information
Opening this now, can describe the PR later, sorry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ straightforward as possible.
 
 ### Changed
 
--
+- MNT: Add repr method to Parachute class [#490](https://github.com/RocketPy-Team/RocketPy/pull/490)
 
 ### Fixed
 
--- a/rocketpy/rocket/parachute.py
+++ b/rocketpy/rocket/parachute.py
@@ -220,6 +220,13 @@ class Parachute:
             self.cd_s,
         )
 
+    def __repr__(self):
+        """Representation method for the class, useful when debugging."""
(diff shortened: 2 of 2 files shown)


giovaniceotto (changes requested):
@Gui-FernandesBR, I understand the usefullness of these codelines. However, I believe it would be a good practice to follow Python conventions for the `__repr__` method, specially since this is mostly for dev use.

From [https://docs.python.org/3/reference/datamodel.html#object.__repr__](https://docs.python.org/3/reference/datamodel.html#object.__repr__):

> If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment). If this is not possible, a string of the form <...some useful description...> should be returned.

Options are:

```
f"<Parachute(name={self.name}, cd_s={self.cd_s}, trigger={self.trigger}, sampling_rate={self.sampling_rate}, lag={self.lag}, noise={self.noise})>"
```

or, less verbose,

```
f"<Parachute name={self.name} (cd_s = {self.cd_s:.4f} m2, trigger = {self.trigger})>"
```

Look at https://stackoverflow.com/questions/44595218/python-repr-for-all-member-variables for more inspiration.

PR #492, opened 2023-11-28
FIX: Broken Function.get_value_opt for N-Dimensional Functions

giovaniceotto:
## Pull request type

## Checklist

## Current behavior

As reported by @styris00, the following code shows an error with `Function.get_value_opt` when used with 2D functions.

```
import numpy as np
from rocketpy import Function

arr = np.array([(1, 1, 2), (3, 6, 8),
            (8, 8, 9)])

myFun = Function(
    arr,
    interpolation="shepard",
    extrapolation="natural",
    title="myFun"
)

myFun.set_get_value_opt()

print(myFun.get_value(1, 1)
print(myFun.get_value_opt(1, 1))
```

Output is:

```
2.0
1.0
```

## New behavior

The bug was fixed.

## Breaking change

- [x] No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,15 @@ straightforward as possible.
 
 -
 
+## [v1.1.3] - 2023-11-29
+
+Here we write upgrading notes for brands. It's a team effort to make them as
+straightforward as possible.
+
+### Fixed
+
+-  FIX: Broken Function.get_value_opt for N-Dimensional Functions [#492](https://github.com/RocketPy-Team/RocketPy/pull/492/files) 
+
 ## [v1.1.2] - 2023-11-25
 
 You can install this version by running `pip install rocketpy==1.1.2`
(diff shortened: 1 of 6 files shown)


Gui-FernandesBR:
I wonder when this bug was introduced. The tests haven't been modified recently.

Gui-FernandesBR (dismissed):

styris00 (approved):

giovaniceotto:
> I wonder when this bug was introduced. The tests haven't been modified recently.

I didn't have enough time to track it down (the snake case commit makes things harder). However, the old test was never going to catch the bug.

What was going on is that, considering x, y, z, `get_value_opt` was returning y instead of z.  And, in the test, y was equal to z, so it passed.

giovaniceotto:
Are we going to make this into a hotfix?
@MateusStano

MateusStano:
> Are we going to make this into a hotfix?
> @MateusStano

Yes we are!

MateusStano (approved):

Gui-FernandesBR (approved):

PR #496, opened 2023-11-30
DOC: Change from % to ! in the first cell to run properly in Colab.

giovaniceotto:
## Pull request type

## Checklist

## Current behavior

The first code cell of our Google Colab Getting Started Notebook fails because it is using `%` instead of `!` to execute shell commands.

## New behavior

1. In the first code cell, `%` was replaced by `!` and `netCDF` was removed from `pip install` since it is already installed when RocketPy is installed now.
2. ``%load_ext autoreload` was removed since it is not relevant for the Getting Started notebook.
3. Text cells were adjusted accordingly. 

## Breaking change

- [x] No

## Additional information

This does not affect the RocketPy package in PyPI, so a new release is not necessary. As it affects only an example referenced in the ReadMe, I suggest merging it directly to master soon.

You can run the new fixed notebook here: https://colab.research.google.com/github/RocketPy-Team/rocketpy/blob/doc/fix-getting-started-colab-notebook-error/docs/notebooks/getting_started_colab.ipynb

### DIFF
--- a/docs/notebooks/getting_started_colab.ipynb
+++ b/docs/notebooks/getting_started_colab.ipynb
@@ -22,10 +22,9 @@
     "We start by setting up our environment. To run this notebook, we will need:\n",
     "\n",
     "- RocketPy\n",
-    "- netCDF4 (to get weather forecasts)\n",
     "- Data files (we will clone RocketPy's repository for these)\n",
     "\n",
-    "Therefore, let's run the following lines of code:\n"
+    "Therefore, let's run the following lines of code:"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


Gui-FernandesBR:
Could we no longer use a git clone command in this notebook?

We are currently downloading more than 1.7 GB of data just to use a drag curve and a thrust curve. It's completely unnecessary.

I would suggest we should download only the two files we need. This would speed up our getting started example by a lot.

Gui-FernandesBR:
The following notebooks also suffer from the same problem:

- dispersion_analysis.ipynb
- parachute_drop_from_helicopter.ipynb
- deployable_payload_example.ipynb


Should we also update them?

Gui-FernandesBR:
The `%` works for me on colab.

image

PR #497, opened 2023-11-30
FIX: changes Generic Motor exhaust velocity to cached property

giovaniceotto:
## Pull request type

## Checklist

## Current behavior

`GenericMotor.exhaust_velocity` is being decorated by `funcify_method`, which makes it (a float essentially) into a `lambda` defined `Function`. This causes everything that depends on it algebraically to become `lambda` defined `Functions` as well.

More specifically, this contaminates `GenericMotor.mass_flow_rate`, which in turn contaminates `GenericMotor.propellant_mass` (which actually includes an `integral_function` of the `lambda` `GenericMotor.exhaust_velocity` 🫣). Well, in the end, this contaminates `Rocket.total_mass`, which is used extensively by the `Flight` class during a simulation.

In summary, this use of `funcify_method` causes the `Flight` class to take forever (literally) to run.

## New behavior

What can I say? Changing the decorator from `funcify_method` to `cached_property` makes the sun shine once again, solving everything.

## Breaking change

- [x] No

### DIFF
--- a/rocketpy/motors/motor.py
+++ b/rocketpy/motors/motor.py
@@ -1214,7 +1214,7 @@ class GenericMotor(Motor):
         """
         return self.propellant_initial_mass
 
-    @funcify_method("Time (s)", "Exhaust velocity (m/s)")
+    @cached_property
     def exhaust_velocity(self):
         """Exhaust velocity by assuming it as a constant. The formula used is
         total impulse/propellant initial mass.
--- a/rocketpy/motors/motor.py
+++ b/rocketpy/motors/motor.py
@@ -1214,7 +1214,7 @@ class GenericMotor(Motor):
         """
         return self.propellant_initial_mass
 
-    @cached_property
+    @funcify_method("Time (s)", "Exhaust velocity (m/s)")
     def exhaust_velocity(self):
         """Exhaust velocity by assuming it as a constant. The formula used is
         total impulse/propellant initial mass.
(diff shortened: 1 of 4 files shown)


phmbressan (dismissed):
Thank you for going through the journey to discover the bottleneck in this class. I was postponning for some time. Great finds!

Won't even start about snowball effect that the `Function` is currently generating that slows down the code so much. I will only say that the sooner this is fixed, the better.

Gui-FernandesBR (approved):
I swear that I tried but I honestly have no idea of how you discovered this error in such a chain of calls.

Gui-FernandesBR:
Before merging, do you see any possibility for new unit tests here? It's quite complicated because the error was only happening due to a "snowball" effect.

Gui-FernandesBR (approved):
LGTM

rocketpy/motors/motor.py
```diff
@@ -1224,7 +1224,9 @@ def exhaust_velocity(self):
         self.exhaust_velocity : Function
             Gas exhaust velocity of the motor.
         """
-        return self.total_impulse / self.propellant_initial_mass
+        return Function(
+            self.total_impulse / self.propellant_initial_mass
+        ).set_discrete_based_on_model(self.thrust)
```
MateusStano, line 1229: Do we need to return a Function here if this is already a `@funcify_method` ?

MateusStano (approved):

rocketpy/motors/motor.py
phmbressan: The Function is called so we can discretize it. If we applied @funcify directly, the Function source would be a numerical value.

The main issue this PR tackles are hidden lambda based Functions that end up cascading the others in arithmetic, making code slow.

Issue #480, opened 2023-11-23
TST: Restructuring the test suite


lucasfourier:
**Is your feature request related to a problem? Please describe.**

RocketPy test suite is currently very slow. Among other factors, this has to do with the fact that unit tests and integration tests (which are being treated as unit tests in many ways) are being kept in the same folder. 

**Describe the solution you'd like**

Restructuring the test suite will bring clarity as well as proper treatment depending on the nature of the test. With regards to the unit tests, they need to be refactored: commented, rewritten and etc.

**Additional context**

Even though some methods had tests written for them, bugs occurred. Of course, this situation might happen even though unit tests are available for a certain method. However, that also indicates that the test suite/testing process is not robust enough. As an example of this, check this bug:
https://github.com/RocketPy-Team/RocketPy/pull/476

- [x] Add a simple documentation or guideline to the "development" section in our documentation, just to ensure people know how to start creating new tests if needed.

Gui-FernandesBR:
Good idea! 

I liked the idea of separating unit and integration tests.

Here some other suggestions (not mandatory):

- https://pypi.org/project/pytest-random-order/
- https://github.com/kevlened/pytest-parallel

Also, the acceptance tests are quite important, we should improve them by bringing more examples of real flights to be tested.

lucasfourier:
Nice! Gonna check the suggestions!

lucasfourier:
------"Algorithm" that will run the reorganization:------

**1.** For each module x in the tests directory, copy x into the new directory unit tests.
**2.** After that, classify those tests which are certainly unit and leave them there. 
**3.** Remove the ones which are not unit and are inside the module. (Observe that they are not deleted forever, you could still find them in the original test_function.py testing module).
**4.** Now, having copied each testing module to the unit test folder, refactor every unit test in every testing module (if needed)
**5.** Finished unit testing organization. Start developing integration tests.

Pull request in each step.

PR #499, opened 2023-12-01
TST: Splitting unit testing module from non unit testes inside the new 'unit' folder for the several testing modules.

lucasfourier:
## Pull request type

## Checklist

## Current behavior

Every testing module is inside the 'tests' folder. 

## New behavior

Inside a newly created folder called 'unit', testing modules which contained only tests classified as unit tests were brought inside the unit folder. There are nine modules inside the 'unit' folder, and all of them should contain only unit tests. This is the beginning of the procedure described in https://github.com/RocketPy-Team/RocketPy/issues/480. Every testing module outside this 'unit' folder contains only tests which are not unit tests. The test_tank.py was not further analyzed because it needs to be worked on more carefully later.

## Breaking change

- [ ] Yes
- [X] No

## Additional information

This PR is part of the effort to reorganize the testing suite. Further information can be found here: https://github.com/RocketPy-Team/RocketPy/issues/480

New behavior:
image


Gui-FernandesBR (changes requested):
I think I finally understand this PR after the discussions we had yesterday.

I have nothing against the new `tst/testing-suite-reestructuring` file.

However, what is the difference between this new file and the currently https://github.com/RocketPy-Team/RocketPy/blob/develop/tests/test_function.py file? I was under the impression you were creating new code but I think you simply copy and past code.

Seems like you duplicated a file. The tests in the tests/test_function.py still exist so you're testing the same thing twice. If that's true, you should use git mv command to move one file to another folder. In that case I suggest you to delete the branch and start this work again.

lucasfourier:
Actually, what was classified as an unit test was kept inside the test_function.py module inside the folder 'unit'. What was not an unit test remained intact in the original test_function.py file.

Thanks a lot to @Gui-FernandesBR which greatly contributed this work.

Gui-FernandesBR (approved):
Looks great to me, Great work @lucasfourier !!

In the future we want to run unit tests before integration tests when triggering the github actions, this will save us some time.

Next step is defining the integration tests schema, I"m still not sure how to do or where to save them.

PR #362, opened 2023-05-08
TST: skip tests for draft PRs and non-Python files

giovaniceotto:
## Pull request type

- [x] ReadMe, Docs and GitHub maintenance

## Pull request checklist

  - [x] Spelling has been verified
  - [x] Code docs are working correctly 

## What is the current behavior?

GitHub Actions runs a PyTest workflow on draft PRs and even when non python files are updated, wasting resources and causing annoyances. 

## What is the new behavior?

- When a PR is in draft mode, no tests are run, saving resources, but the workflow fails indicating that they need to be run later.
- When no python files are updated, no tests are run,

## Does this introduce a breaking change?

- [x] No

## Other information

References:

- https://github.com/orgs/community/discussions/25722
- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onevent_nametypes

PR #500, opened 2023-12-03
MNT: update github actions with logical sequence of tests

Gui-FernandesBR:
## Pull request type

## Checklist
- [ ] `CHANGELOG.md` has been updated (if relevant) (not relevant)

## Current behavior

Important note: The skip-if-draft (https://github.com/RocketPy-Team/RocketPy/pull/362) job is not working currently. I see some execution times up to 12 hours on our organization account, let's try to solve it (e.g. https://github.com/RocketPy-Team/RocketPy/actions/runs/7071607213)

## New behavior
- Updating the github workflows, specially the job of failure if marked as a draft.
- The tests now are ran following a business sequence: first we run the unit test, than the "integration" tests, acceptance, documentation.
- If the PR is going to the master branch, the pytest --runslow will also be triggered.
- Code coverage report is shared among all the pytest executions
- We are now caching the python libraries installed via pip
- no longer automatically ask review from individuals rather than the code-owners

## Breaking change
- [x] No

## Additional information
Extra reading:
- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions
- https://github.com/actions/setup-python#caching-packages-dependencies
- https://stackoverflow.com/questions/58139406/only-run-job-on-specific-branch-with-github-actions
- https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows


Gui-FernandesBR:
Kinda doesn't work as expected. Gonna close it for a while so nobody reviews it, but we can take a deeper look in the short-future.

Gui-FernandesBR:
After almost 4 hours spent on figuring out how the GitHub actions works, I believe I finally finished this one!!
It's open for a review @RocketPy-Team/code-owners !!

- We need to urgently reduce the time spent on "integration" tests, they are consuming too much of our time. This not only drains the limited quota that we have on github actions but also let us waiting for about 15-20 minutes each time we want to validate the checks. 
- I tried to run concurrent jobs to open more runners at once on github in order to reduce the waiting time, but this was tricky since the amount of time speant on each runner will count for the final usage of github actions.
- I also think that it should not be necessary to run pytest in all the 3 operational system every time. Maybe in the future we could trigger the 3 of them only wen targeting to the master branch. But I left as it is for now.
- After approving, the 32 commits can be squashed into a single one, without any issues.

Final comment: This was the PR #500 :)

Gui-FernandesBR:
Also, I wonder why the tests on python 3.12 are much slower than the 3.8, it should be the opposite.

.github/workflows/lint_black.yaml
```diff
@@ -16,10 +16,10 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v1
         with:
-          python-version: 3.8
+          python-version: 3.10.13
```
phmbressan, line 19: Any particular reason to choose this version?

Really small nitpick here, but I would keep the `3.8` so we can ensure that developing `rocketpy` is also possible with the oldest supported `Python` version.  If there is a meaningful reason to not use the `3.8`, I am fine with the change.

phmbressan (approved):

.github/workflows/lint_black.yaml
Gui-FernandesBR: You are right, I fixed it on my last commit. [374eefd](https://github.com/RocketPy-Team/RocketPy/pull/500/commits/374eefd3fe1627d53a10db98fcdc9a6a66b9b4d7)

I didn't pay attention, but black may give different results depending on the python version.

PR #501, opened 2023-12-03
ENH: optimization of get_value_opt in set_get_value_opt of class Function

styris00:
## Pull request type

## Checklist

## Current behavior

Calls to get_value_opt for shepard interpolation were slower.

## New behavior

Calls to get_value_opt for shepard interpolation are quicker (using Numpy vectorized operations).

## Breaking change

- [x] No

### DIFF
--- a/rocketpy/mathutils/function.py
+++ b/rocketpy/mathutils/function.py
@@ -496,7 +496,7 @@ class Function:
                 if len(zero_distance_index) > 0:
                     return y_data[zero_distance_index[0]]
 
-                weights = distances_squared**(-1.5)
+                weights = distances_squared ** (-1.5)
                 numerator_sum = np.sum(y_data * weights)
                 denominator_sum = np.sum(weights)
 
--- a/rocketpy/mathutils/function.py
+++ b/rocketpy/mathutils/function.py
@@ -488,19 +488,18 @@ class Function:
 
             # change the function's name to avoid mypy's error
             def get_value_opt_multiple(*args):
-                x = np.array([[float(x) for x in list(args)]])
-                numerator_sum = 0
-                denominator_sum = 0
-                for i in range(len_y_data):
-                    sub = x_data[i] - x
-                    distance = np.linalg.norm(sub)
-                    if distance == 0:
-                        numerator_sum = y_data[i]
-                        denominator_sum = 1
-                        break
-                    weight = distance ** (-3)
-                    numerator_sum = numerator_sum + y_data[i] * weight
-                    denominator_sum = denominator_sum + weight
+                x = np.array([[float(val) for val in args]])
+                sub_matrix = x_data - x
+                distances_squared = np.sum(sub_matrix**2, axis=1)
+
+                zero_distance_index = np.where(distances_squared == 0)[0]
+                if len(zero_distance_index) > 0:
+                    return y_data[zero_distance_index[0]]
+
+                weights = distances_squared**(-1.5)
+                numerator_sum = np.sum(y_data * weights)
+                denominator_sum = np.sum(weights)
+
                 return numerator_sum / denominator_sum
 
             get_value_opt = get_value_opt_multiple
--- a/tests/unit/test_function.py
+++ b/tests/unit/test_function.py
@@ -123,3 +123,38 @@ def test_integral_function():
     """
     zero_func = Function(0)
     assert isinstance(zero_func, Function)
+
+
+@pytest.mark.parametrize(
+    "x, y, z",
+    [
+        (0, 0, 11.22540929),
+        (1, 2, 10),
+        (2, 3, 15.272727273),
+        (3, 4, 20),
+        (4, 5, 24.727272727),
+        (5, 6, 30),
+        (10, 10, 25.7201184),
+    ],
+)
+def test_get_value_opt(x, y, z):
+    """Test the get_value_opt method of the Function class. Currently only tests
+    a 2D function with shepard interpolation method. The tests are made on
+    points that are present or not in the source array.
+
+    Parameters
+    ----------
+    x : scalar
+        The x-coordinate.
+    y : scalar
+        The y-coordinate.
+    z : float
+        The expected interpolated value at (x, y).
+    """
+    x_data = np.array([1.0, 3.0, 5.0])
+    y_data = np.array([2.0, 4.0, 6.0])
+    z_data = np.array([10.0, 20.0, 30.0])
+    source = np.column_stack((x_data, y_data, z_data))
+    func = Function(source, interpolation="shepard", extrapolation="natural")
+    assert isinstance(func.get_value_opt(x, y), float)
+    assert np.isclose(func.get_value_opt(x, y), z, atol=1e-6)


giovaniceotto:
Just sending the report made by @styris00 here so that all reviewers can see it.

[Optimization_report.pdf](https://github.com/RocketPy-Team/RocketPy/files/13553987/Optimization_report.pdf)

rocketpy/mathutils/function.py
```diff
@@ -488,19 +488,18 @@ def get_value_opt(x):
...
-                        break
-                    weight = distance ** (-3)
-                    numerator_sum = numerator_sum + y_data[i] * weight
-                    denominator_sum = denominator_sum + weight
+                x = np.array([[float(val) for val in args]])
+                sub_matrix = x_data - x
+                distances_squared = np.sum(sub_matrix**2, axis=1)
+
+                zero_distance_index = np.where(distances_squared == 0)[0]
+                if len(zero_distance_index) > 0:
```
Gui-FernandesBR, line 496: when does this condition doesn't get true?

styris00: if len(zero_distance_index) > 0:

- is True when x (which comes from *args) is an element of x_data. In this case you just have to return the corresponding value of y_data.
- is False when x is not exactly an element of x_data. In this case you have to return an output based on the weighted average calculation with distances.


If the tests do not cover the 3 lines below, maybe it's because there are only tests with values directly from the source used to create the Function object.


You can try it with the example : 

```
import numpy as np


def get_value_opt_multiple(*args):
    x = np.array([[float(val) for val in args]])
    sub_matrix = x_data - x
    distances_squared = np.sum(sub_matrix**2, axis=1)

    zero_distance_index = np.where(distances_squared == 0)[0]
    if len(zero_distance_index) > 0:
        print("Input in the source")
        return y_data[zero_distance_index[0]]
    
    print("Input not in the source")
    weights = distances_squared ** (-1.5)
    numerator_sum = np.sum(y_data * weights)
    denominator_sum = np.sum(weights)

    return numerator_sum / denominator_sum


# Sample data
x_data = np.array([[1.0, 2.0], [3.0, 4.0], [5.0, 6.0]])
y_data = np.array([10.0, 20.0, 30.0])


# Input in the source
args = (3.0, 4.0)
result = get_value_opt_multiple(*args)
print("Result:", result)

# Input not in the source
args = (4.0, 5.0)
result = get_value_opt_multiple(*args)
print("Result:", result)
```


The output is : 

```
Input in the source
Result: 20.0
Input not in the source
Result: 24.727272727272723
```

styris00:
> @styris00, can you share the code you used for your execution time reports? Specially the definition of the Function instances.

I can't manage to send a python file so I'll put the code below.

To perform the tests below and compare the two versions of get_value_opt, you have to modify the set_get_value_opt() method so the Function instance have the attributs : 
- get_value_opt1 : a function that correspond to the former get_value_opt attribut
- get_value_opt2 : a function that correspond to the new get_value_opt attribut

(You may have to also change the RocketPy import)

Note that the matrix MATRIX_RESULTS displayed by the tests method is the one we put on the optimization report

```
import sys
sys.path.append('..')

import numpy as np
from RocketPy.rocketpy import Function  # May need to be changed
from timeit import Timer
import random as rd



def create_data(dim, size):  # Create a source to then create a Function Object
    list = []
    for s in range(size):
        tuple = ()
        for d in range(dim): # inputs
            tuple = tuple + (rd.uniform(-1000, 1000),)
        tuple = tuple + (rd.uniform(-50, 50),) # output
        list.append(tuple)
    array = np.array(list)
    return array

def create_point(dim):  # Create a point as a tuple to call get_value_opt
    point = ()
    for d in range(dim):
        point = point + (rd.uniform(-1000, 1000),)
    return point


dimensions = [2, 3, 4, 5]
sizes = [100, 1000, 10000]


def tests(nb = 10):
    MATRIX_RESULTS = [["Data Dim", "Data Length", "Time before", "Time after", "Improvement"]]
    correct_values = 0
    for dim in dimensions:
        for size in sizes:
            print("Dim", dim, "Size:", size)
            correct_values_case = 0
            mean_before = 0
            mean_after = 0
            mean_improvement = 0
            for i in range(nb):
                array = create_data(dim, size)
                point = create_point(dim)
                fun = Function(                           # Definition of the Function instance with array as a source
                    array,
                    interpolation="shepard",
                    extrapolation="natural",
                    title="test function"
                )
                fun.set_get_value_opt() # for fun.get_value_opt AND fun.get_value_opt2 to work

                value_ref = fun.get_value_opt1(*point)
                #print(value_ref)
                value_new = fun.get_value_opt2(*point)
                #print(value_new)
                same_value = np.isclose(value_ref, value_new, rtol = 1e-12) # check if the two values are nearly the same (1e-12)
                #print(same_value)
                if same_value:
                    correct_values += 1
                    correct_values_case += 1

                globals()['fun'] = fun
                globals()['np'] = np

                timer = Timer(lambda: fun.get_value_opt1(*point), 'from __main__ import fun, np')
                execution_time_ref = timer.timeit(number=1)
                mean_before = mean_before + execution_time_ref

                timer = Timer(lambda: fun.get_value_opt2(*point), 'from __main__ import fun, np')
                execution_time_new = timer.timeit(number=1)
                mean_after = mean_after + execution_time_new

                improvement = (execution_time_new/execution_time_ref)*100
                mean_improvement = mean_improvement + improvement

            mean_before = mean_before/nb
            mean_after = mean_after/nb
            mean_improvement = mean_improvement/nb
            LINE_RESULTS = [str(dim), str(size), mean_before, mean_after, str(mean_improvement)+" %"]
            MATRIX_RESULTS.append(LINE_RESULTS)
            print("Correct values", correct_values_case, "/", nb)

    print("CORRECT VALUES :", correct_values,"/",nb*len(dimensions)*len(sizes))
    MATRIX_RESULTS = np.array(MATRIX_RESULTS)
    print(MATRIX_RESULTS) # the matrix printed correspond to the table in the optimization report

tests(100) # performs the tests

```

rocketpy/mathutils/function.py
```diff
@@ -488,19 +488,18 @@ def get_value_opt(x):
...
+                    return y_data[zero_distance_index[0]]
+
+                weights = distances_squared ** (-1.5)
+                numerator_sum = np.sum(y_data * weights)
+                denominator_sum = np.sum(weights)
```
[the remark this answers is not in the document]
Gui-FernandesBR: See #505 that tries to add an unit test to cover these lines.

@styris00 could you review that please?

Gui-FernandesBR (approved):
Well done! I'm happy seeing this speedup on the code.
Your work was amazing in this PR, thanks for contributions.

styris00:
> > Just sending the report made by @styris00 here so that all reviewers can see it.
> > [Optimization_report.pdf](https://github.com/RocketPy-Team/RocketPy/files/13553987/Optimization_report.pdf)
> 
> This report was really nice to read!
> 
> Just wondering, what were the computer specification when running all the tests? (for example: i7-7200U, running python 3.10 on ubuntu jammy)

The table was created with an AMD Ryzen 7 6800H on Windows 11.
The same improvements (%) were found with an Intel i5-1035G1 on Windows 11.

If you've run these tests too, we wonder if you got roughly the same results ?

Gui-FernandesBR:
> > > Just sending the report made by @styris00 here so that all reviewers can see it.
> > > [Optimization_report.pdf](https://github.com/RocketPy-Team/RocketPy/files/13553987/Optimization_report.pdf)
> > 
> > 
> > This report was really nice to read!
> > Just wondering, what were the computer specification when running all the tests? (for example: i7-7200U, running python 3.10 on ubuntu jammy)
> 
> The table was created with an AMD Ryzen 7 6800H on Windows 11. The same improvements (%) were found with an Intel i5-1035G1 on Windows 11.
> 
> If you've run these tests too, we wonder if you got roughly the same results ?

Actually I didn't run the same tests, I am trusting on your analysis.

Gui-FernandesBR:
I'm comfortable enough with this PR, I think you can merge it if you are comfortable too, or you can wait for another review if you think it's necessary. The PR's main goal was achieved.

MateusStano (approved):

PR #507, opened 2023-12-08
ENH: Alternative for State History

MateusStano:
## Pull request type

## Checklist

## Current behavior

The `AirBrakes` implementation had some questionable handling of the `state` attributes. They made the `Controller` Class work only with air brakes, and their implementation was not scalable to the other classes

## New behavior

All air brakes' `state` related attributes and functions were deleted.

Now, logging of changes is handled inside the `Controller` class:

- `Controller` class now has an attribute `observed_variables`, which is a list passed as an argument for the `controller_function`
- `observed_variables` is initialized with the `initial_observed_variables` in the class' constructor
- The `observed_variables` gets appended to with whatever the `controller_function` returns.

These changes make it so no information of the simulation is stored in the Air Brakes class. If the user wanted to see the change of deployed level through time, he would have to return time and the deployed level inside the `controller_function` and then make the plot himself.

This makes the `Controller` class much more flexible in terms of usability.

## Additional information

I have not updated the docs page yet, but there is an updated example in the notebook

This current implementation makes me question the need for the `state_history` argument of the `controller_function`, since the user could just keep what it needs in the `observed_variables`

Need opinions on this @giovaniceotto @phmbressan


rocketpy/control/controller.py
```diff
@@ -57,13 +78,16 @@ def __init__(
...
-        self.observed_objects = observed_objects
+        self.interactable_objects = interactable_objects
         self.controller_function = controller_function
         self.sampling_rate = sampling_rate
         self.name = name
         self.prints = _ControllerPrints
 
+        self.observed_variables = (
+            initial_observed_variables  # list of lists of observed variables
+        )
```
phmbressan, line 89: May I ask, what is the purpose of the brackets here? I believe this line would have the same behavior without brackets.

Nonetheless the goal here seems to create a list of lists, therefore the line likely should be:

```py
self.observed_variables = [initial_observed_variables]
```

Then, differently from a immutable tuple, you can append new states with: `self.observed_variables.append(new_variables)`. What is happening on line 121 is that you are appending to the same list of `initial_observed_variables`.

rocketpy/control/controller.py
```diff
@@ -2,42 +2,58 @@
...
         ----------
-        observed_objects : list
-            A list of objects to be observed by the controller. It can be any
-            python object. This list will be passed to the controller function
-            as positional arguments, meaning that the order of the objects in
-            this list matters. These objects will then be able to be accessed
-            and modified by the controller function.
+        interactable_objects : list
+            A collection of objects that the controller function can access and
+            potentially modify. This list is flexible and can include any Python
```
Gui-FernandesBR, line 26: "This list is flexible and can include any Python object"

this sentence let me a bit lost here. What would be examples of "any Python object"? What do you mean by that?

rocketpy/control/controller.py
```diff
@@ -2,42 +2,58 @@
...
+            3. `state` (list): The state vector of the simulation, structured as
                `[x, y, z, vx, vy, vz, e0, e1, e2, e3, wx, wy, wz]`.
-            4. A list containing the state history of the simulation. The state
-               history is a list of every state vector of every step of the
-               simulation. The state history is a list of lists, where each
-               sublist is a state vector and is ordered from oldest to newest.
-            5. A list containing the objects to be acted upon by the controller.
-               The objects in this list are the same as the objects in the
-               observed_objects list, but they can be modified by the controller.
+            4. `state_history` (list): A record of the rocket's state at each
```
Gui-FernandesBR, line 40: so the `state_history` works like the same as the `Flight.solution` attribute, right?
I wonder if each state could/should be an instance of a class, so we would be saving a list of objects of such class, instead of saving a list of lists.

rocketpy/control/controller.py
```diff
@@ -2,42 +2,58 @@
...
+               step throughout the simulation. The state_history is organized as
+               a list of lists, with each sublist containing a state vector. The
+               last item in the list always corresponds to the previous state
+               vector, providing a chronological sequence of the rocket's
+               evolving states.
+            5. `observed_variables` (list): A list containing the variables that
+               the controller function returns. At every call of the controller
```
Gui-FernandesBR, line 47: I'm sorry but I could understand this sentence: " At every call of the controller function, the variables this list is appended to with the variables that the controller function returns."

Could you rephrase that? or explain it.

rocketpy/rocket/rocket.py
```diff
@@ -1243,7 +1254,11 @@ def add_air_brakes(
...
         )
         controller = Controller(
-            air_brakes, controller_function, sampling_rate, controller_name
+            interactable_objects=air_brakes,
+            controller_function=controller_function,
+            sampling_rate=sampling_rate,
+            initial_observed_variables=initial_observed_variables,
+            name=controller_name,
         )
         self.air_brakes.append(air_brakes)
```
Gui-FernandesBR, line 1263: don't we have a self.add_air_brakes(air_brakes) method? These two rows (1263 and 1264) does the same thing but in two different ways.

Gui-FernandesBR (commented):
Good work, Stano. Overall you reduced the total number of code line, that's always a good improvement.

I wonder why the tests are not passing... Could you take a look please?

rocketpy/control/controller.py
phmbressan: Agreed, likely some misplaced words: my suggestion would be similar to: `The return of each controller function call is appended to the observed_variables list`.

rocketpy/rocket/rocket.py
```diff
@@ -1161,6 +1161,7 @@ def add_air_brakes(
         sampling_rate,
         clamp=True,
         reference_area=None,
+        initial_observed_variables=None,
         name="AirBrakes",
         controller_name="AirBrakes Controller",
     ):
```
phmbressan, line 1167: Regarding the `plots` situation:

- The earlier implementation has the benefit that we control the `observed_variables`, therefore we could make useful plots;
- In this PRs case, we don't have this information due to the generality of the controller. However this allows for a much greater flexibility.

In my opinion, it would be interesting to have the best of both worlds by using the "convenience" methods for controller creation - such as `add_air_brakes` - to obtain enough information for generating the plots.

In general, one looking for a complex control will generate it from scratch and make their own plots. Alternatively, one using the convenience of `add_air_brakes` likely also wants plots from it without too many manual operations.

Therefore I was looking for a way that we can use the context of `add_air_brakes` to generate their plots. This would require imposing the three first parameters of observed variables and mapping the controller to the air brake somehow, but I could not find a reasonable way to do it.

This was just to explore ways to make this implementation easier for the user, not to pinpoint any particular change. I am open to suggestions.

rocketpy/control/controller.py
```diff
@@ -2,42 +2,58 @@
...
-            5. A list containing the objects to be acted upon by the controller.
-               The objects in this list are the same as the objects in the
-               observed_objects list, but they can be modified by the controller.
+            4. `state_history` (list): A record of the rocket's state at each
+               step throughout the simulation. The state_history is organized as
+               a list of lists, with each sublist containing a state vector. The
+               last item in the list always corresponds to the previous state
+               vector, providing a chronological sequence of the rocket's
+               evolving states.
+            5. `observed_variables` (list): A list containing the variables that
```
phmbressan, line 46: @MateusStano, about your comment in the description that the `observed_variables` may substitute the need of the `state`.

- If it were to substitute, every time a variable from the rocket state is needed the rocket would have to be added to the interactable objects;
- Due to the rocket state importance and ubiquity for control computation, I believe it is fair that it is a default argument of the controller function (`state`) for convenience.

You are indeed correct that the observable variables can substitute the `state`, but I do not think it is worth it. Let me know what you think.

phmbressan (commented):
This PR offers much improved flexibility for controlling different rocket components, well done. I will approve it after some comments I have raised are considered.

Every implementation has advantages and disadvantages, the main comments I made concern these:

Advantages:

- Greater flexibility and scalability, does not require history parameters in the controlled objects;
- Support for much deeper/complex control.

Disadvantages:

- There are no automatically generated plots for control results;
- (Perhaps) A bit harder to use/understand when creating a controller from scratch.

rocketpy/control/controller.py
```diff
@@ -2,42 +2,58 @@
...
-            as positional arguments, meaning that the order of the objects in
-            this list matters. These objects will then be able to be accessed
-            and modified by the controller function.
+        interactable_objects : list
+            A collection of objects that the controller function can access and
+            potentially modify. This list is flexible and can include any Python
+            object. The objects listed here are provided to the controller
+            function as the last argument, maintaining the order specified in
+            this list. The controller function gains the ability to interact
+            with and make adjustments to these objects during its execution.
```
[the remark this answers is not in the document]
MateusStano: Yes, changing some attributes will not work correctly. That's why we added setters and getter in the aero surfaces a long time ago, so some things (like roll control) will still work.

We could try to define exactly what attributes can be changed inside the controller function, but I do not see an easy way to do that. Do you have any suggestions?

Also, this is one of the big reasons why, to add a robust controller implementation, we need to change a lot of the code structure. On a first thought, some of the attributes in relevant classes (rocket, aerosurfaces...) should probably be an object of a class that deals with its ''controllability''. Then we would be able to clearly define what is controllable and what is not.

rocketpy/rocket/rocket.py
MateusStano: We can just add the state history for the air brakes again, except that the state history would only save information, and none be passed to the controller function. This might still require adding things in the flight class

However, I generally think that if we do a good job with the documentation, the user should be able to easily plot the data and so on

rocketpy/control/controller.py
MateusStano: Okay, lets keep it then

MateusStano:
> Advantages:
> 
> * Greater flexibility and scalability, does not require history parameters in the controlled objects;
> * Support for much deeper/complex control.
> 
> Disadvantages:
> 
> * There are no automatically generated plots for control results;
> * (Perhaps) A bit harder to use/understand when creating a controller from scratch.

To sum up, I think we can always add the air brakes plots somehow later. It would have to be something similar to the old implementation (which is not ideal) but without making it necessary for any object that can be controlled to have it as well.

Also, with a good documentation (specially in a usage .rst docs page) the user should not have a problem handling the data

rocketpy/control/controller.py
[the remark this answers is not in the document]
phmbressan: Understood. Indeed full control support is not supported right now, but it is good that this PR structure facilitates this refactor in the future.

rocketpy/rocket/rocket.py
phmbressan: Good point. Let's focus on making a proper documentation for this and, if there is the need, we can implement something to plot it automatically.

phmbressan (approved):
Thanks for your relies @MateusStano this solves the comments I had about this PR. Good job!

giovaniceotto:
@MateusStano, can we merge this PR and then I'll do a final review of #426 directly?

rocketpy/control/controller.py
MateusStano: Well, the user can pass any type of object in this list. If they want to use an object of a class that they defined they can.

Since these are passed directly to the controller_function, and the controller_function is completely defined by the user, he can pass whatever he wants in here.

I will try to improve the docs a bit here

rocketpy/control/controller.py
MateusStano: It is an interesting idea. Maybe being a named tuple would be more fitting

It is something to consider whenever we work on some refactoring in the Flight class

rocketpy/rocket/rocket.py
MateusStano: I don't think I understood what you were saying here

Gui-FernandesBR: you have used these two code lines:

`self.air_brakes.append(air_brakes)`

and

`self.add_controllers(controller)`

My question is: why do you append airbakes but you add controllers? Shouldn't they follow the same architecture? The code could be more intuitive

MateusStano: `add_controllers` is a separate method that can be used by itself to add any type of controller to the rocket

the `add_air_brakes` is specific for adding air_brakes. It also adds the controller because it just makes sense for air brakes to be added together with their controller

Gui-FernandesBR: > `add_controllers` is a separate method that can be used by itself to add any type of controller to the rocket
> 
> the `add_air_brakes` is specific for adding air_brakes. It also adds the controller because it just makes sense for air brakes to be added together with their controller

Now I'm the one who thinks you didn't understand...

The point is that in your code you didn't use `add_air_brakes` at all, instead you are mutating a list using the `self.airbrakes.append()`.

Imagine when we are on the Rocket class, we don't do this: `calisto.aero_surfaces.append(...)`.
Instead, we do `calisto.add_surface(...)`.
This is important because there are some function calls inside the add_surface method that make the first approach not valid.

MateusStano: `add_air_brakes` creates the air brakes and their related `controller` and saves them in their appropriate lists.

Air brakes do not generate any lift, so there is no reason for it to be included in `aerodynamic_surfaces` or make any center of pressure-related calculations. So it does not call `add_surface`

MateusStano: Also, air brakes have no position, so it only needs to be saved in a normal list. No need to create a `Component` object

PR #510, opened 2023-12-14
BUG: fin_flutter_analysis doesn't find any fin set

Gui-FernandesBR:
## Pull request type

## Checklist

## Current behavior

This bug was reported through discord yesterday: 
image

Also, the tests for the `fin_flutter_analysis` were deactivated since they were not passing.

## New behavior
- Solved the bug
- Refactored the code to improve readability.
- Improved the tests for flutter analysis.

## Breaking change
- [x] No

## Additional information
This feature works only for trapezoidal fins, that's said.
Also, we should improve the documentation in the future, try to give some more examples.


Gui-FernandesBR:
I am still working on unit tests to prevent from future occasions. Please don't review it yet

rocketpy/utilities.py
```diff
@@ -291,25 +290,9 @@ def _flutter_safety_factor(flight, flutter_mach):
...
-        source=safety_factor,
-        inputs="Time (s)",
-        outputs="Fin Flutter Safety Factor",
-        interpolation="linear",
-    )
-
+    safety_factor = flutter_mach / flight.mach_number
+    safety_factor.set_title("Fin Flutter Safety Factor")
+    safety_factor.set_outputs("Safety Factor")
     return safety_factor
```
phmbressan, line 296: I noticed that in the earlier implementation there was a check and treatment for zero division errors before making it a `Function` input. Does this more straightforward approach have the same results, because there is some treatment for `NaN` values in the `Function` division, but those vary depending on the input source type (lambda, arrays etc).

Furthermore, due to this change, I believe there should be a unit test that specifically tests these two edge cases: when there might be a `NaN` or `inf` in safety factor calculation.

Gui-FernandesBR: Yes the whole zerodivision thing was redundant in this code currently.
The results are simply the same.

Can't add tests as u suggested bc there's no rocket that flies on mach = 0

rocketpy/utilities.py
```diff
@@ -229,52 +229,51 @@ def fin_flutter_analysis(
     ------
     None
     """
+    found_fin = False
 
-    # First, we need identify if there is at least a fin set in the rocket
-    for aero_surface in flight.rocket.aerodynamic_surfaces:
```
Gui-FernandesBR, line 234: Just to make reviewer's life easier: this was the problematic code line. The aerodynamic_surfaces is no longer a list of surfaces, instead it is a list of components;

rocketpy/utilities.py
```diff
@@ -229,52 +229,51 @@ def fin_flutter_analysis(
...
-    try:
-        s = s
-    except NameError:
-        print("There is no fin set in the rocket, can't run a Flutter Analysis.")
+            if not found_fin:
+                found_fin = True
+            else:
+                warnings.warn("More than one fin set found. The last one will be used.")
+    if not found_fin:
+        print("There is no TrapezoidalFins in the rocket, can't run Flutter Analysis.")
```
phmbressan, line 247: Should't we raise an `AttributeError` here?

phmbressan (approved):
Good to see these tests reintroduced, let's hope they don't give the same headaches as before.

rocketpy/utilities.py
Gui-FernandesBR: That's a good suggestion, way better than a print statement.
But the warning above must be kept, bc the code could still work.

Gui-FernandesBR: Solved here: [6f4c3c3](https://github.com/RocketPy-Team/RocketPy/pull/510/commits/6f4c3c306f1744821f640b3a9b07e4584d946a75)

PR #511, opened 2023-12-15
ENH Precalculate Barometric Height

MateusStano:
## Pull request type

## Checklist

## Current behavior

The `find_input` calls to get the height AGL from the pressure value for the parachute trigger was slowing down the simulation drastically:

image

## New behavior

Now a `Function` is created in `Environment` called `barometric_height` and that is called inside the `Flight` class instead:

image

## Additional information

This `find_input` call was added in #345

### DIFF
--- a/rocketpy/environment/environment.py
+++ b/rocketpy/environment/environment.py
@@ -111,6 +111,10 @@ class Environment:
     Environment.pressure : Function
         Air pressure in Pa as a function of altitude. Can be accessed as regular
         array, or called as a Function. See Function for more information.
+    Environment.barometric_height : Function
+        Geometric height above sea level in m as a function of pressure. Can be
+        accessed as regular array, or called as a Function. See Function for
+        more information.
     Environment.temperature : Function
         Air temperature in K as a function of altitude. Can be accessed as
         regular array, or called as a Function. See Function for more
@@ -1315,6 +1319,8 @@ class Environment:
 
         # Save temperature, pressure and wind profiles
         self.pressure = self.pressure_ISA
+        self.barometric_height = self.barometric_height_ISA
+
         self.temperature = self.temperature_ISA
         self.wind_direction = Function(
             0,
@@ -1433,6 +1439,7 @@ class Environment:
         if pressure is None:
             # Use standard atmosphere
             self.pressure = self.pressure_ISA
+            self.barometric_height = self.barometric_height_ISA
         else:
             # Use custom input
             self.pressure = Function(
@@ -1441,6 +1448,11 @@ class Environment:
                 outputs="Pressure (Pa)",
                 interpolation="linear",
             )
+            self.barometric_height = self.pressure.inverse_function().set_discrete(
+                0, max_expected_height, 1000, extrapolation="constant"
+            )
+            self.barometric_height.set_inputs("Pressure (Pa)")
+            self.barometric_height.set_outputs("Height Above Sea Level (m)")
             # Check maximum height of custom pressure input
             if not callable(self.pressure.source):
                 max_expected_height = max(self.pressure[-1, 0], max_expected_height)
@@ -1605,6 +1617,12 @@ class Environment:
             outputs="Pressure (Pa)",
             interpolation="linear",
         )
+        self.barometric_height = Function(
+            data_array[:, (0, 1)],
+            inputs="Pressure (Pa)",
+            outputs="Height Above Sea Level (m)",
+            interpolation="linear",
+        )
         self.temperature = Function(
             data_array[:, (1, 2)],
             inputs="Height Above Sea Level (m)",
@@ -1732,6 +1750,12 @@ class Environment:
             outputs="Pressure (Pa)",
             interpolation="linear",
         )
+        self.barometric_height = Function(
+            data_array[:, (0, 1)],
+            inputs="Pressure (Pa)",
+            outputs="Height Above Sea Level (m)",
+            interpolation="linear",
+        )
 
         # Retrieve temperature from data array
         data_array[:, 2] = data_array[:, 2] + 273.15  # Converts C to K
@@ -1845,6 +1869,7 @@ class Environment:
 
         # Extract pressure as a function of height
         pressure_array = []
+        barometric_height_array = []
         for line in lines:
             # Split line into columns
             columns = re.split(" +", line)[1:]
@@ -1858,7 +1883,9 @@ class Environment:
                     if max(columns) != 99999:
                         # Save value
                         pressure_array.append(columns)
+                        barometric_height_array.append([columns[1], columns[0]])
         pressure_array = np.array(pressure_array)
+        barometric_height_array = np.array(barometric_height_array)
 
         # Extract temperature as a function of height
         temperature_array = []
@@ -1903,6 +1930,14 @@ class Environment:
             pressure_array,
             inputs="Height Above Sea Level (m)",
             outputs="Pressure (Pa)",
+            interpolation="linear",
+        )
+        # construct barometric height array from pressure array
+        barometric_height_array[:, 0] = 10 * barometric_height_array[:, 0]
+        self.barometric_height = Function(
+            barometric_height_array,
+            inputs="Pressure (Pa)",
+            outputs="Height Above Sea Level (m)",
             interpolation="linear",
         )
 
@@ -2272,6 +2307,12 @@ class Environment:
             data_array[:, (1, 0)],
             inputs="Height Above Sea Level (m)",
             outputs="Pressure (Pa)",
+            interpolation="linear",
+        )
+        self.barometric_height = Function(
+            data_array[:, (0, 1)],
+            inputs="Pressure (Pa)",
+            outputs="Height Above Sea Level (m)",
             interpolation="linear",
         )
         self.temperature = Function(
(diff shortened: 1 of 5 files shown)


rocketpy/simulation/flight.py
```diff
@@ -697,9 +697,8 @@ def __init__(
                     parachute.noisy_pressure_signal.append([node.t, pressure + noise])
                     # Gets height above ground level considering noise
                     hAGL = (
-                        self.env.pressure.find_input(
+                        self.env.barometric_height(
                             pressure + noise,
```
Gui-FernandesBR, line 701: Remove the leading comma and run black again
```suggestion
                            pressure + noise
```

rocketpy/simulation/flight.py
```diff
@@ -1006,9 +1005,8 @@ def __init__(
                                     )
                                     # Gets height above ground level considering noise
                                     hAGL = (
-                                        self.env.pressure.find_input(
+                                        self.env.barometric_height(
                                             pressure + noise,
```
Gui-FernandesBR, line 1009: Remove the leading comma and run black again

```suggestion
                                            pressure + noise
```

rocketpy/environment/environment.py
```diff
@@ -111,6 +111,10 @@ class Environment:
     Environment.pressure : Function
         Air pressure in Pa as a function of altitude. Can be accessed as regular
         array, or called as a Function. See Function for more information.
+    Environment.barometric_height : Function
```
Gui-FernandesBR, line 114: pending unit tests for this new attribute

rocketpy/environment/environment.py
```diff
@@ -1441,6 +1448,11 @@ def process_custom_atmosphere(
                 outputs="Pressure (Pa)",
                 interpolation="linear",
             )
+            self.barometric_height = self.pressure.inverse_function().set_discrete(
+                0, max_expected_height, 1000, extrapolation="constant"
```
Gui-FernandesBR, line 1452: why 1000 points? this seems to be too much imo. I think 100 points would be accurate enough and already faster

rocketpy/environment/environment.py
```diff
@@ -1441,6 +1448,11 @@ def process_custom_atmosphere(
...
+            )
+            self.barometric_height.set_inputs("Pressure (Pa)")
+            self.barometric_height.set_outputs("Height Above Sea Level (m)")
```
Gui-FernandesBR, line 1455:
```suggestion
            self.barometric_height.set_outputs("Altitude Above Sea Level (m)")
```

And what about the title of this Function object?

rocketpy/environment/environment.py
```diff
@@ -1905,6 +1932,14 @@ def process_noaaruc_sounding(self, file):
             outputs="Pressure (Pa)",
             interpolation="linear",
         )
+        # construct barometric height array from pressure array
+        barometric_height_array[:, 0] = 10 * barometric_height_array[:, 0]
```
Gui-FernandesBR, line 1936: please explain in the code why you're multiplying by 10, it may not be so trivial to realize it.

Gui-FernandesBR (commented):
Good work, @MateusStano !! Before approving this one, I'd like to see some unit tests for the barometric height, lemme know if you need any help.

Question: What would happen if a given Environment had the same value of pressure at diferent altitude? For example, for really high altitudes.

rocketpy/environment/environment.py
MateusStano: It is automatically set by the Function class, just as the other Environment attributes that are defined here (like `pressure`)

MateusStano:
Hey! @Gui-FernandesBR solved everything you commented. 

However, during the test creation, I identified an issue with the barometric height data. In some environmental models, the pressure remains constant from the surface up to a specific altitude. For example, the pressure assumed in our tests remains constant from ground level to 722 meters.

This is a problem for the `barometric_height` calculation, as it results in a constant height for certain pressures. This can mess up some parachute triggers. For instance, using the pressure data from our tests yields a `barometric_height` function with a minimum height of 722 meters. This inconsistency can disrupt a parachute trigger set at 400 meters, because the function never reaches that value.

To solve this, I've added exatrapolation into the pressure data. I would appreciate a double check on this solution.

Gui-FernandesBR:
I know the implementation seems to be straightforward, but I'd like to ensure the code will have the best behavior under different circumstances...

1. Somethings isn't right in this plot:

image


2. What would happen if someone sets the elevation higher than the first available point? e.g. you have pressure starting from 400m but you set the elevation as 500m. (EDIT: Actually I understood now that you set the pressure at the 0 height level, which is not always the ground elevation. Am I right?). 

4. When we use ensembles, different pressure profiles are available, how does the barometric_height interact with that? If I change the selcted ensemble member, will the barometric pressure change as well? This is important when running monte carlo

5. I thought about an alternative solution here: Try to set the barometric_height's extrapolation method to "natural" rather than the default "constant". This will make the Function have equivalent behavior as in your code. Why is this a better solution? Less code lines, so less chance to get the things wrong.

6. Something that I'm concerned here is that the pressure and the barometric_height might become incompatible somehow. Meaning that the inverse of barometric_height may not necessarly correspond to the pressure profile, as well as the inverse of pressure may not correspond exactly to the barometric_height. Do you see something similar happening here?

rocketpy/environment/environment.py
```diff
@@ -1605,6 +1617,25 @@ def process_windy_atmosphere(self, model="ECMWF"):
...
+        pressure_ground = np.array(
+            [
+                [
+                    columns[1][0]
+                    + (-columns[1][1] / (columns[0][1] - columns[1][1]))
+                    * (columns[0][0] - columns[1][0]),
+                    0,
+                ]
+            ]
+        )
```
phmbressan, line 1631: I don't have much experience to talk about the validity of this extrapolation, but I advise being cautious with it so that things do not lose physical meaning. If necessary, we can find better ways to do it.

As a sidenote, if these code lines are so important, I would make a function in `tools.py` (or a static method here) to acomplish the extrapolation. I also recommend using more intermediate variables for some columns names so the code is easier to understand.

An extra would be also a simple unit test for this extrapolation method.

MateusStano:
Hey, guess I forgot the `extrapolation="natural"` was a thing heheh

I changed the implementation to use this, and it seems more consistent. The error showed by @Gui-FernandesBR has been fixed.

Regarding the "correctness" of the barometric_height, it should not be very wrong. The `barometric_height` is essentially calculated in two possible ways. The first is by simply inverting the data used to define the pressure directly, meaning the `barometric_height` is as "correct" as the pressure curve. The other way is using the ISA modeling, by inverting the ISA pressure model, which is pretty simple so it should not lead to big errors.

The only inconsistency happens with models that consider the pressure to be constant from height 0 to a certain level, like this:

image

This is using the Wyoming Upper Air Soundings example from the environment usage notebook.

Since we are dealing with parachutes that commonly have triggers at low altitudes, simply inverting this curve would not be sufficient for our applications. The resulting `barometric_height` of this is model is:

image

By using `extrapolation="natural"` the `barometric_height` stays consistent.

Note that, in this case, the reason for the pressure to be constant below 720m is because the data is from a weather observation site located at an elevation of 720m ASL.

Also note that the height here can reach negative values. However, unless the user has set the elevation of the environment to be negative, this will never be called in the simulation.

The extrapolation is the most logical way to extend the data from environment models to our use case.

Gui-FernandesBR:
> Hey, guess I forgot the `extrapolation="natural"` was a thing heheh
> 
> I changed the implementation to use this, and it seems more consistent. The error showed by @Gui-FernandesBR has been fixed.
> 
> Regarding the "correctness" of the barometric_height, it should not be very wrong. The `barometric_height` is essentially calculated in two possible ways. The first is by simply inverting the data used to define the pressure directly, meaning the `barometric_height` is as "correct" as the pressure curve. The other way is using the ISA modeling, by inverting the ISA pressure model, which is pretty simple so it should not lead to big errors.
> 
> The only inconsistency happens with models that consider the pressure to be constant from height 0 to a certain level, like this:
> 
> image
> 
> This is using the Wyoming Upper Air Soundings example from the environment usage notebook.
> 
> Since we are dealing with parachutes that commonly have triggers at low altitudes, simply inverting this curve would not be sufficient for our applications. The resulting `barometric_height` of this is model is:
> 
> image
> 
> By using `extrapolation="natural"` the `barometric_height` stays consistent.
> 
> Note that, in this case, the reason for the pressure to be constant below 720m is because the data is from a weather observation site located at an elevation of 720m ASL.
> 
> Also note that the height here can reach negative values. However, unless the user has set the elevation of the environment to be negative, this will never be called in the simulation.
> 
> The extrapolation is the most logical way to extend the data from environment models to our use case.

Thanks for the comments! Gonna review it again as soon as possible just to double check it.

rocketpy/environment/environment.py
```diff
@@ -2965,7 +3025,14 @@ def pressure_function(h):
             outputs="Pressure (Pa)",
         )
 
-        return None
+        # Save international standard atmosphere height by pressure profile
+        # and discretize it. This is done to speed up the calculations in the
+        # trajectory simulation.
```
Gui-FernandesBR, line 3030: Shorten comments are usually better. Also, comment "why", not "what"

```suggestion
        # Discretize Function to speed up the trajectory simulation.
```

Gui-FernandesBR (approved):
Good work here, @MateusStano !

I ran some examples locally and everything seems to be working correctly.
I'm also trusting the unit tests here.

This PRs reminds me that the Environment class has some code repetitions. We could work on a refactor/overhaul in the future. Lot of the code remains the same since the first commits.

Gui-FernandesBR (approved):

PR #512, opened 2023-12-15
ENH: Remove Pydantic from Remaining Classes

MateusStano:
## Pull request type

## New behavior

All Stochastic classes have now been updated to the new architecture.

This can be tested in the dispersion usage notebook.

Nothing has been changed in the `Dispersion` class yet.

## Additional Information

Next dispersion related PR will be changing all the classes names and file structure


rocketpy/rocket/components.py
```diff
@@ -63,27 +63,6 @@ def add(self, component, position):
...
-        list
-            A list of components matching the specified type.
-        """
-        component_type_list = [
-            c.component
-            for c in self._components
-            if isinstance(c.component, component_type)
-        ]
-        return component_type_list
-
```
Gui-FernandesBR, line 86: Hey, I think this method could be useful to other applications rather than the monte carlo analysis. 

Btw what you did in here is a breaking change, formally speaking.

I recommend not removing it.

rocketpy/monte_carlo/mc_flight.py
```diff
@@ -1,90 +1,140 @@
...
+                    "vx_init, vy_init, vz_init, e0_init, e1_init, e2_init, "
+                    "e3_init, w1Init, w2Init, w3Init"
+                )
+                assert all(
+                    [isinstance(i, (int, float)) for i in initial_solution]
+                ), "`initial_solution` must be a tuple of numbers"
+            else:
+                raise TypeError("`initial_solution` must be a tuple of numbers")
+
+    # TODO: these call dict_generator a lot of times unecessaryly
```
Gui-FernandesBR, line 100: why is that happening?

Gui-FernandesBR (commented):
Partial review only. Good to see all these new changes.

rocketpy/monte_carlo/mc_rocket.py
```diff
@@ -26,483 +27,553 @@
...
 
-    def addMotor(self, motor, position):
-        """Adds a motor to the McRocket object.
+        # checks if input is a Motor
+        if not isinstance(motor, (Motor, MotorDispersionModel)):
+            raise AssertionError("`motor` must be a McMotor or Motor type")
+        if isinstance(motor, Motor):
+            # create McMotor
+            # TODO check motor type when hybrids and liquids are implemented
+            motor = McSolidMotor(solid_motor=motor)
```
phmbressan, line 247: Will this work for `GenericMotor`?

phmbressan (commented):
Good, certainly the code is longer without `Pydantic`, but as least we don't have to add another dependency. You made a good job with all these hard validations.

Side-note before approval, you should check out the comment regarding the `components.get_by_type` removal, I also think it is a breaking change.

rocketpy/monte_carlo/mc_rocket.py
MateusStano: Nope, fixed it in the last commit

phmbressan (approved):

rocketpy/monte_carlo/mc_flight.py
Gui-FernandesBR: > This is necessary in the `Dispersion` class to run a simulation.
> 
> To define a `McFlight` you need to pass a `Flight` object, which has a `Rocket` and a `Environment`. However, the `Dispersion` class receives a `McFlight`, `McRocket` and `McEnvironment` and has to construct a brand new flight from those objects. The `McFlight` `create_object` method is never used in the Dispersion class, instead you only need randomized values for rail_length, inclination and heading
> 
> This PR doesn't change anything in the `Dispersion` class, so I simply converted the `McFlight` class to not have Pydantic, but keep it working the same as before.
> 
> Whenever we start working on the `Dispersion` class we should discuss this further. Honestly, I do not think that the `McFlight` class should exist, or at least it should not be passed to the `Dispersion` class

We can re-discuss in the future, the whole architecture of Monte Carlo analysis is something we can optimize before merging to develop.

phmbressan:
There is one thing about this PR that I just noticed that is is worth thinking about.

In general, `assert` statements are not used in `Python` for production code, since they are ignored when running a optimized build `python -O`.

This is not a huge problem for now, since I think your build in `pip install` is a standard one. This behavior of `assert` is a hotly debated one in Python Community, auch as this [discussion thread](https://discuss.python.org/t/stop-ignoring-asserts-when-running-in-optimized-mode/13132).

rocketpy/monte_carlo/mc_solid_motor.py
```diff
@@ -138,18 +208,21 @@ def __init__(
             center_of_dry_mass_position=center_of_dry_mass_position,
             nozzle_position=nozzle_position,
             throat_radius=throat_radius,
+            interpolate=None,
+            coordinate_system_orientation=None,
```
Gui-FernandesBR, line 212: I think nobody will ever change the `coordinate_system_orientation` in a Monte Carlo loop, but I'm not sure what were your intentions here so just double check it please.

Gui-FernandesBR (approved):
Validating user inputs is not an easy task, but I think this PR is dealing this sufficiently. 

Ready to be merge, no?

rocketpy/monte_carlo/mc_solid_motor.py
MateusStano: Yep, these lines are guaranteeing that the interpolation and coordinat_system_orientation will be gotten from the original object. Note that these arguments are not on the class' constructor

PR #514, opened 2023-12-17
ENH: adds new Function.savetxt method

Gui-FernandesBR:
## Pull request type

## Checklist

## Current behavior
If you have a Function object, good luck trying to export this to be used later or to send to someone else.

## New behavior

Inspired by the [numpy.savetxt](https://numpy.org/doc/stable/reference/generated/numpy.savetxt.html) method, I've created the Function.savetxt that is currently working very well for both 1D or N-D Function, with or without discrete sources.

## Breaking change
- [x] No

## Additional information
- In the future we may need a `to_serializable` in other to export different attributes of the Function instances to a .json file. Let's leave this for a future PR.
- The tests are running really fast... thank you numpy for exsisting.
- Also, can I say how better the Function classes are now after we officialized the support for csv files with headers? Amazing feature!


rocketpy/mathutils/function.py
```diff
@@ -2817,6 +2817,76 @@ def compose(self, func, extrapolate=False):
...
+                )
+            # Generate the data points using the callable
+            x = np.linspace(lower, upper, samples)
+            data_points = np.column_stack((x, self.source(x)))
+        else:
+            # If the source is already an array, use it as is
+            data_points = self.source
+
+            if lower and upper and samples:
+                data_points = self.set_discrete(lower, upper, samples).source
```
phmbressan, line 2883: I advise deepcopying with `copy.deepcopy` and then performing discretization over this temporary object, since the `set_discrete` method is a instance mutator: it changes the `Function` source. This is unlikely to be the intention or expected when running this method, I believe.

phmbressan (commented):
Great

rocketpy/mathutils/function.py
```diff
@@ -2817,6 +2817,76 @@ def compose(self, func, extrapolate=False):
                 extrapolation=self.__extrapolation__,
             )
 
+    def savetxt(
+        self,
```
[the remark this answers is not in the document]
MateusStano: Actually, just saw what you commented in the description. I still think it could be more descriptive to use `to_` names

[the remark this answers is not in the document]
Gui-FernandesBR: This is an interesting suggestion.

Between numpy and pandas, I have to say that numpy might be a better inspiration in this case, since pandas is not one of the rocketpy's dependency.

Beyond that, I believe we need to be careful with the `to_` methods, let's see:
- `.savetxt`: takes a Function instance and save its source to a text file, where each row represents one point from the source.
- `.savecsv`: this call the savetxt method above but modify the file extension to .csv
- `.to_json` (or other similar name like ``) takes the Function instance and create a json file with different attributes from the object (e.g. interpolation method, inputs and outputs names, repr method, etc.)

If we change the `savetxt` to `to_txt` we would be creating an opportunity for confusion since the `to_txt` and `to_json` would export totally different contents.
There's a reason for having the `save` methods to save the source and the `to_` methods to convert the *entire object* to something else.

Let me know in case your disagree with this.

The #522 might be connected to this comment too.

rocketpy/mathutils/function.py
Gui-FernandesBR: I'm currently waiting for the #519 resolution.

Gui-FernandesBR: Solved here: [696e709](https://github.com/RocketPy-Team/RocketPy/pull/514/commits/696e709945fea9d957a2615ef72c5f570218260f)

The #519 made my life much easier here, thank you @phmbressan !!

Gui-FernandesBR:
All fixed here.
@MateusStano @phmbressan could you please re-review it? 
Thank you for the past comments btw, I'm happy that you brought my attention to certain points.

phmbressan (approved):
As I mentioned in my last review, really good, this PR makes it really easy to export data in RocketPy. Nice to see that #519 was indeed useful.

PR #515, opened 2023-12-19
ENH: Shepard Optimized Interpolation - Multiple Inputs Support

phmbressan:
## Pull request type

## Checklist

## Current behavior

The optimized method `get_value_opt` for `shepard` interpolation had a great speed up by #501. However, this method did not support (even before than the #501) arrays of inputs.

## New behavior

Inspired by the implementation made in the aforementioned PR, the support for multiple inputs was added in the optimized way, benefiting from the great speed up implemented. Since this also satisfies the requisites for the common `get_value` method while being faster, it was also updated.

## Breaking change

- [X] No


tests/unit/test_function.py
```diff
@@ -48,6 +48,52 @@ def test_integral_spline_interpolation(request, func, a, b):
...
+)
+def test_3d_shepard_interpolation(x, y, z, w_expected):
+    """Test the shepard interpolation method of the Function class."""
+    # Test plane x + y + z + w = 1
+    source = [(1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1)]
+    func = Function(source=source, inputs=["x", "y", "z"], outputs=["w"])
+    w = func(x, y, z)
+    w_opt = func.get_value_opt(x, y, z)
+    assert np.isclose(w, w_opt, atol=1e-8).all()
+    assert np.isclose(w_expected, w, atol=1e-8).all()
```
Gui-FernandesBR, line 94: In the first test, last assert: You test z against z_expected.
In the second test, last assert: You test w_expected against w (the expected value is now the first).

Code is working so no problem, but I believe the sequence might matter here.

tests/unit/test_function.py
```diff
@@ -48,6 +48,52 @@ def test_integral_spline_interpolation(request, func, a, b):
...
+        (0.5, 0.5, 1 / 3),
+        (0.25, 0.25, 25 / (25 + 2 * 5**0.5)),
+        ([0, 0.5], [0, 0.5], [1, 1 / 3]),
+    ],
+)
+def test_2d_shepard_interpolation(x, y, z_expected):
+    """Test the shepard interpolation method of the Function class."""
+    # Test plane x + y + z = 1
+    source = [(1, 0, 0), (0, 1, 0), (0, 0, 1)]
+    func = Function(source=source, inputs=["x", "y"], outputs=["z"])
```
Gui-FernandesBR, line 67: please consider to explicitly defining the interpolation method as the `shepard` method, this may no longer be the only, default option in the future.

Same thing happens in the second test function.

rocketpy/mathutils/function.py
Gui-FernandesBR, line 1: I honestly have no clue on how you bought these changes to the table.
It's not trivial to follow it, but I see there's an optimization here simply by reading it.
Hope one day we can improve `get_value` method using numpy operations as well.

Good work.

tests/unit/test_function.py
Gui-FernandesBR, line 1: Technically saying, the code inside this module should individually test units of the source code.
It seems like you're testing the `get_value_opt` method under a specific case of shepard interpolation, so I think the names of the tests could better reflact that, because you're not exactly testing the "2d_shepard_interpolation" or the "3d_shepard_interpolation".

However, I wouldn't stop this PR merging as the code and the tests are working properly.
In the future we can rediscuss this when we find more free time, I think the priority now is the next release.

Gui-FernandesBR (approved):
Nice job here. Thanks for the contribution.

image

tests/unit/test_function.py
phmbressan: Well noted, the sequence matters only for the error message, not in the assertion itself. Nonetheless, the order is important, I will fix it.

tests/test_function.py
```diff
@@ -215,6 +215,56 @@ def test_multivariable_dataset(a, b):
     assert np.isclose(func(a, b), a + b, atol=1e-6)
 
 
+@pytest.mark.parametrize(
+    "x,y,z_expected",
+    [
+        (0, 0, 1),
```
phmbressan, line 221: Just noticed this test case is repeated, I will fix it soon.

phmbressan: Solved by [acbc14d](https://github.com/RocketPy-Team/RocketPy/pull/515/commits/acbc14dae924d874e6a2803e2f494be92eec7f09).

Gui-FernandesBR (approved):
There was a conflict on the changelog, I solved it and now I'm re-approaving.

Excellent PR, this really demonstrantes how Brillant are some people in our team.
In this case, @phmbressan .

ps.: Thank you `numpy` for existing.

Gui-FernandesBR (approved):

PR #518, opened 2023-12-20
ENH : Add of a line for saving the filtered dataset

kalounis:
## Pull request type

## Checklist

## Current behavior

The code is able to filter data (of a CSV file) but doesn't save the filtered dataset.

## New behavior

Once you filter a dataset, this one is saved in a CSV file.

## Breaking change

- [X] No

## Additional information

Enter text here...

### DIFF
--- a/rocketpy/mathutils/function.py
+++ b/rocketpy/mathutils/function.py
@@ -1086,7 +1086,7 @@ class Function:
             extrapolation="zero",
         )
 
-    def low_pass_filter(self, alpha):
+    def low_pass_filter(self, alpha, file_path = None):
         """Implements a low pass filter with a moving average filter
 
         Parameters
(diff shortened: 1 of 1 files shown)


rocketpy/mathutils/function.py
```diff
@@ -1112,6 +1112,9 @@ def low_pass_filter(self, alpha):
                 alpha * self.source[i] + (1 - alpha) * filtered_signal[i - 1]
             )
 
+        # Save the new csv file with filtered data
+        np.savetxt("filtered_data.csv", filtered_signal, delimiter=",")
```
phmbressan, line 1116: Good feature. One small note is that it might not be the user's will to always save the file and in this location.

Therefore, what I suggest is to add an optional parameter to the method `filepath` (or another name you prefer) that defaults to `None` (in this case nothing will be saved), but may receive the filepath (and filename) string that will go into the `np.savetxt`.

The final step would be describing this behavior on this parameter docstring.

As always, we are open to any other suggestions you might have.

phmbressan (commented):
Thanks for the contributions! Left I quick comment on this feature regarding making the save behavior more versatile.

rocketpy/mathutils/function.py
kalounis: Good idea, thank you! I didn't thought about it, I will try to implement it!

phmbressan (approved):

PR #519, opened 2023-12-20
ENH: Argument for Optional Mutation on Function Discretize

phmbressan:
## Pull request type

## Checklist

## Current behavior

The behavior of `Function.set_discrete` is well documented, but not exactly intuitive: the original `Function` is always mutated. This may not always be the need of the user and it has caused problems before in development, such as #514 .

## New behavior

This PR adds a new optional parameter (whose default does not call a breaking change) so that it may be specified whether the `self` should be discretized.

Tests for the changes have been added.

## Breaking change

- [X] No

## Additional information

Please, on review make a sanity recheck that this is not a breaking change.


phmbressan:
I will work on the codecov report later.

rocketpy/mathutils/function.py
```diff
@@ -544,18 +549,25 @@ def set_discrete(
...
-            ys = self.get_value(xs.tolist()) if one_by_one else self.get_value(xs)
-            self.set_source(np.concatenate(([xs], [ys])).transpose())
-            self.set_interpolation(interpolation)
-            self.set_extrapolation(extrapolation)
-        elif self.__dom_dim__ == 2:
+            ys = func.get_value(xs.tolist()) if one_by_one else func.get_value(xs)
+            func.set_source(np.concatenate(([xs], [ys])).transpose())
+            func.set_interpolation(interpolation)
+            func.set_extrapolation(extrapolation)
+        elif func.__dom_dim__ == 2:
```
Gui-FernandesBR, line 570: Seems like the `set_discrete` method only works for 1-D and 2-D Functions, but this behavior is not documented tho. IMO, this is a problem, the function will be returned as the same. Or is there anything that I'm missing here?

Also, what is the advantage of setting the `one_by_one` to False? 

Not on the scope of this PR tho. I'd suggest to open an issue to improve the method for N-D functions if you agree that it would be nice.

Gui-FernandesBR (approved):
Well done, I don't see any breaking changes happening here.

It would be nice to include tests for 2-D functions.

rocketpy/mathutils/function.py
phmbressan: You are absolutely correct, this method behavior for higher dimensional Functions is really bad: the Function stays the same.

I do think this should be fixed and I am working on it, but this could be technically considered a breaking change.

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ straightforward as possible.
 
 ### Changed
 
+- ENH: Argument for Optional Mutation on Function Discretize [#519](https://github.com/RocketPy-Team/RocketPy/pull/519)
```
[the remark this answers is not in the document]
phmbressan: Fixed in [b7b01a4](https://github.com/RocketPy-Team/RocketPy/pull/519/commits/b7b01a431260576a4783659159f5281729a71700).

Gui-FernandesBR (approved):
LGTM

Gui-FernandesBR:
@phmbressan do you know what is happening with the GitHub actions here?

Gui-FernandesBR (approved):

PR #534, opened 2024-01-18
BUG: Parachute Pressures not being Set before All Info.

phmbressan:
## Pull request type

## Checklist

## Current behavior

The method `export_pressure` incorrectly calls the `lists`: `parachute.clean_pressure_signal` and `noisy_pressure_signal` as if they were functions. Furthermore, this data is not set before the call of the `all_info`, which could lead to invalid exports if it was working.

The issue when trying to export the pressure was: 

```py
   3091 for parachute in self.rocket.parachutes:
   3092     for t in time_points:
-> 3093         p_cl = parachute.clean_pressure_signal(t)
   3094         p_ns = parachute.noisy_pressure_signal(t)
   3095         file.write(f"{t:f}, {p_cl:.5f}, {p_ns:.5f}\n")

TypeError: 'list' object is not callable
```

## New behavior

The employed fix was to call the method that set these values not in the `all_info`, but after `Flight` computations. On top of that, the incorrect `list` calls were substituted by their `_function` counterparts.

Now the data is being exported correctly: 

```
0.000000, 85639.55296, 85652.28657
1.000000, 85228.90121, 85223.88463
...
```
## Breaking change

- [X] No

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -435,6 +435,7 @@ disable=raw-checker-failed,
         use-implicit-booleaness-not-comparison-to-zero,
         no-else-return,
         inconsistent-return-statements,
+        unspecified-encoding,
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ straightforward as possible.
 
 ### Fixed
 
+- BUG: Parachute Pressures not being Set before All Info. [#534](https://github.com/RocketPy-Team/RocketPy/pull/534)
 - BUG: Invalid Arguments on Two Dimensional Discretize. [#521](https://github.com/RocketPy-Team/RocketPy/pull/521)
 
 ## [v1.1.4] - 2023-12-07
--- a/rocketpy/plots/flight_plots.py
+++ b/rocketpy/plots/flight_plots.py
@@ -817,7 +817,6 @@ class _FlightPlots:
         if len(self.flight.parachute_events) > 0:
             for parachute in self.flight.rocket.parachutes:
                 print("\nParachute: ", parachute.name)
-                self.flight._calculate_pressure_signal()
                 parachute.noise_signal_function()
                 parachute.noisy_pressure_signal_function()
                 parachute.clean_pressure_signal_function()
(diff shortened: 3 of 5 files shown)


Gui-FernandesBR (approved):
Well done!

I think we should adopt the `Function.savetxt` to do the same job once we finish its PR.

Nevertheless, for this hotfix I think we are fine with the solution, thanks.

giovaniceotto:
Should we expect any significant increase in simulation execution speed since we are adding a new call inside `Flight.__init__`?

phmbressan:
> Should we expect any significant increase in simulation execution speed since we are adding a new call inside `Flight.__init__`?

That could be tested, of course, but I don't think so. The method basically makes two Function instances per parachute if I recall correctly.

Unless the simulation has unusualy too many points or a lot of parachutes, the time spend should be in the `ms` range.

Again, that could be easily tested later. This solution is not the best one, but I do think it involves the least amount of changes for a Hotfix.

I am open to other suggestions.

Gui-FernandesBR:
> > Should we expect any significant increase in simulation execution speed since we are adding a new call inside `Flight.__init__`?
> 
> That could be tested, of course, but I don't think so. The method basically makes two Function instances per parachute if I recall correctly.
> 
> Unless the simulation has unusualy too many points or a lot of parachutes, the time spend should be in the `ms` range.
> 
> Again, that could be easily tested later. This solution is not the best one, but I do think it involves the least amount of changes for a Hotfix.
> 
> I am open to other suggestions.

I reviewed it again and I don't think it's a problem. The new behavior is not prejudicial in terms of code performance.

Gui-FernandesBR:
In order to prevent future errors, we ideally should implement unit test for the methods we fix in hotfixes.
I did that in my last commit [a3e49a2](https://github.com/RocketPy-Team/RocketPy/pull/534/commits/a3e49a24b593c0548fa209cc28f2cd80aef2d9ab).

@phmbressan please review the commit and feel free to proceed and merge.

phmbressan:
> In order to prevent future errors, we ideally should implement unit test for the methods we fix in hotfixes.
> I did that in my last commit [a3e49a2](https://github.com/RocketPy-Team/RocketPy/pull/534/commits/a3e49a24b593c0548fa209cc28f2cd80aef2d9ab).
> 
> @phmbressan please review the commit and feel free to proceed and merge.

Since I opened the PR GitHub does not allow me to submit an approval.

Gui-FernandesBR (approved):

Gui-FernandesBR (approved):

MateusStano (approved):
Good! The export pressure function has broken a million times before. Its good that we finally have a test for it!

PR #537, opened 2024-01-22
MNT: Encapsulate quaternion conversions

Gui-FernandesBR:
## Pull request type

## Checklist

## Current behavior
We take quaternions (euler parameters) and calculate euler angles inside the flight class, but this code cannot be reused in other applications.

## New behavior
- The quaternions -> angles conversions were moved to the tools.py module, thus allowing for other classes to access these functions.
- The Flight class was adjusted to use the new functions.

## Breaking change

- [x] No (I'm not sure if we should create a new .py file just to put these functions or not.) 

## Additional information
This is going to be quite useful for the `rocketpy.simulation.FlightDataImporter` class.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -436,6 +436,7 @@ disable=raw-checker-failed,
         no-else-return,
         inconsistent-return-statements,
         unspecified-encoding,
+        no-member, # because we use funcify_method decorator
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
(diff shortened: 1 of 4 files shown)


phmbressan (commented):
The documentation helps, but I have a minor issue with the naming schema of the created functions.

You certainly have more experience on quaternions than I, so I would like to hear your opinion:

- Except the `precession` one, the names end up with the angle naming convention used for RocketPy (`phi`,  `theta`).

- I know this is a common angle convention, but I am not sure of it is universal. So I believe it would be better to name them after their "physical angle", such as `precession` and `nutation`.

Gui-FernandesBR:
> The documentation helps, but I have a minor issue with the naming schema of the created functions.
> 
> You certainly have more experience on quaternions than I, so I would like to hear your opinion:
> 
> * Except the `precession` one, the names end up with the angle naming convention used for RocketPy (`phi`,  `theta`).
> * I know this is a common angle convention, but I am not sure of it is universal. So I believe it would be better to name them after their "physical angle", such as `precession` and `nutation`.

You have a good point.
I solved that with the [cdb42ec](https://github.com/RocketPy-Team/RocketPy/pull/537/commits/cdb42ece907a26f6b941eb404118607396214f76).

Can you please re-review it?

MateusStano (approved):

PR #540, opened 2024-01-25
DOC: Installing imageio library on dispersion analysis notebook

Lucas-Prates:
## Pull request type

## Checklist

## Current behavior

The dispersion analysis notebook uses, on its last cell, the library imageio, but does not install it at any point, nor is it a dependency. Executing the whole notebook without having the library prompts a "ModuleNotFoundError" at the last cell.

## New behavior

Install imageio at the 'Install and Load Libraries' section.

## Breaking change

- [x] No

## Additional information

The installation is done exclusively for this notebook, so there are no changes to dependencies.

### DIFF
--- a/docs/notebooks/dispersion_analysis/dispersion_analysis.ipynb
+++ b/docs/notebooks/dispersion_analysis/dispersion_analysis.ipynb
@@ -66,7 +66,7 @@
    },
    "outputs": [],
    "source": [
-    "!pip install rocketpy"
+    "!pip install rocketpy imageio"
    ]
   },
   {


Gui-FernandesBR:
In order to test it on colab: https://colab.research.google.com/github/RocketPy-Team/rocketpy/blob/master/docs/notebooks/dispersion_analysis/dispersion_analysis.ipynb

This link clones from master branch, so we need to modify the pip install command before running all the cells.

Gui-FernandesBR:
The code is running as expected here.

In the future we might wanna change the implementation to fix this warning: 
```stdout
<ipython-input-26-957e0ff3cd5e>:6: DeprecationWarning: Starting with ImageIO v3 the behavior of this function will switch to that of iio.v3.imread. To keep the current behavior (and make this warning disappear) use `import imageio.v2 as imageio` or call `imageio.v2.imread` directly.
  img = imread("dispersion_analysis_inputs/Valetudo_basemap_final.jpg")
```

Gui-FernandesBR (approved):
Works on colab

PR #541, opened 2024-01-25
MNT: delete tools.cached_property (fix #531)

### LINKED ISSUE #531: MNT: delete tools.cached_property

Since we no longer support python 3.7, we don't need to keep our own implementation of the cached_property decorator.

- [x] Delete the cached_property definition inside the tools.py module.
- [x] Adjust all the imports throughout the rocketpy module

In order to avoid breaking changes, I think we could import the `cached_property` in the tools.py module so it doesn't break when we do `from rocketpy.tools import cached_property`

dyu056:
## Pull request type

- [ v] Code changes (bugfix, features)

## Checklist

- [ v] Tests for the changes have been added (if needed)
- [ v] Docs have been reviewed and added / updated
- [ v] Lint (`black rocketpy/ tests/`) has passed locally 
- [ v] All tests (`pytest --runslow`) have passed locally

## Current behavior

Issue 531, current behavior doesn't change too much

## New behavior

Issue 531, current behavior doesn't change too much

## Breaking change

- [ ] Yes
- [ v] No

## Additional information

Enter text here...

### DIFF
--- a/rocketpy/environment/environment_analysis.py
+++ b/rocketpy/environment/environment_analysis.py
@@ -8,6 +8,11 @@ import netCDF4
 import numpy as np
 import pytz
 
+try:
+    from functools import cached_property
+except ImportError:
+    from ..tools import cached_property
+
 from ..mathutils.function import Function
 from ..plots.environment_analysis_plots import _EnvironmentAnalysisPlots
 from ..prints.environment_analysis_prints import _EnvironmentAnalysisPrints
@@ -21,11 +26,6 @@ from ..tools import (
 )
 from ..units import convert_units
 from .environment import Environment
-
-try:
-    from functools import cached_property
-except ImportError:
-    from ..tools import cached_property
 
 # TODO: the average_wind_speed_profile_by_hour and similar methods could be more abstract than currently are
 
--- a/rocketpy/tools.py
+++ b/rocketpy/tools.py
@@ -6,35 +6,12 @@ from bisect import bisect_left
 import pytz
 from cftime import num2pydate
 from packaging import version as packaging_version
+from functools import cached_property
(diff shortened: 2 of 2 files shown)


rocketpy/environment/environment_analysis.py
```diff
@@ -8,6 +8,11 @@
 import numpy as np
 import pytz
 
+try:
+    from functools import cached_property
+except ImportError:
+    from ..tools import cached_property
```
Gui-FernandesBR, line 14: You can actually deleted the try-except statement and simply import using `from functools import cached_property`. This will work fine in any environment with python>3.8, so we are covered.

Also, there are a few more files where we do this kind of try-except.

image

Gui-FernandesBR:
This PR is trying to merge directly into the master branch. This needs to be changed to merge into develop instead.

I think the easiest solution is the cherry picking now: https://stackoverflow.com/questions/36055622/how-to-change-base-branch-in-git

It's important to follow the git flow correctly so we an easily navigate through our git tree in the future. Therefore, only hotfixes and releases should be merged directly to the master branch. New features needs to pass through the `develop` one.

Gui-FernandesBR:
This PR is trying to merge directly into the master branch. This needs to be changed to merge into develop instead.

I think the easiest solution is the cherry picking now: https://stackoverflow.com/questions/36055622/how-to-change-base-branch-in-git

It's important to follow the git flow correctly so we an easily navigate through our git tree in the future. Therefore, only hotfixes and releases should be merged directly to the master branch. New features needs to pass through the `develop` one.

Gui-FernandesBR:
Renamed the PR to make it easier for us.

PR #542, opened 2024-01-27
ENH: Function inputs from CSV file header.

phmbressan:
## Pull request type

## Checklist

## Current behavior

The Function class did not get the inputs and outputs name from the header of CSV file.

## New behavior

If the CSV file has headers, the input and output parameters that are left `None` will be set from the header information.

## Breaking change

- [X] No

### DIFF
--- a/rocketpy/mathutils/function.py
+++ b/rocketpy/mathutils/function.py
@@ -98,7 +98,6 @@ class Function:
         (II) Fields in CSV files may be enclosed in double quotes. If fields are
         not quoted, double quotes should not appear inside them.
         """
-        # Set input and output
         if inputs is None:
             inputs = ["Scalar"]
         if outputs is None:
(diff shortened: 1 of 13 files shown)


rocketpy/mathutils/function.py
```diff
@@ -93,12 +95,18 @@ def __init__(
 
         Notes
         -----
-        (I) CSV files can optionally contain a single header line. If present,
-        the header is ignored during processing.
+        (I) CSV files can optionally contain a single header line. If a header
+        naming each data column is present, it will be used for naming the `inputs`
+        and `outputs` left `None`. If the header is not formatted in this way it
+        will be ignored.
+
```
MateusStano, line 102: This is a little confusing. Suggestion:

```
CSV files may include an optional single header line. If this 
header line is present and contains names for each 
data column, those names will be used to label the inputs
and outputs unless specified otherwise. If there are header 
specified for only a few columns, the remaining unnamed
columns will have the string `None` assigned to it.
```

phmbressan: Thanks for the suggestion. I have reworded it a bit from your comment. Actually if the number of "header columns" does not match the number of inputs/outputs, the header is ignored.

Gui-FernandesBR (approved):
Good job. i  confirm it's working as expected.

Gui-FernandesBR (approved):
Good job. i  confirm it's working as expected.

PR #543, opened 2024-01-28
ENH: Dispersion Class Overhaul

MateusStano:
## Pull request type

## Description

Did a big refactor on the Dispersion Class, it is working perfectly now. Simulations can be ran perfectly.

I did not touch anything relating to plots, prints or tests. So tests will not pass.

Some things for the future: 
 - Something related to the saving of inputs and outputs are affecting simulation. I do not know exactly what it is
 - The whole `exportables` list needs to be reviewed
 
 Next PR will be renaming all the classes.


phmbressan (commented):
Great to see things coming to shape. 

You mentioned something not working on the simulation it the PR description, in the notebook there is an error in the creation of an `mc_rocket.create_object()`. Is that what you meant?

rocketpy/simulation/dispersion.py
```diff
@@ -22,111 +22,109 @@
...
+        Time when the simulation started.
+    _start_cpu_time : float
+        CPU time when the simulation started.
+    _input_file : str
+        String containing the filepath of the input file
+    _output_file : str
+        String containing the filepath of the output file
+    _error_file : str
+        String containing the filepath of the error file
     """
```
Gui-FernandesBR, line 92: What is the point of documenting weakly private attribute? If you're going to document it, then why don't you just go and make them public? Also, all the descriptions here are totally predictable, meaning that if you don't write the docstring I can still understand it.

rocketpy/simulation/dispersion.py
```diff
@@ -153,9 +156,260 @@ def __init__(
...
+        print("Starting monte carlo analysis", end="\r")
+
+        try:
+            while self.iteration_count < self.number_of_simulations:
+                self.__run_single_simulation(input_file, output_file)
+        except (TypeError, ValueError, KeyError, AttributeError) as error:
+            print(f"Error on iteration {self.iteration_count}: {error}")
+            error_file.write(f"{self._inputs_dict}\n")
+            self.__close_files(input_file, output_file, error_file)
+            raise error
```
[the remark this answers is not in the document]
Gui-FernandesBR: General exceptions are not so welcome as they tend to generalize they tend to generalize the behavior.
If an exception is raised without being in the list of expected exceptions, the code will still stop running, buuuut.... 
I'm curious to know what happens if you finish execution without calling the `self.__close_files(input_file, output_file, error_file)` line.

rocketpy/simulation/dispersion.py
```diff
@@ -153,9 +156,260 @@ def __init__(
...
+            output_file=output_file,
+        )
+
+        self.__reprint(
+            f"Current iteration: {self.iteration_count:06d} | "
+            f"Average Time per Iteration: {((process_time() - self.start_cpu_time) / self.iteration_count):2.6f} s | "
+            f"Estimated time left: {int((self.number_of_simulations - self.iteration_count) * ((process_time() - self.start_cpu_time) / self.iteration_count))} s",
+            end="\r",
+            flush=True,
+        )
```
Gui-FernandesBR, line 247: Getting two warnings of "line too long" here, please separate into more lines so it become more readable.

Here is a suggestion.

```suggestion
        average_time = (process_time() - self.start_cpu_time) / self.iteration_count
        estimated_time = int(
            (self.number_of_simulations - self.iteration_count) * average_time
        )

        self.__reprint(
            f"Current iteration: {self.iteration_count:06d} | "
            f"Average Time per Iteration: {average_time:.3f} s | "
            f"Estimated time left: {estimated_time} s",
            end="\r",
            flush=True,
        )
```

rocketpy/simulation/dispersion.py
```diff
@@ -153,9 +156,260 @@ def __init__(
...
+
+    def __finalize_simulation(self, input_file, output_file, error_file):
+        """Finalizes the simulation, closes the files and prints the results."""
+        final_string = (
+            f"Completed {self.iteration_count} iterations. Total CPU time: "
+            f"{process_time() - self.start_cpu_time:.1f} s. Total wall time: "
+            f"{time() - self.start_time:.1f} s\n"
+        )
+
+        self.__reprint(final_string + f"Saving results.", flush=True)
```
Gui-FernandesBR, line 263: You are using a f-string that doesn't have an interpolated value. Make it a simple string or add some content to interpolate. Here is a simple suggestion:

```suggestion
        self.__reprint(final_string + "Saving results.", flush=True)
```

rocketpy/simulation/dispersion.py
```diff
@@ -279,308 +529,16 @@ def set_results(self):
...
-        Returns
-        -------
-        None
-        """
+        parameter available in the results"""
         self.processed_results = {}
         for result in self.results.keys():
             mean = np.mean(self.results[result])
             stdev = np.std(self.results[result])
             self.processed_results[result] = (mean, stdev)
```
Gui-FernandesBR, line 540: This is more pythonic and cleaner:

```suggestion
        for result, values in self.results.items():
            mean = np.mean(values)
            stdev = np.std(values)
            self.processed_results[result] = (mean, stdev)
```

rocketpy/simulation/dispersion.py
```diff
@@ -695,8 +653,6 @@ def import_results(self, filename=None):
         self.import_inputs(filename=filepath)
         self.import_errors(filename=filepath)
 
-        return None
-
     def exportEllipsesToKML(
```
Gui-FernandesBR, line 656: Please consider renaming this method to follow the `snake_case` convention.  

`exportEllipsesToKML` --> `export_ellipses_to_kml`

rocketpy/simulation/dispersion.py
```diff
@@ -786,23 +743,14 @@ def exportEllipsesToKML(
...
-        Returns
-        -------
-        None
-        """
+        """Print information about the monte carlo simulation."""
         self.prints.all_results()
 
-        return None
-
     def allInfo(self):
```
Gui-FernandesBR, line 751: allInfo -> all_info

Gui-FernandesBR (changes requested):
Good refactor here, some of the changes are already contributing to a better understanding of the class in general.

The notebook is not running to the following problem:

```txt
AttributeError: 'Dispersion' object has no attribute 'print_results'
```

We should keep working on the Dispersion/Monte Carlo feature and improve it to release it by version v1.3.0, hopefully soon enough!

rocketpy/simulation/dispersion.py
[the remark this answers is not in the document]
phmbressan: I agree that a general Exception is not in general a good practice. I still need to read this code a bit better, but it would be basically to not stop the execution by a possibly "simple" error.

I am sorry if I am suggesting here something already implemented.

There are many solutions that we could employ to make the behavior of the dispersion in case of an exception convenient:

- Make a count incremented by each exception so that if too many simulations fail sequentially the program stops;

- Allow for saving partial dispersion results to be reused in a possible simulation restart.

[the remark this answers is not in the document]
Gui-FernandesBR: > because we need to detect if a simulation failed (basically by any reason).

I can agree with using the general exception then, given your comment.

rocketpy/simulation/dispersion.py
MateusStano: I guess there is no point in having them. Will remove it

Gui-FernandesBR (approved):
LGTM

PR #547, opened 2024-02-07
ENH: Plotting both power on and off drag curves in a single plot.

juliomachad0:
## Pull request type

## Checklist

## Current behavior

Before the changes, the drag curves were plotted separated.

## New behavior

Now, with the changes made, there is only a plot with both curves.

## Breaking change

- [ ] Yes
- [X] No

## Additional information

The methods power_on_drag and power_off_drag of the _RocketPlots class were removed.

### DIFF
--- a/rocketpy/plots/rocket_plots.py
+++ b/rocketpy/plots/rocket_plots.py
@@ -86,27 +86,31 @@ class _RocketPlots:
 
         return None
 
-    def power_on_drag(self):
-        """Plots power on drag of the rocket as a function of time.
+    def power_on_and_off_drag(self):
+        """Plots power off and on drag curve of the rocket as a function of time.
 
         Returns
         -------
         None
         """
+        x_power_drag_off = self.rocket.power_off_drag.x_array
+        y_power_drag_off = self.rocket.power_off_drag.y_array
+        x_power_drag_on = self.rocket.power_on_drag.x_array
+        y_power_drag_on = self.rocket.power_on_drag.y_array
 
-        self.rocket.power_on_drag()
+        fig, ax = plt.subplots()
+        ax.plot(x_power_drag_on, y_power_drag_on, label="Power on Drag")
+        ax.plot(
+            x_power_drag_off, y_power_drag_off, label="Power off Drag", linestyle="--"
+        )
 
-        return None
-
-    def power_off_drag(self):
-        """Plots power off drag of the rocket as a function of time.
-
-        Returns
-        -------
-        None
-        """
-
-        self.rocket.power_off_drag()
+        ax.set_title("Power on and off Drag")
+        ax.set_ylabel("Drag Coefficient")
+        ax.set_xlabel("Mach")
+        ax.axvspan(0.8, 1.2, alpha=0.3, color="gray", label="Transonic Region")
+        ax.legend(loc="best", shadow=True)
+        plt.grid(True)
+        plt.show()
 
         return None
 
@@ -524,8 +528,7 @@ class _RocketPlots:
         # Drag Plots
         print("Drag Plots")
         print("-" * 20)  # Separator for Drag Plots
-        self.power_on_drag()
-        self.power_off_drag()
+        self.power_on_and_off_drag()
 
         # Stability Plots
         print("\nStability Plots")


rocketpy/plots/rocket_plots.py
```diff
@@ -86,27 +86,31 @@ def stability_margin(self):
...
-        None
-        """
-
-        self.rocket.power_on_drag()
-
-        return None
-
-    def power_off_drag(self):
-        """Plots power off drag of the rocket as a function of time.
+    def power_on_and_off_drag(self):
```
Gui-FernandesBR, line 89: What do you think about `def drag_curves(self)` instead of `def power_on_and_power_off_drag`?

I think it's easier to do `my_rocket.plots.drag_curves()` than `my_rocket.plots.power_on_and_power_off_drag()`.
We keep the same meaning but in a more concise manner.

rocketpy/plots/rocket_plots.py
```diff
@@ -86,27 +86,31 @@ def stability_margin(self):
...
-        self.rocket.power_off_drag()
+        ax.set_title("Power on and off Drag")
+        ax.set_ylabel("Drag Coefficient")
+        ax.set_xlabel("Mach")
+        ax.axvspan(0.8, 1.2, alpha=0.3, color="gray", label="Transonic Region")
+        ax.legend(loc="best", shadow=True)
+        plt.grid(True)
+        plt.show()
 
         return None
```
Gui-FernandesBR, line 115: You don't need this return statement. I think it's better to just delete it.

```suggestion
```

rocketpy/plots/rocket_plots.py
```diff
@@ -86,27 +86,31 @@ def stability_margin(self):
...
-        Returns
-        -------
-        None
-        """
-
-        self.rocket.power_on_drag()
-
-        return None
-
-    def power_off_drag(self):
```
Gui-FernandesBR, line 101: With the new behavior, there will be a problem when somebody tries to do these:

```python
my_rocket.plots.power_on_drag()
my_rocket.plots.power_off_drag()
```
Therefore, the current implementation is leading to a small breaking change, although the _RocketPlots is a private class.

I would recommend to keep the old methods but include a warning like this one: https://github.com/RocketPy-Team/RocketPy/blob/435bd62ff2a218825760f73b8bf9fb11c06bbbed/rocketpy/environment/environment.py#L3200

rocketpy/plots/rocket_plots.py
```diff
@@ -86,27 +86,31 @@ def stability_margin(self):
...
+        y_power_drag_on = self.rocket.power_on_drag.y_array
+
+        fig, ax = plt.subplots()
+        ax.plot(x_power_drag_on, y_power_drag_on, label="Power on Drag")
+        ax.plot(
+            x_power_drag_off, y_power_drag_off, label="Power off Drag", linestyle="--"
+        )
 
-        self.rocket.power_off_drag()
+        ax.set_title("Power on and off Drag")
```
Gui-FernandesBR, line 107: A small suggestion regarding the title (feel free to refuse it): "Drag curves".

Gui-FernandesBR (commented):
Excellent PR, I went through the PR description and everything was transparent enough.

I'm adding some commentaries with concerns that I have, it would be good to discuss them if needed. Also, please let me know if there's anything you don't understand at a first glance.

Welcome to the Matplotlib School of Engineering!

rocketpy/plots/rocket_plots.py
juliomachad0: I think it is a great suggestion ! I tried to be as clear as possible, but I forgot to be a little concise haha

rocketpy/plots/rocket_plots.py
juliomachad0: Ok. I agree, I think it will be better for advanced users who use the methods.

rocketpy/plots/rocket_plots.py
juliomachad0: I agree, it will look better

Gui-FernandesBR (approved):
LGTM,

Thank you for your first contribution. 
It does solve the problem and make our lives better when seeing all those plots.

As a suggestion, mext PR I recommend you to also include a picture of the plot you make so we can have a preview before reviewing it.

phmbressan (approved):
Great! This is much better to compare the drag curves and adding the transonic region is the icing on the cake.

PR #554, opened 2024-02-12
ENH: adds `Function.remove_outliers` method

Gui-FernandesBR:
## Pull request type

## Checklist

## Current behavior
Different datasets present undesired values that we usually have to remove manually.

## New behavior
The Function class now can remove outliers from the source (if array like) using the IQR method.

## Breaking change
- [x] No

## Additional information
- This was saved in my machine for a while. It is simple but has been working without problems.
- It will not be included in the version v1.2.0, but in the 1.3.0 instead.

### DIFF
--- a/rocketpy/mathutils/function.py
+++ b/rocketpy/mathutils/function.py
@@ -1146,7 +1146,7 @@ class Function:
 
     def remove_outliers_iqr(self, threshold=1.5):
         """Remove outliers from the Function source using the interquartile
-        range method.
+        range method. The Function should have an array-like source.
 
         Parameters
         ----------
@@ -1164,8 +1164,10 @@ class Function:
         """
 
         if callable(self.source):
-            print("Cannot remove outliers if the source is a callable object.")
-            return self
+            raise TypeError(
+                "Cannot remove outliers if the source is a callable object."
+                + " The Function.source should be array-like."
+            )
 
         x = self.x_array
         y = self.y_array
--- a/rocketpy/mathutils/function.py
+++ b/rocketpy/mathutils/function.py
@@ -1144,41 +1144,7 @@ class Function:
             title=self.title,
         )
 
-    def remove_outliers(self, method="iqr", **kwargs):
-        """Remove outliers from the Function source using the specified method.
-
-        Parameters
-        ----------
-        method : string, optional
-            Method to be used to remove outliers. Options are 'iqr'.
-            Default is 'iqr'.
-        **kwargs : optional
-            Keyword arguments to be passed to specific methods according to the
-            selected method.
-            If the selected method is the 'iqr', then the following kwargs are
-            available:
-            - threshold : float
-                Threshold for the interquartile range method. Default is 1.5.
-
-        Returns
-        -------
-        Function
-            The new Function object without outliers.
-        """
-        if callable(self.source):
-            print("Cannot remove outliers if the source is a callable object.")
-            return self
-
-        if method.lower() == "iqr":
-            return self.__remove_outliers_iqr(**kwargs)
-        else:
-            print(
-                f"Method '{method}' not recognized. No outliers removed."
-                + "Please use one of the following supported methods: 'iqr'."
-            )
-            return self
-
-    def __remove_outliers_iqr(self, threshold=1.5):
+    def remove_outliers_iqr(self, threshold=1.5):
         """Remove outliers from the Function source using the interquartile
         range method.
 
@@ -1196,6 +1162,11 @@ class Function:
         ----------
         [1] https://en.wikipedia.org/wiki/Outlier#Tukey's_fences
         """
+
+        if callable(self.source):
+            print("Cannot remove outliers if the source is a callable object.")
+            return self
+
         x = self.x_array
         y = self.y_array
         y_q1 = np.percentile(y, 25)
--- a/rocketpy/mathutils/function.py
+++ b/rocketpy/mathutils/function.py
@@ -1144,6 +1144,78 @@ class Function:
             title=self.title,
         )
 
+    def remove_outliers(self, method="iqr", **kwargs):
+        """Remove outliers from the Function source using the specified method.
+
+        Parameters
+        ----------
+        method : string, optional
+            Method to be used to remove outliers. Options are 'iqr'.
+            Default is 'iqr'.
+        **kwargs : optional
+            Keyword arguments to be passed to specific methods according to the
+            selected method.
+            If the selected method is the 'iqr', then the following kwargs are
+            available:
+            - threshold : float
+                Threshold for the interquartile range method. Default is 1.5.
+
+        Returns
+        -------
+        Function
+            The new Function object without outliers.
+        """
+        if callable(self.source):
+            print("Cannot remove outliers if the source is a callable object.")
+            return self
+
+        if method.lower() == "iqr":
+            return self.__remove_outliers_iqr(**kwargs)
+        else:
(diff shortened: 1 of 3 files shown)


phmbressan (changes requested):
Interesting feature, useful for removing spurious points in data.

However, the way this method parameters were set (`kwargs` that change based on another parameter) seems to overcomplicate its usage in order to prevent breaking changes. For instance, I see the following disadvantages:

- I see that you explained everything in the documentation, but it has the potential of getting quite convoluted if we add more types with different `**kwargs` names.

- Code completition in most IDEs will not be available, since the parameters do not have a fixed name.

- It is quite hard use the method without reading its documentation.

- We can add, if needed, the `kwargs` behavior later without a breaking change, but we cannot remove or change it.

Do you have any references for this, or any particular reasoning I might have missed? I recall [solve_ivp](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html) doing something similar, but the method overall clearly has a common set of parameters and the optional ones are generally secondary.

Personally, I would either standardize the parameter names among the different types or make standalone methods.

rocketpy/mathutils/function.py
```diff
@@ -1144,6 +1144,78 @@ def low_pass_filter(self, alpha, file_path=None):
...
+            return self.__remove_outliers_iqr(**kwargs)
+        else:
+            print(
+                f"Method '{method}' not recognized. No outliers removed."
+                + "Please use one of the following supported methods: 'iqr'."
+            )
+            return self
+
+    def __remove_outliers_iqr(self, threshold=1.5):
+        """Remove outliers from the Function source using the interquartile
```
phmbressan, line 1182: Here is the only place that `iqr` is mentioned in full. I believe this should be part of the documentation of the main method, not only the private one.

phmbressan (commented):
** made a comment in another review because of a weird GitHub issue with pending comments.

rocketpy/mathutils/function.py
Gui-FernandesBR: You're 100% right.

Gui-FernandesBR:
> Interesting feature, useful for removing spurious points in data.
> 
> However, the way this method parameters were set (`kwargs` that change based on another parameter) seems to overcomplicate its usage in order to prevent breaking changes. For instance, I see the following disadvantages:
> 
> * I see that you explained everything in the documentation, but it has the potential of getting quite convoluted if we add more types with different `**kwargs` names.
> * Code completition in most IDEs will not be available, since the parameters do not have a fixed name.
> * It is quite hard use the method without reading its documentation.
> * We can add, if needed, the `kwargs` behavior later without a breaking change, but we cannot remove or change it.
> 
> Do you have any references for this, or any particular reasoning I might have missed? I recall [solve_ivp](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html) doing something similar, but in their case the method overall has a common set of parameters and the optional ones are generally secondary.
> 
> Personally, I would either standardize the parameter names among the different types or make them standalone methods.

I understand your points.

Just to check on this first:

- Are you okay with us having multiple methods to perform the same operation but with some different approaches?
- I am sorry but I cannot anticipate what will be the required parameters for eventual future new methods.
- References? Matplotlib does this quite a lot. Actually, they often do not care about describing the kwargs at all: https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.plot.html

phmbressan:
> I understand your points.
> 
> Just to check on this first:
> 
> - Are you okay with us having multiple methods to perform the same operation but with some different approaches?
> - I am sorry but I cannot anticipate what will be the required parameters for eventual future new methods.
> - References? Matplotlib does this quite a lot. Actually, they often do not care about describing the kwargs at all: https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.plot.html

Great questions:

- I am ok having multiple methods having smaller and less coupled chunks of code is generally considered a better practice;
- That is completely normal, sometimes we cannot predict future argument structure;
- The references thing was just to understand if you had any particular implementation in mind or whether I did not foresee a possible common practice in this situation.
    - About the link you mentioned, I don't think this is exactly the same as the issue we are discussing: matplotlib `**kwargs` do not change depending on the value of the others mandatory parameters, i.e., all of the `kwargs` are meaningful;
        - As I mentioned, the issue for me was not the mere existence of `kwargs`, but the possibility that they change based on the selected outlier detection string.

Gui-FernandesBR:
> > I understand your points.
> > Just to check on this first:
> > 
> > * Are you okay with us having multiple methods to perform the same operation but with some different approaches?
> > * I am sorry but I cannot anticipate what will be the required parameters for eventual future new methods.
> > * References? Matplotlib does this quite a lot. Actually, they often do not care about describing the kwargs at all: https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.plot.html
> 
> Great questions:
> 
> * I am ok having multiple methods as smaller and less coupled chunks of code is generally considered a better practice;
> * That is completely normal, sometimes we cannot predict future argument structure;
> * The references thing was just to understand if you had any particular implementation in mind or whether I did not foresee a possible common practice in this situation.
>   
>   * About the link you mentioned, I don't think this is exactly the same as the issue we are discussing: matplotlib `**kwargs` do not change depending on the value of the others mandatory parameters, i.e., all of the `kwargs` are meaningful;
>   * As I mentioned, the issue for me was not the mere existence of `kwargs`, but the possibility that they change based on the selected outlier detection string.

OK! 
I will be refactoring this implementation as soon as possible

rocketpy/mathutils/function.py
```diff
@@ -1144,6 +1144,49 @@ def low_pass_filter(self, alpha, file_path=None):
...
+            The Function with the outliers removed.
+
+        References
+        ----------
+        [1] https://en.wikipedia.org/wiki/Outlier#Tukey's_fences
+        """
+
+        if callable(self.source):
+            print("Cannot remove outliers if the source is a callable object.")
+            return self
```
phmbressan, line 1168: My suggestion here would be to raise an exception, so that this method does not misled the user by returning normally.

Check out `set_discrete_based_on_model`, the `Function` class normally throws an error when the operation cannot be executed instead of returning a default.

rocketpy/mathutils/function.py
```diff
@@ -1144,6 +1144,49 @@ def low_pass_filter(self, alpha, file_path=None):
             title=self.title,
         )
 
+    def remove_outliers_iqr(self, threshold=1.5):
+        """Remove outliers from the Function source using the interquartile
```
phmbressan, line 1148: It should be documented, either here or on a "notes" section (compare `set_discrete_based_on_model`) that the method only works if the `Function` is array based.

Gui-FernandesBR: Good point. Solved here: [5987bd8](https://github.com/RocketPy-Team/RocketPy/pull/554/commits/5987bd8fb267daf95a6ed345849c3b218dcfb838)

rocketpy/mathutils/function.py
Gui-FernandesBR: Solved here: [5987bd8](https://github.com/RocketPy-Team/RocketPy/pull/554/commits/5987bd8fb267daf95a6ed345849c3b218dcfb838)

Gui-FernandesBR:
Thanks for the solid review, @phmbressan .
All comments addressed again, could you take a look and approve?

phmbressan (approved):

Issue #557, opened 2024-02-14
Invalid Center of Pressure for Rockets with Variable Diameter


giovaniceotto:
**Describe the bug**

When setting up a rocket with multiple fuselage sections with different diameters, RocketPy may calculate the wrong value of center of pressure position.

**To Reproduce**

Here is an example:

```python
from rocketpy import Rocket, NoseCone, Tail, TrapezoidalFins

radius_1 = 0.1
radius_2 = 0.15

multiple_diameter_rocket = Rocket(
    radius=radius_2,
    mass=1,
    inertia=(1, 1, 1),
    power_off_drag=1,
    power_on_drag=1,
    center_of_mass_without_motor=1,
    coordinate_system_orientation="nose_to_tail",
)

nose = NoseCone(
    length=0.5,
    kind="vonKarman",
    base_radius=radius_1,
    rocket_radius=radius_1,
    bluffness=None,
    name="Nose Cone",
)

transition = Tail(
    length=0.1,
    top_radius=radius_1,
    bottom_radius=radius_2,
    rocket_radius=radius_2,
    name="Transition",
)

fins = TrapezoidalFins(
    n=4,
    span=0.2,
    root_chord=0.3,
    tip_chord=0.2,
    name="Fins",
    rocket_radius=radius_2,
)

boat_tail = Tail(
    length=0.1,
    top_radius=radius_2,
    bottom_radius=radius_2/2,
    rocket_radius=radius_2,
    name="Tail",
)

multiple_diameter_rocket.add_surfaces([nose, transition, fins, boat_tail], [0, 1.5, 2.5, 3])

print(multiple_diameter_rocket.static_margin(0))
```
**Expected behavior**

The expected static margin for the example above is: **2.71**

The returned value is: **1.75**

**Screenshots**

Current result: 

![image](https://github.com/RocketPy-Team/RocketPy/assets/572483/416f0651-e31c-4b4a-a2c0-680ed1f953e7)

Expected result:
![image](https://github.com/RocketPy-Team/RocketPy/assets/572483/fd68095c-677c-4d47-89d7-b0f14346b313)

**Additional context**

This bug was caught by Faraday Rocketry UPV when attempting to simulate a two-stage rocket.

**Workaround**
The current workaround is to use all `rocket_radius` values as the largest rocket radius. This causes the following side-effects:

- `Rocket.draw` will return a drawing where the fuselage does not meet the nose cone or transitions appropriately.
[...]

PR #558, opened 2024-02-14
BUG: Add reference area factor correction to aero surfaces (solves #557)

giovaniceotto:
## Pull request type

## Checklist

## Current behavior

See #557.

## New behavior

Issue #557 is fixed.

## Breaking change

- [x] No

## TODO:
- [x] Add tests to prevent similar issues from going undetected again.
- [x] Update CHANGELOG.md.
- [ ] Bump up version to 1.2.1 and release hotfix. (Gui: different PR maybe?)

### DIFF
--- a/rocketpy/prints/rocket_prints.py
+++ b/rocketpy/prints/rocket_prints.py
@@ -114,9 +114,13 @@ class _RocketPrints:
         print("\nAerodynamics Lift Coefficient Derivatives\n")
         for surface, position in self.rocket.aerodynamic_surfaces:
             name = surface.name
+            # ref_factor corrects lift for different reference areas
+            ref_factor = (surface.rocket_radius / self.radius) ** 2
             print(
                 name
-                + " Lift Coefficient Derivative: {:.3f}".format(surface.clalpha(0))
+                + " Lift Coefficient Derivative: {:.3f}".format(
+                    ref_factor * surface.clalpha(0)
+                )
                 + "/rad"
             )
 
@@ -134,6 +138,9 @@ class _RocketPrints:
         print("\nStability\n")
         print(
             f"Center of Mass position (time=0): {self.rocket.center_of_mass(0):.3f} m"
+        )
+        print(
+            f"Center of Pressure position (time=0): {self.rocket.cp_position(0):.3f} m"
         )
         print(
             "Initial Static Margin (mach=0, time=0): "
--- a/rocketpy/prints/rocket_prints.py
+++ b/rocketpy/prints/rocket_prints.py
@@ -115,7 +115,7 @@ class _RocketPrints:
         for surface, position in self.rocket.aerodynamic_surfaces:
             name = surface.name
             # ref_factor corrects lift for different reference areas
-            ref_factor = (surface.rocket_radius / self.radius) ** 2
+            ref_factor = (surface.rocket_radius / self.rocket.radius) ** 2
             print(
                 name
                 + " Lift Coefficient Derivative: {:.3f}".format(
--- a/rocketpy/rocket/aero_surface.py
+++ b/rocketpy/rocket/aero_surface.py
@@ -349,12 +349,20 @@ class NoseCone(AeroSurface):
         # If base radius is not given, the ratio between base radius and
         # rocket radius is assumed as 1, meaning that the nose cone has the
         # same radius as the rocket
-        if self.base_radius is None or self.rocket_radius is None:
+        if self.base_radius is None and self.rocket_radius is not None:
             self.radius_ratio = 1
+            self.base_radius = self.rocket_radius
+        elif self.base_radius is not None and self.rocket_radius is None:
+            self.radius_ratio = 1
+            self.rocket_radius = self.base_radius
         # If base radius is given, the ratio between base radius and rocket
         # radius is calculated
-        else:
+        elif self.base_radius is not None or self.rocket_radius is not None:
             self.radius_ratio = self.base_radius / self.rocket_radius
+        else:
+            raise ValueError(
+                "Either base radius or rocket radius must be given to calculate the nose cone radius ratio."
+            )
 
         self.fineness_ratio = self.length / (2 * self.base_radius)
         return None
--- a/rocketpy/rocket/aero_surface.py
+++ b/rocketpy/rocket/aero_surface.py
@@ -357,7 +357,7 @@ class NoseCone(AeroSurface):
             self.rocket_radius = self.base_radius
         # If base radius is given, the ratio between base radius and rocket
         # radius is calculated
-        elif self.base_radius is not None or self.rocket_radius is not None:
+        elif self.base_radius is not None and self.rocket_radius is not None:
             self.radius_ratio = self.base_radius / self.rocket_radius
         else:
             raise ValueError(
--- a/rocketpy/rocket/rocket.py
+++ b/rocketpy/rocket/rocket.py
@@ -901,7 +901,8 @@ class Rocket:
         name : string
             Nose cone name. Default is "Nose Cone".
         base_radius : int, float, optional
-            Nose cone base radius in meters. If not given, use rocket radius.
+            Nose cone base radius in meters. If not given, the rocket radius
+            will be used.
 
         See Also
         --------
@@ -993,7 +994,7 @@ class Rocket:
         radius : int, float, optional
             Reference fuselage radius where the fins are located. This is used
             to calculate lift coefficient and to draw the rocket. If None,
-            which is default, use rocket radius.
+            which is default, the rocket radius will be used.
         airfoil : tuple, optional
             Default is null, in which case fins will be treated as flat plates.
             Otherwise, if tuple, fins will be considered as airfoils. The
@@ -1075,7 +1076,7 @@ class Rocket:
         radius : int, float, optional
             Reference fuselage radius where the fins are located. This is used
             to calculate lift coefficient and to draw the rocket. If None,
-            which is default, use rocket radius.
+            which is default, the rocket radius will be used.
         airfoil : tuple, optional
             Default is null, in which case fins will be treated as flat plates.
             Otherwise, if tuple, fins will be considered as airfoils. The
--- a/rocketpy/rocket/rocket.py
+++ b/rocketpy/rocket/rocket.py
@@ -517,9 +517,13 @@ class Rocket:
         # Calculate total lift coefficient derivative and center of pressure
         if len(self.aerodynamic_surfaces) > 0:
             for aero_surface, position in self.aerodynamic_surfaces:
-                self.total_lift_coeff_der += aero_surface.clalpha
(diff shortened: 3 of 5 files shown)


rocketpy/prints/rocket_prints.py
```diff
@@ -114,9 +114,13 @@ def rocket_aerodynamics_quantities(self):
         print("\nAerodynamics Lift Coefficient Derivatives\n")
         for surface, position in self.rocket.aerodynamic_surfaces:
             name = surface.name
+            # ref_factor corrects lift for different reference areas
+            ref_factor = (surface.rocket_radius / self.rocket.radius) ** 2
```
Gui-FernandesBR, line 118: Defining this property (`ref_factor`) inside the prints module is dangerous.
I recommend defining this directly inside the Rocket class and using the already calculated value here instead of making this multiplication.

There should be no "business logic" within the plots or prints modules.

What do you think?

rocketpy/rocket/rocket.py
```diff
@@ -871,7 +875,9 @@ def add_tail(
         self.add_surfaces(tail, position)
         return tail
 
-    def add_nose(self, length, kind, position, bluffness=0, name="Nose Cone"):
+    def add_nose(
+        self, length, kind, position, bluffness=0, base_radius=None, name="Nose Cone"
+    ):
```
Gui-FernandesBR, line 880: I think this is a breaking change. Although a small one, it is a breaking change and perhaps we should at least mention it. 

I'm sorry that I couldn't execute the steps below before commenting, but here are my thoughts:

An application that uses the following code in rocketpy v1.2.0 and above will stop working:
```python
my_rocket.add_nose(1, "von_karman", 0, 0, "Nose Cone")
```
As you can see, the "Nose Cone" will now be interpreted as the 'base_radius' argument instead of 'name'.

If this is really a breaking change (I'm not sure because you used the `or` operand below and I couldn't run it yet), I see two possible solutions:
- Just ignore the breaking change and assume not a lot of users will suffer from it (which is usually not ok)
- Just move the `base_radius` argument to the last position so it comes after the `name` argument

rocketpy/rocket/aero_surface.py
```diff
@@ -349,12 +349,20 @@ def evaluate_geometrical_parameters(self):
...
+        if self.base_radius is None and self.rocket_radius is not None:
             self.radius_ratio = 1
+            self.base_radius = self.rocket_radius
+        elif self.base_radius is not None and self.rocket_radius is None:
+            self.radius_ratio = 1
+            self.rocket_radius = self.base_radius
         # If base radius is given, the ratio between base radius and rocket
         # radius is calculated
-        else:
+        elif self.base_radius is not None or self.rocket_radius is not None:
```
Gui-FernandesBR, line 360: You can only make the division if both `self.base_radius` **and** `self.rocket_radius` are not None.

```suggestion
        elif self.base_radius is not None and self.rocket_radius is not None:
```

Gui-FernandesBR (commented):
Thank you so much for a fast resolution for this bug. Here comes my first review.

I will try to help with the remaining TODOs, let's coordinate it together.

I would appreciate hearing from you on the comments that I've added below.

rocketpy/rocket/rocket.py
giovaniceotto: I agree with your statement. While the second option of not causing a break change is more appealing, it breaks pattern with similar methods that leave the name as the last argument.

Having said that, I am ok with this pattern break if it means avoiding a breaking change in a hotfix.

rocketpy/prints/rocket_prints.py
giovaniceotto: Makes sense! I'll look into a reasonable way to implement this.

Currently, what I have in mind is that each aero surface has three main relevant information when it comes to the Rocket class: itself, it's position along the rocket, and it's reference area factor. Therefore, I will try to implement this alongside the position info.

giovaniceotto: Tooke a quick look and couldn't find a way to integrate it into the `Components` class without making things look disorganized, since it is also used for other componentes which are not aerodynamic.

Two options:
 1. Find another way to implement this.
 2. Create an `AeroComponents` class the inherets from the `Components` class and adds the reference area factor attribute.

What do you think @Gui-FernandesBR?

rocketpy/rocket/rocket.py
giovaniceotto: Fixed here: [c63cb14](https://github.com/RocketPy-Team/RocketPy/pull/558/commits/c63cb148dbf1e66b162af2a23cabd2896d0bc0e6)

rocketpy/prints/rocket_prints.py
Gui-FernandesBR: Creating a new class just for this seems a bit overkill to me, at least at a first glance.

I would say that perhaps we could save the ref_factor directly in the AeroSurface object itself. Can't we?

Also, would it be a problem if we had the ref_factor in the component class but leave it equals to 1 by default?

giovaniceotto: > Creating a new class just for this seems a bit overkill to me, at least at a first glance.

I agree.

> I would say that perhaps we could save the ref_factor directly in the AeroSurface object itself. Can't we?

We can, but I strongly advise aignst it. Stand focused a lot on making each aero surface as independent from the Rocket as possible. This is important in the architecture that these components can be moved from rocket to rocket without changing the component itself.

Now, the "local" rocket radius was an information that we needed to have in orther to calculate things like interference factor. Some might argue that the local radius is indeed part of the component structure, so it makes sense to have that info.

Having said that, addind a reference to the "maximum" rocket radius to each component crosses the line in my point of view. Each component would end up having an excessive relationship to the rocket.

> Also, would it be a problem if we had the ref_factor in the component class but leave it equals to 1 by default?

This seems like a suboptimal solution for me. In some cases, this attribute will make no sense.

To be honest, a simple division in the print class doesn't seem like such a bad compromise.

Gui-FernandesBR: > This seems like a suboptimal solution for me. In some cases, this attribute will make no sense.

Can you give me some examples?

> To be honest, a simple division in the print class doesn't seem like such a bad compromise.

Given the discussion, I agree with you.

giovaniceotto: > Can you give me some examples?

For the rail button component.

Gui-FernandesBR: Ok, I see it.

After thinking, thinking, and thinking during these past days.
I know think that the AeroComponent class might be a good idea.
Only problem is that the rail buttons should also be treated as a "aero surface", but anyway.

We should have Component (motors, buttons) and AeroComponents (tails, noses, etc.) being added to the rocket.

The AeroComponent could be related to the Component class via inheritance.

I think we could do that in a separate PR, since the current solution is already working and does not make anything wrong. Also, refactoring this should be easy in the future.

phmbressan: I read all the thread and I agree that keeping it as it is seems the best for this hotfix.

In my opinion, the rocket class should be reponsible for storing and mapping each specific surface to their connection parameters (e.g. position, local radius etc).

When dealing with `AeroSurface` we should rely more on its abtraction (base class) than on particular implementations. For instance, about the particular issue raised about `RailButtons`, I believe this is a case that does not matter too much whether the "local radius" has physical meaning: the code should treat it internally as any surface but this parameter would not be exposed to the user.

As a side note, the bigger difference I see when trying to think about an abstraction for surfaces is that some of them are attached to the tube (such as fins) and others goes transversely (like nosecones), which make them quite different geometricaly.

rocketpy/rocket/rocket.py
```diff
@@ -517,9 +517,13 @@ def evaluate_center_of_pressure(self):
...
-                self.total_lift_coeff_der += aero_surface.clalpha
-                self.cp_position += aero_surface.clalpha * (
-                    position - self._csys * aero_surface.cpz
+                # ref_factor corrects lift for different reference areas
+                ref_factor = (aero_surface.rocket_radius / self.radius) ** 2
+                self.total_lift_coeff_der += ref_factor * aero_surface.clalpha
+                self.cp_position += (
+                    ref_factor
+                    * aero_surface.clalpha
+                    * (position - self._csys * aero_surface.cpz)
```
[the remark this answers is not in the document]
Gui-FernandesBR: [duplicated]

rocketpy/rocket/rocket.py
```diff
@@ -894,6 +900,8 @@ def add_nose(self, length, kind, position, bluffness=0, name="Nose Cone"):
             the radius of the base of the ogive.
         name : string
             Nose cone name. Default is "Nose Cone".
+        base_radius : int, float, optional
+            Nose cone base radius in meters. If not given, use rocket radius.
```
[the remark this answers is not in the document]
Gui-FernandesBR: Solved here: [955f61d](https://github.com/RocketPy-Team/RocketPy/pull/558/commits/955f61da83d90bda0d5d8443ff70debc99cd0f27)

Gui-FernandesBR (approved):
Let's open an issue to deal with the AeroComponent and ref_factor thread.

tests/test_rocket.py
```diff
@@ -205,3 +207,76 @@ def test_air_brakes_clamp_off(mock_show, calisto_air_brakes_clamp_off):
     assert air_brakes_clamp_off.deployment_level == 0
 
     assert air_brakes_clamp_off.all_info() == None
+
+
+def test_add_surfaces_different_noses(calisto):
+    """Test the add_surfaces method with different nose cone configurations.
+    More specifically, this will checks the static margin of the rocket with
```
lucasfourier, line 214: this will 'check'

lucasfourier (dismissed):
Great work guys. The only thing left to do I guess would be to refactor the newly created tests. Gonna open an issue for that. I think there is room for creating fixtures, parametrizing, etc. However, for a hotfix this is more than enough.

tests/test_rocket.py
Gui-FernandesBR: Solved: [643ea6e](https://github.com/RocketPy-Team/RocketPy/pull/558/commits/643ea6eeb235311297972b3a1599700cc6709e4f)

Gui-FernandesBR (approved):
LGTM,

All credits go to @giovaniceotto (aka the wizard) for this fix.

lucasfourier (approved):

phmbressan:
Thanks for quickly solving the review comments @Gui-FernandesBR . Great fix by @giovaniceotto .

phmbressan (approved):

rocketpy/prints/rocket_prints.py
Gui-FernandesBR: I added a new issue to the repo so we can follow up this question later: https://github.com/RocketPy-Team/RocketPy/issues/561

Marking this as solved for now.

Issue #560, opened 2024-02-21
Incorrect/incomplete docstring in environment.py


chasgior214:
**Describe the bug**

Incorrect/incomplete documentation of the Environment class's \_\_init__ method.

**To Reproduce**

Open environment.py and go to the \_\_init__ method (also [here](https://docs.rocketpy.org/en/latest/_modules/rocketpy/environment/environment.html#Environment.__init__))

**Expected behavior**

Correct (no reference to launch rail length) and complete documentation.

Gui-FernandesBR:
Hello @chasgior214 , thank you for opening this issue.

I'm a bit lost tho, could you point me exactly what is wrong with the current documentation?
I don't see any reference to launch rail length and the documentation seems complete enough to me.

![image](https://github.com/RocketPy-Team/RocketPy/assets/63590233/755b1b7c-1bad-40df-b5ad-a5b23911cebd)

chasgior214:
Hi @Gui-FernandesBR 

The mention of launch rail length is at the top of the screenshot. There's also two incomplete bits. The first is the rest of that paragraph that has an unfinished sentence ("is loaded until another"). The second is an unfinished sentence in the description of the gravity parameter.

Gui-FernandesBR:
> Hi @Gui-FernandesBR
> 
> The mention of launch rail length is at the top of the screenshot. There's also two incomplete bits. The first is the rest of that paragraph that has an unfinished sentence ("is loaded until another"). The second is an unfinished sentence in the description of the gravity parameter.

Got it. Thanks for specifying it. We will work on that soon enough

Gui-FernandesBR:
@phmbressan assigned to you, boss.

PR #565, opened 2024-02-29
DOC: Improvements of Environment docstring phrasing.

phmbressan:
## Pull request type

## Checklist

## Current behavior

As described in #560, the docstring for the `Environment.__init__` were incomplete and outdated.

## New behavior

The following changes were employed:
- The legacy reference to `rail_length` was removed;
- The incomplete phrasing of `__init__` was improved;
- Minor additions for clarity and references behind Somigliana's formula.

## Breaking change

- [ ] Yes
- [X] No


Gui-FernandesBR:
@phmbressan what happened with the git history in this? oh lord, this looks confusing. 
Maybe you started from master first and then changed (rebased) to develop?

phmbressan:
@Gui-FernandesBR I believe with the `develop` update the commit history seems to have normalized.

rocketpy/environment/environment.py
```diff
@@ -468,8 +474,8 @@ def set_gravity_model(self, gravity):
         ----------
         gravity : None or Function source
```
Gui-FernandesBR, line 475: Something does not make sense here.

In line 290 you say:
```python
gravity : int, float, callable, string, array, optional
```

But here you mention that only None or Function source are allowed. 
Only one can be truth.

Can you provide more examples in the documentation please? A new user you never understand all the different ways that can be done to define the gravity.

rocketpy/environment/environment.py
```diff
@@ -276,15 +276,21 @@ def __init__(
         timezone="UTC",
         max_expected_height=80000.0,
     ):
```
Gui-FernandesBR, line 278: I can't comment on some lines because they were not edited (thanks GitHub).

Here are some points I would like to mention:
* in the date description, does it really mean that the hour is ALWAYS in UTC? What if I use a different timezone? Is my inputted time gonna be converted or not?
* Please mention the default values for latitude and longitude
* Sometimes it says "timezone", and sometimes "time zone" . I searched but could not find the "correct way". But we should keep a standard imo.
* in the datum description, it is not saying that the argument is optional. It should.

rocketpy/environment/environment.py
```diff
@@ -276,15 +276,21 @@ def __init__(
...
-        """Initialize Environment class, saving launch rail length,
-        launch date, location coordinates and elevation. Note that
-        by default the standard atmosphere is loaded until another
+        """Initialize Environment class, saving parameters of the launch location,
+        such as launch date, coordinates and elevation. This class also computes
+        relevant quantities for the Flight simulation, such as air pressure, density
+        and gravitational acceleration.
```
Gui-FernandesBR, line 282: Trying to be more precise with the words here, let me know what you think:

```suggestion
        """Initialize the Environment class, capturing essential parameters of the
        launch site, including the launch date, geographical coordinates, and
        elevation. This class is designed to calculate crucial variables for the
        Flight simulation, such as atmospheric air pressure, density, and
        gravitational acceleration.
```

Gui-FernandesBR (changes requested):
This seems to be a simple PR with small changes, but don't underestimate it.
The Environment class is the first contact that a new user has with RocketPy, so I want to make sure that these docstrings are clear enough.

Thank you so much for your contribution already, and I hope you may consider some of the comments below.

Gui-FernandesBR:
Btw, when solving a bug please link the open PR in the development section (right side of you github page) so the Pr and issue can be connected.

In this case, the #560

rocketpy/environment/environment.py
phmbressan: Fixed in [a2aa714](https://github.com/RocketPy-Team/RocketPy/pull/565/commits/a2aa71467b3be87c8124343cd49768c1ac4d6a9d).

phmbressan:
Thank you for your review @Gui-FernandesBR. I am still going to solve your comment about timezones, but I would like to ask your opinion on one matter:

- I have implemented a long time ago the methods `set_gravity_model` and `set_earth_geometry`. They, paradoxially, do not set anything in the `Enviroment` classes (they are using to make computations to set in the constructor).
- This behavior is extremely unintuitive given its naming and behavior being quite different when comparing them to other setters in the same class.

Do you think is worth changing this behavior to something more standard in a future PR or not risking a small breaking change and only documenting this peculiar behavior.

Gui-FernandesBR:
> Thank you for your review @Gui-FernandesBR. I am still going to solve your comment about timezones, but I would like to ask your opinion on one matter:
> 
> - I have implemented a long time ago the methods `set_gravity_model` and `set_earth_geometry`. They, paradoxically, do not set anything in the `Enviroment` classes (they are using to make computations to set in the constructor).
> - This behavior is extremely unintuitive given its naming and behavior being quite different when comparing them to other setters in the same class.
> 
> This behavior confused me quite some time while writing the docstring and I would almost consider it a code issue. Do you think is worth changing this behavior to something more standard in a future PR or would it be considered a small breaking change and we should only document this peculiar behavior?

I'm ok with that kind of refactor.

Please keep in mind that there are several setter methods in the Environment class, they should follow a common standard.

However, I don't think this is exactly a priority right now, so treat it carefully.

rocketpy/environment/environment.py
phmbressan: Thanks for the suggestion, applied in [6b84c03](https://github.com/RocketPy-Team/RocketPy/pull/565/commits/6b84c03bdaafd96c140067cfbf0a00fe05556d1e).

rocketpy/environment/environment.py
phmbressan: Great catches, those should be fixed by commit [27dd73d](https://github.com/RocketPy-Team/RocketPy/pull/565/commits/27dd73d24fa82f3cbf066d6edb5aa3063b102099).

rocketpy/environment/environment.py
```diff
@@ -276,49 +275,70 @@ def __init__(
...
         latitude : float, optional
             Latitude in degrees (ranging from -90 to 90) of rocket
             launch location. Must be given if a Forecast, Reanalysis
             or Ensemble will be used as an atmospheric model or if
-            Open-Elevation will be used to compute elevation.
+            Open-Elevation will be used to compute elevation. Positive
+            values correspond to the North. Default value is 0, which
+            corresponds to the equator.
         longitude : float, optional
             Longitude in degrees (ranging from -180 to 360) of rocket
```
Gui-FernandesBR, line 317:
```suggestion
            Longitude in degrees (ranging from -180 to 180) of rocket
```

rocketpy/environment/environment.py
```diff
@@ -396,15 +416,57 @@ def set_date(self, date, timezone="UTC"):
...
+        - If the ``date`` is given as a ``Datetime`` object without a time zone,
+          it will be assumed to be in the same time zone as specified by the
+          ``timezone`` parameter. However, if the ``Datetime`` object has a time
+          zone specified in its ``tzinfo`` attribute, the ``timezone``
+          parameter will be ignored.
+
+        Examples
+        --------
+
+        Let's set the launch date as an array:
```
Gui-FernandesBR, line 450: Be careful, what you call an array is actually called "list" in python

rocketpy/environment/environment.py
```diff
@@ -396,15 +416,57 @@ def set_date(self, date, timezone="UTC"):
...
+
+        >>> date = [2000, 1, 1, 13] # January 1st, 2000 at 13:00 UTC+1
+        >>> env = Environment()
+        >>> env.set_date(date, timezone="Europe/Rome")
+        >>> print(env.datetime_date) # Get UTC time
+        2000-01-01 12:00:00+00:00
+        >>> print(env.local_date)
+        2000-01-01 13:00:00+01:00
+
+        Now let's set the launch date as a ``Datetime`` object:
```
Gui-FernandesBR, line 460: Do you agree with me that there's no `Datetime` class? Instead, we use `datetime` objects.
I noticed this mistake a couple of times in the documentation, I'm mostly ignoring it bc time is a constraint, but please take this into consideration. 

```suggestion
        Now let's set the launch date as a ``datetime`` object:
```

rocketpy/environment/environment.py
```diff
@@ -458,23 +520,66 @@ def set_location(self, latitude, longitude):
...
 
         Returns
         -------
         Function
             Function object representing the gravity model.
+
+        Notes
+        -----
+        This method **does not** set the gravity acceleration, it only returns
+        a :class:`Function` object representing the gravity model.
```
Gui-FernandesBR, line 562: Thank you for adding this note!

Gui-FernandesBR (commented):
Thank you for your contribution, @phmbressan .

I left some other comments, really small ones. Can you check that out please?

Additionally, could you please update te CHANGELOG.md file with your PR?
That would be important for us to keep track on the main files changes.

rocketpy/environment/environment.py
phmbressan: You are correct, I have confused syntax between different programming languages.

I will fix it.

phmbressan: Fixed in [9c1a461](https://github.com/RocketPy-Team/RocketPy/pull/565/commits/9c1a4619ee2beb273d81f63fa78343dfa84ca790).

rocketpy/environment/environment.py
phmbressan: Good catch, changed to `list` in [9c1a461](https://github.com/RocketPy-Team/RocketPy/pull/565/commits/9c1a4619ee2beb273d81f63fa78343dfa84ca790).

Gui-FernandesBR (approved):
LGTM

PR #566, opened 2024-02-29
ENH: Eliminating multiple plots for inertia components

dipesh2212:
## Pull request type

## Current behavior

There are multiple inertial tensor plots that are being plotted, it is kinda annoying to deal with those

## New behavior

I've tried to integrate all the plots in one single plot which can compare the inertia tensor in a convenient way

## Breaking change

- [ ] No


Gui-FernandesBR:
I'm still investigating the reason, but I got this error when running `my_solid_motor_example.all_info()`:

image

rocketpy/plots/motor_plots.py
```diff
@@ -135,113 +135,41 @@ def exhaust_velocity(self, lower_limit=None, upper_limit=None):
...
-        Return
-        ------
-        None
-        """
-        self.motor.I_13.plot(lower=lower_limit, upper=upper_limit)
-
-    def I_23(self, lower_limit=None, upper_limit=None):
-        """Plots I_23 of the motor as a function of time.
+        for label in labels:
+            getattr(self.motor, label).plot(
```
Gui-FernandesBR, line 161: I don't think this will going to work as expected.

- The `Function.plot` method does not have a `label` argument. I think you should use the  `Function.plot_1d` instead
- This for loop is simply plotting all the curves one by one in different plots, right? Instead, what we expect from this PR is something very similar to this: https://github.com/RocketPy-Team/RocketPy/pull/547/files#

rocketpy/plots/motor_plots.py
```diff
@@ -135,113 +135,41 @@ def exhaust_velocity(self, lower_limit=None, upper_limit=None):
...
+        ax.set_title("Inertia Tensors Over Time")
+        ax.grid(True)
+        plt.tight_layout()
 
-        Return
-        ------
-        None
-        """
-        self.motor.I_23.plot(lower=lower_limit, upper=upper_limit)
+        return fig
```
Gui-FernandesBR, line 172: No need to return the matplotlib fig. Just use the plt.show() instead.

Gui-FernandesBR (changes requested):
The submitted code in this PR doesn't seem to be working.
@dipesh2212 we should work on 
Here are some comments to guide you:
- Try to read @juliomachad0 's solution, you might benefit from his approach in this case.
- The lines like `self.I_11(*self.motor.burn_time)` in the motor_plots.py file will no longer work, please update the `_MotorPlots.all` method.
- Always run the tests before submitting your code, this prevents you from submitting a code that is not working.
- I think you could easily use the `Function.compare_plots` to easily create a single figure with all the inertia Functions and then display them.

Also, some other "small but important" things that I noticed:
- Rename your PR to start with `ENH: `, thus showing clearly the intention of your PR.
- Please run black before submitting your code. (see the Makefile in the root directory for more details)
- If possible, please also provides screenshot of the plots you have modified, this allows us to understand what is the expected behavior without needing to run the code on our own.

Gui-FernandesBR:
A quick update in the function:

image


I'm getting this result:
image

Gui-FernandesBR:
Here is the final result: 

image

Gui-FernandesBR (approved):
This solves the issue!

PR #568, opened 2024-03-04
ENH: Initial solution from flights gets previous results

Gui-FernandesBR:
## Pull request type

## Checklist

- [ ] Tests for the changes have been added (not needed)

## Current behavior
When you use a Flight object as the initial solution, only the final state of the flight is captured.
See orange plot:
image

## New behavior
Now the whole flight solution array is used in the new flight.
See orange plot now:

image

## Breaking change

- [x] No (all the current script made on top of rocketpy will still run and work, just some plots might change (for better) when using deployables of multi-stage.

## Additional information
Another enhancement allowed by the euroc works that we have been doing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ Attention: The newest changes should be on top -->
 ### Changed
 
 - 
+- ENH: Initial solution from flights gets previous results [#568](https://github.com/RocketPy-Team/RocketPy/pull/568)
 - DOC: New Environment class docs pages [#644](https://github.com/RocketPy-Team/RocketPy/pull/644)
 
 ### Fixed
--- a/rocketpy/simulation/flight.py
+++ b/rocketpy/simulation/flight.py
@@ -1101,6 +1101,7 @@ class Flight:  # pylint: disable=too-many-public-methods
         elif isinstance(self.initial_solution, Flight):
             # Initialize time and state variables based on last solution of
             # previous flight
+            self.solution = self.initial_solution.solution[:-1]
             self.initial_solution = self.initial_solution.solution[-1]
             # Set unused monitors
             self.out_of_rail_state = self.initial_solution[1:]


phmbressan (approved):
Good improvements, plots are much more intuitive to interpret now.

As a future enhancement, I believe it would be interesting to divide this plot different `Flights` with colors and labeling.

Gui-FernandesBR:
> Good improvements, plots are much more intuitive to interpret now.
> 
> As a future enhancement, I believe it would be interesting to divide this plot different `Flights` with colors and labeling. I might open an issue about that later.

I don't think that is possible.
The solution array from the first flight is passed as the initial solution array of the second. After the simulation is executed, there's no easy way to distinguish the first phase from the second.

Gui-FernandesBR:
This is what happens when running the deployable_payload_example.ipynb example:

image

Maybe your idea of creating different colors for the flight might help a little.
Moreover, I think a better idea is to have some kind of "events" attribute and save the initial time with it.
In other words, be able too identify the events that happened in a flight:
FlightEvents(ignition=0, out_of_rail=x, burn_out=y, apogee=z, deploy_chute=w ....)

Anyways, I think it is important to add tests to this PR before merging. I'm not sure if all the plots will work properly after the changes.

Gui-FernandesBR:
PR is not working, needs some debugging still. As this is not so urgent, I will mark as a draft for now.

```
self = <rocketpy.simulation.flight.Flight object at 0x70b7fba580d0>

    @funcify_method("Time (s)", "Freestream Velocity X (m/s)", "spline", "constant")
    def stream_velocity_x(self):
        """Freestream velocity X component as a Function of time."""
        stream_velocity_x = np.column_stack(
>           (self.time, self.wind_velocity_x[:, 1] - self.vx[:, 1])
        )
E       ValueError: operands could not be broadcast together with shapes (389,) (1429,)

rocketpy/simulation/flight.py:2356: ValueError

PR #569, opened 2024-03-07
BUG: wrong rocket mass in parachute u dot method

Gui-FernandesBR:
## Pull request type

## Checklist

## Current behavior
When calculating the parachute forces in the `Flight.u_dot_parachute` method, the code is using the rocket's mass without the motor's dry mass. It should be using the `Rocket.dry_mass` value instead.

## New behavior
Fixed the bug, updated docs and tests.

## Breaking change

- [x] No (but you should expect some slight change in impact velocity values since the v1.0 was out).

## Additional information
- Quick obs.: Acceptance tests should prevent bugs like this, @lucasfourier @GabrielBarberini !!
- This bug was captured today by the Faraday team, Spain. Many thanks to their contribution!
image

### DIFF
--- a/rocketpy/simulation/flight.py
+++ b/rocketpy/simulation/flight.py
@@ -1911,7 +1911,7 @@ class Flight:
         rho = self.env.density.get_value_opt(u[2])
         to = 1.2
         ma = ka * rho * (4 / 3) * np.pi * R**3
-        mp = self.rocket.mass
+        mp = self.rocket.dry_mass
         eta = 1
         Rdot = (6 * R * (1 - eta) / (1.2**6)) * (
             (1 - eta) * t**5 + eta * (to**3) * (t**2)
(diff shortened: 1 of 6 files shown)


rocketpy/simulation/flight.py
```diff
@@ -1911,7 +1911,7 @@ def u_dot_parachute(self, t, u, post_processing=False):
         rho = self.env.density.get_value_opt(u[2])
         to = 1.2
         ma = ka * rho * (4 / 3) * np.pi * R**3
-        mp = self.rocket.mass
+        mp = self.rocket.dry_mass
```
phmbressan, line 1914: I understand that a code might be slightly out of scope of a hotfix, but I advise improving the variable naming here for better clarity.

phmbressan (approved):

rocketpy/simulation/flight.py
Gui-FernandesBR: I agree. This method is hard to read. 

Can we do that in a separate PR please? The Flight class is really old...

lucasfourier (approved):

PR #571, opened 2024-03-07
TST: Environment unit tests refactoring

GabrielBarberini:
## Pull request type

## Checklist

- [X] All tests (`pytest tests -m slow --runslow`) have passed locally --> All tests that were previously passing continued to pass

## Breaking change

- [X] No

## Additional information

- Small refactoring with emphasis on test clarity and DRY;
- FUP on [RFC - RocketPy pytest architecture](https://docs.google.com/document/d/14GBEDzqXjn7YAg4s0ePvL0SW0PoGbqB4UxN0IZ5J-nQ/edit#heading=h.jvat5sscp9g9)

### DIFF
--- a/tests/test_plots.py
+++ b/tests/test_plots.py
@@ -43,7 +43,7 @@ def test_compare(mock_show, flight_calisto):
 
 
 @patch("matplotlib.pyplot.show")
-def test_compare_flights(mock_show, calisto, example_env):
+def test_compare_flights(mock_show, calisto, example_plain_env):
     """Tests the CompareFlights class. It simply ensures that all the methods
     are being called without errors. It does not test the actual plots, which
     would be very difficult to do.
@@ -54,14 +54,14 @@ def test_compare_flights(mock_show, calisto, example_env):
         Mocks the matplotlib.pyplot.show() function to avoid showing the plots.
     calisto : rocketpy.Rocket
         Rocket object to be used in the tests. See conftest.py for more details.
-    example_env : rocketpy.Environment
+    example_plain_env : rocketpy.Environment
         Environment object to be used in the tests. See conftest.py for more details.
 
     Returns
     -------
     None
     """
-    example_env.set_atmospheric_model(
+    example_plain_env.set_atmospheric_model(
         type="custom_atmosphere",
         pressure=None,
         temperature=300,
@@ -77,7 +77,7 @@ def test_compare_flights(mock_show, calisto, example_env):
     for heading in headings:
         for inclination in inclinations:
             flight = Flight(
-                environment=example_env,
+                environment=example_plain_env,
                 rocket=calisto,
                 rail_length=5,
                 inclination=inclination,
--- a/tests/unit/test_environment.py
+++ b/tests/unit/test_environment.py
@@ -107,7 +107,7 @@ def test_environment_export_environment_exports_valid_environment_json(
     example_spaceport_env : rocketpy.Environment
     """
     # Check file creation
-    assert example_spaceport_env.export_environment(filename="environment") == None
+    assert example_spaceport_env.export_environment(filename="environment") is None
     with open("environment.json", "r") as json_file:
         exported_env = json.load(json_file)
     assert os.path.isfile("environment.json")
--- a/tests/unit/test_environment.py
+++ b/tests/unit/test_environment.py
@@ -1,4 +1,3 @@
-import datetime
 import os
 
 import numpy as np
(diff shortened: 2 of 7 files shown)


Gui-FernandesBR:
Good PR.

Tests not passing? That catches my attention. Investigating it right now.

image

GabrielBarberini:
> Good PR.
> 
> Tests not passing? That catches my attention. Investigating it right now.
> 
> image

I summon: works on my machine 😂 

```tests/unit/test_environment.py ..................                                    [100%]``` 

Let me know if you find any issues

**Comment by Gui-FernandesBR** [comment]
> > Good PR.
> > Tests not passing? That catches my attention. Investigating it right now.
> > image
> 
> I summon: works on my machine 😂
> 
> `tests/unit/test_environment.py .................. [100%]`
> 
> Let me know if you find any issues

Fair enough lol,

Are you using MacOS? Which python version?

**Comment by GabrielBarberini** [comment]
> > > Good PR.
> > > Tests not passing? That catches my attention. Investigating it right now.
> > > image
> > 
> > 
> > I summon: works on my machine 😂
> > `tests/unit/test_environment.py .................. [100%]`
> > Let me know if you find any issues
> 
> Fair enough lol,
> 
> Are you using MacOS? Which python version?

MacOS 14.2 Beta (23C5041e) under Python 3.11.5

**Comment by Gui-FernandesBR** [comment]
@GabrielBarberini a problem is that you renamed the variable `example_env_robust` but a lot of tests used it.
Try to run all the tests at once, it doesn't take that long to be executed. (I usually run `make pytest`, see the Makefile in the root directory).

Try replacing `example_env_robust` to `example_spaceport_env`.

More replacements will be necessary I think.

**Comment by GabrielBarberini** [comment]
> @GabrielBarberini a problem is that you renamed the variable `example_env_robust` but a lot of tests used it. Try to run all the tests at once, it doesn't take that long to be executed. (I usually run `make pytest`, see the Makefile in the root directory).
> 
> Try replacing `example_env_robust` to `example_spaceport_env`.
> 
> More replacements will be necessary I think.

Hmmm! You are right. Didn't notice this dependency.

Fixed :)

**Comment by GabrielBarberini** [comment]
All good now @Gui-FernandesBR 

Sorry for that, for some reason I assumed the env fixtures were used only by the env tests.

Lastly, I tried to use fixture a teardown, but since other modules use env fixtures differently, it ended up conflicting with my change. So I set a request.addfinalizer() inside the unit env test that requires teardown (for deleting the `environment.json` file).

For review convenience, I added it as a separate commit.

**Comment by GabrielBarberini** [comment]
Now all good.

File: tests/test_plots.py [hunk 1]
```diff
@@ -54,14 +54,14 @@ def test_compare_flights(mock_show, calisto, example_env):
...
 
     Returns
     -------
     None
     """
-    example_env.set_atmospheric_model(
+    example_plain_env.set_atmospheric_model(
         type="custom_atmosphere",
         pressure=None,
         temperature=300,
```
Gui-FernandesBR, line 67: I believe in the future we should be using a fixture with the custom atmosphere already set. This would avoid us to set the atm model several times, which is really time consuming.

tests/unit/test_environment.py
```diff
@@ -8,109 +7,106 @@
...
 
     Parameters
     ----------
-    example_env_robust : rocketpy.Environment
-        Example environment object to be tested.
+    example_spaceport_env : rocketpy.Environment
     """
-    assert example_env_robust.export_environment(filename="environment") == None
+    assert example_spaceport_env.export_environment(filename="environment") == None
+    assert os.path.isfile("environment.json")
```
Gui-FernandesBR, line 105: Asserting the file exists is already something good!
What do you think about opening the file and checking if the results are still like the expected values? For example, checking if all the keys are available. I worried about generating an empty file.

tests/unit/test_environment.py
```diff
@@ -8,109 +7,106 @@
...
     )
-    assert (
-        example_env.get_elevation_from_topographic_profile(
-            example_env.latitude, example_env.longitude
-        )
-        == 1565
+    computed_elevation = example_plain_env.get_elevation_from_topographic_profile(
+        latitude, longitude
     )
+    assert computed_elevation == theoretical_elevation
```
Gui-FernandesBR, line 94: This seems to be too restricted. What if we use pytest.approx to make this comparison?
It doesn't need to have a large tolerance, it is just to avoid numerical errors like: 5 is not equals to 5.0000000000001

tests/unit/test_environment.py
```diff
@@ -8,109 +7,106 @@
...
     """Tests the export_environment() method of the Environment class.
 
     Parameters
     ----------
-    example_env_robust : rocketpy.Environment
-        Example environment object to be tested.
+    example_spaceport_env : rocketpy.Environment
     """
-    assert example_env_robust.export_environment(filename="environment") == None
+    assert example_spaceport_env.export_environment(filename="environment") == None
```
Gui-FernandesBR, line 104: This is more pythonic:

```suggestion
    assert example_spaceport_env.export_environment(filename="environment") is None
```

Gui-FernandesBR: It happens in more sections of this same file.

tests/unit/test_environment.py
```diff
@@ -8,109 +7,106 @@
...
 
-def test_env_set_date(example_env):
-    """Test that the date is set correctly in the environment object. This
-    basically takes a date and time and converts it to a datetime object, then
-    set the date to the example environment object. The test checks if the
-    datetime object is the same as the one in the example environment object.
+def test_date_naive_set_date_saves_utc_timezone_by_default(
+    example_plain_env, example_date_naive
+):
+    """Tests time zone is set to UTC by default in environment obj given a date_naive input
```
Gui-FernandesBR, line 14: This line is too long (more than 80 columns). I know it is annoying, but can you break it in 2 please?

tests/unit/test_environment.py
```diff
@@ -142,51 +138,53 @@ def test_utm_to_geodesic():
...
     ],
 )
-def test_decimal_degrees_to_arc_seconds(angle, deg, arc_min, arc_sec):
-    """Tests if the conversion from decimal degrees to arc seconds is correct.
-    It takes 3 different angles and their expected results and compares them
-    with the results from the method.
+def test_decimal_degrees_to_arc_seconds_computes_correct_values(
+    angle, theoretical_degree, theoretical_arc_minutes, theoretical_arc_seconds
+):
+    """Tests the conversion from decimal degrees to degrees, arc minutes, and arc seconds.
```
Gui-FernandesBR, line 173: Line too long too.

Gui-FernandesBR (commented):
Easy to review a PR when it is well constructed and fixes one problem at time.

My comments are only related to small convention details.
Code works, PR is clear, that's  what really matters.

GabrielBarberini:
> Easy to review a PR when it is well constructed and fixes one problem at time.
> 
> My comments are only related to small convention details. Code works, PR is clear, that's what really matters.

Thanks for the feedback! I'll address your requests soon

Gui-FernandesBR:
This PR contributes to #480

tests/test_plots.py
GabrielBarberini: I agree this would increase readability by avoiding code repetition. However I could not find a common base for the atmospheric model used in tests, sometimes custom_atm is used, sometimes standard, sometimes other, etc...

Also, I'm not sure that moving it to fixtures will speed up testing as each fixture call creates a new instance to set the atm model (fixture instances are not shared between tests).

tests/unit/test_environment.py
GabrielBarberini: Nice suggestion. BTW this made me find a bug in the environment code.

At some point in Environment code (line 1321), `self.pressure` is set to be equal to `self.pressure_ISA` (which is, indirectly a `Function()`) or to `Function()` directly. This led `export_environment()` to set `atmospheric_model_pressure_profile` as the memory address of the Function() obj, not the expected `pressure_function()` output.

I've fixed it both tests and lib by changing it to 

```
(...)
 ma.getdata(
                self.pressure.get_source()(self.height)
            ).tolist(),
```

A similar issue was happening with:
```
    "atmospheric_model_wind_velocity_x_profile": "<function Function.set_source.<locals>.source_function at 0x112804720>",
    "atmospheric_model_wind_velocity_y_profile": "<function Function.set_source.<locals>.source_function at 0x112804040>"
```

You can check it in detail in the commit BUG: Fixes environment_export

tests/unit/test_environment.py
Gui-FernandesBR: I understand your argument, and honestly I didn't know the difference between `==` and `is` so much. The stackoverflow link really helped.

There's a "programming recommendation" in PEP8: "Comparisons to singletons like None should always be done with is or is not, never the equality operators". See [here](https://peps.python.org/pep-0008/#programming-recommendations).

Addinionally, there's a short thread [here](https://stackoverflow.com/questions/14247373/python-none-comparison-should-i-use-is-or) and [here](https://stackoverflow.com/questions/3257919/what-is-the-difference-between-is-none-and-none).

This is one of those things in Python where, no matter what we choose, code will be working.

If you still want to keep the `==`, no problems at this point. There are many parts of the library and tests where we use `== None`. The first reason for me to comment on this is due to pylint warnings: [C0121](https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/singleton-comparison.html)

tests/test_plots.py
Gui-FernandesBR: I searched in the code and found 2 instances of "example_plain_env.set_atmospheric_model("custom_atmosphere"...." (not written exactly like this) in the tests.

*Probably* a fixture that takes the plain environment and set the custom_atmosphere would be useful.
No need to implement it in this PR tho.

tests/unit/test_environment.py
Gui-FernandesBR: Thank you for your solution.
I think this method is not so used, therefore the bug was never captured before.
Good to see it's better now.

tests/unit/test_environment.py
Gui-FernandesBR: Thank you for your comment.
I was under the impression that the output of this method should be a bi-linearly interpolated value based on that array, that's why I thought about the float value problem.

Let's leave the code as it is right now, we revisit this if we ever face a floating problem.

tests/unit/test_environment.py
Gui-FernandesBR: Btw these comments are also valid for the `== True` occurrences

Gui-FernandesBR:
Thanks again for the contribution, @GabrielBarberini ,
I commented on the conversations, you can mark them as resolved once you feel they are done.

I believe we are close to approve/merge

tests/unit/test_environment.py
Gui-FernandesBR: For your information, @phmbressan .
I think you'd enjoy reading these threads eventually.

GabrielBarberini: Hmmmm I didn't know `None` was a singleton, now it makes sense. Then I agree `is` should be used instead. I will push a correction

Gui-FernandesBR (approved):
LGTM

PR #572, opened 2024-03-07
ENH: adds new stability margin methods to flight

Gui-FernandesBR:
## Pull request type

## Checklist

## New behavior
Use the new properties to access stability margin at specific times.
This will be useful for the Monte Carlo analyses.

## Breaking change

- [x] No

## Additional information
- Once we create these properties, we won't be able to delete them.
- I did not use cached property because I think the computational time for calling the functions is lower than the time spent to create a cached item in this case. Feel free to disagree.


rocketpy/simulation/flight.py
```diff
@@ -2426,6 +2430,26 @@ def min_stability_margin(self):
...
 
+    @property
+    def initial_stability_margin(self):
+        """Stability margin at time 0.
+
+        Returns
+        -------
+        float
+        """
+        return self.stability_margin(0)
```
phmbressan, line 2441: Is the `t = 0` always a appropriate for getting the simulation beginning?

What if my simulation has an `initial_solution` (another `Flight` or a state vector)?

phmbressan (commented):
I have just made a comment regarding the time at which simulation begin and if an initial solution or # change expected behavior of this method.

rocketpy/simulation/flight.py
Gui-FernandesBR: Nice catch,
the solution was really simple: [ee01e9d](https://github.com/RocketPy-Team/RocketPy/pull/572/commits/ee01e9dd9f159d463e642465244ebe08f4038565)

Tests are passing.

MateusStano (changes requested):
The new properties should be added to flight.prints as well!

Gui-FernandesBR:
> The new properties should be added to flight.prints as well!

Thank you for your suggestion. I solved it in [c9cc0de](https://github.com/RocketPy-Team/RocketPy/pull/572/commits/c9cc0deb6393ed38b3dc18b9d0c8dc59ee840e42)

Can you re-review please? I need your approve now that you requested changes

Gui-FernandesBR:
Got some inspiration from this PR and decided to refactor the whole flight_prints.py module directly: https://github.com/RocketPy-Team/RocketPy/pull/579

MateusStano (approved):

PR #575, opened 2024-03-11
ENH:  Introducing local sensitivity analysis

Lucas-Prates:
## Pull request type
- [x] Other (please describe): This PR is a **draft** to implement sensitivity analysis in RocketPy. It is a work in progress and needs validation.
## Checklist

## Current behavior
The Sensitivity Analysis notebook teaches the users how to perform the simulations, plot the distribution
of some flight variables (e.g. apogee), and computes the prediction ellipses for the landing point.

## New behavior
Our goal is to take sensitivity analysis even further. Briefly, we attempt to answer the following question: **Which parameters would reduce the variability of the variable of interest (e.g. apogee) the most if we measured them with greater precision?**

To that end, a bit of theory is developed, check the technical document. What was developed resembles the work of [1], a core reference in sensitivity analysis for engineering. His approach is a *global sensitivity analysis with a full model containing interaction terms*. Our first implementation considers a  *local sensitivity analysis using only first-order terms*.

A quick and dirty test of the functionality of the SensitivityModel class is provided the "sensitivity_model_usage" notebook. **This notebook is currently giving weird results! The linear approximations for the variables are, for some reason I still have to figure out, not good enough. This was not happening at previous experimentations that suggested that this approached worked. I have to look carefully at what is happening, but I did not want to delay the PR.**

The concepts are discussed in-depth in the "sensitivity_analysis_parameter_importance" notebook (**the notebook was not updated to the new SensitivityModel yet!**) 

## Breaking change

- [x] No

## Additional information

[Technical Document](https://drive.google.com/drive/u/1/folders/1C0HbEfwfSUAkPOp0WF3u5Ku2JyxzNDe2)

[1] Sobol, Ilya M. "Global sensitivity indices for nonlinear mathematical models and their Monte Carlo estimates." Mathematics and computers in simulation


Gui-FernandesBR:
Tests are not passing

```text
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/runner/work/RocketPy/RocketPy/rocketpy/__init__.py", line 2, in <module>
    from .environment import Environment, EnvironmentAnalysis
  File "/home/runner/work/RocketPy/RocketPy/rocketpy/environment/__init__.py", line 1, in <module>
    from .environment import Environment
  File "/home/runner/work/RocketPy/RocketPy/rocketpy/environment/environment.py", line [16](https://github.com/RocketPy-Team/RocketPy/actions/runs/9096367936/job/25001616031?pr=575#step:6:17), in <module>
    from ..tools import exponential_backoff
  File "/home/runner/work/RocketPy/RocketPy/rocketpy/tools.py", line 556, in <module>
    parameters_list: list[str],
TypeError: 'type' object is not subscriptable
Error: Process completed with exit code 1.
```

Could you fix it before our review, please? That would help us. @Lucas-Prates

Lucas-Prates:
> Could you fix it before our review, please? That would help us. @Lucas-Prates

Sure, I will fix it briefly. This simplified type hinting started at [python 3.9](https://peps.python.org/pep-0585/#parameters-to-generics-are-available-at-runtime) :P. I will make sure the tests pass this time.

Gui-FernandesBR:
Please be aware of #444, we are not supporting type hinting or annotations yet.

setup.py
Gui-FernandesBR, line 1: Why adding a setup.py file?

We are using the `pyproject.toml` file now. We no longer support the `setup.py`

rocketpy/sensitivity/sensivity_model.py
```diff
@@ -0,0 +1,515 @@
...
+        Parameters
+        ----------
+        parameters_nominal_mean : np.array
+            An array contaning the nominal mean for parameters in the
+            order specified in parameters names at initialization
+        parameters_nominal_sd : np.array
+            An array contaning the nominal standard deviation for
+            parameters in the order specified in parameters names at
+            initialization
+        """
```
MateusStano, line 81: So do you have to set the mean **and** sd simultaneously? 

Also, do you have to set mean and sd of all parameters? Setting for just some of them does not work?

Another thing, to run a Monte Carlo sim, the mean and sd is already given in the Monte Carlo class right? So it would be natural to get them from there automatically

rocketpy/sensitivity/sensivity_model.py
```diff
@@ -0,0 +1,515 @@
...
+        ----------
+        parameters_matrix : np.matrix
+            Data matrix whose columns correspond to parameters values
+            ordered as passed in initialization
+
+        """
+        if parameters_matrix.shape[1] != self.n_parameters:
+            raise ValueError(
+                "Number of columns (parameters) does not match number of parameters passed at initialization."
+            )
```
MateusStano, line 143: Hasn't this already been checked if there is a `parameters_matrix` ?

rocketpy/tools.py
```diff
@@ -549,6 +550,119 @@ def generate_monte_carlo_ellipses_coordinates(
...
+        Output file exported by MonteCarlo.simulate function. Each line is a
+        sample unit described by a dictionary where keys are target variables
+        names and the values are the obtained values from the flight simulation.
+
+    parameters_list : list[str]
+        List of parameters whose values will be extracted.
+
+    target_variables_list : list[str]
+        List of target variables whose values will be extracted.
+
```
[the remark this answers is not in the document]
MateusStano: This also happens in a lot of the other functions in this PR

MateusStano (changes requested):

rocketpy/tools.py
```diff
@@ -549,6 +550,119 @@ def generate_monte_carlo_ellipses_coordinates(
...
+                    inner_dict = unnest_dict(value[0])
+                    inner_dict = {
+                        key + "_" + inner_key: inner_value
+                        for inner_key, inner_value in inner_dict.items()
+                    }
+                    new_dict.update(inner_dict)
+            else:
+                new_dict.update({key: value})
+
+        return new_dict
```
Gui-FernandesBR, line 612: I would define this function outside the `load_monte_carlo_data()` function, this would allow us to re-use this function in other contexts.

Also, the term `flatten` is usually used to describe this kid of "parse a nested dictionary" operation. Maybe it would be a good alternative of name.

rocketpy/__init__.py
```diff
@@ -49,3 +49,5 @@
     StochasticTail,
     StochasticTrapezoidalFins,
 )
+from .sensitivity import SensitivityModel
```
Gui-FernandesBR, line 52: Run isort

rocketpy/sensitivity/sensivity_model.py
```diff
@@ -0,0 +1,515 @@
...
+    order terms, which correspond to linear terms. Albeit the flight
+    function is nonlinear, the linear hypothesis might be adequate when
+    performing *local* sensitivity analysis.
+
+    The model is fit using separate multiple linear regression for each
+    target variable passed and using the parameters as covariates.
+
+    References
+    ----------
+    [1] Sobol, Ilya M. "Global sensitivity indices for nonlinear mathematical models and their Monte Carlo estimates." Mathematics and computers in simulation 55.1-3 (2001): 271-280.
```
Gui-FernandesBR, line 22: Line too long... You can use the `\` sign to continue the text in the next line whithout breaking it. Like this:

"something here \
    continues here, do you see?"

rocketpy/sensitivity/sensivity_model.py
```diff
@@ -0,0 +1,515 @@
...
+            An array contaning the nominal mean for parameters in the
+            order specified in parameters names at initialization
+        parameters_nominal_sd : np.array
+            An array contaning the nominal standard deviation for
+            parameters in the order specified in parameters names at
+            initialization
+        """
+        if len(parameters_nominal_mean) != self.n_parameters:
+            raise ValueError(
+                "Nominal mean array length does not match number of parameters passed at initilization."
```
Gui-FernandesBR, line 84: typo fix. And line is too long.

```suggestion
                "Nominal mean array length does not match number of parameters passed at initialization."
```

(same thing happens in a few lines below)

rocketpy/sensitivity/sensivity_model.py
```diff
@@ -0,0 +1,515 @@
...
+        for i in range(self.n_target_variables):
+            target_variable = self.target_variables_names[i]
+            self.target_variables_info[target_variable]["nominal_value"] = (
+                target_variables_nominal_value[i]
+            )
+
+        self._nominal_target_passed = True
+
+        return
+
```
Gui-FernandesBR, line 126: You can probally remove this return and the code you keep working.

```suggestion
```

Gui-FernandesBR (changes requested):
This is just a partial review. I could not run your codes to validate it is working, but I suggested a few changes that might improve the quality (i.e. readability) of the code.

setup.py
Lucas-Prates: Indeed, I had this file from a previous pull request and thought the project still used `setup.py`, sorry.

rocketpy/sensitivity/sensivity_model.py
```diff
@@ -0,0 +1,515 @@
...
+                )
+
+        return
+
+    def fit(
+        self,
+        parameters_matrix,
+        target_data,
+    ):
+        """Fits sensitivity model
```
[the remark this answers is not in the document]
Lucas-Prates: I understand and agree with your point. However, several data analysis libraries in Python follow this style that I implement: you define the data analysis model object, and then you pass data to fit it. A famous example is the `scikit-learn` library. Another example that is in the code base is  `statsmodels`, we first define statsmodels.OLS and then we fit it.

All that being said, I do not mind changing the style, I implemented that to follow the format I am used to.

rocketpy/sensitivity/sensivity_model.py
Lucas-Prates: It would be great to recover the mean and sd used in the simulation! I tried to find where it was stored, but I might have missed it. Do you know where it is stored?

Lucas-Prates: Indeed, you don't have to provide both, thought that, if you had one of them, you probably had the other.

rocketpy/sensitivity/sensivity_model.py
Lucas-Prates: You are absolutely right, this check and the one in `_estimate_target_nominal` are redundant.

rocketpy/tools.py
```diff
@@ -549,6 +550,119 @@ def generate_monte_carlo_ellipses_coordinates(
...
+    parameters_matrix = np.empty((n_samples, n_parameters))
+    target_variables_matrix = np.empty((n_samples, n_variables))
+
+    for i in range(n_parameters):
+        parameter = parameters_list[i]
+        parameters_matrix[:, i] = parameters_samples[parameter]
+
+    for i in range(n_variables):
+        target_variable = target_variables_list[i]
+        target_variables_matrix[:, i] = target_variables_samples[target_variable]
```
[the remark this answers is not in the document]
Lucas-Prates: Clear and useful, replaced most of the `ranges` in code by `enumerate`. Thanks!

rocketpy/tools.py
Lucas-Prates: `flatten_dict` is definitely a better name for it! With respect to defining it outside the `load_monte_carlo_data`, there is a detail in the implementation that is specific to the way `MonteCarlo` simulation data is saved. I have not defined it outside of the function in this first review round because of this, but I might think of a way to make it general and work with the `MonteCarlo` data in the future.

rocketpy/sensitivity/sensivity_model.py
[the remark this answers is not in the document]
MateusStano: Well, if this is the standard then best to leave it like this

Lucas-Prates:
The PR is ready for review again. I fixed a bug on the computations of the sensitivity score and it explained why the model was failing overall. Now it works for several target variables and for more complex enviroments (e.g. forecasted atmosphere).

The sensitivity model usage and sensitivity simulation notebooks are the most important to review. There are still a couple of things left to do:

- Improve and extend the writing in the sensitivity model usage notebook;
- Understand why the approximation for x_impact and y_impact is failing.

rocketpy/sensitivity/sensivity_model.py
Gui-FernandesBR: I think we could use another PR to better integrate the `MonteCarlo` and the `SensitivityModel` classes.

docs/notebooks/monte_carlo_analysis/sensitivity_model_usage.ipynb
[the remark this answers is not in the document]
Lucas-Prates: Thanks for the review and changes. Regarding your questions:

1) Indeed, the p-values are not the relevant measures used in this framework for sensitivity analysis. I provided them for completeness sake, but I agree they are more deceitful than helpful. Hence, it is a better idea to remove them from the summary. Since the statistical model is still provided in the output of the sensitivity object, a user which wants to get the p-values (for whatever reasons) should be able to do so;

2) I totally agree that this should be a docs page instead of a notebook. Should we go in that direction in this PR, then?

3) Yes, there is a technical document showing the mathematical/statistical reasoning of the method. I provided a link here in this PR, but it is not available anymore for some reason. I have the latex on overleaf and we can think of a better way to provide the .pdf file;

4) You are absolutely right that the commentary is too broad. Truth is, I do not know what is a good threshold. In my understanding, this is something that would be answered with the use of the tool in practice, so we could get a sense of a reasonable threshold.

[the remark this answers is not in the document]
Gui-FernandesBR: Thx for your work, @Lucas-Prates , I have made some quick fixes to this PR, but there're still some points I believe we could address before merging it onto develop.


Remaining tasks to merge this PR:

- [ ] Remove p-values from the summary table
- [ ] Fix too broad commentary regarding large LAE
- [ ] Review documentation pages
- [ ] Delete the sensitivity notebook after the new docs are ready

Other tasks for the future:
- [ ] Create a `MonteCarlo.sensitivity_analysis(target_variables, parameters)` method that returns a SensitivityModel already fitted.
- [ ] Add integration tests (using real values instead of a generic np.array)

**Question**: is it possible to have multiple target variables in a same model?

docs/notebooks/monte_carlo_analysis/sources_of_variation.png
Gui-FernandesBR, line 1: Suggestion: save this .png file in the RocketPy/docs/static folder

docs/notebooks/monte_carlo_analysis/sensitivity_model_usage.ipynb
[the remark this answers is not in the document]
Lucas-Prates: Yes, it is possible to pass multiple target variables, you would pass `["apogee", "x_impact"]` instead of `["apogee"]`.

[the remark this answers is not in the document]
Gui-FernandesBR: Ok, nice. @Lucas-Prates I fixed some issues that I identified on the documentation. You can use this link to validate how the latest commits of this branch are impacting on the documentation: https://docs.rocketpy.org/en/enh-sensitivity_analysis/

Lucas-Prates:
@Gui-FernandesBR and @MateusStano, this PR is ready for review again! I believe the review should be somewhat quick. The code has already been thoroughly review, so I recommend focusing on two files:

- docs/technical/sensitivity.rst
- docs/user/sensitivity.rst

In both cases, my recommendation is to read the text and check if it is clear. I wouldn't mind much about the mathematics, I took a second look at it and "looks fine." Moreover, I would like to know from you if the example I provided makes sense. It is very simplified, but I would hope it somehow creates the impression that sensitivity analysis can be useful.

@Gui-FernandesBR, I tried to address the "LAE is too large" by giving a pragmatic recommendation of how to use the tool. Giving an exact value of what is "too large" is, in my opinion, not easy/feasible right now.

Gui-FernandesBR (approved):
LGTM

Gui-FernandesBR:
> @Gui-FernandesBR and @MateusStano, this PR is ready for review again! I believe the review should be somewhat quick. The code has already been thoroughly review, so I recommend focusing on two files:
> 
> * docs/technical/sensitivity.rst
> * docs/user/sensitivity.rst
> 
> In both cases, my recommendation is to read the text and check if it is clear. I wouldn't mind much about the mathematics, I took a second look at it and "looks fine." Moreover, I would like to know from you if the example I provided makes sense. It is very simplified, but I would hope it somehow creates the impression that sensitivity analysis can be useful.
> 
> @Gui-FernandesBR, I tried to address the "LAE is too large" by giving a pragmatic recommendation of how to use the tool. Giving an exact value of what is "too large" is, in my opinion, not easy/feasible right now.

@Lucas-Prates your proposed example is exactly what we needed when we created the issue. Problem solved, thank you so much!

PR #580, opened 2024-03-24
MNT: Modularize Rocket Draw

MateusStano:
## Pull request type

## Checklist

## Current behavior

The method of `_RocketPlots` relating to the rocket drawing was too big and hard to change/improve/maintain

## New behavior

The `.draw()` method has been separated into several different methods. No code relating to the plot has been changed.

## Breaking change

- [X] No

## Additional information

I was trying to add sensors to the drawing and it was really hard to work with the drawing, so I refactored it.

A quick note:

- There are a few lines of repeated code that could still be turned into functions. These are usually some matplotlib one-liners. I did not create a function for these because it affected readability negatively.

### DIFF
--- a/rocketpy/plots/rocket_plots.py
+++ b/rocketpy/plots/rocket_plots.py
@@ -189,7 +189,6 @@ class _RocketPlots:
             A full list of color names can be found at:
             https://matplotlib.org/stable/gallery/color/named_colors
         """
-        # TODO: we need to modularize this function, it is too big
         if vis_args is None:
             vis_args = {
                 "background": "#EEEEEE",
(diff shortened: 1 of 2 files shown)


Gui-FernandesBR (commented):
Overall, very good PR. I understand you refactored the code without changing the behavior. You probably ran all the example notebooks to check the plots are being rendered correctly.

Some lines in the `_draw_nozzle_tube` function are missing coverage.
Could you please take a look at that? Just to ensure the code never brakes.

rocketpy/plots/rocket_plots.py
```diff
@@ -225,113 +241,105 @@ def draw(self, vis_args=None):
...
+        ax.plot(
+            [x_nosecone[-1], x_nosecone[-1]],
+            [y_nosecone[-1], -y_nosecone[-1]],
+            color=vis_args["nose"],
+            linewidth=vis_args["line_width"],
+        )
+        # Add the nosecone to the list of drawn surfaces
+        drawn_surfaces.append(
+            (surface, x_nosecone[-1], surface.rocket_radius, x_nosecone[-1])
+        )
```
Gui-FernandesBR, line 278: This is strange to me. You add the surface to a list, which was passed as an argument to the function. If you don't return the list with the added surfaces, is this code line really necessary? I'm afraid that everytime you are starting with an empty list.

```
        # Add the nosecone to the list of drawn surfaces
        drawn_surfaces.append(
            (surface, x_nosecone[-1], surface.rocket_radius, x_nosecone[-1])
        )
```

Gui-FernandesBR: This behavior is also present in other parts of the code.

Gui-FernandesBR:
Also, @MateusStano can you please test the implementation in a multi-stage rocket? I think the NDRT example notebook may help you with that.

rocketpy/plots/rocket_plots.py
MateusStano: The `drawn_surfaces` list is created and then appended to the various draw methods I created. That same list is used later to draw the tubes later.

Remember that a list in Python is just a pointer. Changing the name of the list variable does not change the address that it points to.

Gui-FernandesBR:
> > Also, @MateusStano can you please test the implementation in a multi-stage rocket? I think the NDRT example notebook may help you with that.
> 
> The drawing ends up identical to the way it was before. Did you find any problems there?

Nope. Sorry, I just did not have enough time to checkout your branch and run the examples.

If you have tested all the examples, I trust you. Go ahead and merge it.

Gui-FernandesBR (approved):
Good work.

Issue #481, opened 2023-11-23
ENH: _check_user_input method being called multiple times 


lucasfourier:
**Describe the bug**

The `_check_user_input `method is being called multiple times for some unexpected reason during __init__ for the same source file.

**Expected behavior**

Processing should not result in redundant calculations and function calls. 

**Additional context**

This discussion arised from a hotfix. For additional information, one should read the following: https://github.com/RocketPy-Team/RocketPy/pull/478

lucasfourier:
True. Changing the label.

Issue #450, opened 2023-10-30
ENH: use of built-in python loggers instead of prints


Gui-FernandesBR:
**Is your feature request related to a problem? Please describe.**

RocketPy doesn't offer a good way of tracking the execution steps when running a flight simulation. 
There are some prints around the code, but they do not cover all the classes and do not look robust enough for a definitive solution as rocketpy wants to be.

**Describe the solution you'd like**

* Adopt the best practices from the built-in [logging](https://docs.python.org/3/library/logging.html) facility of Python. A basic guide is available here: https://docs.python.org/3/howto/logging.html#logging-basic-tutorial. Further investigation may be needed to understand how exactly this module works.
* Ensure that at least the core classes are supports log messages from logging.
* We should also allow for a log file generation to be accessed by the end of the execution.
* The Flight class could be the first one to receive this new feature.

**Additional context**

I'm guessing this implementation would leverage the debugging process of rocketpy's simulations, saving valuable time for both devs and users.

YuriCastroDev:
Hey @Gui-FernandesBR, I did some research on how other libraries handle this same situation. Here's what I found:

**pandas**: The `.info()` method writes to a `buf` parameter that defaults to `sys.stdout`. No `print()`, no `logging`. The user can redirect by passing any writable buffer (e.g. `StringIO`).

**matplotlib**: Uses pure `logging` throughout with a `set_loglevel()` helper for the user. Silent by default, user configures output. This would be a breaking change for RocketPy since `.info()` calls would produce no output without user configuration.

**scipy**: Uses `print()` directly when the user explicitly requests output via `disp=True` or `verbose=True`. Uses `warnings.warn()` for internal events. Simple and backwards  compatible, but no logging benefits.

**astropy**: Built a custom `AstropyLogger` subclass that adds a `StreamHandler` internally so output appears in the terminal by default. Works well, but violates the official Python recommendation ("do not add any handlers other than NullHandler to your library's loggers").

**statsmodels**: `.summary()` returns an object that implements `__str__`. The user explicitly calls `print(result.summary())`. Clean, but changes the API and would be a breaking change.

---

I also checked the Python logging guide, which makes this distinction:

> "Display console output for ordinary usage of a command line script or program: **print()**"
> "Report events that occur during normal operation: **logger.info()**"

---

I asked Claude to help me analyse the pros and cons of each approach:

**Option A: pandas approach (`buf` parameter)**
- ✅ Backwards compatible
[...]

PR #581, opened 2024-03-25
ENH: Flight simulation speed up

Gui-FernandesBR:
## Pull request type

## Checklist

## Current behavior
Simulation is taking too long to run!

## New behavior
Read the commit history, one by one.

Some comments:
* The `get_value_opt` method can be twice as faster as the normal `get_value` method.
* After the implementation, I'm seeing a total simulation time of around 3s, while before we had a simulation time of ~4s. I will add my tests here soon enough.

## Breaking change
- [x] No

## Additional information
I am still running some tests within this branch.


Gui-FernandesBR:
Ok, here is more detailed comment regarding the performance of flight simulation.

- Everything was only possible due to `cProfile` and `snakeviz`, amazing tools!
- Pieces of code that we still need to improve:
    - `Flight.u_dot_generalized`: this method is at least x2 slower than the traditional u_dot method, still present. This is hard to speedup tho. 
    - `Function.set_source`: it is calling the _check_user_input, which really slows down the whole Function.__init__
    - `Function._check_user_input`: I think it is time to discuss if we really need this method running as default with rocketpy. It is taking a significant time of simulation.
    - I would also suggest the discussion on the usage of `np.arg_sort()` method to sort the x_array in the Function class. This also consumes time in order to avoid problems in the Function data source.
    - `Function.__mul__`: I tried my best, but there is still a lot of `if` statements slowing this function.  I'm open to suggestions here.
    - `Function.get_value_opt`: It is already very optimized, but it is not enough. Now this is even more crucial for the Flight class, this method is responsible for ~8% of simulation time.
- Is there any numerical integration scheme faster than LSODA? It seems to be taking around 60% of the simulation time. Plus it cannot be parallelized. 
- I created a **simple** test to compare the simulation times before and after changes:

```txt
Before the improvements done in this PR:
========================================

All 20 simulations done, it took 95.40774 seconds
Mean time: 4.77039
Standard deviation: 0.71902

All 20 simulations done, it took 92.84953 seconds
Mean time: 4.64248
Standard deviation: 0.66002

After the improvements done in this PR:
=======================================

All 20 simulations done, it took 50.86210 seconds
Mean time: 2.54310
Standard deviation: 0.12359

All 20 simulations done, it took 50.47766 seconds
Mean time: 2.52388
Standard deviation: 0.10919
```

Finally, I think this one is ready for review now, @RocketPy-Team/code-owners

Gui-FernandesBR:
This PR, if merged, should solve the #481 issue.

rocketpy/environment/environment.py
```diff
@@ -3142,22 +3147,26 @@ def add_wind_gust(self, wind_gust_x, wind_gust_y):
...
             "Height (m)",
             "Wind Heading (degrees)",
             extrapolation="constant",
         )
         self.wind_speed = Function(
-            lambda h: (self.wind_velocity_x(h) ** 2 + self.wind_velocity_y(h) ** 2)
+            lambda h: (
+                self.wind_velocity_x.get_value_opt(h) ** 2
+                + self.wind_velocity_y.get_value_opt(h) ** 2
+            )
```
Gui-FernandesBR, line 3163: TODO: Analize if we could set np.ndarray-like Functions instead of using lambdas and callable in these Environment assignments.

tests/fixtures/environment/environment_fixtures.py
```diff
@@ -54,8 +55,8 @@ def env_analysis():
     EnvironmentAnalysis
     """
     env_analysis = EnvironmentAnalysis(
-        start_date=datetime.datetime(2019, 10, 23),
-        end_date=datetime.datetime(2021, 10, 23),
+        start_date=datetime(2019, 10, 23),
+        end_date=datetime(2021, 10, 23),
```
Gui-FernandesBR, line 59: This is a bug fix. The problem was introduced in a previous PR so slow tests were not passing.

rocketpy/mathutils/function.py
```diff
@@ -2998,6 +2953,8 @@ def savetxt(
             file.write(header_line + newline)
             np.savetxt(file, data_points, fmt=fmt, delimiter=delimiter, newline=newline)
 
+    # Input validators
+
     @staticmethod
     def _check_user_input(
         source,
```
MateusStano, line 2863: So, `_check_user_input` is really only validating `source`, but it also calls other validation methods inside it.

I really dislike this, the way this validation is structured is really confusing. Here is a suggestion:

1. In `__init__` we call `set_source`, `set_inputs`, `set_interpolation`, `set_extrapolation` and `set_title`.
2. Inside each of these methods, the respective validation method is called, e.g. in `set_source` call `_validate_source`

Currently it is very confusing to understand what is being validated and when.

rocketpy/mathutils/function.py
```diff
@@ -3073,114 +3030,138 @@ def _check_user_input(
...
+
+        if isinstance(source, (int, float)):
+            # Convert number source into vectorized lambda function
+            temp = 1 * source
+
+            def source_function(_):
+                return temp
+
+            source = source_function
+        return source, inputs, outputs, interpolation, extrapolation
```
MateusStano, line 2973: The logic here checks if the source is a `Function` and gets its source, then goes on to validate that. However, that source has already been validated since it came from a Function object. 

Some redundant checks are happening here

rocketpy/mathutils/function.py
```diff
@@ -3073,114 +3030,138 @@ def _check_user_input(
...
+
+    @staticmethod
+    def _validate_inputs_outputs(inputs, outputs):  # None | st | list[str]
+        if inputs is None:
```
MateusStano: Any particular reason for this to be a `@staticmethod`?

rocketpy/mathutils/function.py
```diff
@@ -396,14 +353,12 @@ def set_get_value_opt(self):
...
-        elif self.__extrapolation__ == "constant":
-            extrapolation = 2  # Extrapolation is constant
-        else:
-            raise ValueError(f"Invalid extrapolation type {self.__extrapolation__}")
+        try:
+            extrapolation = self.EXTRAPOLATION_TYPES[self.__extrapolation__]
+        except KeyError as err:
+            raise ValueError(
+                f"Invalid extrapolation type {self.__extrapolation__}"
+            ) from err
```
MateusStano: If there is validation for setting `self.__extrapolation__`  this try-except should never be needed

rocketpy/mathutils/function.py
```diff
@@ -369,7 +369,7 @@ def get_value_opt(x):
                 # Interval found... interpolate... or extrapolate
                 if x_min <= x <= x_max:
                     # Interpolate
-                    x_interval = x_interval if x_interval != 0 else 1
+                    x_interval = max(x_interval, 1)
```
MateusStano, line 369: This line behaves differently than the previous version. Is this correct?

rocketpy/mathutils/function.py
```diff
@@ -2054,54 +2054,45 @@ def __mul__(self, other):
...
+        ):
+            ys = self.y_array * other.y_array
+            xs = self.x_array
+            source = np.concatenate(([xs], [ys])).transpose()
+            outputs = f"({self.__outputs__[0]}*{other.__outputs__[0]})"
+            return Function(source, inputs, outputs, interp, extrap)
+        elif isinstance(other, NUMERICAL_TYPES):
+            if not self_source_is_array:
+                return Function(lambda x: (self.get_value_opt(x) * other), inputs)
+            ys = np.multiply(self.y_array, other)
```
MateusStano: Does `np.multiply` have the same behaviour as `*` for all `NUMERICAL_TYPES`?

tests/test_flight.py
```diff
@@ -729,7 +729,7 @@ def test_velocities(flight_calisto_custom_wind, flight_time, expected_values):
         ("t_initial", (1.6542528, 0.65918, -0.067107)),
         ("out_of_rail_time", (5.05334, 2.01364, -1.7541)),
         ("apogee_time", (2.35291, -1.8275, -0.87851)),
-        ("t_final", (0, 0, 159.2212)),
+        ("t_final", (0, 0, 159.3292416824044)),
```
MateusStano, line 732: What caused this change?

tests/unit/test_utilities.py
```diff
@@ -153,7 +153,7 @@ def test_fin_flutter_analysis(flight_calisto_custom_wind):
     assert np.isclose(flutter_mach(np.inf), 1.0048188594647927, atol=5e-3)
     assert np.isclose(safety_factor(0), 64.78797, atol=5e-3)
     assert np.isclose(safety_factor(10), 2.1948620401502072, atol=5e-3)
-    assert np.isclose(safety_factor(np.inf), 61.64222220469224, atol=5e-3)
+    assert np.isclose(safety_factor(np.inf), 61.62131849619694, atol=5e-3)
```
MateusStano: What caused this change?

rocketpy/mathutils/function.py
```diff
@@ -262,7 +262,7 @@ def set_source(self, source):
 
             # Finally set source, update extrapolation and interpolation
             self.source = source
-            self.set_extrapolation(extrapolation)
```
MateusStano: If extrapolation changes, then `get_value_opt` has to change as well. Why would it not be needed here?

rocketpy/environment/environment.py
```diff
@@ -1507,7 +1507,7 @@ def wind_heading_func(h):
         )
 
         def wind_direction(h):
-            return (wind_heading(h) - 180) % 360
+            return (wind_heading_func(h) - 180) % 360
```
MateusStano, line 1510: Was this a bug? Or was this changed just to speed up code?

Both versions should return the same result

rocketpy/simulation/flight.py
```diff
@@ -3074,17 +3074,13 @@ def _calculate_pressure_signal(self):
...
-            )
             parachute.noise_signal_function = Function(
                 parachute.noise_signal, "Time (s)", "Pressure Noise (Pa)", "linear"
             )
-
-        return None
+            parachute.noisy_pressure_signal_function = (
+                parachute.clean_pressure_signal_function
+                + parachute.noise_signal_function
+            )
```
MateusStano, line 3083: This `Function` should be identical to `Function(parachute.noisy_pressure_signal,"Time (s)","Pressure - With Noise (Pa)","linear",)`.

Can you guarantee that it is?

MateusStano:
Incredible PR with really necessary changes. Just want to point out that most commits and the whole PR is not a Maintanence (MNT) type, but Enhancement. Most commits change the behavior of the classes somewhat

> * I created a **simple** test to compare the simulation times before and after changes:

Can you also check in this test if the simulation results are all returning the same values? There are a lot of small changes to the function class and I just want to make sure none of them have any unintended behavior.


> * `Function.set_source`: it is calling the _check_user_input, which really slows down the whole Function.**init**
> * `Function._check_user_input`: I think it is time to discuss if we need this method running as default with rocketpy. It is taking a significant time for simulation.
> * I would also suggest the discussion on the usage of the `np.arg_sort()` method to sort the x_array in the Function class. This also consumes time to avoid problems in the Function data source.
> * `Function.__mul__`: I tried my best, but there are still a lot of `if` statements slowing this function.  I'm open to suggestions here.

I think the `Function` class is actually misused by us. Currently, it serves two distinct purposes:

1. Used to facilitate the simulation by taking care of discretization, interpolation, and extrapolation (a purely mathematical tool)
2. Used for post-simulation data analysis (plots, prints, and so on)

The first needs to be as fast and minimal as possible, with it being used only internally in the code.
The second benefits from it being robust at the cost of speed, and it potentially interacts with the user.

If we want to truly speed the simulation up by changing the `Function` class, some architectural changes are needed. Currently, though, I'd say that we should be focusing the `Function` class on its internal use in the simulation, and maybe all these validation functions (and `_check_user_inputs`) might not be a good implementation in the current state

rocketpy/mathutils/function.py
MateusStano: Im working on this. Should open another PR soon

rocketpy/mathutils/function.py
```diff
@@ -2067,20 +2067,15 @@ def __mul__(self, other):
         if (
             self_source_is_array
             and other_source_is_array
-            and self.__dom_dim__ == other.__dom_dim__
```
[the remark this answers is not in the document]
Gui-FernandesBR: The last condition check is already covering this.

If the dimensions are not same, the np.equal will fail already. There's no need to check this twice.

rocketpy/mathutils/function.py
```diff
@@ -2054,54 +2054,45 @@ def __mul__(self, other):
...
+            and np.array_equal(self.x_array, other.x_array)
+        ):
+            ys = self.y_array * other.y_array
+            xs = self.x_array
+            source = np.concatenate(([xs], [ys])).transpose()
+            outputs = f"({self.__outputs__[0]}*{other.__outputs__[0]})"
+            return Function(source, inputs, outputs, interp, extrap)
+        elif isinstance(other, NUMERICAL_TYPES):
+            if not self_source_is_array:
+                return Function(lambda x: (self.get_value_opt(x) * other), inputs)
```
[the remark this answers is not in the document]
Gui-FernandesBR: Since it's a lambda defined Function, there's no need to set interpolation and extrapolation methods.

rocketpy/mathutils/function.py
Gui-FernandesBR: It's actually the sane behavior vut without a if statement.

The x_interval will never be negative, it's either 0 or some positive integer value. If it's 0, it should be changed to 1.

rocketpy/mathutils/function.py
Gui-FernandesBR: It is static because of 2 reasons:

1 - It does not use class attributes
2 - This way we allow someone to call the method without instantiating the class. Simply call YourClass.your_static_method() and it already works.

Lucas-Prates:
Great PR, looking forward to working on it. Before delving into technical details, as I have to read and work on the PR itself, I would like to make a suggestion. 

To guarantee that our performance assertions are valid, should we create a unified testing benchmark for performance? Creating a testing structure whose goal is not to find bugs or anything, but rather establish performance baselines so we can compare our progress in this PR and future PRs. Some kind of structure that goes into a tests/performance/ folder and is run optionally.

To give a practical example of how it would work, we could agree upon a fixed valid set of simulations and extract basic performance metrics (e.g. mean and sd time of execution as @Gui-FernandesBR  provided) on such simulations. These metrics could then be checked (maybe even reported?) on new releases.

I do not know much about testing workflows, so maybe this suggestion does not fit inside the package test suite. However, it is still a good idea for all of us to have a unified performance benchmark/workflow.

rocketpy/simulation/flight.py
Gui-FernandesBR: For each parachute, the `noisy_pressure_signal` is defined as the sum of the clean_pressure_signal and the noise_signal itself. That's a definition, should be reasonable.

The modified code is doing the very same operation as before.

rocketpy/environment/environment.py
Gui-FernandesBR: sorry this is one small flaw of this PR: some changes I'm doing are more to improve readability, while only others are actually speeding up the code. 

I'm having a hard time trying to separate the modifications as much as possible, I appreciate reviewers patience for cases like this one.

rocketpy/mathutils/function.py
```diff
@@ -110,20 +110,19 @@ def __init__(
...
         # initialize variables to avoid errors when being called by other methods
         self.get_value_opt = None
         self.__polynomial_coefficients__ = None
         self.__akima_coefficients__ = None
         self.__spline_coefficients__ = None
+        self.EXTRAPOLATION_TYPES = {
+            "zero": 0,
+            "natural": 1,
+            "constant": 2,
+        }
```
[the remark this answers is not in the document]
Gui-FernandesBR: In this case I preferred to save it in the class because it only makes sense for the Function class, cannot be used out of it as a global variable.

Honestly, it is not necessary and I don't like it. I think it was used to compare `something == 1` instead of `something == "some_string"`, in a desperate trial of speeding up the code.

tests/test_flight.py
MateusStano: This breaking the tests by the way. The previous value works fine

Gui-FernandesBR:
Hello, thank you for the comments so far.

I still have more changes that contribute to both speed and readability. I plan to publish than here as soon as I test them.

Addressing some comments:
- @MateusStano I look forward to receiving your PR on the `_check_user_input` method. We can work together on this if you want.
- Regarding tests, I hope we don't have to change any tests results, let's see it. I wouldn't add new tests unless the coverage report warns us about modified lines not being tested.
- @Lucas-Prates I think you suggestion for creating a performance benchmark is quite valuable. Let's evaluate this idea and try to implement it.

I understand this PR may be hard to review, but I'd like to add that it is also a complicated task to separate the several speedup additions into smaller PRs, given that each modification in a method also usually requires changes in other parts of the code. I'm trying to isolate the commits as much as possible, at least.

tests/test_flight.py
Gui-FernandesBR: Can you update the value in your local environment?

I think I had to change the value here due to another change, which I didn't upload to origin  yet.

Please ignore the changes in the tests for now.

This PR is ready for review, not for merging (sorry for this joke hehehehe).

rocketpy/mathutils/function.py
MateusStano: https://github.com/RocketPy-Team/RocketPy/pull/582

Gui-FernandesBR:
Thank you for your solid review and contributions, @MateusStano .

After hours of debugging, I finally fixed the problems within the Flight class.
I believe this PR is ready for a review now. We should be merging it soon.

The Function and Flight classes were the main focus of this PR.
I see there's still an improvement opportunity in the Environment class, but let's leave it for another PR.
There are some new TODOs in the code, hope you don't mind it.

rocketpy/simulation/flight.py
```diff
@@ -1860,28 +1818,7 @@ def u_dot_generalized(self, t, u, post_processing=False):
...
-            )
-            self.density_list.append([t, self.env.density.get_value_opt(z)])
-            self.dynamic_viscosity_list.append(
-                [t, self.env.dynamic_viscosity.get_value_opt(z)]
-            )
-            self.pressure_list.append([t, self.env.pressure.get_value_opt(z)])
-            self.speed_of_sound_list.append(
-                [t, self.env.speed_of_sound.get_value_opt(z)]
-            )
+            u_dot.extend([R1, R2, R3, M1, M2, M3])
```
MateusStano, line 1821: I disagree with the way this is implemented. Changing the return of the udot methods is confusing since it needs to return only the states for its use in the integrator. 

I think appending a list here would be cleaner and faster. However, differently from the old code, it should be just one list similar to what you have done here

MateusStano (changes requested):

rocketpy/simulation/flight.py
```diff
@@ -608,38 +602,49 @@ def __init__(
...
 
         # Create known flight phases
-        self.FlightPhases = FlightPhases()
-        self.FlightPhases.add_phase(
+        self.flight_phases = self.FlightPhases()
+        self.flight_phases.add_phase(
             self.t_initial, self.initial_derivative, clear=False
         )
-        self.FlightPhases.add_phase(self.max_time)
+        self.flight_phases.add_phase(self.max_time)
```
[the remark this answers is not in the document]
Gui-FernandesBR: I agree this is a breaking change, but I believe this is for the best.

`Flight.FlightPhases` is a nested class 
`Flight.flight_phases` is an instance of the above class

I think having these two different names will make the code easier to be understood.

It's fair to say that it's not aligned with the PR name, sorry for that.

rocketpy/simulation/flight.py
```diff
@@ -651,46 +656,45 @@ def __init__(
...
-            # print('\tMin Step: ', self.min_time_step)
-            # print('\tMax Step: ', self.max_time_step)
-            # print('\tTolerances: ', self.rtol, self.atol)
+            # print(f"\n\tSolver Initialization Details:")
+            # print(f"\t\tInitial Time (t): {phase.t}")
+            # print(f"\t\tInitial State (y_sol): {self.y_sol}")
+            # print(f"\t\tTime Bound: {phase.time_bound}")
+            # print(f"\t\tMin Step: {self.min_time_step}")
+            # print(f"\t\tMax Step: {self.max_time_step}")
+            # print(f"\t\tTolerances: rtol = {self.rtol}, atol = {self.atol}")
```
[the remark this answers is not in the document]
Gui-FernandesBR: I agree with you.

I have used the code (after uncommenting) during the debugging, and after that process I re-commented again. That's why you see the new changes. I would agree to delete commented code lines.

Just one thing is to remind of issue #450 , maybe these lines will be served as inspiration when adding logging to the Flight class.

rocketpy/simulation/flight.py
```diff
@@ -2019,6 +1961,21 @@ def z(self):
...
+
+    @property
+    def apogee_y(self):
+        """Rocket y position at apogee."""
+        return self.y.get_value_opt(self.apogee_time)
+
+    @property
+    def apogee(self):
+        """Rocket z position at apogee."""
+        return self.z.get_value_opt(self.apogee_time)
```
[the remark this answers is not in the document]
Gui-FernandesBR: I was trying to reduce the number of variables being set during the simulation, in this case, removing the `self.apogee = ...`.

I didn't think about the Monte Carlo loop, but I see you're right, it will probably slow down the whole process.

I will roll this back.

rocketpy/simulation/flight.py
Gui-FernandesBR: Understood.

After my changes, every time we call the `u_dot` with `post_process=True`, the code is doing 1 extend, some replacements, but no appending. Your suggestion is to do 1 append only.

On the tip of my head I can't really tell which one is faster. Although your suggestion seems faster due to less operations,  I wonder if the extend could beat the append in terms of time.

I have to think/test, but I will probably accept your suggestion.

rocketpy/mathutils/function.py
```diff
@@ -2996,189 +2872,334 @@ def savetxt(
...
+            Value of the function evaluated at the first point.
+        yp0 : float
+            Value of the derivative of the function evaluated at the first
+            point.
+        y1 : float
+            Value of the function evaluated at the second point.
+        yp1 : float
+            Value of the derivative of the function evaluated at the second
+            point.
+
```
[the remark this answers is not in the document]
Gui-FernandesBR: Given these methods are being used in the Flight class, we would never move it to utilities.py, but maybe tools.py!

I thought about it in the beginning.

Once again, I think the Function class (and everything related to it) should be positioned in another repository. Mathematical operations, either directly related to Function or not, should be placed in that same repo.
Until that point, putting these codes as static methods of Function or isolated functions in tools.py will not make a giant difference imo.

I believe that what should be taken into consideration is the possibility of future breaking changes. Removing this from Function class in the future would be more painful than removing it from the tool module.

My concern is that the tools.py file is becoming way too large, containing functions related to several different aspects.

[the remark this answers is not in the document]
MateusStano: We could divide the `tools.py` file into several different ones. Or have a `mathtools.py` file in the `mathutils` folder would make sense too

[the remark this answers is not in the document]
Gui-FernandesBR: Fixed here: [8d09864](https://github.com/RocketPy-Team/RocketPy/pull/581/commits/8d0986490fff76726ce09aa64715427db225a0ce)

rocketpy/simulation/flight.py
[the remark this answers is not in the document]
Gui-FernandesBR: All deleted now!

rocketpy/simulation/flight.py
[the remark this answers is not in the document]
Gui-FernandesBR: Solved: [505275a](https://github.com/RocketPy-Team/RocketPy/pull/581/commits/505275a134f06e5a471b331aa021bcda74024a02)

rocketpy/mathutils/function.py
Gui-FernandesBR: Removing the np.multiply make *all* the test break.

I do not recommend us to spend much time trying to figure out it.
Closing this comment, let's leave the np.multiply then.

tests/unit/test_utilities.py
Gui-FernandesBR: I think it was due to the debugging I was doing earlier.

The change is under the error threshold so in the end it didn't affect the tests.

rocketpy/tools.py
```diff
@@ -42,6 +50,150 @@ def tuple_handler(value):
...
+    # Alternative method (same results)
+
+    # Q = (3 * a * c - b**2) / (9 * a**2)
+    # R = (9 * a * b * c - 27 * a**2 * d - 2 * b**3) / (54 * a**3)
+    # S = (R + (Q**3 + R**2)**0.5)**(1 / 3)
+    # T = (R - (Q**3 + R**2)**0.5)**(1 / 3)
+
+    # x1 = S + T - b / (3 * a)
+    # x2 = -(S + T) / 2 - b / (3 * a) + 1j * (S - T) * (3**0.5) / 2
+    # x3 = -(S + T) / 2 - b / (3 * a) - 1j * (S - T) * (3**0.5) / 2
```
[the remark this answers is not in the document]
Gui-FernandesBR: The reason to leave it commented instead of deleting is the famous: "Mmmm it took so long to implement it, maybe I can re-use in the future."

rocketpy/simulation/flight.py
```diff
@@ -823,24 +793,23 @@ def __init__(
...
-                        a = float(-(2 * y1 - yp1 * D - c * D - 2 * d) / (D**3)) + 1e-5
+                        a, b, c, d = calculate_cubic_hermite_coefficients(
+                            0,
+                            float(phase.solver.step_size),
+                            y0,
+                            yp0,
+                            y1,
+                            yp1,
+                        )
+                        a += 1e-5  # TODO: why??
```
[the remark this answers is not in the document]
Gui-FernandesBR: This makes sense. But the code is really odd.

rocketpy/tools.py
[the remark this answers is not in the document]
phmbressan: Do you think adding this alternative implementation in the docstring in a "Notes" section (e.g. "this alternative application of Cardano's formula yields the same results...") is a good alternative or unnecessary cluttering?

rocketpy/simulation/flight.py
```diff
@@ -1059,57 +979,77 @@ def __init__(
...
                                         )
 
         self.t_final = self.t
-        self._calculate_pressure_signal()
+        self.__transform_pressure_signals_lists_to_functions()
         if verbose:
-            print("Simulation Completed at Time: {:3.4f} s".format(self.t))
+            print(f"\n>>> Simulation Completed at Time: {self.t:3.4f} s")
+
+    def __calculate_and_save_pressure_signals(self, parachute, t, z):
```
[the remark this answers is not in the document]
Gui-FernandesBR: This is a private method that will  never (ever!) be used outside of Flight class, and I wanted to reinforce it by triggering the mangling system.

rocketpy/simulation/flight.py
Gui-FernandesBR: Ok no problem, 

I pushed the [2013f16](https://github.com/RocketPy-Team/RocketPy/pull/581/commits/2013f1607b23548cdc309eedda51570097086d5e) commit with a few changes here.

There are "weak breaking changes". I'm no longer using the "R1_list" atributes in the Flight class.

I believe the code is cleaner now! But a few tests had to be changed due to the old acceleration initialization.

Gui-FernandesBR:
Hey, I'm requesting new reviews from @MateusStano and @phmbressan !! 
I believe we are closer to merging this PR now.

I hope you like it.

MateusStano (approved):
Pretty good!

Question: How much is the simulation sped up now?

Gui-FernandesBR:
> Pretty good!
> 
> Question: How much is the simulation sped up now?

Running 100 simulations with the Calisto rocket:

**Before this PR:**
* Mean: 5.24839 s
* Std deviation: 1.82163 s

**After this PR:**
* Mean: 1.45942 s
* Std deviation: 0.45128 s

Overall, code much faster now! 
I even added a new if statement to force the code to use the old `u_dot` method in case the rocket motor is a SolidMotor.
I remember that we agreed with this in a previous meeting.
@MateusStano can you approve it again please?

Tests are not passing due to a problem of hdf5 in python3.8 running on macOS. We will need to discuss it this week.

MateusStano (approved):

MateusStano:
> I even added a new if statement to force the code to use the old `u_dot` method in case the rocket motor is a SolidMotor.

This might be breaking the tests btw

Gui-FernandesBR:
> > I even added a new if statement to force the code to use the old `u_dot` method in case the rocket motor is a SolidMotor.
> 
> This might be breaking the tests btw

Yes, this is exactly what is breaking these tests.

The way I see there are 2 options:
- Revert this change, let it be implemented in a future PR.
- Change the tests values to accommodate these changes. Apparently, the results from u_dot and u_dot_generalized are not identical, although they give very similar results.

I started thinking about option 2, but IMO this PR is long enough already, I see option 1 as more beneficial for now.

Gui-FernandesBR:
I believe the tests should be passing now.

@phmbressan @MateusStano I would appreciate a last approve/review so we can finally merge this one.
Moving forward I will be able to work on the Monte Carlo PR again.

phmbressan (approved):
Amazing work done here, simulations are much quicker.

I re-reviewed the PR and the whole "extend vs append" discussion and I think the implementation here pretty good.

rocketpy/simulation/flight.py
```diff
@@ -1944,20 +1846,9 @@ def u_dot_parachute(self, t, u, post_processing=False):
...
-            # Atmospheric Conditions
-            self.wind_velocity_x_list.append([t, self.env.wind_velocity_x(z)])
-            self.wind_velocity_y_list.append([t, self.env.wind_velocity_y(z)])
-            self.density_list.append([t, self.env.density(z)])
-            self.dynamic_viscosity_list.append([t, self.env.dynamic_viscosity(z)])
-            self.pressure_list.append([t, self.env.pressure(z)])
-            self.speed_of_sound_list.append([t, self.env.speed_of_sound(z)])
+            return self.__post_processed_variables.append(
+                [t, ax, ay, az, 0, 0, 0, Dx, Dy, Dz, 0, 0, 0]
+            )
```
MateusStano, line 1851:
```suggestion
            self.__post_processed_variables.append(
                [t, ax, ay, az, 0, 0, 0, Dx, Dy, Dz, 0, 0, 0]
            )
```

MateusStano (changes requested):

rocketpy/simulation/flight.py
```diff
@@ -3045,17 +2812,40 @@ def _calculate_pressure_signal(self):
...
+            current_derivative = phase.derivative
+            for callback in phase.callbacks:
+                callback(self)
+            for step in self.solution:
+                if init_time < step[0] <= final_time or (
+                    init_time == self.t_initial and step[0] == self.t_initial
+                ):
+                    current_derivative(step[0], step[1:], post_processing=True)
+
+        return np.array(self.__post_processed_variables)
```
MateusStano, line 2848: This gives wrong values for the accelerations of the rail phase.
In the old code there used to be two `retrieve_arrays` methods because the accelerations were treated differently from the forces/moments

I would suggest merging this like this though, since its pretty easy to fix with the code on #586

MateusStano: Fixing this probably makes https://github.com/RocketPy-Team/RocketPy/pull/581/commits/0eb4ae8b9c0779d232b126de58f8394771ff3435 reversable

MateusStano: And also https://github.com/RocketPy-Team/RocketPy/commit/79be71439d1f43f9d6b2a5e20a9a8eef701807d9

Gui-FernandesBR: I understand the problem now.

I can agree with a roll back here, but I wanna state that I don't think this behavior is good enough: instead of calling u_dot_rail1 with post_process=True, we are calling the u_dot_generalized with post_process=True. 

In the future I would like to see an enhancement here.

Gui-FernandesBR: Fixed here: [92e93a7](https://github.com/RocketPy-Team/RocketPy/pull/581/commits/92e93a7f616f9c7a02b281fba09418e96f387b61)

Gui-FernandesBR:
Hello, I'd like to request re-review @MateusStano or @phmbressan .

Everything seems to be working correctly now. Tests were only slightly changed in this PR, returned values are practically identical.

MateusStano (approved):

PR #582, opened 2024-03-28
ENH: Function Validation Rework & Swap `np.searchsorted` to `bisect_left` 

MateusStano:
## Pull request type

## Checklist

## Current behavior

1. The input parameters of the Funciton class are all checked in a very disorganized way. This would lead to redundant checks and overall slowing down of the class
2. all interpolations used in `get_value_opt` were using `np.searchsorted`, which was slow

## New behavior

1. Validation has been reworked as to follow the order: `source` -> `interpolation` and `extrapolation` -> `inputs`, `outputs` and `title`. The `get_value_opt` function is now modularized  and is defined according to the current `interpolation` and `extrapolation`
2. `bisect_left` is now used instead of `np.searchsorted`. This cuts the execution time by half in a `u_dot_generalized` call.

The `Function` class should behave exactly the same as before.

## Breaking change

- [ ] Yes
- [x] No

## Additional information

- There is an attribute called `__img_dim__` that was never used and was always equal to 1. I changed it so its defined at the `__init__` for backward compatibility purposes, but it is still never used
- Further optimizations can be done in `__mul__`, `__sub__`, `__truediv__` and other operations by avoiding going through all the validation checks when re-initializing the Functions

### DIFF
--- a/rocketpy/mathutils/function.py
+++ b/rocketpy/mathutils/function.py
@@ -20,6 +20,14 @@ except ImportError:
     from ..tools import cached_property
 
 NUMERICAL_TYPES = (float, int, complex, np.ndarray, np.integer, np.floating)
+INTERPOLATION_TYPES = {
+    "linear": 0,
+    "polynomial": 1,
+    "akima": 2,
+    "spline": 3,
+    "shepard": 4,
+}
+EXTRAPOLATION_TYPES = {"zero": 0, "natural": 1, "constant": 2}
 
 
 class Function:
@@ -110,27 +118,25 @@ class Function:
         (II) Fields in CSV files may be enclosed in double quotes. If fields
         are not quoted, double quotes should not appear inside them.
         """
-
-        inputs, outputs = Function._validate_inputs_outputs(inputs, outputs)
-
         # initialize variables to avoid errors when being called by other methods
-        self.get_value_opt = None
         self.__polynomial_coefficients__ = None
         self.__akima_coefficients__ = None
         self.__spline_coefficients__ = None
-        self.EXTRAPOLATION_TYPES = {
-            "zero": 0,
-            "natural": 1,
-            "constant": 2,
-        }
 
-        # store variables
-        self.set_inputs(inputs)
-        self.set_outputs(outputs)
+        # initialize parameters
+        self.source = source
+        self.__inputs__ = inputs
+        self.__outputs__ = outputs
         self.__interpolation__ = interpolation
         self.__extrapolation__ = extrapolation
-        self.set_source(source)
-        self.set_title(title)
+        self.title = title
+        self.__img_dim__ = 1  # always 1, here for backwards compatibility
+
+        # args must be passed from self.
+        self.set_source(self.source)
+        self.set_inputs(self.__inputs__)
+        self.set_outputs(self.__outputs__)
+        self.set_title(self.title)
 
     # Define all set methods
     def set_inputs(self, inputs):
@@ -145,8 +151,7 @@ class Function:
         -------
         self : Function
         """
-        self.__inputs__ = [inputs] if isinstance(inputs, str) else list(inputs)
-        self.__dom_dim__ = len(self.__inputs__)
+        self.__inputs__ = self._validate_inputs(inputs)
         return self
 
     def set_outputs(self, outputs):
@@ -161,8 +166,7 @@ class Function:
         -------
         self : Function
         """
-        self.__outputs__ = [outputs] if isinstance(outputs, str) else list(outputs)
-        self.__img_dim__ = len(self.__outputs__)
+        self.__outputs__ = self._validate_outputs(outputs)
         return self
 
     def set_source(self, source):
@@ -211,20 +215,10 @@ class Function:
         self : Function
             Returns the Function instance.
         """
-        source, inputs, outputs, interpolation, extrapolation = self._check_user_input(
-            source,
-            self.__inputs__,
-            self.__outputs__,
-            self.__interpolation__,
-            self.__extrapolation__,
-        )
-        # updates inputs and outputs (could be modified due to csv headers)
-        self.set_inputs(inputs)
-        self.set_outputs(outputs)
+        source = self._validate_source(source)
 
         # Handle callable source or number source
         if callable(source):
-            self.source = source
             self.get_value_opt = source
             self.__interpolation__ = None
             self.__extrapolation__ = None
@@ -232,13 +226,16 @@ class Function:
             # Set arguments name and domain dimensions
             parameters = signature(source).parameters
             self.__dom_dim__ = len(parameters)
-            if self.__inputs__ == ["Scalar"]:
+            if self.__inputs__ is None:
                 self.__inputs__ = list(parameters)
 
         # Handle ndarray source
         else:
+            # Evaluate dimension
+            self.__dom_dim__ = source.shape[1] - 1
+
             # Check to see if dimensions match incoming data set
-            new_total_dim = len(source[0, :])
+            new_total_dim = source.shape[1]
             old_total_dim = self.__dom_dim__ + self.__img_dim__
 
             # If they don't, update default values or throw error
@@ -250,20 +247,25 @@ class Function:
             # if Function is 1D, sort source by x. If 2D, set z
             if self.__dom_dim__ == 1:
                 source = source[source[:, 0].argsort()]
+                self.x_array = source[:, 0]
+                self.x_initial, self.x_final = self.x_array[0], self.x_array[-1]
+                self.y_array = source[:, 1]
+                self.y_initial, self.y_final = self.y_array[0], self.y_array[-1]
+                self.get_value_opt = self._get_value_opt_1d
             elif self.__dom_dim__ == 2:
+                self.x_array = source[:, 0]
+                self.x_initial, self.x_final = self.x_array[0], self.x_array[-1]
+                self.y_array = source[:, 1]
+                self.y_initial, self.y_final = self.y_array[0], self.y_array[-1]
                 self.z_array = source[:, 2]
                 self.z_initial, self.z_final = self.z_array[0], self.z_array[-1]
+                self.get_value_opt = self._get_value_opt_nd
+            else:
+                self.get_value_opt = self._get_value_opt_nd
 
-            # Set x and y arrays (common for 1D or multivariate)
-            self.x_array = source[:, 0]
-            self.x_initial, self.x_final = self.x_array[0], self.x_array[-1]
-            self.y_array = source[:, 1]
-            self.y_initial, self.y_final = self.y_array[0], self.y_array[-1]
-
-            # Finally set source, update extrapolation and interpolation
-            self.source = source
-            self.__extrapolation__ = extrapolation  # to avoid calling set_get_value_opt
-            self.set_interpolation(interpolation)
+        self.source = source
+        self.set_interpolation(self.__interpolation__)
+        self.set_extrapolation(self.__extrapolation__)
         return self
 
     @cached_property
@@ -303,18 +305,27 @@ class Function:
         -------
         self : Function
         """
-        self.__interpolation__ = method
+        if not callable(self.source):
+            self.__interpolation__ = self._validate_interpolation(method)
+            self._update_interpolation_coefficients(self.__interpolation__)
+            self._set_interpolation_function()
+        return self
+
+    def _update_interpolation_coefficients(self, method):
+        """Update interpolation coefficients for the given method."""
         # Spline, akima and polynomial need data processing
         # Shepard, and linear do not
-        if method == "spline":
-            self.__interpolate_spline__()
-        elif method == "polynomial":
+        if method == "polynomial":
             self.__interpolate_polynomial__()
+            self._coeffs = self.__polynomial_coefficients__
         elif method == "akima":
             self.__interpolate_akima__()
-
-        self.set_get_value_opt()
-        return self
+            self._coeffs = self.__akima_coefficients__
+        elif method == "spline" or method is None:
+            self.__interpolate_spline__()
+            self._coeffs = self.__spline_coefficients__
+        else:
+            self._coeffs = []
 
     def set_extrapolation(self, method="constant"):
         """Set extrapolation behavior of data set.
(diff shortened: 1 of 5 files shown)


tests/test_flight.py
```diff
@@ -729,7 +729,7 @@ def test_velocities(flight_calisto_custom_wind, flight_time, expected_values):
         ("t_initial", (1.6542528, 0.65918, -0.067107)),
         ("out_of_rail_time", (5.05334, 2.01364, -1.7541)),
         ("apogee_time", (2.35291, -1.8275, -0.87851)),
-        ("t_final", (0, 0, 159.3292416824044)),
```
MateusStano, line 732: This change was a rollback from https://github.com/RocketPy-Team/RocketPy/commit/9a637a834e443736e7fa988830d1586aaf1f9201

rocketpy/mathutils/function.py
```diff
@@ -13,13 +13,22 @@
 import matplotlib.pyplot as plt
 import numpy as np
 from scipy import integrate, linalg, optimize
+from bisect import bisect_left
```
Gui-FernandesBR, line 16: The bisect, a native python function should be imported before 3rd party libraries. That's for convention and readability.

Do you run `isort` before submitting a PR?
You can do that by running `make isort` in your terminal, our simply `isort .`

rocketpy/mathutils/function.py
```diff
@@ -13,13 +13,22 @@
...
 
 NUMERICAL_TYPES = (float, int, complex, np.ndarray, np.integer, np.floating)
+INTERPOLATION_TYPES = {
+    "linear": 0,
+    "polynomial": 1,
+    "akima": 2,
+    "spline": 3,
+    "shepard": 4,
+}
+EXTRAPOLATION_TYPES = {"zero": 0, "natural": 1, "constant": 2}
```
Gui-FernandesBR, line 31: When you asked before I said I preferred these dictionaries to be save within the Function class since they will not (probably) be used in other contexts.

HOWEVER, now I think it's a good idea to do what you did here, because these constants will be defined only once in te execution time. If we save them as class attributes, they would be defined again and again every time we initiate a Function.

rocketpy/mathutils/function.py
```diff
@@ -110,27 +119,25 @@ def __init__(
...
         are not quoted, double quotes should not appear inside them.
         """
-
-        inputs, outputs = Function._validate_inputs_outputs(inputs, outputs)
-
         # initialize variables to avoid errors when being called by other methods
-        self.get_value_opt = None
         self.__polynomial_coefficients__ = None
         self.__akima_coefficients__ = None
         self.__spline_coefficients__ = None
```
Gui-FernandesBR, line 125: Do you think we still need these initializations here?
What if we simply remove them?

rocketpy/mathutils/function.py
```diff
@@ -110,27 +119,25 @@ def __init__(
...
+        self.__inputs__ = inputs
+        self.__outputs__ = outputs
         self.__interpolation__ = interpolation
         self.__extrapolation__ = extrapolation
-        self.set_source(source)
-        self.set_title(title)
+        self.title = title
+        self.__img_dim__ = 1  # always 1, here for backwards compatibility
+
+        # args must be passed from self.
```
Gui-FernandesBR, line 136: why?

rocketpy/mathutils/function.py
```diff
@@ -250,20 +248,25 @@ def set_source(self, source):
             # if Function is 1D, sort source by x. If 2D, set z
```
Gui-FernandesBR, line 248: This comment is a bit ambiguous. 

The correct sentence should be: "set x and y. If Function is 2D, also set z."

Gui-FernandesBR: Reading the code again...

If Function is 5D, for instance, the x and y array won't be set. Is that expected?

rocketpy/mathutils/function.py
```diff
@@ -333,142 +345,173 @@ def set_extrapolation(self, method="constant"):
...
+        the attribute _interpolation_func."""
+        interpolation = INTERPOLATION_TYPES[self.__interpolation__]
+        if interpolation == 0:  # linear
+
+            def linear_interpolation(x, x_min, x_max, x_data, y_data, coeffs):
+                x_interval = bisect_left(x_data, x)
+                x_left = x_data[x_interval - 1]
+                y_left = y_data[x_interval - 1]
+                dx = float(x_data[x_interval] - x_left)
+                dy = float(y_data[x_interval] - y_left)
```
Gui-FernandesBR, line 367: Do we really need to cast the calculations here? Can we just assume the inputs will always be numbers? I know it may not impact performance that much, but I would avoid redundant operations anyways.

```suggestion
                dx = x_data[x_interval] - x_left
                dy = y_data[x_interval] - y_left
```

rocketpy/mathutils/function.py
```diff
@@ -333,142 +345,173 @@ def set_extrapolation(self, method="constant"):
...
-                            x = x - x_data[0]
-                        else:
-                            a = coeffs[:, -1]
-                            x = x - x_data[-2]
-                        y = a[3] * x**3 + a[2] * x**2 + a[1] * x + a[0]
-                    else:  # Extrapolation is set to constant
-                        y = y_data[0] if x < x_min else y_data[-1]
+            def polynomial_interpolation(x, x_min, x_max, x_data, y_data, coeffs):
+                y = np.sum(coeffs * x ** np.arange(len(coeffs)))
+                return y
```
Gui-FernandesBR, line 377: Use one line instead of two. Faster to read, both to users and interpreter. Also avoids creating another variable in memory.

```suggestion
            def polynomial_interpolation(x, x_min, x_max, x_data, y_data, coeffs):
                return np.sum(coeffs * x ** np.arange(len(coeffs)))
```

rocketpy/mathutils/function.py
```diff
@@ -333,142 +345,173 @@ def set_extrapolation(self, method="constant"):
...
+
+            self._interpolation_func = polynomial_interpolation
+
+        elif interpolation == 2:  # akima
+
+            def akima_interpolation(x, x_min, x_max, x_data, y_data, coeffs):
+                x_interval = bisect_left(x_data, x)
+                x_interval = x_interval if x_interval != 0 else 1
```
Gui-FernandesBR, line 385: I wonder if the following suggestion of comparison would run faster.
I understand that x_interval can never be negative, it is either 0 or positive integer.

```suggestion
                x_interval = x_interval if x_interval > 0 else 1
```

rocketpy/mathutils/function.py
```diff
@@ -333,142 +345,173 @@ def set_extrapolation(self, method="constant"):
...
+                a = coeffs[4 * x_interval - 4 : 4 * x_interval]
+                y = a[3] * x**3 + a[2] * x**2 + a[1] * x + a[0]
                 return y
```
Gui-FernandesBR, line 388: Same as previous comment

```suggestion
                return a[3] * x**3 + a[2] * x**2 + a[1] * x + a[0]
```

rocketpy/mathutils/function.py
```diff
@@ -333,142 +345,173 @@ def set_extrapolation(self, method="constant"):
...
+
+        elif interpolation == 3:  # spline
+
+            def spline_interpolation(x, x_min, x_max, x_data, y_data, coeffs):
+                x_interval = bisect_left(x_data, x)
+                x_interval = max(x_interval, 1)
+                a = coeffs[:, x_interval - 1]
+                x = x - x_data[x_interval - 1]
+                y = a[3] * x**3 + a[2] * x**2 + a[1] * x + a[0]
+                return y
```
Gui-FernandesBR, line 400: suggestion, same reason as above:

```suggestion
                return a[3] * x**3 + a[2] * x**2 + a[1] * x + a[0]
```

rocketpy/mathutils/function.py
```diff
@@ -333,142 +345,173 @@ def set_extrapolation(self, method="constant"):
...
-                return y
+                def natural_extrapolation(x, x_min, x_max, x_data, y_data, coeffs):
+                    if x < x_min:
+                        a = coeffs[:, 0]
+                        x = x - x_data[0]
+                    else:
+                        a = coeffs[:, -1]
+                        x = x - x_data[-2]
+                    y = a[3] * x**3 + a[2] * x**2 + a[1] * x + a[0]
+                    return y
```
Gui-FernandesBR, line 458:
```suggestion
                    return a[3] * x**3 + a[2] * x**2 + a[1] * x + a[0]
```

rocketpy/mathutils/function.py
```diff
@@ -333,142 +345,173 @@ def set_extrapolation(self, method="constant"):
...
+        -------
+        y : scalar
+            Value of the Function at the specified point.
+        """
+        # Retrieve general info
+        x_data = self.x_array
+        y_data = self.y_array
+        x_min, x_max = self.x_initial, self.x_final
+        coeffs = self._coeffs
+        if x_min <= x <= x_max:
```
Gui-FernandesBR, line 502: Are you sure we should use `<=` in both the interval's extremes?
I was under impression at least the right side would receive a `<` instead.

I will have to think more about this point.

rocketpy/mathutils/function.py
```diff
@@ -333,142 +345,173 @@ def set_extrapolation(self, method="constant"):
...
+        else:
+            y = self._extrapolation_func(x, x_min, x_max, x_data, y_data, coeffs)
+        return y
+
+    def _get_value_opt_nd(self, *args):
+        """Evaluate the Function at a single point (x, y, z). This method is
+        used when the Function is N-D."""
+        # always use shepard for N-D functions
+        y = self.__interpolate_shepard__(list(args))
+        return y
```
Gui-FernandesBR, line 513: Quick suggestion: 

```suggestion
    def _get_value_opt_nd(self, *args):
        """Evaluate the Function at a single point (x, y, z). This method is
        used when the Function is N-D."""
        # always use shepard for N-D functions
        return self.__interpolate_shepard__(list(args))
```

Btw do we really need to call list(args)? Can't we just pass `*args` instead?

Gui-FernandesBR:
@MateusStano I made all the commentaries that I found important to be done in this first review.
Overall, code is much better and intuitive. Let's follow the discussions in the next days so we can merge this PR really soon.

Gui-FernandesBR:
Also, @MateusStano  can we make all the new methods completely private (i.e. use `def __your_private_method`) instead of using a single underscore? 

These new methods won't be used out of the Function class, plus using double score seems better, it allows python to mangle the methods.

rocketpy/mathutils/function.py
phmbressan: Both solutions have specific CPU opcodes and translate directly into really fast hardware evaluations. Therefore both of them are really fast and the preference should be either by correctness (should it be different or greater than zero?) or readability.

Of course we are just discussing it for the sake of discussing, but these are absolutely not the bottleneck here.

rocketpy/mathutils/function.py
MateusStano: Seems that removing them does not affect the code. Removed them in https://github.com/RocketPy-Team/RocketPy/pull/582/commits/7cd270b956fa020305a6a6bdc45d52cab7283ce7

rocketpy/mathutils/function.py
MateusStano: The `_check_user_input` had a check to guarantee that `len(outputs)==1`. So this would always end up being one.

rocketpy/mathutils/function.py
MateusStano: I'll change it so it keeps consistency with old code. Done in https://github.com/RocketPy-Team/RocketPy/pull/582/commits/0bd6f6396e8716fb0c76025dd2e29380528cfa0d

rocketpy/mathutils/function.py
MateusStano: I fell like that is something added to avoid some typing problems with numpy. Changing this might create some bugs

rocketpy/mathutils/function.py
MateusStano: It is either this or we could do:

```python
if x_min == x or x == x_max:
    return x
elif x_min < x < x_max:
    interpolate()
else:
    extrapolate()
```

rocketpy/mathutils/function.py
```diff
@@ -2969,86 +2946,78 @@ def _check_user_input(
...
+            The validated inputs parameter.
+        """
+        if self.__dom_dim__ == 1:
+            if inputs is None:
+                return ["Scalar"]
+            if isinstance(inputs, str):
+                return [inputs]
+            if isinstance(inputs, list):
+                if len(inputs) == 1 and isinstance(inputs[0], str):
+                    return inputs
```
phmbressan, line 2975: This will not accept `tuples` I think.

Personally, I would not go that strict into checking if the inner type of the Iterable is actually a string. This is the type of check that I do think that is more idiomatic to duck type.

rocketpy/mathutils/function.py
```diff
@@ -333,141 +338,163 @@ def set_extrapolation(self, method="constant"):
...
-            def get_value_opt_multiple(*args):
-                return self.__interpolate_shepard__(args)
-
-            get_value_opt = get_value_opt_multiple
+        coeffs = self._coeffs
+        if x_min <= x <= x_max:
+            y = self._interpolation_func(x, x_min, x_max, x_data, y_data, coeffs)
+        else:
+            y = self._extrapolation_func(x, x_min, x_max, x_data, y_data, coeffs)
+        return y
```
phmbressan, line 491: Excellent. When thinking about `Function` class refactors, one of the goals was making this method look like this.

rocketpy/mathutils/function.py
phmbressan: I agree with Stanos implementation, since the interpolation should keep all source points untouched, the x-extrema should be "interpolated".

Gui-FernandesBR:
@MateusStano good work solving the majority of commentaries!

Could you add tests for the lines missing coverage?

Gui-FernandesBR:
Also, @MateusStano could you please add your PR to the CHANGELOG?

rocketpy/mathutils/function.py
```diff
@@ -20,6 +21,14 @@
...
 
 NUMERICAL_TYPES = (float, int, complex, np.ndarray, np.integer, np.floating)
+INTERPOLATION_TYPES = {
+    "linear": 0,
+    "polynomial": 1,
+    "akima": 2,
+    "spline": 3,
+    "shepard": 4,
+}
+EXTRAPOLATION_TYPES = {"zero": 0, "natural": 1, "constant": 2}
```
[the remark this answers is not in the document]
phmbressan: I mentioned defining the `dict` at the end of file just so as to not create it everytime on `__init__`. It is indeed an odd approach to squeeze the last bit of performance. The conventional way would be to dispatch the method from a instance bound hashtable:

```py
self.interpolation_map[self.interpolation]()
```

The `match` structure is generally faster than `ifs`, but, as you mentioned, we cannot use it yet. Nonetheless, I still think that both the `ifs` and the `match` solution with the numbers are a workaround that should be a direct dispatch.

There is no need to get the method number that `if` or `match` it to the correct method. You already know which method it the correct one.

On top of that, the major bottleneck of the `Function` are the `lambda` preference on arithmetic and the lack of vectorization. Therefore this discussion is more to think and brainstorm if we need the interpolation type mapping numbers, than an actual performance critical suggestion.

PR #585, opened 2024-04-11
BUG: Swap `rocket.total_mass.differentiate` for `motor.total_mass_flow rate`

MateusStano:
## Pull request type

## Checklist

## Description

If a simulation is using the new EOMs and has a time node at the motor burn-out time, an outlier acceleration value might be calculated:

image

This is due to the second-order differentiation of `rocket.total_mass` blowing up due to numerical errors. Swapping this to `motor.total_mass_flow_rate` gives the following:

image

This makes `u_dot_generalized` more similar to the old udot

## Breaking change

- [ ] Yes
- [x] No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- BUG: Swap rocket.total_mass.differentiate for motor.total_mass_flow rate [#585](https://github.com/RocketPy-Team/RocketPy/pull/585)
 - BUG: export_eng 'Motor' method would not work for liquid motors. [#559](https://github.com/RocketPy-Team/RocketPy/pull/559) 
 
 ## [v1.2.2] - 2024-03-22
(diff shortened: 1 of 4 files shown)


Gui-FernandesBR:
Nice one.

Let me make a question, why not creating a `Rocket.total_mass_flow_rate` in the Rocket class and then accessing it in the `u_dot_generalized`? .

This would allow for a better readability in the flight class, plus it would allow us to access the mass flow rate directly from the rocket object.

Gui-FernandesBR:
More questions:
* Do you agree that this PR solves a bug? I believe the current behavior is a bug, should not be working that way.
* Could you please add your PR to the CHANGELOG.md?

MateusStano:
> Let me make a question, why not creating a `Rocket.total_mass_flow_rate` in the Rocket class and then accessing it in the `u_dot_generalized`? .

Sounds like something nice to have. Added it in https://github.com/RocketPy-Team/RocketPy/pull/585/commits/8ee2c95f6961e94f93ef569303588c643647778f

Gui-FernandesBR (approved):
LGTM

PR #586, opened 2024-04-15
BUG: Air Brakes Deployment Constant at Post Process

MateusStano:
## Pull request type

## Checklist

## Current behavior

The calculations of post-processed variable (i.e. accelerations, forces and moments) are considering only the last state of the air brakes, instead of reconsidering its deployment level.

## New behavior

Fixing this issue was quite challenging since we wanted to maintain the freedom of the `Controller` class without big/breaking changes. The complexity here has to do with the fact that whatever object is being controlled (in this case the air brakes) has a state that varies throughout the simulation. There are two ways to go about dealing with this:

1. Save the control inputs at every control function call
    - Requires that we save exactly what we need from each possible controllable object.
    - For the air brakes, we need only the deployment level.
    - This would force us to have clearly defined what is controllable and what is not, and we would need a way for the user to specify what he is controlling and what he is not. This would be a different approach from the current one, where anything is controllable inside the controller function.
2. Save accelerations and forces during simulation (run post-processing during simulation)
    - This increases simulation time by one udot call plus the time to append values to all necessary arrays per time step
    - There is no need to do post process after simulation because everything calculated is a `@cached_property`

The second option was implemented. The following was changed on the code:

- A `_reset` method was added to the `AirBrakes` class that resets the deployment level to its initial value. This guarantees that running the air brakes will behave equally if two `Flight` initializations are run consecutively
- A warning was added for when a simulation with controllers is run without `time_overshoot=False`
- A `__retrieve_arrays` method was created that unites `retrieve_acceleration_arrays` and `retrieve_temporary_values_arrays` into only one loop and optimizes it. This was done so that the in-simulation post-processing for controllers could be done in only one udot call.

## Breaking change

- [ ] Yes
- [x] No

## Additional information

One thing was changed that changes simulation results. The accelerations used to always be initialized with.

        ax, ay, az = [[0, 0]], [[0, 0]], [[0, 0]]
        alpha1, alpha2, alpha3 = [[0, 0]], [[0, 0]], [[0, 0]]

I do not know what this was for and it seems wrong at a first glance. I changed it to be initialized with empty lists instead and all tests are passing.

The initialization with `[[0, 0]]` was introduced in #295, commit https://github.com/RocketPy-Team/RocketPy/pull/295/commits/291d479124bba13599040cd7a673d41cf3f8f721. Maybe @Gui-FernandesBR knows what this was for?


rocketpy/simulation/flight.py
```diff
@@ -607,6 +606,20 @@ def __init__(
...
+                warnings.warn(
+                    "time_overshoot is set to True, but controllers are present. "
+                    "Controllers will not work properly. "
+                    "Consider setting time_overshoot=False in the Flight object "
+                    "to use controllers."
+                )
+            # reset controllable object to initial state (only airbrakes for now)
+            for air_brakes in self.rocket.air_brakes:
+                air_brakes._reset()
+
```
Gui-FernandesBR, line 622: This code block (from "# Controller initialization" to "air_brakes._reset") is too long. To simplify the `__init__` method, I would create a new `__init_controllers` method and invoke it right here.

Also, regarding your message "Controllers will not work properly", I understand that the controllers will not work at all if you set time_overshoot to true. Is that correct?

What is better:
1. Throwing a warning that the simulation will not work as expected, but run the simulation anyways. 
2. Throwing a warning that the time_overshoot is being changed to `False` and then the simulation?

rocketpy/simulation/flight.py
```diff
@@ -1296,6 +1347,17 @@ def udot_rail1(self, t, u, post_processing=False):
...
+            # Use u_dot post processing code for forces, moments and env data
+            self.u_dot_generalized(t, u, post_processing=True)
+            # Save feasible accelerations
+            self.ax_list[-1] = [t, ax]
+            self.ay_list[-1] = [t, ay]
+            self.az_list[-1] = [t, az]
+            self.alpha1_list[-1] = [t, 0]
+            self.alpha2_list[-1] = [t, 0]
+            self.alpha3_list[-1] = [t, 0]
+
```
Gui-FernandesBR, line 1360: What about other lists? R1_list, R2_list, R3_list, M1_list, M2_list, M3_list, pressure_list, density_list, dynamic_viscosity_list,speed_of_sound_list,wind_velocity_x_list, wind_velocity_y_list

I understand the code might be working, but these vectors will have different lengths. Could that be a problem somehow? Did you think about it?

Gui-FernandesBR: Btw in the future the atmospheric properties will no longer be calculated by post-process. Since we only need the z coordinate use the environment class to calculate all the atmospheric properties, we can be creating different cached_property instead of using appending() within the simulation loop.

rocketpy/rocket/aero_surface.py
```diff
@@ -2023,6 +2025,12 @@ def deployment_level(self, value):
                 )
         self._deployment_level = value
 
+    def _reset(self):
+        """Resets the air brakes to their initial state. This is ran at the
```
Gui-FernandesBR, line 2029: Why did you make this method (weakly) private? 
Are you concerned that we eventully introduces breaking changes here? Or do you want to provent users from invoking this method (ever)?

If the answer is the second option, I would suggest to put a second leading underscore to the name of the method.

rocketpy/simulation/flight.py
```diff
@@ -1089,6 +1108,25 @@ def __init_post_process_variables(self):
...
+        self.R3_list = []
+        self.M1_list = []
+        self.M2_list = []
+        self.M3_list = []
+        self.pressure_list = []
+        self.density_list = []
+        self.dynamic_viscosity_list = []
+        self.speed_of_sound_list = []
+        self.wind_velocity_x_list = []
+        self.wind_velocity_y_list = []
```
Gui-FernandesBR, line 1129: It's said that all of this exists... But it is even worse that they are not private attributes.

Gui-FernandesBR (commented):
Good PR, thanks for the implementations.

I've added some comments, and I'm open to discuss all of them before approving your PR.

Regarding the initialization of acceleration arrays with zeros, I believe it was done to avoid errors when initializing a flight using another flight as the initial solution. 
Could you check if all the examples in the examples page are running correctly after you PR?

Also, sorry for being the CHANGELOG guru again, but could you please add your PR to the list? It should become an habit to us.

Gui-FernandesBR (commented):
Trying to build the docs here... look what I've got:
```
File ~\Documents\github\RocketPy\rocketpy\mathutils\function.py:3388, in funcify_method.<locals>.funcify_method_decorator.__get__(self, instance, owner)
   3384 except KeyError:
   3385     # If cache is not ready, create it
   3386     try:
   3387         # Handle methods which return Function instances
-> 3388         val = self.func(instance).reset(*args, **kwargs)
   3389     except AttributeError:
   3390         # Handle methods which return a valid source
   3391         source = self.func(instance)

File ~\Documents\github\RocketPy\rocketpy\simulation\flight.py:2449, in Flight.stream_velocity_x(self)
   2445 @funcify_method("Time (s)", "Freestream Velocity X (m/s)", "spline", "constant")
   2446 def stream_velocity_x(self):
   2447     """Freestream velocity X component as a Function of time."""
   2448     stream_velocity_x = np.column_stack(
-> 2449         (self.time, self.wind_velocity_x[:, 1] - self.vx[:, 1])
   2450     )
   2451     return stream_velocity_x

ValueError: operands could not be broadcast together with shapes (2066,) (2067,)
```

rocketpy/simulation/flight.py
MateusStano: I prefer the second option. However, is forcing a parameter to be something a good practice?
In this case it seems to make sense, but maybe it is worth the discussion.

rocketpy/rocket/aero_surface.py
MateusStano: Yes, I was trying to avoid breaking changes. I did not want to make it double score because it gets used in other classes.
If you think double is better I could change it no problem.

rocketpy/simulation/flight.py
MateusStano: Agreed

rocketpy/simulation/flight.py
MateusStano: > What about other lists? R1_list, R2_list, R3_list, M1_list, M2_list, M3_list, pressure_list, density_list, dynamic_viscosity_list,speed_of_sound_list,wind_velocity_x_list, wind_velocity_y_list

They are appended to inside `u_dot_generalized` with the correct values since it has `post_processing=True`. The accelerations are also append to there, but those values are incorrect since it is not considering the launch rail reaction forces. The correct acceleration values comes from `udot_rail1`, and so the last item in the acceleration arrays are overwritten.

The vectors end up being the same length

MateusStano:
> Trying to build the docs here... look what I've got:

This seems to be due to `terminate_on_apogee` making the simulation stop before calling the extra udot with post processing. I will fix it soon

rocketpy/simulation/flight.py
Gui-FernandesBR: It is not a good practice if you don't let the user know it. But since we are throwing an warning, I believe it is more than OK.

Gui-FernandesBR:
This PR may present a conflict with the #581 , given that both deal with the postprocessing methods.
I think we might need to merge some ideas here.

MateusStano:
@Gui-FernandesBR ready for a new review!


> This PR may present a conflict with the #581 , given that both deal with the postprocessing methods.
> I think we might need to merge some ideas here.

The best option is to merge your PR first, and then I can change what is necessary here. Its not a big conflict

Gui-FernandesBR:
@MateusStano please let me know when this gets ready for a re-review

MateusStano:
@Gui-FernandesBR its ready

Gui-FernandesBR:
> @Gui-FernandesBR its ready

Ok, I want to test it locally before approving, so I will take a little longer time to review it.

rocketpy/simulation/flight.py
```diff
@@ -1120,6 +1134,23 @@ def __init_equations_of_motion(self):
...
+        self._controllers = self.rocket._controllers[:]
+        if self._controllers:
+            if self.time_overshoot == True:
+                self.time_overshoot = False
+                warnings.warn(
+                    "time_overshoot has been set to False due to the presence of controllers. "
+                )
+            # reset controllable object to initial state (only airbrakes for now)
+            for air_brakes in self.rocket.air_brakes:
+                air_brakes._reset()
```
phmbressan, line 1148: Right now I am not completely against this implementation, since this is a bugfix. However, this pose for us some scalability issues that we have tried to avoid while implementing the controllers.

This code block requires that the `Flight` knows everything that is controllable in a "hardcoded" manner.

Ideally, we would request to the controller of those objects to `reset` them. This resonates quite well with "real life" IMHO.

@MateusStano just to raise a bit of discussion, would it work implementing a `reset` method in the `Controller` class that calls the `controller_function` with `initial_observed_variables` to reset everything this controller controls?

I understand this might have unintended repercutions, since the `controller_function` is user implemented, but I would like your opinion on this because it would bring great architectural benefits in the future.

rocketpy/simulation/flight.py
```diff
@@ -1120,6 +1134,23 @@ def __init_equations_of_motion(self):
...
+
+    def __cache_post_process_variables(self):
```
Gui-FernandesBR, line 1150: What is the benefit of tis `__cache_post_process_variables` method?

Wouldn't it get to the same result if we simply called "np.array(self.__post_processed_variables)" directly?

I'm sorry but I had a hard time understanding all these caches and variables storages.

rocketpy/simulation/flight.py
MateusStano: > This code block requires that the `Flight` knows everything that is controllable in a "hardcoded" manner.

This particular point is not an issue I think. When adding controllable objects to the Rocket class we would need to keep track of it there. This `for air_brakes in self.rocket.air_brakes` would just become `for c in self.rocket.controllables` or something similar.

> , would it work implementing a `reset` method in the `Controller` class that calls the `controller_function` with `initial_observed_variables` to reset everything this controller controls?

No it wouldn't. We can't call the controller function because the user could have dynamic variables that change at every call, and having an extra call not related to the simulation would be confusing I think


> Ideally, we would request to the controller of those objects to `reset` them. This resonates quite well with "real life" IMHO.

This makes more sense though. I think maybe the `._reset` method should actually be in the controller class, and the controller should receive the initial values and what params it has to reset. I think that should be part of the changes coming to the controller class, to support controlling everything besides just air brakes. That change would require deleting the `AirBrakes._reset` method though. Do you think it is okay leaving the `_reset` method in just for this bug fix? It would be a lot of work to add the other way right now

rocketpy/simulation/flight.py
MateusStano: You are right. Right now it changes nothing. I think it was needed before the PR refactoring the post-process

fixed https://github.com/RocketPy-Team/RocketPy/pull/586/commits/464f2017c46367857582b33b81c7379d4d39cd80

rocketpy/simulation/flight.py
phmbressan: Let's keep the way it is right now. In future PR we implement the `reset` on the controller.

rocketpy/simulation/flight.py
```diff
@@ -996,8 +998,15 @@ def __simulate(self, verbose):
...
 
+                    # If controlled flight, post process must be done on sim time
+                    if self._controllers:
+                        phase.derivative(self.t, self.y_sol, post_processing=True)
+
         self.t_final = self.t
         self.__transform_pressure_signals_lists_to_functions()
+        if self._controllers:
+            # cache post process variables
+            self.__evaluate_post_process = np.array(self.__post_processed_variables)
```
Gui-FernandesBR, line 1009: I know this question may sound you strange, but are you sure this is working?

I think the `Flight.__evaluate_post_process` is a property, so you should not be able to set its value. How is this working? 

Probably if you remove the leading underscores the whole code would break.

```suggestion
            self.__evaluate_post_process = np.array(self.__post_processed_variables)
```

MateusStano: This makes it so the code inside `def __evaluate_post_process()` is skipped whenever that property is accessed anywhere else in the code. This is probably because of the `@cached_property` decorator. I don't know much else about it, but it works

Gui-FernandesBR: Ok, it seems like the cached decorator does not forbids us from settings.

Gui-FernandesBR (approved):

PR #588, opened 2024-04-19
ENH: Exponential backoff decorator (fix #449)

### LINKED ISSUE #449: ENH: Exponential backoff retry strategy

Currently the Environment class practices a bad retry by iterating inside a while loop to catch API response within a certain number of maximum attempts. This can be problematic because the user can get blocked by the server or the API may not have enough time to process the new request after . A good standard is to apply exponential backoff, meaning that the execution will give longer pauses as the number of attempts increases.

* Dedicated python libraries may be considered. I'm leaving here two alternatives: https://pypi.org/project/backoff/ and https://stackoverflow.com/a/35636367/19971645
* I'm not against being a bit more verbose with those API recalls. I'd like to know what is happening when trying to retrieve data from te different APIs
* Implementing such features would let the code safer and more reliable.
* I found 6 occurrences of such problem within the Environment class.

**Additional context**

- To read more about [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff#:~:text=An%20exponential%20backoff%20algorithm%20is,later%2C%20then%204%2C%20etc.)
- Example of problematic code lines: 
image

Gui-FernandesBR:
## Pull request type

## Checklist

## Current behavior
It is annoying that several tests fail due to bad approach of fetching external APIs data.

## New behavior
Fixed the issue #449 . Code is running ok, I also testes slow tests.

## Breaking change
- [x] No

## Additional information
I was really tired of seeing the Wyoming atmosphere model failing in our tests.

### DIFF
--- a/rocketpy/environment/environment.py
+++ b/rocketpy/environment/environment.py
@@ -3,7 +3,7 @@ import json
 import re
 import warnings
 from collections import namedtuple
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 
 import numpy as np
 import numpy.ma as ma
@@ -13,6 +13,7 @@ import requests
 from ..mathutils.function import Function, funcify_method
 from ..plots.environment_plots import _EnvironmentPlots
 from ..prints.environment_prints import _EnvironmentPrints
+from ..tools import exponential_backoff
 
 try:
     import netCDF4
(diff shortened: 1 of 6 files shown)


rocketpy/environment/environment.py
```diff
@@ -3538,6 +3466,110 @@ def set_earth_geometry(self, datum):
...
+                self.process_ensemble(file, dictionary)
+                success = True
+            except OSError:
+                attempt_count += 1
+        if not success:
+            raise RuntimeError(
+                "Unable to load latest weather data for GEFS through " + file
+            )
+
+    @exponential_backoff(max_attempts=5, base_delay=2, max_delay=60)
```
phmbressan, line 3549: Was there any specific reasoning for picking the number of attempts and base delay values among the `fetch` methods? Just curious.

phmbressan (approved):
Good job, I don't have many comments here. Certainly the file is much more readable and resilient agains sporadic server downs.

rocketpy/environment/environment.py
Gui-FernandesBR: Nope, I didn't even think twice about it, that was arbitrary.

In the future we can do a sensitivity analysis if needed.

I guess I was lucky that this has worked in the first attempt.

PR #593, opened 2024-05-03
ENH: Allow for time variable density to Tank Fluids.

phmbressan:
## Pull request type

## Checklist

## Current behavior

Currently, the `Tank` class assumes in its computations that its fluid densities are a constant in time. This approximation can be rather inaccurate, specially for tanks that experiment large variations in fluid density (e.g. due to changes in temperature and pressure) during the flow.

## New behavior

This PR essentially allows for time variable densities when defining the `Fluid` and the `Tank` subclasses discretize these values in time for code speed.

## Breaking change

- [ ] Yes
- [X] No

## Additional Information

The tests for the `Tank` class would benefit a lot from a refactor. However, I preferred to keep this PR focused on the new functionalities.


rocketpy/motors/fluid.py
```diff
@@ -12,12 +16,18 @@ class Fluid:
...
     name : str
         Name of the fluid.
-    density : float
-        Density of the fluid in kg/m³.
+    density : int, float, callable, string, array, Function
+        Density of the fluid in kg/m³. If a int or float is given,
+        it is considered a constant in time. A callable, csv file
+        or an list of points can be given to express the density
+        as a function of time. The parameter is used as a ``Function``
+        source.
```
Gui-FernandesBR, line 24: I believe it makes no sense to have a fluid density depending on the time. How can someone predict these function in the first place?

Wouldn't be more appropiate to have this density parameter depending on the temperature and pressure? To create a function like this:

```python
def density(T, P):
    # so some interpolation
    return value
```

This could be created with the assistance of 2D Function objects and .csv files.

We could even include some "common" fluids csv files in the package, so the user could just call the function with the fluid name and the function would load the csv file and return the 2D Function object. 

In case the user still wants to use a constant value, it would still be possible.

rocketpy/motors/fluid.py
```diff
@@ -12,12 +16,18 @@ class Fluid:
...
+
+        See more on :class:`rocketpy.Function` source types.
     """
 
     name: str
-    density: float
+    density: Union[int, float, str, list, Function]
```
Gui-FernandesBR, line 31: Python functions could also be accepted, right? (for example the lambda you applied in the example)

I would let callable be a valid type for the density parameter instead of just rocketpy.Function

```suggestion
    density: Union[int, float, str, list, callable]

```

rocketpy/motors/tank.py
```diff
@@ -485,6 +488,13 @@ def underfill_height_exception(param_name, param):
             elif (height < bottom_tolerance).any():
                 underfill_height_exception(name, height)
 
+    def _discretize_fluids(self):
+        """Discretizes the fluid densities according to the flux time and the
```
Gui-FernandesBR, line 492:
```suggestion
    def _discretize_fluidd_density(self):
        """Discretizes the fluid densities according to the flux time and the
```

Gui-FernandesBR (commented):
Thank you for proposing these changes.
I think everyone is uncomfortable with the current situation regarding constant density.

Since we are proposing a variable fluid density, how about plotting the density evolution somewhere?

rocketpy/motors/fluid.py
phmbressan: Thanks for your suggestion, I considered it at first, but I was hesitant about it for the following reasons. The Tanks module currently does not calculate any thermodynamical quantities.

Correct me if I am mistaken, but unless the Temperature and Pressure in your comment depend on time, we cannot compute anything with it.

At this point, if the Temperature and Pressure are know time variable quantities, the density is also known as a time variable quantity.

Furthermore, the Pressure and Temperature are not used anywhere is Tanks simulation, we would be introducing two new quantities that are not necessary.

I can see two ways of getting the Fluid density in a static fire:

- Having sensors that compute two thermodynamical properties (such as temperature and pressure) and inferring the others from thermodynamical models;
- Having sensors for mass flow and fluid level inside the tank, then the density can be computed.

In my view, in both ways it rather easy to get the density as a function of time. If we implement it from pressure and temperature, we limit ourserves to a particular case (but a really common one) of the first option.

rocketpy/motors/tank.py
phmbressan: Much better, thanks for the suggestion, I will implement it.

rocketpy/motors/fluid.py
Gui-FernandesBR: Can we make Fluid objects motor-agnostic? If fluid's properties depend on time, this time might be associated with a particular motor, so the same fluid couldn't be used in other tanks.

I was thinking about this: 
- You have the volume of the tank, this is always a fixed value.
- By each time, you calculate the mass of fluid multiplying the volume by the density,
- This density would actually depend on pressure and temperature.

I know that different types of liquid motors will be defined based on different types of input (liquid level, pressure, mass, etc.), but that would be something to be carefully worked on.

What really complicates the whole thing is that we have both liquids and gases defined as a Fluid, but those may present quite different behaviors.

Issue #131, opened 2022-02-12
ENH: Different ways of differentiating


Gui-FernandesBR:
We need to improve the Function.differentiate method. As we know, its implementation is not ready for production and should be enhanced

See #122 as an example of where it is mentioned

giovaniceotto:
Suggestions:

- Complex step differentiation
  - [Reference 1](https://mdolab.engin.umich.edu/wiki/guide-complex-step-derivative-approximation)
  - [Reference 2](https://blogs.mathworks.com/cleve/2013/10/14/complex-step-differentiation/)
- Automatic differentiation
  - [Reference 1](https://en.wikipedia.org/wiki/Automatic_differentiation)
  - [Reference 2](https://towardsdatascience.com/build-your-own-automatic-differentiation-program-6ecd585eec2a)
- Finite differences schemes as a fall back (let the user specify the order)
  - [Reference 1](https://web.media.mit.edu/~crtaylor/calculator.html)

PR #594, opened 2024-05-05
ENH: Complex step differentiation

Gui-FernandesBR:
## Pull request type

## Checklist

## Current behavior
The `Function.differentiate()` method is the only method that can be used to derivate Function objects.

## New behavior
The new, powerful, and fast `Function.differentiate_complex_step()` method is here!
This method will be used to speed up simulations in future PRs.

## Breaking change

- [x] No

## Additional information
This PR will close the #131 issue, a really old feature request of our repository.

### DIFF
--- a/rocketpy/mathutils/function.py
+++ b/rocketpy/mathutils/function.py
@@ -860,7 +860,7 @@ class Function:
             # if the function is 1-D:
             if self.__dom_dim__ == 1:
                 # if the args is a simple number (int or float)
-                if isinstance(args[0], (int, float)):
+                if isinstance(args[0], (int, float, complex)):
                     return self.source(args[0])
                 # if the arguments are iterable, we map and return a list
                 if isinstance(args[0], Iterable):
@@ -869,7 +869,7 @@ class Function:
             # if the function is n-D:
             else:
                 # if each arg is a simple number (int or float)
-                if all(isinstance(arg, (int, float)) for arg in args):
+                if all(isinstance(arg, (int, float, complex)) for arg in args):
                     return self.source(*args)
                 # if each arg is iterable, we map and return a list
                 if all(isinstance(arg, Iterable) for arg in args):
--- a/rocketpy/mathutils/function.py
+++ b/rocketpy/mathutils/function.py
@@ -2427,6 +2427,30 @@ class Function:
                 + self.get_value_opt(x - dx)
             ) / dx**2
 
+    def differentiate_complex_step(self, x, dx=1e-6):
+        """Differentiate a Function object at a given point using the complex
+        step method. This method can be faster than ``Function.differentiate``
+        since it requires only one evaluation of the function. However, the
+        evaluated function must accept complex numbers as input.
+
+        Parameters
+        ----------
+        x : float
+            Point at which to differentiate.
+        dx : float, optional
+            Step size to use for numerical differentiation, by default 1e-6.
+
+        Returns
+        -------
+        float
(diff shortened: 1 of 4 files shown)


rocketpy/mathutils/function.py
```diff
@@ -2427,6 +2427,30 @@ def differentiate(self, x, dx=1e-6, order=1):
                 + self.get_value_opt(x - dx)
             ) / dx**2
 
+    def differentiate_complex_step(self, x, dx=1e-6):
+        """Differentiate a Function object at a given point using the complex
+        step method. This method can be faster than ``Function.differentiate``
```
MateusStano, line 2432: Shouldn't `dx` be much smaller for this to be efficient?

MateusStano (commented):
Really good!

Questions:

1. Are there any benefits to using the old differentiation method? Or is this one just better?
2. Have you tried plugging this in the flight class? Does it make a difference in terms of performance? If so, why not do it in this PR?
3. Could you add second-order differentiation to this?

rocketpy/mathutils/function.py
Gui-FernandesBR: Idk I didn't have much time to investigate it.

Copilot suggested 1e-20 but I changed to 1e-6 to make it more similar to the differentiate method.

All in all it is an user choice, I just set the standard value.
From the tests I implemented, we could see that it has to be sufficient. 

Would you still reduce the dx default value even more?

Gui-FernandesBR:
> Really good!
> 
> Questions:
> 
> 1. Are there any benefits to using the old differentiation method? Or is this one just better?

Yes, this new method only works if the function accepts complex numbers that presents an imaginary part. Also, it is required that the function should be "differentiable in complex dimension", but this is not practical insight. Therefore, the old method should still be used sometimes.

> 2. Have you tried plugging this in the flight class? Does it make a difference in terms of performance? If so, why not do it in this PR?

Yes, it helps the flight class to go faster. The goal of this PR was to implement a complex step differentiation method, nothing more.
As I mentioned in the PR description: **"This method will be used to speed up simulations in future PRs."**

> 3. Could you add second-order differentiation to this?
- *First answer*: I have no idea.
- *After googling for 5min*: Yes, I think it is possible! See https://www.acsu.buffalo.edu/~johnc/complex_step08.pdf 

Let me try to implement it the next weekend.

rocketpy/mathutils/function.py
Gui-FernandesBR: Now I see the recommendation to use 1e-200, seems cool.

Gui-FernandesBR: Solved here: [1356248](https://github.com/RocketPy-Team/RocketPy/pull/594/commits/13562483f646d8cb61f2105d1fa72f892aa0a37c)

Gui-FernandesBR:
@MateusStano let's forget about the second order differentiation, it was giving some errors that I really think are not worth it to debug right now.

The tests are already covering basic scenarios and I know for a fact that the flight simulations are running correctly when using the complex step method (1st order).

I would leave the second order for a future PR.

Could you approve this one please?

MateusStano:
> The tests are already covering basic scenarios and I know for a fact that the flight simulations are running correctly when using the complex step method (1st order).
> 
> I would leave the second order for a future PR.
> 
> Could you approve this one please?

Okay. One more thing that would be good here is using complex diff for `derivative_function`. That might make a big difference for liquids/hybrids simulations

MateusStano (approved):

PR #596, opened 2024-05-05
MNT: refactor u_dot parachute method

Gui-FernandesBR:
## Pull request type

## Checklist

- [ ] Tests for the changes have been added (not needed)

## Current behavior
Let's be honest, reading this function isn't that easy.

## New behavior
Changes names of variables and avoid some redundant operations.

## Breaking change
- [x] No

## Additional information
- This can help: https://downloads.hindawi.com/journals/ijae/2012/182907.pdf
- In the future, we should be adding new attributes to the parachute class (e.g. parachute radius, ka, etc.) and using them to improve the u_dot_parachute results.


rocketpy/simulation/flight.py
```diff
@@ -1768,34 +1768,42 @@ def u_dot_parachute(self, t, u, post_processing=False):
...
+        R = 1.5  # Parachute radius
+        # to = 1.2
+        # eta = 1
+        # Rdot = (6 * R * (1 - eta) / (1.2**6)) * (
+        #     (1 - eta) * t**5 + eta * (to**3) * (t**2)
+        # )
+        # Rdot = 0
+
+        # Calculate added mass
+        ma = ka * rho * (4 / 3) * 3.141592653589793 * R**3
```
phmbressan, line 1796: Use `np.pi` or `math.pi`

rocketpy/simulation/flight.py
```diff
@@ -1768,34 +1768,42 @@ def u_dot_parachute(self, t, u, post_processing=False):
...
+
+        # Define constants
+        ka = 1  # Added mass coefficient (depends on parachute's porosity)
+        R = 1.5  # Parachute radius
+        # to = 1.2
+        # eta = 1
+        # Rdot = (6 * R * (1 - eta) / (1.2**6)) * (
+        #     (1 - eta) * t**5 + eta * (to**3) * (t**2)
+        # )
+        # Rdot = 0
```
phmbressan, line 1793: If these comments are important to understand the code or a legacy implementation, this should mentioned. As it is now, it seems to me an unfinished code block.

rocketpy/simulation/flight.py
Gui-FernandesBR: It was using np.pi before, I thought it would be faster to use the value directly. Any objections?

rocketpy/simulation/flight.py
Gui-FernandesBR: Honestly I don't know. Could be something unfinished or not. 

Old code was not making sense: you calculate Rdot, then u assign Rdot=0.

What I did was just comment it so we don't waste time on these operations.

I wouldn't delete it because I see that we could use it in the future.

rocketpy/simulation/flight.py
Gui-FernandesBR: ok, done: [f07be5e](https://github.com/RocketPy-Team/RocketPy/pull/596/commits/f07be5ee023ec1750c6017f3b97ab7b81f161169)

rocketpy/simulation/flight.py
Gui-FernandesBR: Let's keep it as it is.

Gui-FernandesBR:
@phmbressan could you re-approve it please?

phmbressan (approved):

PR #597, opened 2024-05-07
ENH: Implement optional plot saving

### LINKED ISSUE #564: ENH: Allow saving plots instead of showing them

All matplotlib plots generated by RocketPy (for example via [`Rocket.draw`](https://docs.rocketpy.org/en/latest/reference/classes/Rocket.html#rocketpy.Rocket.draw) or [`flight.all_info`](https://docs.rocketpy.org/en/latest/reference/classes/Flight.html#rocketpy.Flight.all_info)) are shown using matplotlib's [`show`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.show.html) function. This does not really allow for automatic export of these plots (as is the case when not using Jupyter Notebook).

Add a parameter to the info functions (and other plot-generating functions) for selecting between immediately showing the plot and saving the plot to a file. Internally, this can be done using matplotlib's [`savefig`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html) function.

nalquas:
## Pull request type

## Checklist

## Current behavior

Currently, the draw functions responsible for most plots use matplotlib's `show` function. There is currently no (supported) way to automatically save the plots to disk. See #564 

## New behavior

The functions generating plots now have a `filename` parameter, by default `None`. If the filename is `None`, the plot is shown as before using matplotlib's `show`. If the filename is a file path string (e.g. `"/absolute/path/to/file.png"` or `"relative/path/to/file.jpg"`), the plot is saved to the specified file. The file ending in the given filename decides which format the plot will be saved in, with matplotlib supporting: eps, jpg, jpeg, pdf, pgf, png, ps, raw, rgba, svg, svgz, tif, tiff and webp.

## Breaking change

- [x] No (only added optional parameters)

## Additional information

I did not modify functions calling several plot-drawing functions, e.g. functions such as `all(...)`. The reason for this is that those are probably only used for manual testing/analysis, e.g. when using Jupyter Notebook. Users wanting automatic output probably prefer calling the individual plot functions directly as it allows for more fine-control. Additionally, this saves me from additional implementation effort 😅, and reduces complexity of this pull request. If this functionality is still wanted, this could be implemented in a future pull request.

Also, I may have missed some plotting functions. If you spot anything obvious missing, feel free to notify me and I will look into it.

I look forward to any feedback during code review.

Fixes #564


Gui-FernandesBR:
Hey, thank you so much for your draft PR @nalquas , it looks good already.

I will take a look at it by the end of this week and see if I find any improvement opportunity.

Nice job.

nalquas:
The PR is now ready for review. I did not modify the tests, though, and I am not sure about what needs to be done to get the docs updated.

Looking forward to your feedback!

rocketpy/mathutils/function.py
```diff
@@ -1088,13 +1090,15 @@ def __call__(self, *args):
             evaluated at all points in the list and a list of floats will be
             returned. If the function is N-D, N arguments must be given, each
             one being an scalar or list.
+        filename : str | None, optional
+            The path the plot should be saved to. By default None, in which case the plot will be shown instead of saved. Supported file endings are: eps, jpg, jpeg, pdf, pgf, png, ps, raw, rgba, svg, svgz, tif, tiff and webp.
```
Gui-FernandesBR, line 1094: In docstrings, please break your lines at around 80 characters. This makes it easier to read the docstrings in the source code and avoid errors with the documentation builder (Sphinx).

Gui-FernandesBR: Do you think it is worth to mention that these supported file formats are the same as the ones supported by matplotlib?

rocketpy/mathutils/function.py
```diff
@@ -16,6 +16,8 @@
 import numpy as np
 from scipy import integrate, linalg, optimize
 
+from ..plots.plot_helpers import show_or_save_plot
+
```
Gui-FernandesBR, line 20: Very good usage of encapsulation, I liked it.

rocketpy/plots/plot_helpers.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    if filename == None:
+        plt.show()
+    else:
+        # Warn if file ending is not supported
+        file_ending = Path(filename).suffix
+        if file_ending not in get_matplotlib_supported_file_endings():
+            warnings.warn(
+                "Warning: Unsupported file ending '" + file_ending + "'!", UserWarning
+            )
```
Gui-FernandesBR, line 26: Python f-strings are usually faster, and also easier to read and write.

```suggestion
            warnings.warn(
                f"Warning: Unsupported file ending '{file_ending}'!", UserWarning
            )

```

What do you think> This is just a suggestion tho, feel free to not apply it if you don't like it.

rocketpy/plots/plot_helpers.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def get_matplotlib_supported_file_endings():
+    # Get matplotlib's supported file ending and return them (without descriptions, hence only keys)
+    filetypes = plt.gcf().canvas.get_supported_filetypes().keys()
+
+    # Ensure the dot is included in the filetype endings
+    filetypes = ["." + filetype for filetype in filetypes]
+
+    return filetypes
```
Gui-FernandesBR, line 70: This `get_matplotlib_supported_file_endings` function is used only in this module and is not documented, so I initially thought it should be converted to a private function. However, I think there's a potential to use this function in other modules in the future, so I am wondering now if we should move this particular function to the `tools.py` module, which works as a utility module for the rocketpy package. This way, we could use this function in other modules more easily, and without the need to import it from the `rocketpy.plots` module. 

A second opinion would be nice here. @MateusStano @phmbressan?

rocketpy/plots/motor_plots.py
```diff
@@ -236,12 +268,14 @@ def I_23(self, lower_limit=None, upper_limit=None):
         upper_limit : float
             Upper limit of the plot. Default is None, which means that the plot
             limits will be automatically calculated.
+        filename : str | None, optional
+            The path the plot should be saved to. By default None, in which case the plot will be shown instead of saved. Supported file endings are: eps, jpg, jpeg, pdf, pgf, png, ps, raw, rgba, svg, svgz, tif, tiff and webp.
 
-        Return
+        Returns
         ------
         None
```
Gui-FernandesBR, line 276:
```suggestion

        Returns
        -------
        None
```

rocketpy/plots/flight_plots.py
```diff
@@ -3,6 +3,13 @@
 import matplotlib.pyplot as plt
 import numpy as np
 
+try:
+    from functools import cached_property
+except ImportError:
+    from ..tools import cached_property
+
```
Gui-FernandesBR, line 10: Why did you have to add these lines of code?

The `cached_property` can be directly imported on python versions 3.8 and above, no need for this try/except block. 

```suggestion
```

Gui-FernandesBR (commented):
Absolutely amazing PR, we really appreciate your contribution. The fact all the tests are passing, including the slow tests, is already a good sign.

I have a few comments about it, we can discuss on top of them.

There are 2 other things I want/need to do before approving your PR: (1) build the documentation and verify it is working correctly, (2) test the getting started notebook one more time and verify the plots can be saved properly.

I understand your reason for not adding save options to the `all_info` methods. Maybe we should wait until someone requests it as a new feature before we actually implement it.

Gui-FernandesBR:
I also would like to run code coverage on this PR, the workflow did not worked because the base branch is coming from a fork. I need to investigate it better.

Gui-FernandesBR:
Btw @nalquas could you please run `isort` to sort the imports in the files you modified?

You could use this command: `isort --profile black rocketpy/ tests/ docs/`

You can run `pip install isort` if needed

rocketpy/plots/flight_plots.py
nalquas: Ah, that block must have sneaked in during a rebase to a (then newer) version of the develop branch. I will remove it again

rocketpy/mathutils/function.py
nalquas: > Do you think it is worth to mention that these supported file formats are the same as the ones supported by matplotlib?

Yes, I think that would be better. Do you think the explicit list of supported formats should still stay in there, or should we remove it? Otherwise, if matplotlib changes their supported formats, someone would have to update each docstring mentioning them

rocketpy/plots/plot_helpers.py
nalquas: Good point, I will apply this

nalquas:
I have applied the suggestions, updated the docstrings, moved get_matplotlib_supported_file_endings into tools.py and run isort/black. I have also rebased the PR onto the current develop branch.

Feel free to review again 🙂

rocketpy/plots/plot_helpers.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    if filename == None:
+        plt.show()
+    else:
+        # Warn if file ending is not supported
+        file_ending = Path(filename).suffix
+        if file_ending not in get_matplotlib_supported_file_endings():
+            warnings.warn(
+                f"Warning: Unsupported file ending '{file_ending}'!", UserWarning
+            )
```
Lucas-Prates, line 31: Since an unsupported file ending will cause an error, should we not raise a ValueError instead? For some reason, the warnings are only being shown once for me, all calls that follow throw a ValueError by `plt.savefig`.

rocketpy/plots/plot_helpers.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    if filename == None:
+        fig.show()
+    else:
+        # Warn if file ending is not supported
+        file_ending = Path(filename).suffix
+        if file_ending not in get_matplotlib_supported_file_endings():
+            warnings.warn(
+                f"Warning: Unsupported file ending '{file_ending}'!", UserWarning
+            )
```
Lucas-Prates, line 62: Since an unsupported file ending will cause an error, should we not raise a ValueError instead? For some reason, the warnings are only being shown once for me, all calls that follow throw a ValueError by `plt.savefig`.

rocketpy/plots/plot_helpers.py
```diff
@@ -0,0 +1,68 @@
...
+    Parameters
+    ----------
+    fig : matplotlib.figure.Figure
+        The figure to be saved or shown.
+    filename : str | None, optional
+        The path the figure should be saved to, by default None. Supported file
+        endings are: eps, jpg, jpeg, pdf, pgf, png, ps, raw, rgba, svg, svgz,
+        tif, tiff and webp (these are the formats supported by matplotlib).
+    """
+    if filename == None:
```
Lucas-Prates, line 54:
```suggestion
    if filename is None:
```

Lucas-Prates (commented):
Great PR, @nalquas ! Overall, it is looking very good. The one issue that might be troublesome is saving the plots from the `Function` class with non-default arguments. I believe it would require some non-trivial changes.

rocketpy/plots/plot_helpers.py
phmbressan: @Gui-FernandesBR any updates here?

rocketpy/plots/compare/compare_flights.py
```diff
@@ -1294,8 +1290,9 @@ def trajectories_3d(self, figsize=(7, 7), legend=None, filename=None):
             must be in the form (width, height).
         legend : boolean, optional
             Whether legend will or will not be included. Default is True
-        savefig : string, optional
-            If a string is passed, the figure will be saved in the path passed.
+        filename : string, optional
+            If a filename is passed, the figure will be saved in the current
+            directory. The default is None.
```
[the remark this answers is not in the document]
nalquas: That makes sense, I guess I missed a few docstrings of functions that had a filename argument before already? I will go through and edit the remaining docstrings to be consistent

rocketpy/mathutils/function.py
```diff
@@ -1088,13 +1090,18 @@ def __call__(self, *args):
...
+            and webp (these are the formats supported by matplotlib).
 
         Returns
         -------
         ans : None, scalar, list
         """
-        if len(args) == 0:
-            return self.plot()
+        if len(args) == 0 or (len(args) == 1 and filename != None):
+            return self.plot(filename=filename)
```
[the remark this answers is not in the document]
phmbressan: Great comments! I will check out some of the issues you have raised. You are absolutely correct that some of our treatment so far of the `call -> plot` parameters is a bit unclear.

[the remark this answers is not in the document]
nalquas: > (1) I believe that `is not None` is a better practice than `!= None`

Makes sense, I didn't know that convention before.

> (2) Why are we checking `len(args) == 1` and not passing `*args` to the plot function? What is the desired behavior for this?

I tried to copy the previous existing behavior. Before my changes, this code block was:

```py
if len(args) == 0:
    return self.plot()
else:
    return self.get_value(*args)
```

As you can see, the plot was only returned when no args were given. So to keep with the old behavior, I instead check if no args are given *or* exactly one arg, which is the filename. Now thinking about this a bit more I notice this is stupid since len(arg) is not affected by filename, so I don't need to check the arg count at all. I will modify this again 🙃

I'm not 100% sure if all this works, I was confused due to the function using an arbitrary amount of arguments, which honestly makes development more difficult. From my testing with a Python interpreter, it seems like the filename can only be set if callers of the function need to explicitly call `filename="something"` to set it, and all other arguments are assigned to *args. So in my opinion having filename in addition to *args should work.

> (3) Due to current implementations, `return self.plot(*args, filename=filename)` **will introduce errors!**

The args were not passed previously either, see the already mentioned code block. I think this is a different issue not related to this pull request.

nalquas:
The PR is (again) ready for review as the requested changes have been implemented.

I look forward to your feedback

nalquas:
> > The PR is (again) ready for review as the requested changes have been implemented.
> > I look forward to your feedback
> 
> Thank you @nalquas . The review will be much faster now. We are almost ready to merge this one.
> 
> Thanks for the contribution

Is there any remaining interest for this functionality from your side? There have been several hundred commits in your repo since then, so I would have to rebase and rewrite everything again.

Gui-FernandesBR:
> > > The PR is (again) ready for review as the requested changes have been implemented.
> > > I look forward to your feedback
> > 
> > 
> > Thank you @nalquas . The review will be much faster now. We are almost ready to merge this one.
> > Thanks for the contribution
> 
> Is there any remaining interest for this functionality from your side? There have been several hundred commits in your repo since then, so I would have to rebase and rewrite everything again.

Actually @phmbressan has already rebased it but he mentioned he was not able to push to your branch since you are working on a fork. We had to move our development to other tasks.

I guess if you could check if you can give him permission to ush to your branch, it would be really nice.

We are totally interested in the feature, the problem is just that we prioritized other tasks in the past few months.

nalquas:
> I guess if you could check if you can give him permission to ush to your branch, it would be really nice.

Ah, that makes sense. I activated the "Allow edits by maintainers" checkbox now.

CHANGELOG.md
Gui-FernandesBR, line 1: The changelog is outdated. This is non-block, we could fix it later.

rocketpy/mathutils/function.py
```diff
@@ -22,6 +22,8 @@
     RBFInterpolator,
 )
 
+from ..plots.plot_helpers import show_or_save_plot
```
Gui-FernandesBR, line 25: it warns me a bit that you are creating the first rocketpy dependency on the Function class.
Ideally, the Function class should use no function or method from the rocketpy package. Do you agree?

Do we really need this function here? What would be the negative consequences if we exclude the Function class from this PR?

tests/unit/test_plots.py
```diff
@@ -38,3 +41,51 @@ def test_compare(mock_show, flight_calisto):  # pylint: disable=unused-argument
...
+
+    Parameters
+    ----------
+    filename : str
+        Name of the file to save the plot. If None, the plot will be
+        shown instead.
+    """
+    fig, _ = plt.subplots()
+
+    fig.show = MagicMock()
```
Gui-FernandesBR, line 84: nice!

rocketpy/motors/hybrid_motor.py
```diff
@@ -601,16 +601,19 @@ def add_tank(self, tank, position):
...
-
-    def info(self):
-        """Prints out basic data about the Motor."""
-        self.prints.all()
-        self.plots.thrust()
-
-    def all_info(self):
-        """Prints out all data and graphs available about the Motor."""
-        self.prints.all()
-        self.plots.all()
```
Gui-FernandesBR, line 616: deleted info and all_info ?

rocketpy/motors/liquid_motor.py
```diff
@@ -463,21 +463,19 @@ def add_tank(self, tank, position):
...
+    def draw(self, filename=None):
+        """Draw a representation of the LiquidMotor.
 
-    def info(self):
-        """Prints out basic data about the Motor."""
-        self.prints.all()
-        self.plots.thrust()
-
-    def all_info(self):
-        """Prints out all data and graphs available about the Motor.
```
Gui-FernandesBR, line 476: deleted info and all_info ?

Gui-FernandesBR (commented):
Partial review only. I'd like more time to test the feature during the weekend.

Nice job here, I think we are very close to merge this one.

rocketpy/motors/liquid_motor.py
phmbressan: They are already present in the `Motor` parent class. I see no need for repeating them unless a child requires a different implementation. They were all the same.

rocketpy/mathutils/function.py
phmbressan: Given that this import is essentially a "standalone tool", I don't see that being a huge problem compared to, let's say, we imported a `staticmethod` from the `Flight`. One could simply move the whole show or save function to the `Function` module and import from there to the rest of rocketpy.

The import is used to get the show or save functionality shortcut to the `Function` class, e.g.:

```py
my_motor.thrust() # plots, as usual
my_motor.thrust(filename="thrust_curve.png") # saves the plot
```

Gui-FernandesBR: Ok, agreed. 

#665 related.

Gui-FernandesBR:
Adding a validation here would be nice: 
image

I mean, a more specific message should help users. Something like "could not understand the file format, please check ..."

Gui-FernandesBR (commented):
@phmbressan something I also missed is an example of usage. We could do something really simple, maybe by the end of the "first simulation" section.

CHANGELOG.md
phmbressan: Fixed by [bd0400a](https://github.com/RocketPy-Team/RocketPy/pull/597/commits/bd0400a9fff82445bf4510373a5ae4b8991b8805).

phmbressan:
Thanks for your comments @Gui-FernandesBR . I have added a `Exception` message if the file format is unsupported and a small documentation section for the file saving. Should you have any other comments, don't hesitate.

Gui-FernandesBR (approved):
Really nice work, I believe rocketpy users will benefit a lot from this new feature!

Thank you, @nalquas @phmbressan !!

PR #598, opened 2024-05-10
BUG: Fix minor type hinting problems

nalquas:
## Pull request type

## Checklist

## Current behavior

Assume basic type checking is active in VSCode. If you try to assign a float to one of the following function parameters, you get a warning that float is not compatible with int:

- Environment constructor: latitude, longitude, elevation
- SolidMotor constructor: nozzle_position
- Rocket.add_trapezoidal_fins: cant_angle
- Flight constructor: inclination, heading

The reason for the warning seems to be that the default values for those parameters is an integer, so the type checker assumes only ints are supported.

## New behavior

No warning is given when a float is assigned to the listed parameters. Furthermore, no warning is given when assigning an int (just like before), because casting an int to a float does not warrant a warning for the type checker.

## Breaking change

- [x] No

## Additional information

When not using type hinting, the changes should not have any effect, as I have only added a ".0" to the end of the affected ints.

The docs/comments in the code even say that the mentioned parameters support float, but the type checker in VSCode didn't care 😅.

There are likely many more such cases; this PR only covers parameters that I came across so far (so I don't have to see those warnings in the project I work on). A proper solution would be to actually add type hints to everything in the repository.

I am aware of #444, where it was decided that type hinting will not be actively supported in this repository due to too much development overhead.


Gui-FernandesBR (approved):
LGTM,

Thank you for your contribution, @nalquas !

I hope one day we give a second change to the idea of using type hinting in this RocketPy repo, this could help us during the development phases. But I understand that is complicated to maintain.

PR #599, opened 2024-05-15
BUG: plot drag curves when function source is callable

Gui-FernandesBR:
## Pull request type

## Checklist

- [ ] Tests for the changes have been added (if needed) (no idea on how to do that smartly)

## Current behavior
When you use a callable object as the source of your power_on or power_off curves in the Rocket class, the `all_info()` method will break because `self.rocket.power_on_drag.x_array` is inexistent

## New behavior
Solved the issue by  adding a try/except block that will prevent code from breaking in the situation 

## Breaking change
- [x] No

## Additional information
None

### DIFF
--- a/rocketpy/plots/rocket_plots.py
+++ b/rocketpy/plots/rocket_plots.py
@@ -134,10 +134,22 @@ class _RocketPlots:
         None
         """
 
-        x_power_drag_off = self.rocket.power_off_drag.x_array
-        y_power_drag_off = self.rocket.power_off_drag.y_array
-        x_power_drag_on = self.rocket.power_on_drag.x_array
-        y_power_drag_on = self.rocket.power_on_drag.y_array
+        try:
+            x_power_drag_on = self.rocket.power_on_drag.x_array
+            y_power_drag_on = self.rocket.power_on_drag.y_array
+        except AttributeError:
+            x_power_drag_on = np.linspace(0, 2, 50)
+            y_power_drag_on = np.array(
+                [self.rocket.power_on_drag.source(x) for x in x_power_drag_on]
+            )
+        try:
+            x_power_drag_off = self.rocket.power_off_drag.x_array
+            y_power_drag_off = self.rocket.power_off_drag.y_array
+        except AttributeError:
+            x_power_drag_off = np.linspace(0, 2, 50)
+            y_power_drag_off = np.array(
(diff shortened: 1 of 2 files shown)


rocketpy/plots/rocket_plots.py
```diff
@@ -134,10 +134,22 @@ def drag_curves(self):
...
+                [self.rocket.power_on_drag.source(x) for x in x_power_drag_on]
+            )
+        try:
+            x_power_drag_off = self.rocket.power_off_drag.x_array
+            y_power_drag_off = self.rocket.power_off_drag.y_array
+        except AttributeError:
+            x_power_drag_off = np.linspace(0, 2, 50)
+            y_power_drag_off = np.array(
+                [self.rocket.power_off_drag.source(x) for x in x_power_drag_off]
+            )
```
phmbressan, line 152: I believe it would be simpler to check whether the `source` is a `callable` and use `set_discrete(mutate_self=False)`.

Another option, perhaps a better use of the `Function` inner plotting mechanisms, would be the `plot(return_object=True)`.

MateusStano (commented):
Could you also fix this line? https://github.com/RocketPy-Team/RocketPy/blob/b0e34e978e327880c27529d96f941e5fb4ac1d93/rocketpy/prints/aero_surface_prints.py#L138

The airfoil prints will run into the same bug if it is callable defined

rocketpy/plots/rocket_plots.py
Gui-FernandesBR: I'm not in favor of using the set_discrete method due to 2 reasons:

1 - We can't (or shouldn't) modify the source just to plot this data
2 - Even if we don't mutate the source, we would still take longer times to discretize the Function rather then just evaluating it in a few data points.

I understand the importance of reusing the code and features provided from Function. But im this case I prefer the fastest, easier solution, which is to simply add this try/except in the code.

Gui-FernandesBR:
> Could you also fix this line? https://github.com/RocketPy-Team/RocketPy/blob/b0e34e978e327880c27529d96f941e5fb4ac1d93/rocketpy/prints/aero_surface_prints.py#L138
> 
> The airfoil prints will run into the same bug if it is callable defined

This is a very good comment. Let's fix it in another PR.

I think it's important to add a test to this PR so we prevent the same error from happening again in the future

phmbressan (approved):

PR #601, opened 2024-05-16
DOC: Adds prometheus data, Spaceport America 2022

Gui-FernandesBR:
## Pull request type

## Checklist

## New behavior
A new flight has been added to our "Flight Examples" page!

## Breaking change
- [x] No

## Additional information
Many thanks to Giorgio Chassikos and the Western Engineering Rocketry Team


MateusStano (commented):
The rocket is from Spaceport 2022, but the environment on the notebook is using 2023 data. Is this correct?

Also, something is wrong here:

image

Gui-FernandesBR:
> The rocket is from Spaceport 2022, but the environment on the notebook is using 2023 data. Is this correct?
> 
> Also, something is wrong here:
> 
> image
> 

I was running late and wanted publish it as fast as possible, so I used the environment of 2023, which was already saved in our repo, rather than download the 2022 data from Copernicus.

Do you think it's so important to update with the 2022 environment data?

Regarding the acceleration, this is interesting, right?
Seems like the telemetry data is presenting the acceleration as the derivative of speed. After burnout, the rocket constantly reduces its speed, and the telemetry plot represents it.

RocketPy's acceleration will always be in absolute value, no matter if the rocket is speeding up or down. Probably because the acceleration is calculated as the sqrt of ax**2 + ay**2 + az**2

If you want the plots to match perfectly, we could apply the abs() function to the telemetry data. Is this important?

I see this as a beautiful thing.

chasgior214:
Just in case you hadn't seen it yet, the dataset I sent @Gui-FernandesBR from our other flight computer (the TeleMetrum) had a full set of speed data (unlike [the TeleMega data](https://github.com/RocketPy-Team/RocketPy/blob/7a1584693617102f3f7fb6f6d9710b32a378a747/data/prometheus/2022-06-24-serial-6583-flight-0003-TeleMega.csv)). It could be used for a speed comparison to RocketPy. The data from that flight computer is also a lot smoother than the data from the TeleMega.

Gui-FernandesBR:
I swapped the telemetry files and applied a normalization in the data. This is the final plot:

image

Gui-FernandesBR:
@MateusStano ready for review please

MateusStano (approved):

Issue #475, opened 2023-11-20
ENH: Support for power series nose cones


Gui-FernandesBR:
**Is your feature request related to a problem? Please describe.**

The current supported nosecone kinds are: conical, ogive, lvhaack, tangent, vonkarman, elliptical, parabolic. The parabolic shape is just one particular case of the power series, see https://en.wikipedia.org/wiki/Nose_cone_design for more details.

**Describe the solution you'd like**

Support for power series nosecones. 

**Additional context**

Feature request from discord.
![image](https://github.com/RocketPy-Team/RocketPy/assets/63590233/b31b4d03-efd5-431b-9be8-0d22d93f6c15)

lucasfourier:
Gonna try this one

PR #603, opened 2024-05-19
ENH: implement powerseries NoseCones (#475)

Lucas-Prates:
## Pull request type

## Checklist

## Current behavior
This PR implements [power series nosecones](https://en.wikipedia.org/wiki/Nose_cone_design#Power_series), as requested in #475 .

## New behavior
The implementation required adding an extra argument, 'power', to the initialization of the 'NoseCone' class and the 'add_nose' method on the Rocket class.

I tried to maintain the implementation style using setters and property decorators.

I calculated the _enigmatic_ 'self.k' attribute, used to compute the Center of Pressure, in the same way other methods calculate it:

$k = 1 - V/(2\pi R^2 L) $.

$V$ can be computed using basic calculus on the volume of 

$V = \int_{0}^{L} \pi (R\frac{x}{L}) ^ {2n} dx$.

Since this is a simple integral, the final result (if I computed things correctly):

$k = \frac{2n}{2n + 1}$

## Breaking change
- [x] No

## Additional information

I tested it on the example rocket in the Getting Started notebook. Here are some drawings with different values for the 'power' parameter:

$n = 0.2$
powerseries_draw_0-2

$n = 0.5$
powerseries_draw_0-5

$n = 0.8$
powerseries_draw_0-8

### DIFF
--- a/rocketpy/rocket/aero_surface.py
+++ b/rocketpy/rocket/aero_surface.py
@@ -101,7 +101,7 @@ class NoseCone(AeroSurface):
         Nose cone length. Has units of length and must be given in meters.
     NoseCone.kind : string
         Nose cone kind. Can be "conical", "ogive", "elliptical", "tangent",
-        "von karman", "parabolic" or "lvhaack".
+        "von karman", "parabolic", "powerseries" or "lvhaack".
     NoseCone.bluffness : float
         Ratio between the radius of the circle on the tip of the ogive and the
         radius of the base of the ogive. Currently only used for the nose cone's
@@ -156,6 +156,7 @@ class NoseCone(AeroSurface):
         base_radius=None,
         bluffness=None,
         rocket_radius=None,
+        power=None,
         name="Nose Cone",
     ):
         """Initializes the nose cone. It is used to define the nose cone
@@ -167,7 +168,7 @@ class NoseCone(AeroSurface):
             Nose cone length. Has units of length and must be given in meters.
         kind : string
             Nose cone kind. Can be "conical", "ogive", "elliptical", "tangent",
-            "von karman", "parabolic" or "lvhaack".
+            "von karman", "parabolic", "powerseries" or "lvhaack".
         base_radius : float, optional
             Nose cone base radius. Has units of length and must be given in
             meters. If not given, the ratio between ``base_radius`` and
@@ -183,6 +184,10 @@ class NoseCone(AeroSurface):
             The reference rocket radius used for lift coefficient normalization.
             If not given, the ratio between ``base_radius`` and
             ``rocket_radius`` will be assumed as 1.
+        power : float, optional
+            Factor that controls the bluntness fo the shape for a power series
+            nose cone. Must be between 0 and 1. It is ignored when other nose
+            cone types are used.
         name : str, optional
             Nose cone name. Has no impact in simulation, as it is only used to
             display data in a more organized matter.
@@ -202,6 +207,17 @@ class NoseCone(AeroSurface):
                     f"Bluffness ratio of {bluffness} is out of range. It must be between 0 and 1."
                 )
         self._bluffness = bluffness
+        if kind == "powerseries":
+            if power is None:
+                raise ValueError(
+                    "Parameter 'power' cannot be None when using a nose cone kind 'powerseries'."
+                )
+
+            if power > 1 or power < 0:
+                raise ValueError(
+                    f"Power value of {power} is out of range. It must be between 0 and 1."
+                )
+        self._power = power
         self.kind = kind
 
         self.evaluate_lift_coefficient()
@@ -243,6 +259,14 @@ class NoseCone(AeroSurface):
         self._length = value
         self.evaluate_center_of_pressure()
         self.evaluate_nose_shape()
+
+    @property
+    def power(self):
+        return self._power
+
+    @power.setter
+    def power(self, value):
+        self._power = value
 
     @property
     def kind(self):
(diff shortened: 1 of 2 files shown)


Gui-FernandesBR:
Nice one! 

@Lucas-Prates can you mark the PR 475 in the "Development" section at the top right corner of your PR? 
This ensures the issue will be automatically closed when this PR gets to the default branch (i.e. the `master`).

image

rocketpy/rocket/aero_surface.py
```diff
@@ -183,6 +184,10 @@ def __init__(
             The reference rocket radius used for lift coefficient normalization.
             If not given, the ratio between ``base_radius`` and
             ``rocket_radius`` will be assumed as 1.
+        power : float, optional
+            Factor that controls the bluntness fo the shape for a power series
+            nose cone. Must be between 0 and 1. It is ignored when other nose
+            cone types are used.
```
MateusStano, line 190:
```suggestion
        power : float, optional
            Factor that controls the bluntness of the shape for a power series
            nose cone. Must be between 0 and 1. It is ignored when other nose
            cone types are used.
```

rocketpy/rocket/aero_surface.py
```diff
@@ -244,6 +260,14 @@ def length(self, value):
...
         self.evaluate_nose_shape()
 
+    @property
+    def power(self):
+        return self._power
+
+    @power.setter
+    def power(self, value):
+        self._power = value
+
```
MateusStano, line 270: This setter should behave similar to the `bluffness` setter

MateusStano (changes requested):
Hey! This is great! The center of pressure values seem to be correct, so just a few comments:

- Can you add `NoseCone.power` to the class attributes docs?
- The parameter `bluffness` should not be used with `power` I think. It would be good to have a check on this. The drawing breaks if you add `bluffness` to a powerseries nose cone.

rocketpy/rocket/aero_surface.py
Lucas-Prates: Thanks for the review! I will make the suggested changes tonight or tomorrow. With respect to the `bluffness` parameter, should I raise an error if it is not `None` when passing a power series nose cone?

rocketpy/rocket/aero_surface.py
```diff
@@ -167,7 +168,7 @@ def __init__(
             Nose cone length. Has units of length and must be given in meters.
         kind : string
             Nose cone kind. Can be "conical", "ogive", "elliptical", "tangent",
-            "von karman", "parabolic" or "lvhaack".
+            "von karman", "parabolic", "powerseries" or "lvhaack".
```
Gui-FernandesBR, line 171: In this section, I would describe it a bit more... 
Maybe adding something like:  If "powerseries" is used, the "power" argument must be assigned to a value between 0 and 1.

```
            "von karman", "parabolic", "powerseries" or "lvhaack".
```

Gui-FernandesBR (commented):
Very good PR, simple and effective, congratulations!

This PR made me to realize that we do not have a docs page for each aerodynamic surface. maybe this is something to be added to the documentation tasks. @phmbressan

Gui-FernandesBR:
@Lucas-Prates your PR looks nice to me.

Just to attention points:
- The CHANGELOG was not updated with your PR
- There is no test too cover the new feature. Indeed, there are no such thing as a "test_aero_surfaces.py" file in the repo, which make everything harder. 

The changelog needs to be updated before merging the PR, it is really simple so it is a must.
However, the tests are discussable... How good is your understand about unit/integration/e2e tests right now?
Would you prefer to merge this PR first and open up another PR with new tests **OR** to implement the tests here directly? 

We can help you with theses tests, no rocket science, don't worry.

Lucas-Prates:
Thanks a lot for the review, @Gui-FernandesBR. With respect to the tests, I have extremely basic experience with it. However, I do not mind implementing the tests in this PR since it seems to be a good opportunity to learn.

Also, I do not know what a changelog is or how to adequately update it, so I will need guidance with that as well.

Lucas-Prates:
One minor issue: the documentation states that the `power` argument must be between $0$ and $1$. However, it could also be one of these extreme values. When `power` equals $1$, we have the same result as a 'conical' nose cone, and the code runs normally.

When `power` equals $0$, we currently raise a `ValueError` since the drawing breaks because the front end of the nose cone does not close ($0 ^ 0 = 1$ for Python). However, we could understand $0$ as the limit case where the nose cone becomes a cylinder.

Should we still raise a `ValueError` if the user passes $0$ as input for `power` or should we handle this input differently? For instance, we could set `power` to a value close to $0$, e.g. 1e-6, so its behavior is equivalent but without breaking the drawing.

Here are some drawings to understand.

`power` $= 0.01$
powerseries_1e-2

`power` $= 1e-6$
powerseries_1e-6

`power` $= 0$
powerseries_0

rocketpy/rocket/aero_surface.py
MateusStano: I think ignoring the bluffness and raising a warning would be best

MateusStano:
> Should we still raise a `ValueError` if the user passes  0  as input for `power` or should we handle this input differently? For instance, we could set `power` to a value close to  0 , e.g. 1e-6, so its behavior is equivalent but without breaking the drawing.

Passing 0 should raise an error since it does not make physical sense. Passing 1 should work normally

Gui-FernandesBR (approved):

Gui-FernandesBR (commented):
Good job with this PR @Lucas-Prates!! I believe we are ready to merge it already, we will just need to fix a simple conflict within the CHANGELOG.md file

MateusStano (approved):

Issue #133, opened 2022-02-13
ENH: Variation of CL with Mach (nose and tail)


Gui-FernandesBR:
Currently we have calculation of Lif Coefficient in function of Mach Number for our fins, but in the future will be important having the same approach for the nosecone and tail

Gui-FernandesBR:
There's no recommendation to use the Prandtl correction factor (beta) for nose or boat tail, at least not on Barrowman.
We will have to base it in some other literature.

PR #609, opened 2024-05-23
ENH: Add Prandtl-Glauert transformation to NoseCone and Tail

MateusStano:
## Pull request type

## Checklist

## Current behavior

Prandtl-Glauert transformation is only applied to `Fins`. This was done because on https://ntrs.nasa.gov/citations/20010047838, Barrowman only applies it to fins

## New behavior

Prandtl-Glauert transformation is now also applied to `NoseCone` and `Tail`. This is done because Barrowman does in a more recent paper: https://arc.aiaa.org/doi/10.2514/6.1979-504

## Breaking change

- [ ] Yes
- [x] No


rocketpy/rocket/aero_surface.py
```diff
@@ -34,6 +34,30 @@ def __init__(self, name):
         self.name = name
         return None
 
+    # Defines beta parameter
+    def _beta(_, mach):
```
Gui-FernandesBR, line 38: This `_beta` method doesn't use any class attribute. It should be either a static method or a an isolated function commonly defined in tools.py. This would also avoid making the method private.

I don't really see this function being applied outside of the context of aero surfaces. Therefore, I would convert it into a `@staticmethod`

rocketpy/rocket/aero_surface.py
```diff
@@ -34,6 +34,30 @@ def __init__(self, name):
...
+            Value that characterizes flow speed based on the mach number.
+        """
+
+        if mach < 0.8:
+            return np.sqrt(1 - mach**2)
+        elif mach < 1.1:
+            return np.sqrt(1 - 0.8**2)
+        else:
+            return np.sqrt(mach**2 - 1)
+
```
Gui-FernandesBR, line 60: Where did you get this from? Probably other software, right?

Should we explicit our references?
If you took it from other software, might be a good idea to search for their references as well.

tests/test_flight.py
Gui-FernandesBR, line 1: Do you think we should increase tolerance (maybe to 2 decimals) along with the update in the assertion values?

Maybe these assertions are too tight... maybe not? 

I propose we reflect on that.

Gui-FernandesBR (commented):
This PR is really good, simple and effective.

My only "strong suggestion" is: can you include your references <somewhere> in the code? I remember I read the first document you shared in your PR description and I got the impression we shouldn't add the `beta` correction. Now I understand it.

This PR will close the following issue: https://github.com/RocketPy-Team/RocketPy/issues/133

rocketpy/rocket/aero_surface.py
MateusStano: https://github.com/RocketPy-Team/RocketPy/pull/609/commits/0b44194de7ad19d6f914ce84b0ef1a4e5c8c7ea5

Gui-FernandesBR (approved):

Issue #610, opened 2024-05-24
BUG: TypeError when creating nose cones with bluffness between 0 and 1


Lucas-Prates:
**Describe the bug**

A "TypeError: unsupported operand type(s) for ** or pow(): 'list' and 'int'" is raised when creating any nose cone with bluffness different of None or $0$.

**To Reproduce**

Create any nose cone with bluffness between $0$ and $1$. Minimal code to reproduce:

```
from rocketpy import NoseCone
NoseCone(length=1, kind="conical", base_radius=1, bluffness=0.5)
```

PR #611, opened 2024-05-24
BUG: Fixes nose cone bluffness issue #610

Lucas-Prates:
## Pull request type

Bug fix issue #610 .

## Checklist

## Current behavior
The `find_radius` method inside the `NoseCone` class computes the square of its argument `x`. The lambda function, however, passes a list to it, causing a `TypeError`.

## New behavior
Just pass the value instead of the list.

## Breaking change

- [x] No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- BUG: Fixes nose cone bluffness issue #610 [#611](https://github.com/RocketPy-Team/RocketPy/pull/611)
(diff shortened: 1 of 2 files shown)


Lucas-Prates:
As suggested by @Gui-FernandesBR, this might be a good indication to expand unit tests for aero_surface classes. This seems like an error that would have been caught by tests.

MateusStano (approved):

Gui-FernandesBR:
> As suggested by @Gui-FernandesBR, this might be a good indication to expand unit tests for aero_surface classes. This seems like an error that would have been caught by tests.

Don't worry, we can add more robust tests for aero_surfaces classes in the future.
I think @lucasfourier should be aware of this.

Thank you for your fast solution, @Lucas-Prates !! Nice job, we are merging this one.

PR #552, opened 2024-02-09
BUG: Update flight trajectory plot axes limits

Gui-FernandesBR:
## Pull request type

- [x] Code changes (bugfix, features)

## Checklist

- [ ] Tests for the changes have been added (if needed)
- [ ] Docs have been reviewed and added / updated
- [x] Lint (`black rocketpy/ tests/`) has passed locally 
- [x] All tests (`pytest --runslow`) have passed locally
- [x] `CHANGELOG.md` has been updated (if relevant)

## Current behavior
The limits for the projections and flight path in the 3D map is a bit odd since some time ago. This is unexpected.

image

## New behavior
Improves the plot and restore a better, more intuitive behavior that was present since previous versions.

image

## Breaking change
- [x] No

## Additional information
- This is another PR to be merged BEFORE the #536

PR #623, opened 2024-06-16
MNT: Fix warnings in test suite and adds support for numpy 2.0

Gui-FernandesBR:
## Pull request type

## Checklist

## Current behavior
Numpy 2.0 was officially released today, but we cannot run rocketpy with that new version because some methods were deprecated.
Also, running pytest in our repo usually raises a lot of annoying warnings.

## New behavior
Fixed, fixed, fixed! Please see the commit history.

## Breaking change

- [x] No

## Additional information
- https://numpy.org/devdocs/numpy_2_0_migration_guide.html#
- The optional dependency `timezonefinder` cannot be installed with numpy 2.0. See this issue: https://github.com/jannikmi/timezonefinder/issues/234


Gui-FernandesBR:
This PR was actually partially problematic: https://github.com/RocketPy-Team/RocketPy/pull/552/files 
The 3D trajectory is currently being distorted. Fixed too!

Gui-FernandesBR:
Something is wrong here. The aritmetic priority is mixed, especially when running tests on macOS.
I can't find an easy solution right now.

Gui-FernandesBR:
> Something is wrong here. The aritmetic priority is mixed, especially when running tests on macOS. I can't find an easy solution right now.

Ok, I have narrowed down the problem, it is a bit clearer now. However, I don't have a solution yet.

The documentation tests are breaking on Python 3.12 with numpy 2.0. Essentially, this is a type promotion problem that was possibly caused by NEP50. This may help to find a solution: https://numpy.org/neps/nep-0050-scalar-promotion.html#alternatives

However, I think the easiest solutions for us would be the removal of the problematic tests.

Below we have an example: 

```bash
(.venv12) PS RocketPy> pytest rocketpy --doctest-modules
======================================================= test session starts ========================================================
platform win32 -- Python 3.12.4, pytest-8.2.2, pluggy-1.5.0
rootdir: RocketPy
configfile: pyproject.toml
plugins: cov-5.0.0
collected 34 items

rocketpy\environment\environment.py F.F                                                                                       [  8%]
rocketpy\mathutils\function.py FF....                                                                                         [ 26%]
rocketpy\mathutils\vector_matrix.py ................                                                                          [ 73%]
rocketpy\tools.py .........                                                                                                   [100%]

============================================================= FAILURES =============================================================
_________________________ [doctest] rocketpy.environment.environment.Environment.calculate_density_profile _________________________
3084         None
3085
3086         Examples
3087         --------
3088         Creating an Environment object and calculating the density
3089         at Sea Level:
3090
3091         >>> env = Environment()
3092         >>> env.calculate_density_profile()
3093         >>> env.density(0)
Expected:
    1.225000018124288
Got:
    np.float64(1.225000018124288)

RocketPy\rocketpy\environment\environment.py:3093: DocTestFailure
_____________________________ [doctest] rocketpy.environment.environment.Environment.set_gravity_model _____________________________
556         a :class:`Function` object representing the gravity model.
557
558         Examples
559         --------
560         Let's prepare a `Environment` object with a constant gravity
561         acceleration:
562
563         >>> g_0 = 9.80665
564         >>> env_cte_g = Environment(gravity=g_0)
565         >>> env_cte_g.gravity([0, 100, 1000])
Expected:
    [9.80665, 9.80665, 9.80665]
Got:
    [np.float64(9.80665), np.float64(9.80665), np.float64(9.80665)]

RocketPy\rocketpy\environment\environment.py:565: DocTestFailure
_____________________________________ [doctest] rocketpy.mathutils.function.Function.get_value _____________________________________
829         >>> f2.get_value([1, 2, 3], [1, 2, 3])
830         [2, 8, 18]
831         >>> f2.get_value([5], [5])
832         [50]
833
834         Testing with ndarray source (1 dimension):
835         >>> f3 = Function(
836         ...    [(0, 0), (1, 1), (1.5, 2.25), (2, 4), (2.5, 6.25), (3, 9), (4, 16)]
837         ... )
838         >>> f3.get_value(2)
Expected:
    4.0
Got:
    np.float64(4.0)

RocketPy\rocketpy\mathutils\function.py:838: DocTestFailure
_______________________________ [doctest] rocketpy.mathutils.function.Function.is_strictly_bijective _______________________________
2617         -------
2618         result : bool
2619             True if the Function is "strictly" bijective, False otherwise.
2620
2621         Examples
2622         --------
2623         >>> f = Function([[0, 0], [1, 1], [2, 4]])
2624         >>> f.isbijective()
2625         True
2626         >>> f.is_strictly_bijective()
Expected:
    True
Got:
    np.True_
```

phmbressan:
The new promotion rules make it rather difficult to support doctest of both `numpy<2.0` and the newest one, since there is a type difference in the `__repr__` check: `True` (for older numpy) and `np.True_`.

The PR #624 bumps the minimum Python version to 3.9, allowing all the docstest to be executed with `numpy>=2.0` allowing for consistency. The code itself (and the tests) still work fine on `numpy<2.0`.

Therefore, that PR should be merged first.

rocketpy/environment/environment.py
```diff
@@ -682,7 +682,7 @@ def set_elevation(self, elevation="Open-Elevation"):
         #     self.elevation = elev
 
         elif self.latitude is not None and self.longitude is not None:
-            self.elevation = self.__fetch_open_elevation()
+            self.elevation = float(self.__fetch_open_elevation())
```
phmbressan, line 685: Is this to avoid breaking changes regarding the type of `self.elevation`?

rocketpy/mathutils/function.py
```diff
@@ -1875,7 +1883,9 @@ def __add__(self, other):
                 return Function(lambda x: (self.get_value_opt(x) + other(x)))
         # If other is Float except...
         except AttributeError:
-            if isinstance(other, NUMERICAL_TYPES):
+            if isinstance(other, NUMERICAL_TYPES) or self.__is_single_element_array(
+                other
+            ):
```
phmbressan, line 1888: What was the old behavior that introduced the reason of this change?

To be clear, I am in favor of the change, just asking to understand the context.

rocketpy/environment/environment.py
Gui-FernandesBR: that was just me trying to debug the problem.

It doesn't really changes the code behavior, we can ignore it.

rocketpy/mathutils/function.py
Gui-FernandesBR: The problem is that we had `np.ndarray` in the `NUMERICAL_TYPES` list. This was problematic because we cannot say that `np.array([(0,0), (0, 1)])` is numeric. What we would like to say, actually, is: "numpy arrays can be considered numeric if they have only a single value".

I updated the NUMERICAL_TYPES but had to add a special comparison "is_single_element_array". 

I think the code is safer now.

phmbressan: Agreed.

rocketpy/environment/environment.py
phmbressan: Alright, I am favor rolling back this change here, but this is a non blocking issue for me.

phmbressan (approved):

PR #634, opened 2024-07-08
MNT: Refactor `AeroSurfaces`

MateusStano:
## Pull request type

## Checklist

## Changes

- Separate each `AeroSurface` class into a different file
- Made a few shared attributes part of the `AeroSurface` class

## Breaking change

- [ ] Yes
- [x] No


rocketpy/rocket/aero_surface.py
Gui-FernandesBR, line 1: This file deletion will probably create merge conflicts with the following PR: #597

rocketpy/rocket/aero_surface/__init__.py
Gui-FernandesBR, line 1: classes `Fins` and `AeroSurface` are abstract classes and therefore cannot be instantiated.
They are only exposed here if the user decides to create a new class based on them.
With that being said, I think we should avoid re-exposing them in the init files.

I don't want someone doing something like this:

```python
from rocketpy import Fins

fins = Fins(...) # raises an error!!
```

An experienced user would still be able to import the class using its absolute path, like this:

```python
from rocketpy.rocket.aero_surface.fins import Fins,

class NewFins(Fins):
    ...
```

rocketpy/rocket/aero_surface/air_brakes.py
```diff
@@ -0,0 +1,207 @@
+import warnings
+
+import numpy as np
+
+from rocketpy.mathutils.function import Function
+from rocketpy.plots.aero_surface_plots import _AirBrakesPlots
+from rocketpy.prints.aero_surface_prints import _AirBrakesPrints
+
+from .aero_surface import AeroSurface
```
Gui-FernandesBR, line 9: Any particular reason for not using absolute path here?

```suggestion
from rocketpy.rocket.aero_surface.aero_surface import AeroSurface
```

All the other 3 imports are absolute.

rocketpy/rocket/aero_surface/nose_cone.py
Gui-FernandesBR, line 1: 500 lines just for the NoseCone class, omg

rocketpy/simulation/flight.py
```diff
@@ -1723,8 +1722,8 @@ def u_dot_generalized(
                 position - self.rocket.center_of_dry_mass_position
             ) * self.rocket._csys - aero_surface.cpz
             comp_cp = Vector([0, 0, comp_cpz])
-            surface_radius = aero_surface.rocket_radius
-            reference_area = np.pi * surface_radius**2
+            reference_area = aero_surface.reference_area
+            reference_length = aero_surface.reference_length
```
Gui-FernandesBR, line 1726: is the reference length **always** equals to `aero_surface.rocket_radius` ?

Gui-FernandesBR (approved):
Good job.

It will be much easier to add new `AeroSurface` now. 

The `flight.py` file is the only one that warns me a bit. Please double check this file to ensure that you are not making any bad operation.

rocketpy/rocket/aero_surface/air_brakes.py
MateusStano: Its just easier to understand what you are importing since its from a file in the same folder

PR #645, opened 2024-07-20
DOC: Fix documentation warnings

Gui-FernandesBR:
## Pull request type

## Checklist

## New behavior
Solves warnings and deprecations.

The main file you should care when reviewing is the `conf.py` file. All the rest are just tiny details that I've adjusted to fix the warning that are displayed by the documentation building process.

## Breaking change

- [x] No


rocketpy/motors/motor.py
```diff
@@ -503,9 +501,9 @@ def I_11(self):
         The e_1 direction is assumed to be the direction perpendicular to the
         motor body axis. Also, due to symmetry, I_11 = I_22.
 
-        References
-        ----------
-        .. [1] https://en.wikipedia.org/wiki/Moment_of_inertia#Inertia_tensor
+        See Also
+        --------
+        https://en.wikipedia.org/wiki/Moment_of_inertia#Inertia_tensor
```
phmbressan, line 506: Any reason to prefering the ``See Also``? I believe the warnings that were raised by the ``References`` section were caused since this ``[1]`` was not being referenced in the docstring text.

rocketpy/rocket/parachute.py
```diff
@@ -238,3 +212,11 @@ def all_info(self):
...
         self.info()
         # self.plots.all() # Parachutes still doesn't have plots
+
+        # .. note::
+
+        #     The drag force exerted on the parachute is computed by \
+        #     the equation F = ((1/2)*rho*V^2)*cd_s, where rho is the \
+        #     freestream air density, ``V`` is the relative velocity between \
+        #     the rocket and the air, and ``cd_s`` is the drag coefficient \
+        #     times the reference area of the parachute.
```
phmbressan, line 222: It seems a bit odd to have these lines at the end of the class in the info method. Is this the intended location?

phmbressan (approved):
Good job on tackling so many warnings. The changes regarding math inputs, ``LaTeX``, text clarity etc are much welcome.

rocketpy/motors/motor.py
Gui-FernandesBR: The problem was actually different: multiple references were being added with the number 1.
Unfortunatelly sphinx shares these numbers within the whole documentation project.

Also, the See Also is a special component that gets redered as a snippet after building the html docs.

rocketpy/rocket/parachute.py
Gui-FernandesBR: This was certainly not made on purpose.

PR #647, opened 2024-07-30
BUG: Time Node Merge Not Including Controllers

MateusStano:
## Pull request type

## Checklist

## Description

Found by team RED on Discord.

Controller time nodes were being ignored and discarded  in the merging of the time nodes, now they are correctly considered

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@ Attention: The newest changes should be on top -->
 
 ### Fixed
 
-- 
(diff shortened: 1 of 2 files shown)


Gui-FernandesBR (approved):

phmbressan (approved):
Good job on quickly solving the issue. 

Is there any way we could add an assertion to the `air_brakes` tests to prevent this from happening again? At first, I don't think so since the way the merging is implemented requires manually adding things.

PR #649, opened 2024-08-03
MNT: Refactor Parallel MonteCarlo and Stochastic Seeding

phmbressan:
## Pull request type

## Checklist

## Current behavior

Enter text here...

## New behavior

Enter text here...

## Breaking change

- [ ] Yes
- [X] No

## Additional information

Enter text here...


rocketpy/tools.py
```diff
@@ -232,32 +232,39 @@ def bilinear_interpolation(x, y, x1, x2, y1, y2, z11, z12, z21, z22):
     ) / ((x2 - x1) * (y2 - y1))
 
 
-def get_distribution(distribution_function_name):
+def get_distribution(distribution_function_name, rng_generator=None):
```
Gui-FernandesBR, line 235: Seems like you chose using implicit names rather than a full description.

`rng_generator` -> `random_number_generator`

Seems more readable. What do you think?

rocketpy/tools.py
```diff
@@ -232,32 +232,39 @@ def bilinear_interpolation(x, y, x1, x2, y1, y2, z11, z12, z21, z22):
...
     """Sets the distribution function to be used in the monte carlo analysis.
 
     Parameters
     ----------
     distribution_function_name : string
         The type of distribution to be used in the analysis. It can be
         'uniform', 'normal', 'lognormal', etc.
+    rng_generator : np.random.Generator, optional
+        The random number generator to be used. If None, the default generator
+        is used.
```
Gui-FernandesBR, line 245: Let it cleaerer that the default generator is the numpy default gerator.

rocketpy/tools.py
Gui-FernandesBR: If you REALLY want to use RNG acronym, I suggest you to re-think if we really need the "_generator" part.

How do you read this variable name? "Random generator number underscore generator"

rocketpy/stochastic/stochastic_model.py
```diff
@@ -63,9 +66,24 @@ def __init__(self, obj, **kwargs):
 
         self.obj = obj
         self.last_rnd_dict = {}
+        self.__stochastic_dict = kwargs
+        self._set_stochastic()
```
Gui-FernandesBR, line 70:
```suggestion
        self._set_stochastic(seed)
```

Gui-FernandesBR: I wonder whether you made a mistake

rocketpy/simulation/monte_carlo.py
```diff
@@ -288,43 +279,54 @@ def __run_in_serial(self):
...
+
+                sim_monitor.update_status(sim_monitor.count)
+
+            sim_monitor.finalize_status()
 
         except KeyboardInterrupt:
-            print("Keyboard Interrupt, files saved.")
-            error_file.write(json.dumps(self._inputs_dict, cls=RocketPyEncoder) + "\n")
+            MonteCarlo._reprint("Keyboard Interrupt, files saved.")
+            with open(self._error_file, "a", encoding="utf-8") as file:
```
Gui-FernandesBR, line 313: Any reason for enforcing utf-8 ? I was once told to not do that specially in an open-source library where people of different languages and operatoional systems could benefit form it.

rocketpy/simulation/monte_carlo.py
```diff
@@ -1764,28 +1319,36 @@ def reprint(self, sim_idx, end="\n", flush=False):
...
+
+        Parameters
+        ----------
+        show_process_time : bool, optional
+            If True, the process time will be shown. Default is True.
+
+        Returns
+        -------
+        None
+        """
```
Gui-FernandesBR, line 1347: The docstring is outdated

```suggestion
        """Prints the final status of the simulation."""
```

I also suggest renaming the method. "print_final_status" seems cooler.

Gui-FernandesBR (approved):
Outstanding work. Code is much cleaner now, we really appreciate your effort.

I look forward to seeing this feature being released

rocketpy/stochastic/stochastic_model.py
phmbressan: It is a mistake, nice catch.

rocketpy/tools.py
phmbressan: Great point, fixed on [d18408e](https://github.com/RocketPy-Team/RocketPy/pull/649/commits/d18408eb0679f06f89b42a2999b6a3d8c5c5698f).

rocketpy/tools.py
phmbressan: Made it clearer on [d18408e](https://github.com/RocketPy-Team/RocketPy/pull/649/commits/d18408eb0679f06f89b42a2999b6a3d8c5c5698f).

rocketpy/stochastic/stochastic_model.py
phmbressan: Solved by [abe5747](https://github.com/RocketPy-Team/RocketPy/pull/649/commits/abe574774fa7bc8e16b5fcc62ab4c9e4dfc40afa).

rocketpy/simulation/monte_carlo.py
phmbressan: My main concern was standardization, since we might have to read these files later and knowing the encoding will make things easier. The file saving overall will be refactored with better encoding support, but there are generally two approaches to save files:

- Prioritize human readability / editing: in this case the encoding is important and is a good practice to have at least a standard. There are a plethora of writing systems in the world, of course `utf-8` does not cover everything, but it has become an "Internet standard". From [python docs](https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files): "Because UTF-8 is the modern de-facto standard, encoding="utf-8" is recommended unless you know that you need to use a different encoding.".
- Prioritize I/O operations: this generally requires saving things as bytes, which does not require encoding. This makes all files "compatible" (reading simulations). However, they are generally not human readable and it is very hard to edit them in common editors.

I am more inclined to the current first option, which makes easier to read and make changes to the files.

rocketpy/simulation/monte_carlo.py
phmbressan: Solved in [fe7bad3](https://github.com/RocketPy-Team/RocketPy/pull/649/commits/fe7bad3f0ec74927083290a3436b8ac5aaf0afef).

rocketpy/simulation/monte_carlo.py
Gui-FernandesBR: That's ok. I'm ok with this.

PR #678, opened 2024-08-30
MNT: Refactor Tank's testing Assertion with CAD data.

phmbressan:
## Pull request type

## Checklist

## Current behavior

Currently the Tank's testing module present the following issues:

- The module has too complex logic inside the tests;
- There are analytical functions that perform geometrical validation:
    - Those, even if correct, are hard to maintain and may, for instance, repeat mistakes from the code.

## New behavior

This PR aims to bring the following changes to code:

- CAD related data by `cadquery` module for geometrical validation;
- Refactor of the complex logic inside the Tanks tests;
    - There is still room for improvement here, but the code is better organized now.
- Implement plotting and printing functionality to the Tanks class (this can be moved to another PR if needed). 

## Breaking change

- [ ] Yes
- [X] No

## Additional information

Check out [CadQuery docs](https://cadquery.readthedocs.io/en/latest/) for further information and examples.

An important notice is that the CAD module is not executed in every test for the two reasons:

- The CAD library has a large install size (> 200MB);
- Some models are slow to be used in parametrized tests.

Therefore I have added the necessary code into a Jupyter notebook for easy recalculation of the expected values, but this can be discussed further on this PR.


rocketpy/plots/tank_plots.py
```diff
@@ -98,3 +150,10 @@ def all(self):
         -------
         None
         """
+        self.draw()
+        self.tank.fluid_mass.plot(*self.flux_time)
+        self.tank.net_mass_flow_rate.plot(*self.flux_time)
+        self.fluid_height()
+        self.fluid_volume()
+        self.fluid_center_of_mass()
```
MateusStano, line 158: You should add an `.all_info` and `.info` method to the `Tanks` class and `Fluids` class now that the plots and prints are good

tests/fixtures/motor/data/cad_tank_geometry.ipynb
MateusStano, line 1: I actually quite like this solution, seems to avoid some annoyances while being very pratical. I would suggest that you rename this file to something like `tank_cad_validation_data_generator.ipynb` or something of the sort for clarity purposes.

Or maybe just add a markdown cell with one or two sentences explaining what the notebook is for.

phmbressan: Great comment. I will do that.

The notebook is still a bit raw in terms of explanations, since I opted to wait for opinions on whether we'd keep it.

Gui-FernandesBR (commented):
a few tests are not passing locally...

```python
def test_all_info(mock_show, flight_calisto_robust):  # pylint: disable=unused-argument
def test_motor(mock_show, cesaroni_m1670):  # pylint: disable=unused-argument
```

PR #679, opened 2024-08-30
ENH: Expansion of Encoders Implementation for Full Flights.

phmbressan:
## Pull request type

## Checklist

- [X] Tests for the changes have been added (partially done)
- [ ] Docs have been reviewed and added / updated (future PR)

## Current behavior

Currently the `_encoding` module is just a stub and cannot still encode full simulations into a JSON like format. 

## New behavior

This PR expand this module bringing the following:
- Custom Encoding for more types;
- Customized behavior for some required classes with methods `to_dict`;
- Support for `Function` encoding.

This PR also introduces decoding functionalities. The way this works is:

- If the is a `to_dict` or `from_dict` method, they are used to perform the (de)encoding;
- Otherwise, a default encoding of all vars() and constructor based decoding is done.

## Breaking change

- [ ] Yes
- [X] No

## Additional information

There is much to discuss here on the implementation and maintainability side of things. Main discussion points:

- Should we add a `to_dict` and (in the future) a `from_dict` method to some (or all) rocketpy classes?
    - My Opinion: I think adding the methods to everything might be the only way out. But there is some work and maintainability considerations.
- The current status of the encoding can be fully deserialized?
    - My Opinion: All the information is there, but the typing should be improved to know which objects to decode (deserialize) into. For that reason, I kept the `_encoders` private.
- Are there any other modules to make our work easier?
    - From my research, I have tested the following:
        - `simplejson`: brings some more types and speed into encoding, but does not help much with custom types;
        - `jsonpickle`: really great for the typing handler (already writes by default each object type) and is able to serialize. The problem is that is does not handle `functions/lambdas` and I did not find a good way to implement custom handlers for our use case;
        - `pickle / dill`: the most plug and play solution, however the output is not human readable and not generally compatible between different rocketpy versions. 

Of course, this is what I understood upon researching and testing, feel free to make any comments or suggest other modules.  

## Remarks and Future Considerations

Here are some of the steps that could be improved:

- [X] Verify why the test for hybrid motor flight requires higher tolerances;
    - Corrected, no higher tolerances needed anymore.
- [ ] Test if all of the atmospheric models of Environments can be decoded without requiring re-read of files or API;
- [X] Evaluate whether we should encode outputs or only the minimum required information to decode the class from its __init__;
    - Option for including outputs or not was added.
- [ ] (Optional) Expand implementation to `AirBrakes` + `Controllers` and new classes.

Some important remarks:

- The simple getting started simulation is essentially identical comparing the store & load, the only difference is the random lag and noise from Parachute class. I could not found a simple solution to reset those and always get the same results.


requirements.txt
```diff
@@ -5,3 +5,4 @@ netCDF4>=1.6.4
 requests
 pytz
 simplekml
+dill
```
Gui-FernandesBR, line 8: have you searched about this lib before adding it to the requirements? is this stable enough?

Also, you must edit the pyproject.toml file otherwise this change is useless.

rocketpy/mathutils/function.py
```diff
@@ -3269,6 +3272,52 @@ def __validate_extrapolation(self, extrapolation):
...
+            "source": source,
+            "title": self.title,
+            "inputs": self.__inputs__,
+            "outputs": self.__outputs__,
+            "interpolation": self.__interpolation__,
+            "extrapolation": self.__extrapolation__,
+        }
+
+    @classmethod
+    def from_dict(cls, func_dict):
```
Gui-FernandesBR, line 3298: I`d like to see an example of you loading a class from the dict.

rocketpy/rocket/components.py
```diff
@@ -186,3 +186,18 @@ def sort_by_position(self, reverse=False):
         None
         """
         self._components.sort(key=lambda x: x.position, reverse=reverse)
+
+    def to_dict(self):
+        """Return a dictionary representation of the components.
```
Gui-FernandesBR, line 191: You`re gonna have to make a lot of `to_dict` methods. My recommendation is to avoid documenting all of them. 
It's still experimental feature, so let's be "lazy" this time.

Gui-FernandesBR (approved):
Nothing to complain about, good work.

requirements.txt
phmbressan: The `pyproject` file imports the default requirements one, so I don't think I need to change there. This library seems to be stable enough, they release about every six months consistently and have a large userbase / star count.

phmbressan:
I have just noticed that I made a mistake when merging CHANGELOG conflicts for this PR.

I will fix it before merging. This does not block reviewing.

- [ ] Fix `CHANGELOG.md` merging errors.

rocketpy/_encoders.py
```diff
@@ -33,11 +39,153 @@ def default(self, o):
...
+    Parameters
+    ----------
+    signature : str
+        Signature of the class.
+
+    Returns
+    -------
+    type
+        Class defined by the signature.
+    """
```
Gui-FernandesBR, line 129: I still cannot know what is a "class signature", is it the signature of the init class?

what do you have in the "signature" dictionary (which bte the docs says it is a str) exactly?

rocketpy/motors/fluid.py
```diff
@@ -60,3 +60,10 @@ def __str__(self):
         """
 
         return f"Fluid: {self.name}"
+
+    def to_dict(self, _):
```
Gui-FernandesBR, line 64: I usually see this behavior in other packages, what do you think about it? I dont really understand the reasons of that tho.

```suggestion
    def to_dict(self, *):
```

rocketpy/_encoders.py
phmbressan: You are right, I changed what is a `signature` a few times during development. I will improve the docs.

rocketpy/motors/fluid.py
phmbressan: Well, actually I think I will simply revert to having the parameter `include_outputs` in all of them, and disable the warning if the method does not use it. That brings better standardization, since the call `to_dict(include_outputs=...)` would always work.

The `**` would be a better option than I used here, since it allows for the same interface as I mentioned. However, it also allows for other parameter names, and changing it in the future would be a breaking change.

MateusStano (commented):
Overall comments:

- `to_dict` and `from_dict` methods are missing docstrings in all classes

rocketpy/_encoders.py
phmbressan: Done in [c8e1682](https://github.com/RocketPy-Team/RocketPy/pull/679/commits/c8e168290dd97e7a4b4699e7c2c285ceea4a16f3).

rocketpy/motors/tank_geometry.py
```diff
@@ -373,6 +400,7 @@ def __init__(self, radius, height, spherical_caps=False, geometry_dict=None):
         """
         geometry_dict = geometry_dict or {}
         super().__init__(geometry_dict)
+        self.__input_radius = radius
```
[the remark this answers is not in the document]
phmbressan: Tough question, I revisited the naming and the issue is that `radius` is a `property` of the parent class. If it were a common method I would think it is a poor name, but it is reasonable for a `property`, but we can discuss changing it. The bigger issue to me is the parameter name being `radius`, which causes the conflict.

For now, I don't think there is another way than creating the `self.__input_radius` attribute. Even if we decide to change the name, it would be necessary a deprecation notice period, which still necessitates the `self.__input_radius`.

rocketpy/_encoders.py
```diff
@@ -1,15 +1,21 @@
...
-
 
 class RocketPyEncoder(json.JSONEncoder):
-    """NOTE: This is still under construction, please don't use it yet."""
+    """Custom JSON encoder for RocketPy objects. It defines how to encode
+    different types of objects to a JSON supported format."""
+
+    def __init__(self, *args, **kwargs):
+        self.include_outputs = kwargs.pop("include_outputs", True)
+        super().__init__(*args, **kwargs)
```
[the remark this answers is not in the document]
phmbressan: Changed in [3591837](https://github.com/RocketPy-Team/RocketPy/pull/679/commits/3591837f1d7b9220613dbaa11c18891ac4092749).

rocketpy/motors/tank_geometry.py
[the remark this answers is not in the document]
Gui-FernandesBR: @phmbressan would you mind opening an issue (or adding a TODO) so we keep following this `radius` naming problem?

rocketpy/environment/environment.py
```diff
@@ -2737,6 +2743,96 @@ def decimal_degrees_to_arc_seconds(angle):
...
+    def to_dict(self, include_outputs=False):
+        env_dict = {
+            "gravity": self.gravity,
+            "date": self.date,
+            "latitude": self.latitude,
+            "longitude": self.longitude,
+            "elevation": self.elevation,
+            "datum": self.datum,
+            "timezone": self.timezone,
+            "_max_expected_height": self.max_expected_height,
```
Gui-FernandesBR, line 2755:
```suggestion
            "max_expected_height": self.max_expected_height,
```

rocketpy/environment/environment.py
```diff
@@ -2737,6 +2743,96 @@ def decimal_degrees_to_arc_seconds(angle):
...
+            "atmospheric_model_type": self.atmospheric_model_type,
+            "pressure": self.pressure,
+            "barometric_height": self.barometric_height,
```
Gui-FernandesBR, line 2758: Isn't this an output? 

```suggestion
```

rocketpy/environment/environment.py
```diff
@@ -2737,6 +2743,96 @@ def decimal_degrees_to_arc_seconds(angle):
...
+            "temperature": self.temperature,
+            "wind_velocity_x": self.wind_velocity_x,
+            "wind_velocity_y": self.wind_velocity_y,
+            "wind_heading": self.wind_heading,
+            "wind_direction": self.wind_direction,
+            "wind_speed": self.wind_speed,
```
Gui-FernandesBR, line 2764: These are also outputs, no?
```suggestion
```

Do not accept the suggestion directly, remember to add the attributes in the includ_outputs section

Gui-FernandesBR (commented):
One last test we could run is: Create a Monte Carlo simulation with 100 simulations and verify if the output .txt files are still readable enough.

rocketpy/environment/environment.py
phmbressan: They are necessary to rebuild the `environment` without going back to the API IIRC, aren't they?

Gui-FernandesBR: I don't know about that, if you have the wind_y and wind_x, shouldn't we be able to calculate the wind direction, speed and heading? I guess that depends on the way you are rebuilding the Environment object.

All in all, it is late here, so I don't mind keeping the way it is. Let's go.

phmbressan:
> One last test we could run is: Create a Monte Carlo simulation with 100 simulations and verify if the output .txt files are still readable enough.

Good point. I have added some customization options to MonteCarlo so that one can choose whether to include all `Function` data or keep the way it is currently on `develop` that is only the `Function` string.

rocketpy/mathutils/function.py
Gui-FernandesBR: Solved with tests

rocketpy/environment/environment.py
Gui-FernandesBR: This is a blocker

Gui-FernandesBR (changes requested):
@phmbressan just a few more conversations need to be solved.

The PR looks good btw.

rocketpy/environment/environment.py
phmbressan: This is already solved. I forgot to resolve the conversation.

rocketpy/motors/tank_geometry.py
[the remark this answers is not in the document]
phmbressan: I have just created #759.

Gui-FernandesBR (approved):
LGTM, amazing work @phmbressan .

@GabrielBarberini should be interested in this feature. Please notice this is not a part of the public API!!

PR #681, opened 2024-09-06
DOC : Halcyon Flight Example

LUCKIN13:
## Pull request type

## Checklist

## Description
 Added files for Halcyon flight example from Aerospace Team Graz. 

## Additional Information

Seems like the data for the atmospheric model is not equal to the registered in the flight simulation before the launch.


Gui-FernandesBR:
Great work, @LUCKIN13 !!

I've upload the `.nc` file that will allow us to simulate all the flights for the EuRoC 2023. I think it is better to use these data rather than relying on Windy. I downloaded the data from Copernicus (ECMWF) server earlier this year, and it is the same file used to create the EuRoC 2023 post-flight report.

Gui-FernandesBR:
I think in this case it is worth it to add the second plot as well. Can you give it a try, @LUCKIN13 ? Now you should be able to make the GNSS data comparison using the .nc file

image

LUCKIN13:
Thanks for your input, Already added the latitude and longitude plots and they look great when compared to the reality. However, the descent phase of the flight seems to happen quicker. 
I triple checked all the parameters and it's all equal to the ones used in the last simulation.

Gui-FernandesBR (approved):
Great addition to our example gallery!

PR #689, opened 2024-09-13
ENH: generic surfaces pylint

Lucas-Prates:
## Pull request type

## Checklist

## Current behavior

Fixes pylint for the generic surfaces PR. 

## New behavior

There are two major changes:

- The `__init__` of `GenericSurface` and `LinearGenericSurfaces` have been reworked to receive a single argument for the coefficients instead of a multitude of arguments.

- I cleaned up the arguments from `compute_forces_and_moments` method a bit. I kept the `*args` because of the way the function is called in the `Flight` class . Basically, the class iterates over all aerodynamic surfaces and calls this function for each of them. In turn, this function has different signatures depending on which aerodynamic surface we have. As I see it, there are three approaches:
  1) Put `*args` to ignore some arguments for functions with smaller signatures;
  2) Do a flow control (if/else) and match the arguments accordingly;
  3) Use some tools from the `inspect` module to get the signature before hand, create an accordingly dictionary with arguments as keys and its values, and pass to the method.

Approach 2) seems a bit ugly, approach 3) seems unnecessarily complicated and might difficult maintaining the code in the future. Hence, I chose 1). If someone has a better suggestion, I am eager to see it!

## Breaking change

- [ ] Yes
- [x] No


rocketpy/rocket/aero_surface/generic_surface.py
```diff
@@ -17,12 +17,7 @@ def __init__(
         self,
         reference_area,
         reference_length,
-        cL=0,
-        cQ=0,
-        cD=0,
-        cm=0,
-        cn=0,
-        cl=0,
+        coefficients="all_null",
```
MateusStano, line 20: With this change, there is no need to make the `coefficients` parameter optional. Lets just leave it mandatory and remove the `"all_null"` option.

rocketpy/rocket/aero_surface/generic_surface.py
```diff
@@ -85,12 +84,55 @@ def __init__(
...
+
+    def _complete_coefficients(self, input_coefficients, default_coefficients):
+        """Completes coefficients dictionary from user input in __init__
+
+        Parameters
+        ----------
+        input_coefficients : str, dict
+            Coefficients dictionary passed by the user. If the user only specifies some
+            of the coefficients, the
+
```
MateusStano, line 120: Unfinished docs

rocketpy/rocket/aero_surface/generic_surface.py
```diff
@@ -85,12 +84,55 @@ def __init__(
...
+        """
+
+        if input_coefficients == "all_null":
+            coefficients = default_coefficients
+        else:  # complete user dictionary with null values for the coefficients
+            coefficients = input_coefficients
+            for coeff, value in default_coefficients.items():
+                if coeff not in coefficients.keys():
+                    coefficients[coeff] = value
+
```
MateusStano, line 134: I believe `coefficients = input_coefficients` does nothing here

rocketpy/simulation/flight.py
```diff
@@ -1671,6 +1671,7 @@ def u_dot_generalized(
         v = Vector([vx, vy, vz])  # CDM velocity vector
         e = [e0, e1, e2, e3]  # Euler parameters/quaternions
         w = Vector([omega1, omega2, omega3])  # Angular velocity vector
+        omega = [omega1, omega2, omega3]
```
MateusStano, line 1674: We already have `w` here, no need to create the `omega` list.

`Vector`s can be accessed just like lists with brackets (e.g. `[0]`), so using it shouldn't break anything

rocketpy/rocket/aero_surface/generic_surface.py
Lucas-Prates: Indeed, we need a deepcopy here.

rocketpy/rocket/aero_surface/generic_surface.py
Lucas-Prates: If it becomes mandatory, then we would have to pass coefficients even if we wanted the default with everything zero. Is it okay?

MateusStano: At least one coefficient should not be zero for the class to be useful, otherwise it doesn't do much. So I think this is the best way

rocketpy/rocket/aero_surface/generic_surface.py
MateusStano, line 134: I think this loop might be a little deficient.

What happens if the dict has a key that is not one of the coefficients names?

Also, since you are looping through the names, you should raise an error in case any keys in the user dict does not have an accepted string as key

Lucas-Prates: Added a `_check_coefficients` method to raise an error in such cases.

rocketpy/rocket/aero_surface/fins/fins.py
```diff
@@ -373,13 +373,9 @@ def compute_forces_and_moments(
         stream_mach,
         rho,
         cp,
-        _,
-        omega1,
-        omega2,
-        omega3,
+        omega,
```
Gui-FernandesBR, line 376: I would rename it to `w`.

Btw the docstring of this function is incomplete. We should either remove the papameters section or complete it

rocketpy/rocket/aero_surface/generic_surface.py
```diff
@@ -194,8 +258,8 @@ def compute_forces_and_moments(
             Center of pressure coordinates in the body frame.
         reynolds : float
             Reynolds number.
-        omega1, omega2, omega3 : float
-            Angular velocities around the x, y, z axes.
+        omega: tuple of float
```
Gui-FernandesBR, line 261:
```suggestion
        omega: tuple[float, float, float]
```

rocketpy/rocket/aero_surface/generic_surface.py
```diff
@@ -85,12 +85,80 @@ def __init__(
...
+        ------
+        ValueError
+            Raises a value error if the input coefficient has an invalid key
+        """
+        for key in input_coefficients.keys():
+            if key not in default_coefficients.keys():
+                raise ValueError(
+                    "Invalid coefficient name used in key! Check the "
+                    + "documentation for valid names."
+                )
```
Gui-FernandesBR, line 161: I won`t make it a suggestion, but just for you to think about, I think this could make the code a bit faster:

```python
invalid_keys = set(input_coefficients) - set(default_coefficients)
if invalid_keys:
    raise ValueError(
        f"Invalid coefficient name(s) used in key(s): {', '.join(invalid_keys)}. "
        "Check the documentation for valid names."
    )
```

Gui-FernandesBR (approved):
Good work, Lucas!! 

In the future I think we should make all aerosurfaces inherit from the generic surface class.

rocketpy/rocket/aero_surface/generic_surface.py
Lucas-Prates: Good suggestion, especially because it is more informative on which keys are invalid!

rocketpy/rocket/aero_surface/fins/fins.py
Lucas-Prates: Line 1513 of the `Flight` class sets `w = Vector([omega1, omega2, omega3])`. It seems to me that `omega` is a better name, so perhaps a better choice is to rename `w` to `omega`?

Gui-FernandesBR: I think `w` is used in our paper, but the user should really be reading "omega" instead of "double u".

I agree with us using `omega` from now on, but I don't think we should waster time changing that within this PR

PR #694, opened 2024-09-17
ENH: Free-Form Fins

MateusStano:
## Pull request type

## Checklist

## Description

Free Form Fins have been added!

Every calculation that is related to the free-form fins shape points (essentially almost everything that is calculated inside `evaluate_geometrical_parameters`) comes directly from Open Rocket. 

This means that the free-form fins in RocketPy are modeled exactly like Open Rocket, and only deviate when it comes to some effects related to Mach number variations (especially at supersonic speeds).

## Breaking change

- [ ] Yes
- [x] No


Gui-FernandesBR (commented):
Really nice feature, this is quite important for us.

1 - a good test would be to create a FreeFormFins set with the same size as a trapezoidal fins so we can validate that the center of pressure is the same when using both Trapezoidal and FreeForm fins
2 -  can you provide a minimum example of using the class? It is still a bit abstract to me; i.e. I can`t visualize how to use this new feature.

Well done!

rocketpy/plots/aero_surface_plots.py
```diff
@@ -380,6 +380,67 @@ def draw(self):
...
+
+    # pylint: disable=too-many-statements
+    def draw(self):
+        """Draw the fin shape along with some important information, including
+        the center line, the quarter line and the center of pressure position.
+
+        Returns
+        -------
+        None
+        """
```
Lucas-Prates, line 394: The draw method is not closing the fin draw when the last point is different than the initial point, as shown in the figure below

image

rocketpy/rocket/aero_surface/fins/free_form_fins.py
```diff
@@ -0,0 +1,379 @@
...
+        Function which defines the lift coefficient as a function of the angle
+        of attack and the Mach number. Takes as input the angle of attack in
+        radians and the Mach number. Returns the lift coefficient.
+    FreeFormFins.clalpha : float
+        Lift coefficient slope. Has units of 1/rad.
+    FreeFormFins.mac_length : float
+        Mean aerodynamic chord length of the fin set.
+    FreeFormFins.mac_lead : float
+        Mean aerodynamic chord leading edge x coordinate.
+    """
```
Lucas-Prates, line 100: It seems to me that some of these attributes do not make sense for free form fins. For instance, does `sweep_length` makes sense for `FreeFormFins`? It looked to me that it is something related to the geometry of a `TrapezoidalFins`.

rocketpy/rocket/aero_surface/fins/free_form_fins.py
```diff
@@ -0,0 +1,379 @@
...
+        for i in range(1, len(shape_points)):
+            if shape_points[i][1] > shape_points[i - 1][1] and down:
+                warnings.warn(
+                    "Jagged fin shape detected. This may cause small inaccuracies "
+                    "center of pressure and pitch moment calculations."
+                )
+                break
+            if shape_points[i][1] < shape_points[i - 1][1]:
+                down = True
+            i += 1
```
Lucas-Prates, line 161: What is a jagged fin? I looked at [OpenRocket documentation](https://openrocket.sourceforge.net/techdoc.pdf), but I am not sure what they define as a jagged fin. For instance, the image below show three different fins. For Fin 1, we get the jagged fin warning. For Fins 2 and 3, no warning is issued.
<details>
<summary>Fin 1 - Jagged Fin warning</summary>

image

</details>

<details>
<summary> Fin 2 - No Jagged fin warning</summary>

image

</details>

<details>
<summary> Fin 3 - No Jagged fin warning</summary>

image

</details>

According to the raised error, and OR documentation, jagged fins are have unstable/unreliable calculations. Are not Fin 2 and Fin 3 also problematic? Should they not be "jagged?"

If that is the case, then is "jagged" a synonym for non-convex?

In any case, I think we should separate this logic into a function in `mathutils` module depending on what is jagged:

- If jagged is not a synonym of non-convex, then we could create a `check_jagged_region` which just implements this, without the warning, and returns a boolean;
- If jagged means non-convex, I can implement a `check_convex_region` that returns `True` if the region is convex or `False` otherwise.

The class would then just call any of these functions and raise the warning if the condition is `False`.

rocketpy/plots/aero_surface_plots.py
```diff
@@ -380,6 +380,67 @@ def draw(self):
...
+        )
+
+        ax.add_line(yma_line)
+        ax.scatter(*cp_point, label="Center of Pressure", color="red", s=100, zorder=10)
+        ax.scatter(*cp_point, facecolors="none", edgecolors="red", s=500, zorder=10)
+
+        # Plot settings
+        ax.set_xlabel("Root chord (m)")
+        ax.set_ylabel("Span (m)")
+        ax.set_title("Trapezoidal Fin Cross Section")
```
Lucas-Prates, line 437:
```suggestion
        ax.set_title("Free Form Fin Cross Section")
```

Incorrect fin form in titles.

rocketpy/plots/aero_surface_plots.py
Gui-FernandesBR: how are we going to close it? linearly? we should provide an example (in a future PR maybe) showing case when the fin is set at a non-planar surface. for example when the fins are placed on the tail/transition

Lucas-Prates: Well, I assumed that the Fin is defined by a closed polygon. Indeed, the documentation states that "the last point should be the first one." So, if the user "forgot" to make sure the last point is the first one, we either should raise an error or just close it ourselves.

rocketpy/rocket/aero_surface/fins/free_form_fins.py
```diff
@@ -0,0 +1,379 @@
...
+        """
+        # pylint: disable=invalid-name
+        # pylint: disable=too-many-locals
+        # Calculate the fin area (Af) using the Shoelace theorem (polygon area formula)
+        Af = 0
+        for i in range(len(self.shape_points) - 1):
+            x1, y1 = self.shape_points[i]
+            x2, y2 = self.shape_points[i + 1]
+            Af += (y1 + y2) * (x1 - x2)
+        Af = abs(Af) / 2
```
Lucas-Prates, line 220: Could be a separated function named `compute_polygon_area` in `mathutils` module.

rocketpy/rocket/aero_surface/fins/free_form_fins.py
```diff
@@ -0,0 +1,379 @@
...
+        )  # Chord length for each spanwise section
+
+        # Discretize fin shape and calculate chord length, leading, and trailing edges
+        for p in range(1, len(self.shape_points)):
+            x1, y1 = self.shape_points[p - 1]
+            x2, y2 = self.shape_points[p]
+
+            # Compute corresponding points along the fin span (clamp to valid range)
+            prev_idx = int(y1 * 1.0001 / self.span * (points_per_line - 1))
+            curr_idx = int(y2 * 1.0001 / self.span * (points_per_line - 1))
```
Lucas-Prates, line 260: Why `1.0001`?

rocketpy/plots/aero_surface_plots.py
MateusStano: The model assumes that it is closed linearly. I will add the last line in the plot as well

rocketpy/rocket/aero_surface/fins/free_form_fins.py
MateusStano: So Open Rocket only checks the "jaggedness" of the first case because I believe that is the only case that makes sense for fins. It would totally be valid to add better checks just like you mentioned

I am not sure though if the assumptions that Barrowman makes for the cp position and clalpha would still hold for case 2 and 3. I personally think they would be too inaccurate.

I don't think this is needed for this PR, but those extra checks you suggested would be great to have in the future. Opening an issue describing those cases would be useful

rocketpy/rocket/aero_surface/fins/free_form_fins.py
MateusStano: I believe this is something related to numerical issues in Java. This comes directly from Open Rocket.

I removed it because I guess it is not needed for us

PR #695, opened 2024-09-22
ENH: Simulation Save and Load in JSON Files

phmbressan:
## Pull request type

## Checklist

## New behavior

This PR improves rocketpy encoding and introduces decoding functionalities. The way this works is:

- If the is a `to_dict` or `from_dict` method, they are used to perform the (de)encoding;
- Otherwise, a default encoding of all vars() and constructor based decoding is done.

## Breaking change

- [ ] Yes
- [X] No

## Additional information

PR #679 should be merged first.

Here are some of the steps that could be improved:

- [X] Verify why the test for hybrid motor flight requires higher tolerances;
    - Corrected, no higher tolerances needed anymore.
- [ ] Test if all of the atmospheric models of Environments can be decoded without requiring re-read of files or API;
- [X] Evaluate whether we should encode outputs or only the minimum required information to decode the class from its __init__;
    - Option for including outputs or not was added.
- [ ] (Optional) Expand implementation to `AirBrakes` + `Controllers` and new classes.

Some important remarks:

- The simple getting started simulation is essentially identical comparing the store & load, the only difference is the random lag and noise from Parachute class. I could not found a simple solution to reset those and always get the same results.


GabrielBarberini (commented):
LGTM

Gui-FernandesBR:
@phmbressan , as previously discussed, can you leave only one of the #679 and #695 open please?

PR #696, opened 2024-09-23
DEV: Prepare Master for v1.6

MateusStano:


rocketpy/simulation/flight.py
```diff
@@ -2058,37 +2007,44 @@ def get_solution_at_time(self, t, atol=1e-3):
...
 
     @funcify_method("Time (s)", "Z (m)", "spline", "constant")
     def z(self):
-        """Rocket z position as a Function of time."""
+        """Rocket z position relative to the launch pad as a Function of
+        time."""
         return self.solution_array[:, [0, 3]]
 
     @funcify_method("Time (s)", "Altitude AGL (m)", "spline", "constant")
     def altitude(self):
```
Gui-FernandesBR, line 2027: in the future the `env.elevation` parameter should be a function of x and y, or even a function of lat and lon

rocketpy/simulation/flight.py
Gui-FernandesBR, line 1: we really need to add a test to cover the `u_dot` method.
It could be a simple integration method with no special assertions.

rocketpy/simulation/flight.py
```diff
@@ -2113,88 +2069,94 @@ def e3(self):
...
+        """Aerodynamic moment in the rocket y-axis as a Function of time.
+        Sometimes referred to as yaw moment."""
         return self.__evaluate_post_process[:, [0, 11]]
 
-    @funcify_method("Time (s)", "M3 (Nm)", "spline", "zero")
+    @funcify_method("Time (s)", "M3 (Nm)", "linear", "zero")
     def M3(self):
-        """Aerodynamic bending moment in the same direction as the rocket's
-        axis of symmetry as a Function of time."""
+        """Aerodynamic moment in the rocket z-axis as a Function of time.
```
Gui-FernandesBR, line 2158: any reason for removing the bending term? Was the description wrong? just trying to understand

```suggestion
        """Aerodynamic bending moment in the rocket z-axis as a Function of time.
```

rocketpy/simulation/flight.py
```diff
@@ -2536,6 +2507,7 @@ def max_dynamic_pressure(self):
     # Total Pressure
     @funcify_method("Time (s)", "Total Pressure (Pa)", "spline", "zero")
     def total_pressure(self):
+        """Total pressure as a Function of time."""
```
Gui-FernandesBR, line 2510: What is a total pressure? 

Do you see the equation? I don't completely get it

rocketpy/simulation/flight.py
```diff
@@ -2639,35 +2615,105 @@ def drag_power(self):
...
+        # Stream velocity in standard aerodynamic frame
+        stream_velocity = -self.stream_velocity_body_frame
+        beta = np.arctan2(
+            -stream_velocity[:, 0],
+            stream_velocity[:, 2],
+        )  # x-z plane
+        return np.column_stack([self.time, np.rad2deg(beta)])
+
+    @funcify_method("Time (s)", "Angle of Attack (°)", "spline", "constant")
+    def angle_of_attack2(self):
```
Gui-FernandesBR, line 2708: what is the difference between the `angle_of_attack` and the `angle_of_attack2`?

rocketpy/rocket/rocket.py
```diff
@@ -525,18 +536,50 @@ def evaluate_center_of_pressure(self):
...
+        -------
+        self.surfaces_cp_to_cdm : dict
+            Dictionary mapping the relative position of each aerodynamic
+            surface center of pressure to the rocket's center of mass.
+        """
+        for surface, position in self.aerodynamic_surfaces:
+            self.evaluate_single_surface_cp_to_cdm(surface, position)
+        return self.surfaces_cp_to_cdm
+
+    def evaluate_single_surface_cp_to_cdm(self, surface, position):
```
Gui-FernandesBR, line 569: I'm under the impression this second method should be private.

The reason: why should we provide the public API with 2 different methods that does basically the same thing? It is similar to the discussion we had in the past with `add_surface(s)` methods

```suggestion
    def __evaluate_single_surface_cp_to_cdm(self, surface, position):
```

rocketpy/rocket/rocket.py
```diff
@@ -952,16 +996,21 @@ def add_surfaces(self, surfaces, positions):
...
+                    position = Vector(position)
                 self.aerodynamic_surfaces.add(surface, position)
+                self.evaluate_single_surface_cp_to_cdm(surface, position)
+        except TypeError:
+            if not isinstance(positions, (Vector, tuple, list)):
+                positions = Vector([0, 0, positions])
+            else:
+                positions = Vector(positions)
+            self.aerodynamic_surfaces.add(surfaces, positions)
+            self.evaluate_single_surface_cp_to_cdm(surfaces, positions)
```
Gui-FernandesBR, line 1014: I feel this is not the only place where we will have to deal with """vectorizing""" the position argument.
Also, the code is a bit hard to understand.

I like the idea of creating a "to_vector" method in another file and importing it in here. Smth like this:


```python
from collections.abc import Iterable

def to_vector(value):
    if isinstance(value, Vector):
        return value
    elif isinstance(value, Iterable):
        return Vector(value)
    else:
        return Vector([0, 0, value])

def process_surface(self, surface, position):
    position = to_vector(position)
    self.aerodynamic_surfaces.add(surface, position)
    self.evaluate_single_surface_cp_to_cdm(surface, position)

if isinstance(surfaces, Iterable) and isinstance(positions, Iterable):
    for surface, position in zip(surfaces, positions):
        process_surface(self, surface, position)
else:
    process_surface(self, surfaces, positions)
```

rocketpy/rocket/rocket.py
```diff
@@ -1274,6 +1323,84 @@ def add_elliptical_fins(
...
+        position,
+        cant_angle=0.0,
+        radius=None,
+        airfoil=None,
+        name="Fins",
+    ):
+        """Create a free form fin set, storing its parameters as part of the
+        aerodynamic_surfaces list. Its parameters are the axial position along
+        the rocket and its derivative of the coefficient of lift in respect to
+        angle of attack.
```
Gui-FernandesBR, line 1339:
```suggestion
        """Create a free form fin set, storing its parameters as a part of the
        ``aerodynamic_surfaces`` list. Its parameters are the axial position along
        the rocket and the derivative of its lifit coefficient with respect to the
        angle of attack.
```

rocketpy/rocket/rocket.py
```diff
@@ -1274,6 +1323,84 @@ def add_elliptical_fins(
...
+
+        Parameters
+        ----------
+        n : int
+            Number of fins, from 2 to infinity.
```
Gui-FernandesBR, line 1344:
```suggestion
            Number of fins, must be greater or equal to 2.
```

rocketpy/rocket/rocket.py
```diff
@@ -1274,6 +1323,84 @@ def add_elliptical_fins(
...
+        shape_points : list
```
Gui-FernandesBR, line 1345:
```suggestion
        shape_points : list[tuple[float, float]]
```

rocketpy/rocket/aero_surface/fins/free_form_fins.py
```diff
@@ -0,0 +1,368 @@
...
+                    chord_length[i] += x
+
+        # Replace infinities and handle invalid values in chord_lead and chord_trail
+        for i in range(points_per_line):
+            if (
+                np.isinf(chord_lead[i])
+                or np.isinf(chord_trail[i])
+                or np.isnan(chord_lead[i])
+                or np.isnan(chord_trail[i])
+            ):
```
Gui-FernandesBR, line 285: https://numpy.org/doc/stable/reference/generated/numpy.nan_to_num.html

docs/user/rocket/generic_surface.rst
```diff
@@ -0,0 +1,436 @@
...
+Where:
+
+- :math:`L_{ref}` is the reference length used to calculate the coefficients.
+  Commonly the rocket's diameter is used as the reference length.
+
+
+Aerodynamic angles
+~~~~~~~~~~~~~~~~~~
+
+There aerodynamic angles are defined in two different ways in RocketPy:
```
Gui-FernandesBR, line 117:
```suggestion
The aerodynamic angles are defined in two different ways in RocketPy:
```

Gui-FernandesBR (approved):

rocketpy/simulation/flight.py
MateusStano: Bending moment is usually used for solid mechanics analysis and might be misleading in this case since we are just calculating the moment generated by aerodynamic forces.

It might still be equal to the bending moment, but that could lead to the thought that something related to solid mechanics was involved in its calculation, which it wasn't

rocketpy/simulation/flight.py
MateusStano: I do not know where the expression comes from. I just added the missing docs

rocketpy/rocket/rocket.py
MateusStano: I'll add a TODO on the code for this

Gui-FernandesBR (approved):

PR #698, opened 2024-10-01
BUG: `savetxt` Not Accepting lambda Functions

MateusStano:
## Description

Currently, lambda defined `Function` are not correctly saved when using `savetxt`. The problem was that a np.array was being passed for the evaluation of the lambda.

The solution was to use `Function`'s  `get_value_opt` to call on the lambdas.


Gui-FernandesBR (approved):
Tests passing?

MateusStano:
> Tests passing?

they are passing locally. The problem on actions is unrelated to the PR

PR #700, opened 2024-10-01
BUG: Single Point Functions Can Not Be Defined

MateusStano:
## Description

Functions of a single point (`Function([[1,2]]`) do not currently work because of an added validation for source shape.

I have corrected the validation to account for one point Functions


Gui-FernandesBR:
Are tests really passing?

Gui-FernandesBR (approved):

MateusStano:
> Are tests really passing?

locally yes

PR #701, opened 2024-10-03
BUG: Rail Buttons Not Accepted in Add Surfaces

MateusStano:
In v1.6.0 rail buttons are no longer accepted in the `add_surfaces` method of the rocket class. This PR reintroduces this functionality as to avoid breaking changes

This was unfortunately something added in previous versions. Rail buttons do not behave like aerodynamic surfaces and should not be grouped together with them. A rocket can only have one set of rail buttons, this is why we have a `set_rail_buttons` and not a `add_rail_buttons` method.

Tests are passing locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@ Attention: The newest changes should be on top -->
 
 ### Fixed
 
-- 
+- BUG: Vector encoding breaks MonteCarlo export. [#704](https://github.com/RocketPy-Team/RocketPy/pull/704)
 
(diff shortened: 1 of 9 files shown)


Gui-FernandesBR (approved):
The CHANGELOG needs to be updated!!! Is the develop branch already updated from master?

Please wait for the tests to be all checked.

We should pack a new release soon.

Gui-FernandesBR (approved):

Gui-FernandesBR:
LGTM

Issue #699, opened 2024-10-01
ENH: Callback function for collecting additional data from Monte Carlo sims


emtee14:
**Is your feature request related to a problem? Please describe.**

When running Monte Carlo sims currently you are limited to the predefined variables when exporting the results. For certain applications, I'd like to be able to get other data points that aren't already in the default list or generate my own variables from the Flight class after the sim has run and add them to the exported dict. My primary use for this would be to export environmental data as at the moment when aggregating multiple sims together with multiple different dates/times or weather forecasts the only way to differentiate is by embedding the data in the file name which becomes quite arduous when you have a lot of additional data you want to add. It can also be useful to export sensor data since it can then be used for sensor mocking.

**Describe the solution you'd like**

My idea to fix this would be a function you pass to the MonteCarlo class which takes a Flight as an argument and returns a dict which can then be appended to the output dict of the simulation. This would allow other data points to be exported like environmental data and sensor data from the rocket.

**Additional context**

emtee14:
Also to add myself or someone from LURA is happy to implement this

Gui-FernandesBR:
related: #269

Gui-FernandesBR:
This is coming as a new feature in the next release!

PR #702, opened 2024-10-06
ENH: Callback function for collecting additional data from Monte Carlo sims

emtee14:
## Pull request type

## Checklist

## Current behaviour

Implementation for this feature suggestion #699 . At the moments users are limited to set variables when exporting data from Monte Carlo experiments.

## New behaviour
Now a function can be passed to the MonteCarlo class which takes a Flight object as an argument and returns a dict to be appended to the default outputs.

## Breaking change

- [x] No

## Additional information

At the moment the implementation works but there is an existing issue with the RocketPyEncoder for writing the inputs to the .txt file as JSON strings. I've also made it check if the function overwrites existing variables though I feel this may not be necessary. I also had to fix how the previous results are imported as they are errors since text results can now be exported, see line 671.


Gui-FernandesBR:
Thanks for opening the PR @emtee14 , we will get back to you after the EuRoC competition, when our development team will be more available.

Lucas-Prates:
@emtee14 thank you very much for opening this PR! I have been assigned to help you out with this feature. Your description in #699 and the (nice) implementation here give me a good idea of what you want to achieve. 

To be more thorough, could you provide me an example of how you want to use it? Possibly a snippet of the code creating the `export_function` and calling the `MonteCarlo` class. It would be even better if it is an actual (maybe simplified?) example of what you are doing, e.g. exporting the environmental data mentioned in #699. If you want to give an example that is easy to run, consider the _monte_carlo_class_usage.ipynb_ notebook and assume that this snippet would replace the code in the cell that calls the `MonteCarlo` class. 

This would ensure that the design is best suited for the goal, and we could also add it as an example in the Monte Carlo usage notebook.

emtee14:
Hi yes, I'll make an example that can be added to the docs. Another one of the uses we've found for it is exporting sensor data which we then use to run on our flight computer emulator allowing us to test firmware.

Lucas-Prates:
This PR is ready for review. Along with the callback implementation provided by @emtee14, some error checks and tests were implemented. I provided a first simplified example on how to use this feature in the end of _monte_carlo_class_usage.ipynb_ notebook, but am eager to see the example of @emtee14. Of course, his example can be provided in another PR as well, so there is no need to rush.

rocketpy/simulation/monte_carlo.py
```diff
@@ -359,6 +371,22 @@ def __export_flight_data(
...
+                raise ValueError(
+                    "An error was encountered running your custom export function. "
+                    "Check for errors in 'export_function' definition."
+                ) from e
+
+            for key in additional_exports.keys():
+                if key in self.export_list:
+                    raise ValueError(
+                        f"Invalid export function, returns dict which overwrites key, '{key}'"
+                    )
```
Gui-FernandesBR, line 387: We only need to do this once, I recommend moving this validation to somewhere in the init method so we don't need to re-run it every time.

rocketpy/simulation/monte_carlo.py
```diff
@@ -654,9 +682,15 @@ def set_processed_results(self):
         """
         self.processed_results = {}
         for result, values in self.results.items():
-            mean = np.mean(values)
-            stdev = np.std(values)
-            self.processed_results[result] = (mean, stdev)
+            try:
+                if isinstance(values[0], float):
```
Gui-FernandesBR, line 686: I think checking for float is a bit dangerous here.
Isn't there any other types that would need to be included here? @phmbressan what do you think about it?

Btw, @Lucas-Prates for future PRs, we may want to start considering other statistics as well, such as median, min, max, etc.

rocketpy/simulation/monte_carlo.py
```diff
@@ -104,6 +110,11 @@ def __init__(
             `out_of_rail_stability_margin`, `out_of_rail_time`,
             `out_of_rail_velocity`, `max_mach_number`, `frontal_surface_wind`,
             `lateral_surface_wind`. Default is None.
+        export_function : callable, optional
+            A function which gets called at the end of a simulation to collect
+            additional data to be exported that isn't pre-defined. Takes the
+            Flight object as an argument and returns a dictionary. Default is None.
```
Gui-FernandesBR, line 116: I recommend specifying the dictionary structure. What are the keys and values? Can you give an example to anticipate users' curiosity?

Other thing is that once I read the name `export_function` I automatically imagine a callback function that returns a float, for example:

```python
def mach_number(flight: Flight) -> float:
    return ...

export_function= mach_number
```

Therefore, unless this is only me that has such impression, I believe we could propose a better name for the argument. Maybe `data_collector`or `additional_data_collector`.

Gui-FernandesBR (commented):
have you thought about the possibility of using both `export_list` and `export_function` at the same time?

rocketpy/simulation/monte_carlo.py
Gui-FernandesBR: @Lucas-Prates can you open up an issue so we tackle that point later? Thx

Gui-FernandesBR:
> > have you thought about the possibility of using both `export_list` and `export_function` at the same time?
> 
> You mean using both in an example or replacing `export_list` functionality completely using `export_function`?

have you thought about what would happen if someone uses both the `export_list` and `export_function` arguments at the same time (i.e. simultaneously)? Would that be a problem?

rocketpy/simulation/monte_carlo.py
Gui-FernandesBR: > It is a good idea to run this only once, but moving it to the init method might be problematic in the current implementation. The reasoning is as follows: to validate that there are no keys being overwritten, we have to call `export_list` so we can have the dictionary outputted by the callback. However, in order to make that call, we have to have the data from the simulations. I see two options:
> 
> 1. Create a "mock dictionary", or at least its keys, from `export_function` and make the test in the init method;
> 2. Create a flag to check if `export_function` was validated. On the first call of these lines (inside `__export_flight_data`, we make the test and set the flag, hence halting further validations.

@Lucas-Prates if we have a `__check_export_list` method, I think we should also have a `__check_export_function` as well. I like the idea of a `mock_flight`, but I don't know how hard it would be to make it happen.

I believe my other comment is related to this. If instead of a "callback function" that evaluates and returns a dictionary, maybe we should receive a dictionary with different callback functions. That way we could easily get the expected key names.

Other option is to not make any validation, let the user deal with the fact they will be overwriting the values if use the same name of variables in the dictionary.

Lucas-Prates: I like the idea of separating the callback functions. Thinking more carefully, it would even allow for more descriptive error raises when there is an implementation error by the user.

Lucas-Prates:
No problem at all! Indeed, the example I provided in the notebook uses both `export_list` and `export_function` simultaneously.

rocketpy/simulation/monte_carlo.py
phmbressan, line 387: This `Exception` message could be much clearer. I believe it would be interesting to specify that a key from the `export_list` conflicts with a key from the `export_function` dictionary. Then the user know exactly which parameters to look.

phmbressan: My two cents on the matter:

1. My preference would be not to validate and document that `export_function` overwrites identical keys of `export_list`;
2. Another option would be using the `flight` parameter from `__init__` as the discussed `mocked_flight`, it would also compute whether the `export_function` as a whole works.

rocketpy/simulation/monte_carlo.py
```diff
@@ -654,9 +682,15 @@ def set_processed_results(self):
...
-            self.processed_results[result] = (mean, stdev)
+            try:
+                if isinstance(values[0], float):
+                    mean = np.mean(values)
+                    stdev = np.std(values)
+                    self.processed_results[result] = (mean, stdev)
+                else:
+                    self.processed_results[result] = (None, None)
+            except TypeError:
+                self.processed_results[result] = (None, None)
```
phmbressan, line 693: Now I understand a bit better the discussion on "Is the average even meaningful?". I would simply compute the average if the value "allows" it. Like:

```py
try:
    self.processed_results[result] = (np.mean(values), np.std(values))
except TypeError:
    self.processed_results[result] = (None, None)
```

Checking for a `float` seems odd to me, since `int`, `Decimal` etc could absolutely be desired for averaging things.

Gui-FernandesBR: I agree. Overall, it is not our responsibility to ensure every statistic output makes sense or not. Users should be responsible to critique what they want to average or not, regardless of the type of the variable.

Lucas-Prates:
Sorry for the delay, but this PR is ready for review again. The following changes were made:
1) The `export_function` argument had its name changed to `data_collector`;
2) `data_collector` is a now a `dict` of python `callable`s, reflecting the idea that we have several distinct callback functions instead of a "big" callback function as implemented previously;
3) The check if computing the summary made sense by checking for a `float` was removed;
4) The refactoring allowed for the validation of the `data_collector` to be done in the initialization of the `MonteCarlo` class;
5) Docs and tests were updated accordingly.

rocketpy/simulation/monte_carlo.py
```diff
@@ -143,7 +152,8 @@ def __init__(
         self._last_print_len = 0  # used to print on the same line
 
         self.export_list = self.__check_export_list(export_list)
-        self.export_function = export_function
+        self._check_data_collector(data_collector)
```
phmbressan, line 155: Optional fix, but to keep consistency with the `__check_export_list`, I would double underscore the method to `__check_data_collector`.

phmbressan (approved):

rocketpy/simulation/monte_carlo.py
Lucas-Prates: I implemented with double underscores, but could not call them on the test suite.

phmbressan:
Also, @Lucas-Prates could you add this PR to the `CHANGELOG.md`?

PR #703, opened 2024-10-06
BUG: Initialize `_Controller` Init Parameters 

MateusStano:
Some arguments passed to `__init__` of the `_Controller` class were not being saved. This PR fixes this issue

This is important for controller/airbrakes serialization.

Should be merged after #701

### DIFF
--- a/rocketpy/control/controller.py
+++ b/rocketpy/control/controller.py
@@ -84,6 +84,7 @@ class _Controller:
         None
         """
         self.interactive_objects = interactive_objects
+        self.base_controller_function = controller_function
         self.controller_function = self.__init_controller_function(controller_function)
         self.sampling_rate = sampling_rate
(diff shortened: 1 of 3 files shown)


rocketpy/control/controller.py
```diff
@@ -84,8 +84,10 @@ def __init__(
         None
         """
         self.interactive_objects = interactive_objects
+        self.base_controller_function = controller_function
```
Gui-FernandesBR, line 87: @MateusStano I have no idea why you want to save this attribute as `base_controller_function`instead of simply `controller_function`. I'm approving this since the class is still private.

Gui-FernandesBR (approved):
@MateusStano I know you don't like the CHANGELOG.md file.
But keep in mind that, when you don't update it, someone else you have to do it for you.
In the case of last 3 PRs, that one was me.

PR #710, opened 2024-10-09
BUG: Fix Motor Zero Dry Mass Check

laurapgp:
fixed zero mass check for motor dry_mass. Local tests are passing.


Gui-FernandesBR (approved):
Please wait for the tests on CI,

Please update the CHANGELOG.


The setter is not checking for negative mass values. I understand it's not the focus of this PR. We can improve it in the future.

PR #719, opened 2024-11-02
DEV: updates workflows to include python 3.13

Gui-FernandesBR:
with also a few other additions

### DIFF
--- a/.github/workflows/test-pytest-slow.yaml
+++ b/.github/workflows/test-pytest-slow.yaml
@@ -4,6 +4,14 @@ on:
   schedule:
     - cron: "0 17 * * 5"  # at 05:00 PM, only on Friday
       timezone: "America/Sao_Paulo"
+  push:
+    branches:
+      - main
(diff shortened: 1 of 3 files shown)


Gui-FernandesBR:
@phmbressan apparently we cannot support python 3.13 yet ? Tests are failling.

I will turn this PR into a draft

phmbressan:
> apparently we cannot support python 3.13 yet ? Tests are failling.
> 
> I will turn this PR into a draft

When I tested I ran the standard test suite, forgot about the doctest. The MacOS `pytest` seems to be fine, which is good news. 

I will check if there is anything we can do to get the same results from the doctests as before, but it seems to be a numerical precision change.

phmbressan:
Locally, the same doctest is passing fine on Linux with 3.13. I believe we should round the values, since the difference is at the limit of numerical precision. Hard to do that with a "simple one liner" for the doctest.

PR #722, opened 2024-11-07
DEV: Updates master for v1.6.2

Gui-FernandesBR:


rocketpy/tools.py
```diff
@@ -374,126 +373,127 @@ def inverted_haversine(lat0, lon0, distance, bearing, earth_radius=6.3781e6):
...
+    apogee_x=np.array([]),
+    apogee_y=np.array([]),
+    impact_x=np.array([]),
+    impact_y=np.array([]),
+    n_apogee=[1, 2, 3],
+    n_impact=[1, 2, 3],
+    apogee_rgb=(0, 1, 0),
+    impact_rgb=(0, 0, 1),
+    opacity=0.2,
+):  # pylint: disable=dangerous-default-value
```
phmbressan, line 442: Is it alright to simply disable this warning instead of changing the defaults? I understand that is not very used outside of internal code, I am asking just to confirm the reasoning.

rocketpy/tools.py
```diff
@@ -515,44 +515,49 @@ def generate_monte_carlo_ellipses_coordinates(
...
+            origin_lon,
+            math.sqrt(x**2 + y**2),
+            math.degrees(math.atan2(x, y)),
+            earth_radius=6.3781e6,
+        )
+        for x, y in points
+    ]
+
+
+def __generate_ellipse_points(ellipse, resolution: int):
```
phmbressan, line 543: Small type hint. I do not oppose adding them in a module whole module, but since this is the only parameter that has it, I believe it is a "typo".

rocketpy/tools.py
Gui-FernandesBR: Yes it is more than OK,
I just don`t even know why I didnt disable this pylint rule yet. 
Leave it there.

phmbressan (approved):

PR #728, opened 2024-11-09
DOC: fixed documentation about spherical caps

alessio26gas:
## Pull request type

- [ ] Other

## Current behavior

The existing documentation for the ``spherical_caps`` parameter currently states:
> The ``spherical_caps`` parameter is optional and defaults to ``False``. If set
> to ``True``, the tank will be defined as a cylinder with flat caps.
> If True, the tank will have spherical caps.

## New behavior

This should be updated to:
> The ``spherical_caps`` parameter is optional and defaults to ``False``. If set
> to ``True``, the tank will have spherical caps at the top and bottom with the
> same radius as the cylindrical part. If set to ``False``, the tank will
> be defined as a cylinder with flat caps.

## Breaking change

- [ ] Yes
- [x] No

### DIFF
--- a/docs/user/motors/tanks.rst
+++ b/docs/user/motors/tanks.rst
@@ -93,8 +93,9 @@ The predefined ``CylindricalTank`` class is easy to use and is defined as such:
 
 .. note::
   The ``spherical_caps`` parameter is optional and defaults to ``False``. If set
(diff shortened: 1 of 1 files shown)


Gui-FernandesBR (approved):
Nice catch!

PR #731, opened 2024-11-15
GIT: Updates master to v1.7.0

Gui-FernandesBR:
Quick and easy approval


phmbressan (dismissed):

Gui-FernandesBR:
@phmbressan I was reflecting on the `filename` addition and got to a question. I'd like to ask your opinion on that.

If we release this version as it is, in the future we may suffer when adding new arguments to methods that has the filename argument. What will happen is that the `filename` will no longer be the last argument if we add new arguments to these methods.

I wonder if this is the case where we would benefit from doing something like this:

```python
def thios_method(self, arg1, arg2, *, filename=None):
```

That way we could (potentially) add new arguments before the filename, this enforcing the filename always to be the last argument present in a method.

Gui-FernandesBR:
image

phmbressan:
> If we release this version as it is, in the future we may suffer when adding new arguments to methods that has the filename argument. What will happen is that the `filename` will no longer be the last argument if we add new arguments to these methods.
> 
> I wonder if this is the case where we would benefit from doing something like this:
> 
> ```python
> def thios_method(self, arg1, arg2, *, filename=None):
> ```
> 
> That way we could (potentially) add new arguments before the filename, this enforcing the filename always to be the last argument present in a method.

Well noticed. This seems like a good example of a future situation that we might get stuck with a breaking change. The solution you provided is alright, although in this case I find more straightforward to put it in the `**kwargs`. In both cases, I believe using it positionally only would cause an error.

```python
def this_method(self, arg1, arg2, **kwargs):
```

I believe we should change the current behavior to one of the two solutions. Do you have any preferences?

Gui-FernandesBR:
@phmbressan just to be clear... we would be removing the `filename` argument to `**kwargs` ?

phmbressan:
> just to be clear... we would be removing the `filename` argument to `**kwargs` ?

That is one possibility. It makes more sense if we had many paramters, like matplotlib does here (scroll down to the kwargs section):
 https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.plot.html

The way you solved it should also work, if you prefer. The difference between them is that kwargs would allow any names to be a "named parameter".

Gui-FernandesBR:
> > just to be clear... we would be removing the `filename` argument to `**kwargs` ?
> 
> That is one possibility. It makes more sense if we had many parameters, like matplotlib does here (scroll down to the kwargs section): https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.plot.html
> 
> The way you solved it should also work, if you prefer. The difference between them is that kwargs would allow any names to be a "named parameter".

@phmbressan I reflected on both options, they both are pretty similar to be honest. The idea is to enforce that no one ever uses the filename as a positional argument. 

The only advantage I see in using the `*` operator is for type hinting purposes, since it would allow us to specify the type of the inputs. I know we do not use type hints in this project yet, but I have a dream one day someone will convince @MateusStano of using that!

Aside that fact, I see no other reason for any preference here.
What do you think? Do you agree?

I believe we should make a new PR implementing the solution we agree on. I am more inclined to use the `*` now, but honestly I would easily accept PRs with the alternative solution as well.

rocketpy/simulation/monte_carlo.py
```diff
@@ -104,6 +113,17 @@ def __init__(
...
+            A dictionary whose keys are the names of the exported variables
+            and the values are callback functions. The keys (variable names) must not
+            overwrite the default names on 'export_list'. The callback functions receive
+            a Flight object and returns a value of that variable. For instance
+
+            .. code-block:: python
+                custom_data_collector = {
+                    "max_acceleration": lambda flight: max(flight.acceleration(flight.time)),
+                    "date": lambda flight: flight.env.date,
+                }
```
Gui-FernandesBR, line 126: I wonder how this code block is being rendered on our docs page

Gui-FernandesBR:
> > > just to be clear... we would be removing the `filename` argument to `**kwargs` ?
> > 
> > 
> > That is one possibility. It makes more sense if we had many parameters, like matplotlib does here (scroll down to the kwargs section): https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.plot.html
> > The way you solved it should also work, if you prefer. The difference between them is that kwargs would allow any names to be a "named parameter".
> 
> @phmbressan I reflected on both options, they both are pretty similar to be honest. The idea is to enforce that no one ever uses the filename as a positional argument.
> 
> The only advantage I see in using the `*` operator is for type hinting purposes, since it would allow us to specify the type of the inputs. I know we do not use type hints in this project yet, but I have a dream one day someone will convince @MateusStano of using that!
> 
> Aside that fact, I see no other reason for any preference here. What do you think? Do you agree?
> 
> I believe we should make a new PR implementing the solution we agree on. I am more inclined to use the `*` now, but honestly I would easily accept PRs with the alternative solution as well.

@phmbressan we need to address this filename/kwargs situation. It is currently blocking us from our next release.
Are we really going to change the filename arguments? How are we going to do so, and who is assigned to address it?
I don't have time to work on this right now, sadly.

Ideally, the next release should happen still in November.

rocketpy/simulation/monte_carlo.py
Gui-FernandesBR: Docs are breaking. @Lucas-Prates 

https://readthedocs.org/projects/rocketpyalpha/builds/26380046/

Gui-FernandesBR: fixed

PR #732, opened 2024-11-18
DOC: improvements to developers documentation

Gui-FernandesBR:
Our development documentation really needs to be updated.
The last major additions to these pages were done more than 3 years ago.

I have re-strucctured the dev docs pages, but I'd appreciate if we could split these pages among our team of developers so we can have a more diverse point of view throughout our documentation.

These are the pages that still needs to be finished and reviewed :
- [x] setting_up.rst
- [x] Running RocketPy as a Developer <rocketpy_as_developer.rst>
- [x] Your First Pull Request (PR) <first_pr.rst>
- [x] Solving git conflicts <conflicts.rst>
- [x] Style Guide <style_guide>
- [x] Testing Guidelines <testing.rst>
- [x] RocketPy with Docker <docker.rst>
- [x] Building the Documentation <build_docs.rst>
- [x] Pro Tips <pro_tips.rst>

PR #733, opened 2024-11-20
DOC: RED Flight Example 

LUCKIN13:
## Pull request type

## Checklist

## Additional information
 This branch still needs some data reviews since the simulation file was developed in an older version of RocketPy. However, reviews can already be made, since the example should only suffer minor changes, if needed. 
Changelog and  Index for examples will be updated in other flight example, to prevent conflicts.


Gui-FernandesBR (changes requested):
Once again, naming the folder after the team name is risky, since this may not be the last rocket we use as an example from this team.

The apogee should be quite near 3000m, since the rocket has an airbrake model. Are you sure the air brakes have worked during this simulation? 8.6% of accuracy is too much, I was expecting something below 2% for this model (of course I could be wrong about this.

I personally recommend you to not mention the "last simulated apogee before launch", instead just mention the simulation we have obtained right now.

Plus, we need to add this new rocket to the flight examples comparison plot.

Gui-FernandesBR:
@LUCKIN13 before merging this PR, I believe we should add it to the CHANGELOG file.
#732 is similar.

LUCKIN13:
> The apogee should be quite near 3000m, since the rocket has an airbrake model. Are you sure the air brakes have worked during this simulation? 8.6% of accuracy is too much, I was expecting something below 2% for this model (of course I could be wrong about this.

The airbrakes did work, since the apogee was 3300 meters and is now 3180m. I have an idea for an improvement to that value, based on using the actuation curve of the controller designed by the team.

Gui-FernandesBR (commented):
@LUCKIN13 it looks like you have duplicated the example notebook.

image

LUCKIN13:
@Gui-FernandesBR thanks for calling out my distraction! I think it's finally ready!

Gui-FernandesBR (approved):
LGTM

PR #743, opened 2024-12-02
BUG: Correctly update atmospheric conditions after changing date and location

WilliamArmst:
## Pull request type

Bugfix

## Checklist

## Current behavior

If the environment date is updated after setting the atmospheric model, the function call to update the model to that date did not pass ``self.atmospheric_model_type`` when it is required by the function.  This led to the file being passed in as the type.

## New behavior

Behaves as expected

## Breaking change

- [ ] Yes
- [x] No

## Additional information

N/A

### DIFF
--- a/rocketpy/environment/environment.py
+++ b/rocketpy/environment/environment.py
@@ -688,7 +688,9 @@ class Environment:
             "Ensemble",
         ]:
             self.set_atmospheric_model(
-                type=self.atmospheric_model_type, file=self.atmospheric_model_file, dictionary=self.atmospheric_model_dict
+                type=self.atmospheric_model_type,
+                file=self.atmospheric_model_file,
+                dictionary=self.atmospheric_model_dict,
             )
 
     def set_location(self, latitude, longitude):
@@ -726,7 +728,9 @@ class Environment:
             "Ensemble",
         ]:
             self.set_atmospheric_model(
-                type=self.atmospheric_model_type, file=self.atmospheric_model_file, dictionary=self.atmospheric_model_dict
+                type=self.atmospheric_model_type,
+                file=self.atmospheric_model_file,
+                dictionary=self.atmospheric_model_dict,
             )
 
     def set_gravity_model(self, gravity=None):
--- a/rocketpy/environment/environment.py
+++ b/rocketpy/environment/environment.py
@@ -688,7 +688,7 @@ class Environment:
             "Ensemble",
         ]:
             self.set_atmospheric_model(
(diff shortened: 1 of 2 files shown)


rocketpy/environment/environment.py
```diff
@@ -688,7 +688,7 @@ def set_date(self, date, timezone="UTC"):
             "Ensemble",
         ]:
             self.set_atmospheric_model(
-                self.atmospheric_model_file, self.atmospheric_model_dict
+                self.atmospheric_model_type, self.atmospheric_model_file, self.atmospheric_model_dict
```
Gui-FernandesBR, line 691: My suggestion is for us to prevent the same error from happening again:

```suggestion
                type=self.atmospheric_model_type, file=self.atmospheric_model_file, dictionary=self.atmospheric_model_dict
```

After that, we need to run black again

rocketpy/environment/environment.py
```diff
@@ -726,7 +726,7 @@ def set_location(self, latitude, longitude):
...
```
Gui-FernandesBR, line 729:
```suggestion
                type=self.atmospheric_model_type, file=self.atmospheric_model_file, dictionary=self.atmospheric_model_dict
```

Gui-FernandesBR (changes requested):
Nice catch!

Can you run black again please? @WilliamArmst 

I am wondering whether we should release this BUG in a minor version (v1.7.1).

WilliamArmst:
I'm not sure what running black again means, if it's the local tests I'll have to wait until I get back from classes.

This is a pretty small fix and even without it, I think updating the date sometimes works anyways for some reason?  It doesn't make sense to me, but it only became an issue for me when I tried doing batch simulations with multiple dates.  Ultimately up to you if you want to put it into a bugfix.

WilliamArmst:
Is it failing because the linter wants this?

```
self.set_atmospheric_model(
        type=self.atmospheric_model_type,
        file=self.atmospheric_model_file,
        dictionary=self.atmospheric_model_dict,
)
```

Gui-FernandesBR:
> Is it failing because the linter wants this?
> 
> ```
> self.set_atmospheric_model(
>         type=self.atmospheric_model_type,
>         file=self.atmospheric_model_file,
>         dictionary=self.atmospheric_model_dict,
> )
> ```

Yes!

The code line should not exceeds 88 length positions.

Black is one of our linters. Running `make format` should work. you can learn more about it here: https://docs.rocketpy.org/en/latest/development/style_guide.html

Gui-FernandesBR (approved):

Issue #664, opened 2024-08-19
ENH: Create a dataset of pre-registered motors.


Gui-FernandesBR:
Something OpenRocket does pretty well is to provide a wide range of available motors already on the software.
Maybe we could do something similar with .json files.

There's already na open PR solving this kind of issue in the first rocketpy hackathon (2020)

Gui-FernandesBR:
<img width="737" height="671" alt="Image" src="https://github.com/user-attachments/assets/f0110da6-08d1-4fcf-8ca7-44b03d1a7be9" />

ting-hong-shieh:
I checked the current packaging state at `cb6106a717207dd8fc2dfe1446d80ff75022f21b` to narrow down what remains after #744.

The source tree contains 64 `.eng` files under `data/motors`, but a wheel built from that commit contains none of them:

| Check | Result |
| --- | ---: |
| Source `.eng` files | 64 |
| Wheel entries | 138 |
| `.eng` files in the wheel | 0 |

Command used:

```console
$ uv build --wheel --out-dir /tmp/rocketpy-dist-664
Successfully built /tmp/rocketpy-dist-664/rocketpy-1.13.0-py3-none-any.whl

$ unzip -Z1 /tmp/rocketpy-dist-664/rocketpy-1.13.0-py3-none-any.whl | grep -E '(^|/)data/motors/|\.eng$'
# no output
```

`pyproject.toml` currently declares only `rocketpy.plots/assets/*.stl` as package data. This leaves the motor files unavailable to users who install RocketPy from a wheel.

Could a maintainer confirm which remaining scope is intended?

1. Package the existing catalog, add a small manifest with source and license information, expose a discovery API, and verify that every bundled file can be parsed; or
2. Define a richer schema containing enough metadata to construct complete `SolidMotor` objects.

I recommend option 1 as a narrow first PR. It fixes the distribution gap without committing the project to a larger motor metadata schema. If that matches the intended scope, I can implement it with source-tree and installed-wheel tests.

ting-hong-shieh:
I reviewed the image attached when this issue was reopened. It makes the intended direction more specific than the two scopes I listed above: move the catalog to a separate `rocketpy-motor-dataset` repository/package, add RocketPy optional extras for that package, and load its resources through `importlib.resources`.

[...]

PR #744, opened 2024-12-02
ENH: create a dataset of pre-registered motors. See #664

caioessouza:
I followed the recommendation "Download and save several .eng files in the repo so we can install it along with the rocketpy package itself".

The website thrustcurve.org was very useful to search for some .eng files. I mainly focused in some of the main brands on the market: Cesaroni, Aero Tech, Animal Motors and Loki. And also focused on classes K to M, because this is the main range of total impulse that I'm used to seeing in rocketry. I tried to pick motors with a difference of about 300~600Ns in total impulse.

Some more improvements than can also be made following this issue are expanding the dataset for whole SolidMotor objects, more than only thrust curves. I think this would be what the recommendation "Save .json files with all the information we may find available on internet" could mean. I decided to go for the simple for now, but having the thrust curves is a good first step to implementing that in the future, which I would totally be able to do!

## Pull request type

- [x] Data updates

## Checklist

- [x] Data have been added

## Current behavior

I've worked on issue #664.

## New behavior

Now there's a larger dataset of thrust curves available in the library.

## Breaking change

- [ ] Yes
- [x] No


Gui-FernandesBR (approved):
LGTM

PR #748, opened 2024-12-06
ENH: Allow for Alternative and Custom ODE Solvers.

phmbressan:
## Pull request type

## Checklist

## Current behavior

Currently only `LSODA` is used to integrate the equations of motion of the `Flight` class, with no option for other solvers or custom code.

## New behavior

This PR adds the possibility of using different solvers from `scipy.integrate` API.

## Breaking change

- [ ] Yes
- [X] No

## Additional information

`LSODA`, of course, remains the default option and the strongly recommended one. Quick testing showed that other solvers might match `LSODA` speed, but require more tuning to avoid instabilities.

### DIFF
--- a/rocketpy/simulation/flight.py
+++ b/rocketpy/simulation/flight.py
@@ -7,7 +7,7 @@ from functools import cached_property
 
 import numpy as np
 import simplekml
-from scipy import integrate
+from scipy.integrate import BDF, DOP853, LSODA, RK23, RK45, OdeSolver, Radau
 
 from ..mathutils.function import Function, funcify_method
 from ..mathutils.vector_matrix import Matrix, Vector
@@ -24,8 +24,19 @@ from ..tools import (
     quaternions_to_spin,
 )
 
+ODE_SOLVER_MAP = {
+    'RK23': RK23,
+    'RK45': RK45,
+    'DOP853': DOP853,
+    'Radau': Radau,
+    'BDF': BDF,
+    'LSODA': LSODA,
+}
 
-class Flight:  # pylint: disable=too-many-public-methods
+
+# pylint: disable=too-many-public-methods
+# pylint: disable=too-many-instance-attributes
+class Flight:
     """Keeps all flight information and has a method to simulate flight.
 
     Attributes
@@ -506,6 +517,7 @@ class Flight:  # pylint: disable=too-many-public-methods
         verbose=False,
         name="Flight",
         equations_of_motion="standard",
+        ode_solver="LSODA",
     ):
         """Run a trajectory simulation.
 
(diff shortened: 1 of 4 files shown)


rocketpy/simulation/flight.py
```diff
@@ -24,8 +24,19 @@
...
 )
 
+ODE_SOLVER_MAP = {
+    'RK23': RK23,
+    'RK45': RK45,
+    'DOP853': DOP853,
+    'Radau': Radau,
+    'BDF': BDF,
+    'LSODA': LSODA,
+}
```
Gui-FernandesBR, line 34: Could this possible become an Enum? Seems more suitable

rocketpy/simulation/flight.py
```diff
@@ -1185,6 +1213,19 @@ def __init_solver_monitors(self):
...
+            try:
+                self._solver = ODE_SOLVER_MAP[self.ode_solver]
+            except KeyError as e:
+                raise ValueError(
+                    f"Invalid ``ode_solver`` input: {self.ode_solver}. "
+                    f"Available options are: {', '.join(ODE_SOLVER_MAP.keys())}"
+                ) from e
+
+        self.__is_lsoda = hasattr(self._solver, "_lsoda_solver")
+
```
Gui-FernandesBR, line 1228: Quite verbose...

Maybe we should create a separate method for this validation.

rocketpy/simulation/flight.py
```diff
@@ -581,10 +593,23 @@ def __init__(  # pylint: disable=too-many-arguments,too-many-statements
...
+            documentation [1]_.
+
 
         Returns
         -------
         None
+
+        References
+        ----------
+        .. [1] https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html
```
Gui-FernandesBR, line 612: Does this link work? Does this throw a warning?

Gui-FernandesBR (approved):

rocketpy/simulation/flight.py
phmbressan: I don't see any warnings as far as I checked and the hyperlinks are working. One easy mistake is forgetting the `..` before the reference, then it will bring a warning.

rocketpy/simulation/flight.py
phmbressan: Fixed by [26ceab9](https://github.com/RocketPy-Team/RocketPy/pull/748/commits/26ceab9dd04f5676dd357e134e9dc88bd96c4c5c).

rocketpy/simulation/flight.py
phmbressan: Certainly could be an Enum, however I prefer keeping it like this to better mirror [`scipy` usage](https://github.com/scipy/scipy/blob/v1.14.1/scipy/integrate/_ivp/ivp.py#L12-L17), I could even have imported their dictionary directly, but that is not guaranteed to not have breaking changes, so I preferred to simply build the dict myself.

PR #735, opened 2024-11-23
BUG: forecast and reanalysis models - move wind_speed to correct position

WilliamArmst:
## Pull request type

Bugfix

## Checklist

- [x] Lint (`black rocketpy/ tests/`) has passed locally 
- [x] All tests (`pytest tests -m slow --runslow`) have passed locally

## Current behavior

Using [ERA5 Reanalysis](https://docs.rocketpy.org/en/latest/user/environment/1-atm-models/reanalysis.html#era5) weather data, the wind speed function was in the incorrect position.  See [line 1899](https://github.com/RocketPy-Team/RocketPy/compare/master...WilliamArmst:RocketPy:master?expand=1#diff-a7018336c61275e0580c3eb23f78df42963b8d0f6460a21e37aada51828a2702R1899), where the ``__set_wind_speed_function`` references ``data_array[7]`` for the wind speed.

## New behavior

Wind speed is now calculated correctly.

## Breaking change

- [ ] Yes
- [x] No

## Additional information

N/A

PR #752, opened 2024-12-11
DOC: Lince (Team STAR) 2023 flight sim

LUCKIN13:
## Pull request type

## Checklist

## Additional information
Unfortunately I couldn't find any GNSS flight data from the payload. However, this is still a really good example of a simulation model with deployable payload.


docs/examples/lince_flight_sim.ipynb
Gui-FernandesBR, line 1: can you double check the plots in the `## Flight Simulation` section please?
Seems odd to me that both 3D plots look the same

LUCKIN13: Double checked all the parameters.  They are indeed pretty similar, but I did some tests and the difference makes sense according to the difference in weight and parachute cds.

data/weather/euroc_2023_fixed.nc
Gui-FernandesBR, line 1: @LUCKIN13  why do you need this file? Isn't there another .nc file for euroc 2023?

Gui-FernandesBR (commented):
This plot isn't correct.
The flight should stop at the main deployment height.
You should set a max_time for the first flight object.

image

Also, how did you generate the initial solution for other flights?

Gui-FernandesBR (changes requested):
Also, please add this flight to the comparison plot on the index.rst file

data/weather/euroc_2023_fixed.nc
LUCKIN13: @Gui-FernandesBR unfortunately, the other file seems to have some kind of error. I noticed, while working on the other examples, that the wind velocities were around 100–200 m/s, which is unrealistic. I made a new request to Copernicus, and I’m considering updating all the EUROC 2023 examples to use the new file and deleting the older one (if it is no longer needed for anything else).

LUCKIN13:
> Also, please add this flight to the comparison plot on the index.rst file

Sorry for forgetting that. Will do that asap

data/weather/euroc_2023_fixed.nc
Gui-FernandesBR: Are other examples also wrong?

I don't like the name of the new file at all. We should either delete the older or the new file, keeping only one is the best approach.

LUCKIN13: I was checking and looks like all the flight examples have wind speeds over 150 m/s. I gave that name knowing that would be temporary (either stick with the old one, or change)

docs/examples/lince_flight_sim.ipynb
Gui-FernandesBR: Good.

data/weather/euroc_2023_fixed.nc
Gui-FernandesBR: I think #735 solved that!

Btw we should create more acceptance tests in order to fix that.

Gui-FernandesBR:
@LUCKIN13 try to read my last commit, I think the #735 solved the problem with the wind speeds.
The problematic code was related to the environment class, not the .nc file, so I've deleted your new file and preserved only the old one.

Gui-FernandesBR (approved):
LGTM

LUCKIN13:
> @LUCKIN13 try to read my last commit, I think the #735 solved the problem with the wind speeds. The problematic code was related to the environment class, not the .nc file, so I've deleted your new file and preserved only the old one.

Great! Makes sense, since all the other examples also add some problems. Thanks a lot for your help!

PR #756, opened 2024-12-15
TST: increase test coverage

Gui-FernandesBR:
## Pull request type

## Checklist

## Current behavior

* Coverage without slow tests: 76%

## New behavior

* Coverage without slow tests: 80%
* Coverage with slow tests: 92%

## Breaking change

- [x] No

## Additional information
* I'm not sure if I will have time to add more tests, so I guess the PR is ready for review
* Please review carefully the logics behind the tests.

### DIFF
--- a/tests/unit/test_function.py
+++ b/tests/unit/test_function.py
@@ -787,3 +787,107 @@ def test_low_pass_filter(alpha):
             f"The filtered value at index {i} is not the expected value. "
             f"Expected: {expected}, Actual: {filtered_func.source[i][1]}"
         )
+
+
+def test_average_function_ndarray():
+
+    dummy_function = Function(
+        source=[
+            [0, 0],
+            [1, 1],
+            [2, 0],
+            [3, 1],
+            [4, 0],
+            [5, 1],
+            [6, 0],
+            [7, 1],
+            [8, 0],
+            [9, 1],
+        ],
+        inputs=["x"],
+        outputs=["y"],
+    )
+    avg_function = dummy_function.average_function()
+
+    assert isinstance(avg_function, Function)
+    assert np.isclose(avg_function(0), 0)
+    assert np.isclose(avg_function(9), 0.5)
+
+
+def test_average_function_callable():
+
+    dummy_function = Function(lambda x: 2)
+    avg_function = dummy_function.average_function(lower=0)
+
+    assert isinstance(avg_function, Function)
+    assert np.isclose(avg_function(1), 2)
+    assert np.isclose(avg_function(9), 2)
+
+
+@pytest.mark.parametrize(
+    "lower, upper, sampling_frequency, window_size, step_size, remove_dc, only_positive",
+    [
+        (0, 10, 100, 1, 0.5, True, True),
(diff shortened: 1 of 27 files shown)


Gui-FernandesBR:
These are the top 10 files we should be worried about and try to increase coverage:

image

Lucas-Prates:
@Gui-FernandesBR if you do not mind, I can take a closer look at `sensitivity_model` and `sensitivity_prints`. I recall that you implemented tests back in the PR, but they were failing for some reason.

Gui-FernandesBR:
> @Gui-FernandesBR if you do not mind, I can take a closer look at `sensitivity_model` and `sensitivity_prints`. I recall that you implemented tests back in the PR, but they were failing for some reason.

Yes please, I'd like some help on that

Lucas-Prates:
100% cover for both `sensitivity_model` and `sensitivity_print`. Also removed a piece of duplicate code inside `sensitivity_model`.

tests/unit/test_function.py
```diff
@@ -787,3 +787,107 @@ def test_low_pass_filter(alpha):
...
+    "lower, upper, sampling_frequency, window_size, step_size, remove_dc, only_positive",
+    [
+        (0, 10, 100, 1, 0.5, True, True),
+        (0, 10, 100, 1, 0.5, True, False),
+        (0, 10, 100, 1, 0.5, False, True),
+        (0, 10, 100, 1, 0.5, False, False),
+        (0, 20, 200, 2, 1, True, True),
+    ],
+)
+def test_short_time_fft(
```
Gui-FernandesBR, line 837: @Lucas-Prates could yo review this test please? `test_short_time_fft`

Gui-FernandesBR:
Ok, I think this PR is already in a good point.
Of course that we can do even better. Ideally we should see 90% of code coverage without slow tests, and up to 10% of slowed tests.
But for today we are fine already.

@RocketPy-Team/code-owners ready for review! Pleas notice that the code coverage has increased 4 pp, but I didn't spend much time trying to find the best validation possible for each function method. I opted to be simple.

Gui-FernandesBR:
Once again I am sharing the top 10 worst files in terms of coverage.
Perhaps you could contribute a little to the function.py file, @phmbressan ?

image

tests/unit/test_function.py
Lucas-Prates: Looks okay to me! Now, if you want to include a "numeric assertion" test, since you are using pure signal (i.e. sine function), all the result functions should be a Dirac delta function, i.e. a pulse function, at the given frequency, 5 Hz in this case. Something like this

image

So, if you want to, you could include an extra argument `true_frequency`, and then

```
signal = np.sin(2 * np.pi * true_frequency * t)
.
.
.
for f in stft_results:
    pytest.approx(1, 0.01) == f[true_frequency - 1][1]
```

Gui-FernandesBR: Thx for helping. I will keep the test as simple as the current version. This PR already took me too  much time.

PR #764, opened 2025-02-08
ENH: unstable rocket error

Gui-FernandesBR:
## Pull request type

## Checklist

## Current behavior

A common mistake on rocketpy is to try running a flight simulation when the rocket is unstable.
The simulation will likely fail or never stops.
It is true that not always it will fail, but I believe we should be more restricted in order to avoid these case scenarios;

## New behavior
* Throws a `UnstableRocketException` everytime a flight is instantiated using a rocket with static margin lower than 0.
* Added rocketpy official exceptions

## Breaking change

- [x] Yes and No. As you can see by the tests, some code may break. However, no current function will be 

## Additional information

Future PRs should add more custom exceptions if needed.

In this PR:
- Update docs: (1) explaining about rocketpy exceptions and (2) explaining about the stability check.
- CHANGELOG and slow tests.


rocketpy/exceptions.py
```diff
@@ -0,0 +1,21 @@
...
+Custom exceptions for RocketPy library.
+"""
+
+
+class RocketPyException(Exception):
+    """All RocketPy custom exceptions should inherit from this class."""
+
+
+class UnstableRocketError(RocketPyException):
+    """Raised when the rocket jas negative static margin."""
```
GabrielBarberini, line 11: has*

rocketpy/simulation/flight.py
```diff
@@ -664,6 +666,15 @@ def __repr__(self):
...
 
+    def __validate_rocket_static_margin(self):
+        """
+        Avoids running a flight simulation if the rocket stability margin is
+        negative. This is a common mistake that can lead to unstable flights,
+        which usually runs indefinitely.
+        """
+        if (s := self.rocket.static_margin(0)) < 0:
+            raise UnstableRocketError(s)
+
```
MateusStano, line 677: I think this error check might not be suitable right now. I see a few points as to why:

- If a user wants to simulate an unstable rocket, for any reason, we should not make that impossible. There could always be a rocket that has a negative static margin, but the stability margin could quickly become positive with fuel burn or other effects

- When in monte carlo simulations, what would happen if one of the iterations has an unstable rocket? Will all the simulations break because of it?

- The generic aerodynamic coefficients don't calculate static margin correctly. This essentially means that all rockets with custom coefficients will have a negative static margin.

Maybe raising a warning is the best we can do?

phmbressan: I agree with the first point you raised: not all rockets that are instantaneously unstable are necessarily unsimulatable.

A possible compromise would be letting the user set a variable, such as `ALLOW_UNSTABLE`.

GabrielBarberini: I like the Idea of having a flag with default to false (so current API is respected).

Then we can simply turn It on for infinity

MateusStano: Would `ALLOW_UNSTABLE` be by default `True` or `False`?

And if the default is `False` would that be a breaking change? Since someone who has a sim with an unstable rocket might start getting an error with this SM check

GabrielBarberini: Sorry, I meant default `true`

Gui-FernandesBR: Converting from an error to a warning is the easiest option right now.

However, I think we should better understand the cases where an unstable rocket can fly. 
Perhaps the problem/bug happens inside the u_dot function (after rail exit) ?

If setting ALLOW_UNSTABLE to false, we should describe this behavior in the error message.
Setting ALLOW_UNSTABLE to true would not make much help.

GabrielBarberini: I think setting It to `true` by default mantains current behavior while allowing clients to deliberatively switch.

IMHO a simple warning would not be easy to consume for safe guarding purposes

Gui-FernandesBR: I understand. 

My point is that maybe we should not keep the current behavior at all.

Gui-FernandesBR: I will think more about this during the week

Gui-FernandesBR:
Gotta think a better way of solving this problem.

PR #775, opened 2025-03-03
DEV: Runs make format and updates makefile

Gui-FernandesBR:


Makefile
Gui-FernandesBR, line 1: it would be nice to test these new commands on ubuntu before merging

Gui-FernandesBR: tested and working fine

Issue #773, opened 2025-02-21
Bug Report: Runtime Error in RocketPy's add_motor Function


vkhansen:
Issue Summary:The program fails when executing the example_rocket.add_motor(hybrid_motor, position=0) command. This issue occurs even on previously working code, indicating a potential problem with internal calculations in the RocketPy library.

Steps to Reproduce:

Run a previously working RocketPy script that includes the following command:

example_rocket.add_motor(hybrid_motor, position=0)

The error occurs at runtime, preventing further execution.

Expected Behavior:

The motor should be added to the rocket without errors.

The static margin evaluation should proceed without encountering invalid numerical values.

Actual Behavior:

A RuntimeWarning is triggered due to division by zero and invalid scalar multiplications in mathutils/function.py:

C:\Users\User\Documents\GitHub\rocket_trajectory\venv\Lib\site-packages\rocketpy\mathutils\function.py:2378:
RuntimeWarning: divide by zero encountered in scalar divide
return Function(lambda x: (self.get_value_opt(x) / other(x)))

C:\Users\User\Documents\GitHub\rocket_trajectory\venv\Lib\site-packages\rocketpy\mathutils\function.py:2311:
RuntimeWarning: invalid value encountered in scalar multiply
return Function(lambda x: (self.get_value_opt(x) * other(x)), inputs)

This is followed by a traceback leading to a ValueError:

ValueError: array must not contain infs or NaNs

The error occurs while evaluating the static margin of the rocket, specifically within RocketPy’s interpolation and function evaluation methods.

```
Traceback:

Traceback (most recent call last):
  File "C:\Users\User\Documents\GitHub\rocket_trajectory\6IN_OLD.py", line 149, in <module>
    example_rocket.add_motor(hybrid_motor, position=0)
  File "C:\Users\User\Documents\GitHub\rocket_trajectory\venv\Lib\site-packages\rocketpy\rocket\rocket.py", line 991, in add_motor
    self.evaluate_static_margin()
  File "C:\Users\User\Documents\GitHub\rocket_trajectory\venv\Lib\site-packages\rocketpy\rocket\rocket.py", line 655, in evaluate_static_margin
    self.static_margin.set_discrete(
  File "C:\Users\User\Documents\GitHub\rocket_trajectory\venv\Lib\site-packages\rocketpy\mathutils\function.py", line 718, in set_discrete
    func.set_source(np.column_stack((xs, ys)))
[...]

PR #776, opened 2025-03-03
MNT: add error handling for static margin calculation

Gui-FernandesBR:
## Pull request type

## Checklist

## Current behavior
See https://github.com/RocketPy-Team/RocketPy/issues/773

## New behavior

Simple message added,  I hope this helps future users

## Breaking change

- [x] No

## Additional information

Enter text here...

### DIFF
--- a/rocketpy/rocket/rocket.py
+++ b/rocketpy/rocket/rocket.py
@@ -657,11 +657,11 @@ class Rocket:
             self.static_margin.set_discrete(
                 lower=0, upper=self.motor.burn_out_time, samples=200
             )
-        except ValueError as err:
+        except ValueError as err:  # pragma: no cover
             raise ValueError(
-                "For some reason the static margin could not be properly calculated. "
-                "Please check the center of mass and center of pressure functions "
-                "and verify if they are well defined in the 0s to the burn out time range."
+                "Static margin calculation failed. "
+                "Verify the center of mass and center of pressure functions are "
+                "correctly defined within the time range from 0 to burn out time."
             ) from err
         return self.static_margin
 
--- a/rocketpy/rocket/rocket.py
+++ b/rocketpy/rocket/rocket.py
@@ -652,9 +652,17 @@ class Rocket:
         self.static_margin.set_inputs("Time (s)")
         self.static_margin.set_outputs("Static Margin (c)")
         self.static_margin.set_title("Static Margin")
-        self.static_margin.set_discrete(
-            lower=0, upper=self.motor.burn_out_time, samples=200
-        )
+
+        try:
+            self.static_margin.set_discrete(
+                lower=0, upper=self.motor.burn_out_time, samples=200
+            )
+        except ValueError as err:
+            raise ValueError(
+                "For some reason the static margin could not be properly calculated. "
+                "Please check the center of mass and center of pressure functions "
+                "and verify if they are well defined in the 0s to the burn out time range."
+            ) from err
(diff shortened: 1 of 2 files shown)


rocketpy/rocket/rocket.py
```diff
@@ -652,9 +652,17 @@ def evaluate_static_margin(self):
...
         self.static_margin.set_outputs("Static Margin (c)")
         self.static_margin.set_title("Static Margin")
-        self.static_margin.set_discrete(
-            lower=0, upper=self.motor.burn_out_time, samples=200
-        )
+
+        try:
+            self.static_margin.set_discrete(
+                lower=0, upper=self.motor.burn_out_time, samples=200
+            )
```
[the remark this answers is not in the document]
Gui-FernandesBR: I believe you're "overcomplicating" the situation a little bit.

I like to think this way: Regression! If someone faced an issue while running a simulation, in order to avoid future users to face the same problem, I'm adding a more delicate message suggesting what is probably wrong in the code.

The reason we don't do that for every input? Simple: because no one has complained about other inputs yet.
Try to read #773, you're going to see that the source of the error is quite silent, so a message like this can guide the user to quickly find where to look in the code.
If the input is clearly wrong, for example, someone sets a negative mass to the rocket, I think that would be the case we wouldn't require much effort from our team (or the user) to identify the error source. However the case of this PR is particularly trickly given that everything was properly set in the simulation, but rocketpy only crashed because we couldn't perform a single operation.

Additionally, the try/catch block barely changes the execution time, so I see almost no problem of doing so.

Lemme know ehat you think about it.

MateusStano:
In the end, what exactly was causing the computation to break? Simulating with burn time longer then the provided curves should not break things in theory right?

We really have to turn the static margin calculation into something lazy. There is no need to have this calculated for the simulation, it should only be computed when needed

Gui-FernandesBR:
> In the end, what exactly was causing the computation to break? Simulating with burn time longer then the provided curves should not break things in theory right?
> 
> We really have to turn the static margin calculation into something lazy. There is no need to have this calculated for the simulation, it should only be computed when needed

@MateusStano the way the user have set the hybrid motor made the center of mass function undefined after ~14s.
I think it is a particular case, it is almost as if he cooked the example to break the static margin on rocketpy

Gui-FernandesBR:
@phmbressan @MateusStano I'm re-requesting the review on this PR. I believe we are "almost done" here

rocketpy/rocket/rocket.py
```diff
@@ -652,9 +652,17 @@ def evaluate_static_margin(self):
...
+        except ValueError as err:  # pragma: no cover
+            raise ValueError(
+                "Static margin calculation failed. "
+                "Verify the center of mass and center of pressure functions are "
+                "correctly defined within the time range from 0 to burn out time."
+            ) from err
```
MateusStano, line 665: I find it wrong to not allow a Rocket to be created because of the static margin. The static margin is not used for the simulation, so it should not be required

I think just a warning here is fine. If we want to add this exception we should look into the root of what was breaking the SM and add the exception there (probably somewhere related to motor CM calculations)

MateusStano (changes requested):

rocketpy/rocket/rocket.py
Gui-FernandesBR: I disagree but I will close the PR bc I don't have enough time to further elaborate this PR.

PR #783, opened 2025-03-15
BUG: fix the wind velocity factors usage and better visualization of uniform distributions in Stochastic Classes

kevin-alcaniz:
## Pull request type

- [ x] Code changes (bugfix, features)

## Checklist

## Current behavior

- Wind Velocity Factors in the EnvironmentStochastic class:

The attributes assigned to the Evironment class are env.wind_velocity_x_factor and env.wind_velocity_y_factor because the 'key' variable is used in 'setattr' without modification. This doesn't allow to vary the wind profile when using constant values for the wind velocity components, as env.wind_velocity_x and env.wind_velocity_y (which are used in the simulation) remain unchanged.

- visualize_attributes Function in the StochasticModel Class:

The function does not correctly display the values used for uniform distributions. In a uniform distribution, the 'nominal value' acts as the lower bound, and the 'standard deviation' acts as the upper bound (np.random.uniform(nominal_value, std_dev)). However, these distinctions are not reflected in the display.

For example, for a uniform distribution with wind_velocity_x_factor = (-2, 2, 'uniform'), the visualize_attributes function displays:
wind_velocity_x_factor -2.00000 ± 2.00000 (uniform).

This suggests that values are sampled from the range [-4, 0], whereas the actual range is [-2, 2].

## New behavior

- Wind Velocity Factors in the EnvironmentStochastic class:

Now, the 'key' name is modified inside the 'factor' condition so that the attributes assigned to the Environment class are wind_velocity_x and wind_velocity_y instead of wind_velocity_x_factor and wind_velocity_y_factor.

- visualize_attributes Function in the StochasticModel Class:

A new condition has been added to the format_condition function for tuple inputs. In this condition, the uniform distribution's mean and half-range are calculated and displayed instead of the nominal_value and std_dev, respectively. The new display for the previous example would be: wind_velocity_x_factor 0.00000 ± 2.00000 (uniform). 

## Breaking change

- [ ] Yes
- [x ] No

## Additional information

This is my first pull request ever, so I'm looking forward to your feedback. Thank you!!


kevin-alcaniz:
I don't know what codecov/patch fail means. I welcome help 😂

Gui-FernandesBR (commented):
The code looks good, I think we just have to test if everything is running properly locally

rocketpy/stochastic/stochastic_model.py
```diff
@@ -514,11 +514,20 @@ def format_attribute(attr, value):
...
-                    f"{std_dev:.5f} ({dist_func.__name__})"
-                )
+                if callable(dist_func) and dist_func.__name__ == "uniform":
+                    mean = (std_dev + nominal_value) / 2
+                    half_range = (std_dev - nominal_value) / 2
+                    return (
+                        f"\t{attr.ljust(max_str_length)} "
+                        f"{mean:.5f} ± "
+                        f"{half_range:.5f} ({dist_func.__name__})"
+                    )
```
MateusStano, line 524: I think using `±` here makes it a bit confusing.

I suggest just putting a comma like:

```suggestion
                if callable(dist_func) and dist_func.__name__ == "uniform":
                    lower_bound = nominal_value
                    upper_bound = std_dev
                    return (
                        f"\t{attr.ljust(max_str_length)} "
                        f"{lower_bound:.5f}, {upper_bound:.5f} ({dist_func.__name__})"
                    )
```

MateusStano (changes requested):
Hey! Awesome find as always @kevin-alcaniz. I just suggested one quick change and then I think this PR is all good!

kevin-alcaniz:
Good idea, @MateusStano ! I already did the commit.

Gui-FernandesBR (approved):
LGTM, 

I believe we can already merge this PR. Great work, @kevin-alcaniz !

MateusStano (approved):

Issue #769, opened 2025-02-11
ENH: Add pressure thrust


caioessouza:
**Is your feature request related to a problem? Please describe.**

Rocketpy curetly does not take into account the thrust dependece on the ambient pressure: $(p_e-p_a)A_e$.

It's a important behaviour considering two situations: correction for thrust curves obtained from a static hot fire test performed in a altitude much different from the launch or in-flight corrections for high target altitude flights.

**Describe the solution you'd like**

Implement a optional method for take that behaviour into acount without crashing the current library behaviour.

**Additional context**

![Image](https://github.com/user-attachments/assets/0d4cf472-f8bf-4bdf-baac-5345002cb1a3)

**To do list**

- [x] Review theory
- [x] Check which parameter to add to the init of all motor classes
- [x] Implemet a function to split the input thrust curve in mass flow thrust and pressure thrust
- [x] Impelement the thrust correction in the flight simulation

caioessouza:
Based on #148

kevin-alcaniz:
At Faraday, we've made some changes to account for pressure thrust and correct the thrust force as the rocket ascends. Would you like me to open a PR?

caioessouza:
> At Faraday, we've made some changes to account for pressure thrust and correct the thrust force as the rocket ascends. Would you like me to open a PR?

Sure, I just saw what you made at the issue #789 and it seems very good. Feel free to open the PR, thank you for the contribuition!

PR #789, opened 2025-03-19
ENH: Introduce Net Thrust with pressure corrections

kevin-alcaniz:
## Pull request type

## Checklist

## Current behavior

RocketPy currently does not take into account the thrust dependence on the ambient pressure.

It's a important behaviour considering two situations: correction for thrust curves obtained from a static hot fire test performed in a altitude much different from the launch or in-flight corrections for high target altitude flights.

This PR is related to an Issue: ENH: Add pressure thrust #769 

## New behavior

The theory applied to this method is explained in the image:

image

- Motor classes:

A new optional attribute, reference_pressure, has been introduced. If not defined, RocketPy's behavior remains unchanged. Given the reference pressure and the thrust curve, the vacuum thrust curve can be computed using the new get_vacuum_thrust function.

- Flight class:

If a reference pressure is provided, a new method is implemented to calculate the net thrust. Otherwise, the net thrust remains equal to the current thrust.

Additionally, a new funcify_method has been created for net thrust, and thrust power is now calculated based on net thrust as well.

## Breaking change

- [ ] Yes
- [x] No

## Validation

The results obtained after implementing this new feature are shown below:

image

The Y-axis represents Thrust (N) and the X-axis represents Time (s). As you can see, at the very beginning, both curves (raw and net thrust) are the same because the reference pressure is set to sea level pressure. As the rocket ascends, the net thrust increases because the negative term in equation 5 decreases.

I have also attempted to validate this behavior with ASTOS, and the results obtained are shown below:

WhatsApp Image 2025-03-19 at 11 02 52_564937b2

In this plot, the vacuum thrust curve (blue) is input directly into ASTOS. The net thrust curve (black) is evaluated by ASTOS. Finally, the calculated thrust curve (red) was evaluated by me using the same method implemented in this PR.


kevin-alcaniz:
There are other modifications related to another PR (BUG: fix the wind velocity factors usage and better visualization of uniform distributions in Stochastic Classes https://github.com/RocketPy-Team/RocketPy/pull/783)

rocketpy/motors/empty_motor.py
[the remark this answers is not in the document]
Gui-FernandesBR: @kevin-alcaniz I now it changes a lot the code, but the current way it is implemented, we cannot merge it, the proposed solution would break current codes that use positional arguments (i.e. code which instantiate motor classes without defining the name of each parameter).

This is a very common problem we see from contributors, it's ok, but since it's a breaking change, we prefer to refactor the code and avoid releasing a v2.0.0 of rocketpy just because of this break.

I hope you can understand our point of view, lemme know if there's anything still unclear.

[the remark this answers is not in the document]
kevin-alcaniz: If I add the new `reference_pressure` parameter at the end of the parameters list, wouldn't that solve the problem?

Are you going to refactor the code, or do you expect me to do it??

What would be necessary to merge this change??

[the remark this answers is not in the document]
kevin-alcaniz: Okay, @Gui-FernandesBR . I've moved over the `reference_pressure` option to the end of all motor class initializers. This should prevent the breaking change. Please, if you find out another problem, don't hesitate to tell me.

rocketpy/motors/motor.py
```diff
@@ -302,15 +316,27 @@ class Function. Thrust units are Newtons.
...
+                self.thrust.source, reference_pressure
+            )
+        self.vacuum_thrust_source = vacuum_thrust_source
+        self.vacuum_thrust = Function(
+            vacuum_thrust_source,
+            "Time (s)",
+            "Vacuum Thrust (N)",
+            self.interpolate,
+            "zero",
+        )
```
phmbressan, line 333: @Gui-FernandesBR and @MateusStano, I would appreciate a second opinion on the matter here:

IMHO, the separation of `thrust` in two variables, despite correct, is not really necessary here. My view on the role of the `reference_pressure` parameter would be pretty similar to the way the `reshape_thrust_curve` already operates: changing the thrust curve instead of creating a new attribute.

My proposal would be to have something like this:

```python
if reference_pressure is not None:
    self.thrust = Motor.correct_thrust_pressure(self.thrust, reference_pressure, self.nozzle_area)
```

The `correct_thrust_pressure` would be a static method that would perform essentially the same computation as `get_vacuum_thrust`. The changes to `flight.py` would be also the same, but using directly the `thrust` instead.

The main potential advantages I see with the similar behavior as the `reshape_thrust_curve`:
- Consistency: similarly to reshaped thrust, the thrust source would be modified instead of having a new attribute;
- Code independence: higher level code (e.g. rocket, flight etc) would not have to know whether to call `vacuum_thrust` or `thrust`, we would only have `thrust`;
- Static calls: using a static method would allow someone that still does not have a `Motor` to correct thrust data. 

Possible disadvantages with my suggestion:
- Changing the input `thrust` might sound unintuitive, but we already do that with discretization, clipping and reshaping thrust. As a reminder, the exact input thrust is always saved to `self.thrust_curve` (different from `self.thrust`), so it is not lost.

phmbressan (commented):
Amazing work @kevin-alcaniz ! I made some comments regarding the split into two thrust attributes that I would like to hear some second opinions from the team before proceeding; of course, feel free to also participate on the discussions.

rocketpy/motors/motor.py
Gui-FernandesBR: @phmbressan what you are suggesting is not feasible and violates the principle of what a pressure thrust correction really means. It is not possible to modify the thrust curve using only the reference pressure given in the motor class.

The difference occurs due to difference of pressure, which only exists as the rocket flies vertically. Before defining a flight, reshaping the curve is not sufficient, given that the atm. pressure will vary during flight.

But I do understand your point, the way it is implemented, a lot of confusions could be made.

Skipping further comments, my suggestion would be:

```python



class Motor(ABC):

    def __init__(self, ..., reference_pressure):


    @cached_property
    def vacuum_thrust(self):
        # Only added here if the user wants to analyze it, but we won't use it in the simulation

        if self.reference_pressure is None:
            print("Reference pressure is not set, cannot calculate vacuum thrust")
            return self.thrust

        return self.thrust + self.reference_pressure * self.nozzle_area

    def calculate_pressure_thrust(self, pressure):

        if self.reference_pressure is None:
            return 0

        return (self.reference_pressure - pressure) * self.nozzle_area


class Flight():

    ...

    def u_dot(self, ...):

        total_thrust = self.motor.thrust + self.motor.calculate_pressure_thrust(env_pressure)

```

What do you think, @kevin-alcaniz @phmbressan ?

rocketpy/motors/empty_motor.py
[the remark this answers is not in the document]
Gui-FernandesBR: thank you, @kevin-alcaniz !!

rocketpy/motors/motor.py
phmbressan: I should have been more clear, my point is not to only have the correction on the `Motor` class. 

When I meant that the `flight.py` would be the same, I meant the same as this PR implementation: we still need to apply the correction, but we would call `motor.thrust` instead of `vacuum_thrust`.

Gui-FernandesBR: @phmbressan, I understand your point when saying we would still need to apply the correction **during flight*. However it's not clear to me how exactly you expect to do that.

In my suggestion above the code is pretty clear: keep the thrust source as it is, during simulation, use the thrust + a delta of pressures times the nozzle's area.

It's clean, the motor satys flight agnostic and environment agnostic.

If you still plan to modify the thrust curve *inplace* right after Motor's initialization, how exactly do you plan to do that? Will the thrust become the vacuum thrust? If so, that would not make so much sense to me, given you will still need to evaluate the delta of pressure during flight and, most importantly, the motor.thrust would not really represent the real thrust a motor can provide to the rocket.


Ps: I know it's a lot of text, but I'm trying to be as specific as possible since we don't have a meeting to discuss it.

kevin-alcaniz: I understand that what @phmbressan  is proposing is to follow my method exactly, but instead of defining a new property called `vacuum_thrust`, the corrections would be applied directly to the `thrust` attribute.

However, wouldn’t it be more intuitive for the user to keep all thrust concepts separate? Having distinct attributes for raw thrust, vacuum thrust, and net thrust provides greater clarity and flexibility. This way, users can analyze, plot, and export all of them individually, making it easier to compare results and verify that the corrections are being applied correctly.

I see your point about keeping things simple, but in the context of a next-generation simulator for high-power rockets, wouldn't it be more beneficial to prioritize transparency and usability over minimalism? If all corrections are applied to `self.thrust`, users might not immediately understand why their thrust values change significantly after adding reference pressure—this could introduce confusion, especially for those new to the software.

Additionally, separating the attributes makes debugging and validation easier. If a user encounters unexpected results, they can inspect each thrust value separately to pinpoint where the issue arises, rather than having to reverse-engineer corrections embedded into a single attribute.

That’s my perspective! What do you think?

kevin-alcaniz: @Gui-FernandesBR I like your approach—it adds new and useful information (`pressure_thrust`, `vacuum_thrust`, and `total_thrust`) while keeping things simpler, as the Flight class will always use the same expression and the changes are minimal. Great idea!

phmbressan: Understood, your argument favoring this PR implementation makes sense to me.

Just some last questions I have to make sure we are on the same page:
- Besides plotting, is there any usage inside RocketPy for the plain `self.thrust`?
- Should there be any corrections on the computation of the `exhaust_velocity` of `LiquidMotors and `thrust_power` from `Flight`?
- The `thrust_to_weight_ratio`, in theory, should consider the ambient pressure (since the launch might be at a different place from the static fire), right? In practice, I find this difference negligible, since it is not even a value used in the simulation. But should we use `.thrust` or `vacuum_thrust` here?

Gui-FernandesBR: > Understood, your argument favoring this PR implementation makes sense to me.
> 
> Just some last questions I have to make sure we are on the same page:
> 
> * Besides plotting, is there any usage inside RocketPy for the plain `self.thrust`?

Yes, the self.thrust function is accessed during flight.

> * Should there be any corrections on the computation of the `exhaust_velocity` of `LiquidMotors and `thrust_power`from`Flight`?

Interesting, the `thrust_power`, yes, I believe so! Kevin already did that.
Regarding the exhaust_velocity... I don't know, this attribute is only used within the motor class, right? Given that, it is impossible to apply any "correction", since the motor class is environment-agnostic.

> * The `thrust_to_weight_ratio`, in theory, should consider the ambient pressure (since the launch might be at a different place from the static fire), right? In practice, I find this difference negligible, since it is not even a value used in the simulation. But should we use `.thrust` or `vacuum_thrust` here?

Since the `thrust_to_weight_ratio` is an attribute of the motor class (not the flight), we should keep using the `thrust` value.

___________

To make it 100% clear: we will never use the `vaccum_thrust`, it is not a physical attribute, it does not represent the real thrust of the motor, and it is not the focus of this PR.

Please let me know if any of my answers are not clear to you, or if I'm going to the wrong direction here.

Gui-FernandesBR: @kevin-alcaniz , whenever you want to, please update this branch with the above proposed solution.
Let me know if you need any help.

kevin-alcaniz: Okay! I'll do it as soon as possible!

phmbressan:
Good work fixing the review comments @kevin-alcaniz ! I plan to give this PR a final look tomorrow, sorry for the delay.

kevin-alcaniz:
@phmbressan Thank you! Everything is ready on my end for the merge. Let me know if you find any issues.

Gui-FernandesBR (approved):
LGTM.

Waiting for @phmbressan 's review

kevin-alcaniz:
One minute, I'm going to update de CHANGELOG. I always forget 😂

kevin-alcaniz:
Done!

phmbressan (commented):
After reviewing a bit of literature on the topic and the excellent info that @kevin-alcaniz provided in the PR's description, I believe the root of my questions regarding the `exhaust_velocity` are related to these two concepts:

- _exhaust velocity_: the actual exhaust velocity of the gases exiting the nozzle, denoted by $V_9$;
- _effective exhaust velocity_: a measure of efficiency that can be computed by $\frac{T_{\text{net}}}{\dot{m}}$ (or $\frac{T_{\text{ref}}}{\dot{m}}$ in case of a static fire).

If I understood correctly, I believe the current `exhaust_velocity` method for liquid motors actually computes the _effective exhaust velocity_ of the reference thrust curve (e.g. static fire). A similar reasoning apply to the `exhaust_velocity` methods of the other motors, but assuming the mass flow rate as constant.

In my view, this behavior is ok, but it should be documented (likely a Notes section in the docstring) that the value being computed is _not_ the $V_9$ in the formula, but the _effective exhaust velocity_ of the reference curve provided.

What do you think @kevin-alcaniz , did I misunderstand anything?

Gui-FernandesBR:
> If I understood correctly, I believe the current `exhaust_velocity` method for liquid motors actually computes the _effective exhaust velocity_ of the reference thrust curve (e.g. static fire). A similar reasoning apply to the `exhaust_velocity` methods of the other motors, but assuming the mass flow rate as constant.
> 
> In my view, this behavior is ok, but it should be documented (likely a Notes section in the docstring) that the value being computed is _not_ the V 9 in the formula, but the _effective exhaust velocity_ of the reference curve provided.
> 
> What do you think @kevin-alcaniz , did I misunderstand anything?

@kevin-alcaniz what do you think about it?

kevin-alcaniz:
Sorry, @phmbressan , for the delay — I’ve been quite busy these past few days.

Yes, I think you're right. To accurately calculate the actual exhaust velocity, we’d need the thermodynamic properties of the gas upstream of the nozzle as well as the nozzle geometry. That’s not feasible with the current set of parameters used in RocketPy.

If you'd like, I can improve the documentation for this parameter. I propose adding a disclaimer in both the class attribute descriptions and within the exhaust velocity functions themselves.

Would that work for you?

rocketpy/motors/motor.py
```diff
@@ -1039,6 +1049,41 @@ def import_eng(file_name):
...
+        """Calculate the vacuum thrust from the raw thrust and the reference
+        pressure at which the thrust data was recorded.
+
+        Returns
+        -------
+        vacuum_thrust : Function
+            The rocket's thrust in a vaccum.
+        """
+        if self.reference_pressure is None:
+            print("Reference pressure is not set, cannot calculate vacuum thrust")
```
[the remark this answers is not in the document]
kevin-alcaniz: I’m not sure I fully understood what you meant by that.
Would you like me to clarify in the print that the vacuum thrust refers to the actual thrust the rocket would produce in a vacuum?
Or are you suggesting that I should add a warning indicating that this thrust value won’t be used in the actual simulation?
Could you let me know what you had in mind, please?

phmbressan:
> If you'd like, I can improve the documentation for this parameter. I propose adding a disclaimer in both the class attribute descriptions and within the exhaust velocity functions themselves.
> 
> Would that work for you?

Yes, a description in the style you mentioned should be good to clarify everything.

rocketpy/motors/motor.py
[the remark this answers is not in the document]
phmbressan: My suggestion is that, since we cannot compute the vacuum thrust when the reference pressure is None, we should raise a warning in this case to inform that the vacuum_thrust method will simply return the thrust.

kevin-alcaniz:
@phmbressan I've already applied the requested changes. I added a warning for the vacuum thrust when no reference pressure is defined, and improved the documentation for the exhaust velocity, both in the class attributes and the related functions.

If you come across anything else that needs attention, feel free to let me know!

phmbressan (approved):

Issue #668, opened 2024-08-19
ENH: Create a RocketPy file (.rpy) to store the project


Gui-FernandesBR:
There's already some draft cod in the first hackathon repo

PR #800, opened 2025-04-02
ENH: Create a rocketpy file to store flight simulations

ArthurJWH:
## Pull request type

## Checklist

## Current behavior

Issue #668

## New behavior

ENH: The `save_to_rpy` and `load_from_rpy` methods are implemented. With these new methods it is possible to use the `resimulate` attribute to quickly save the simulation's solution and reload it into the `Flight` object (with `info` and `all_info` methods)

## Breaking change

- [ ] Yes
- [ ] No


rocketpy/utilities.py
```diff
@@ -685,3 +690,71 @@ def get_instance_attributes(instance):
...
+            data,
+            f,
+            cls=RocketPyEncoder,
+            indent=2,
+            include_outputs=include_outputs,
+        )
+
+
+def load_from_rpy(filename: str, resimulate = False):
+    """Loads the saved data from the .rpy into a Flight object.
```
Gui-FernandesBR, line 730:
```suggestion
    """Loads the saved data from a .rpy file into a Flight object.
```

rocketpy/utilities.py
```diff
@@ -685,3 +690,71 @@ def get_instance_attributes(instance):
...
+
+    Parameters
+    ----------
+    filename : str
+        Path where the file to be loaded is
+    resimulate : bool, optional
+        If True, the function will resimulate the object Flight,
```
Gui-FernandesBR, line 737:
```suggestion
        If True, the function will resimulate the Flight object,
```

rocketpy/utilities.py
```diff
@@ -685,3 +690,71 @@ def get_instance_attributes(instance):
...
+                              f"RocketPy (v{data["version"]}), the current",
+                              f"imported module is v{version('rocketpy')}")
+            simulation = json.dumps(data["simulation"])
+            flight = json.loads(
+                simulation,
+                cls=RocketPyDecoder,
+                resimulate=resimulate)
+        return flight
+    else:
+        raise ValueError(f"Invalid file extension: {ext}. Allowed: .rpy")
```
Gui-FernandesBR, line 760: You could invert the order of testing here: first you raise the exception, then you run your code.

```suggestion
    if ext != ".rpy":
        raise ValueError(f"Invalid file extension: {ext}. Allowed: .rpy")

    with open(filename, "r") as f:
        data = json.load(f)
        if data["version"] > version("rocketpy"):
            warnings.warn("The file was saved in an updated version of",
                        f"RocketPy (v{data["version"]}), the current",
                        f"imported module is v{version('rocketpy')}")
        simulation = json.dumps(data["simulation"])
        flight = json.loads(
            simulation,
            cls=RocketPyDecoder,
            resimulate=resimulate)
    return flight
```
This way you don't have to ident all our code.

rocketpy/simulation/flight.py
```diff
@@ -3433,24 +3433,42 @@ def to_dict(self, include_outputs=False):
             "time_overshoot": self.time_overshoot,
             "name": self.name,
             "equations_of_motion": self.equations_of_motion,
+            "solution": self.solution, ##
```
Gui-FernandesBR, line 3436: What are these `#` comments? 

```suggestion
            "solution": self.solution,
```

rocketpy/utilities.py
```diff
@@ -685,3 +690,71 @@ def get_instance_attributes(instance):
...
+        by default False
+        
+    Returns
+    -------
+    None
+    """
+    file = os.path.splitext(filename)[0] + ".rpy"
+    with open(file, "w") as f:
+        data = {"date": str(date.today()), "version": version("rocketpy")}
+        print(flight)
```
Gui-FernandesBR, line 716: Do you really need the print() here? Was this used only for debugging purposes?

```suggestion
```

rocketpy/simulation/flight.py
```diff
@@ -3433,24 +3433,42 @@ def to_dict(self, include_outputs=False):
...
                 {
                     "time": self.time,
-                    "out_of_rail_time": self.out_of_rail_time,
                     "out_of_rail_velocity": self.out_of_rail_velocity,
                     "out_of_rail_state": self.out_of_rail_state,
-                    "apogee": self.apogee,
-                    "apogee_time": self.apogee_time,
                     "apogee_x": self.apogee_x,
                     "apogee_y": self.apogee_y,
                     "apogee_state": self.apogee_state,
```
Gui-FernandesBR, line 3471: Any reasons for moving all these attributes from outputs to the standard returned dictionary?

rocketpy/utilities.py
```diff
@@ -685,3 +690,71 @@ def get_instance_attributes(instance):
...
+        Path where the file will be saved in
+    include_output : bool, optional
+        If True, the function will include extra outputs into the file,
+        by default False
+        
+    Returns
+    -------
+    None
+    """
+    file = os.path.splitext(filename)[0] + ".rpy"
```
Gui-FernandesBR, line 713: ```from pathlib import Path``` 

```suggestion
    file = Path(filename).with_suffix(".rpy")

```

rocketpy/utilities.py
```diff
@@ -685,3 +690,71 @@ def get_instance_attributes(instance):
...
+
+    Parameters
+    ----------
+    flight : rocketpy.Flight
+        Flight object containing the rocket's flight data
+    filename : str
+        Path where the file will be saved in
+    include_output : bool, optional
+        If True, the function will include extra outputs into the file,
+        by default False
```
Gui-FernandesBR, line 707: Did you consider setting the default value to True instead?

Gui-FernandesBR (commented):
I'm impressed by what I see in this PR.
Almost perfect PR, including tests, documentation, etc.
Great work.

There are a few comments that I think we should discuss on.

rocketpy/simulation/flight.py
ArthurJWH: Those were to keep track of the new data to be stored, I forgot to remove it

rocketpy/simulation/flight.py
ArthurJWH: These attributes are necessary for the `info` and `all_info` methods in the `Flight` object, so I thought it was more convenient to have it in the standard dictionary

rocketpy/utilities.py
ArthurJWH: It was for debugging purpose, I just forgot to remove it. Thanks

rocketpy/simulation/flight.py
Gui-FernandesBR: > These attributes are necessary for the `info` and `all_info` methods in the `Flight` object, so I thought it was more convenient to have it in the standard dictionary

Could you add a small comment to the code so we avoid future questions just like the one I did?

rocketpy/utilities.py
ArthurJWH: I left the default value as `False` to keep it consistent to the `RocketPyEncoder` class. Also, one of the main purposes of these methods is to create a quicker way to save and load `Flight` within the team. Thus, it was thought to be better having `include_outputs` and `resimulate` defaulted to `False`

rocketpy/simulation/flight.py
ArthurJWH: Would it be good to create a small docstring for this method to clearly present this information?

Gui-FernandesBR: > Would it be good to create a small docstring for this method to clearly present this information?

Not mandatory.
If you create a docstring, the method will be susceptible to breaking changes, I guess that's why we haven't documented it yet.

A commentary would be enough.

However, if you want to proceed and create a complete docstring for the method, I'm not against it either.

rocketpy/utilities.py
Gui-FernandesBR: Pathlib is more specialized to dealing with paths, and it is a part of the standard lib, so it's more recommended in this case.
I was afraid of possible errors from the `+` operation. Perhaps the first part may not always be a string.

rocketpy/_encoders.py
```diff
@@ -84,7 +87,50 @@ def object_hook(self, obj):
...
+                    new_flight.function_evaluations = obj["function_evaluations"]
+                    new_flight.alpha1 = obj["alpha1"]
+                    new_flight.alpha2 = obj["alpha2"]
+                    new_flight.alpha3 = obj["alpha3"]
+                    new_flight.R1 = obj["R1"]
+                    new_flight.R2 = obj["R2"]
+                    new_flight.R3 = obj["R3"]
+                    new_flight.M1= obj["M1"]
+                    new_flight.M2 = obj["M2"]
+                    new_flight.M3 = obj["M3"]
```
Gui-FernandesBR, line 131: there should be a better way of doing this unpacking operation... I just cannot think of one right now.

Gui-FernandesBR (commented):
@ArthurJWH , could you please update the CHANGELOG file?

rocketpy/_encoders.py
ArthurJWH: I just searched up. Do you think looping through the attributes and using` setattr` would work?

Gui-FernandesBR: > I just searched up. Do you think looping through the attributes and using` setattr` would work?

I would - personally - be more confortable with the loop+setattr solution. It seems more pythonic.

Gui-FernandesBR (approved):
LGTM

rocketpy/utilities.py
```diff
@@ -688,3 +694,68 @@ def get_instance_attributes(instance):
...
+        If True, the function will resimulate the Flight object,
+        by default False
+
+    Returns
+    -------
+    rocketpy.Flight
+        Flight object containing simulation information from the .rpy file
+    """
+    ext = os.path.splitext(os.path.basename(filename))[1]
+    if ext != ".rpy":
```
Gui-FernandesBR, line 748:
```suggestion
    if ext != ".rpy":  # pragma: no cover
```

rocketpy/_encoders.py
```diff
@@ -75,6 +77,7 @@ class RocketPyDecoder(json.JSONDecoder):
     different types of objects from a JSON supported format."""
 
     def __init__(self, *args, **kwargs):
+        self.resimulate = kwargs.pop("resimulate", False)
```
Gui-FernandesBR, line 80: Do you realize that `.pop()` will remove the "resimulate" will remove the item from the dict?
Is it a problem in this case?

Remind there's a more common method called `.get` which gets the element without deleting (or "popping out") the element from the dict

Gui-FernandesBR (commented):
As discussed in the meeting yesterday, it would be nice really to have a Flight.save_rpy and Flight.load_from_rpy method.

I'm just not entirely sure whether we should add it to this PR or not. If so, we should move the new functions from utilities to the tools module.

But I was thinking about it and a question has raised: What would happen if we would like to store a multi-stage rocket flight? in this case we would need a composition of multiple Flight objects.
Do you have any opinion on this matter, @ArthurJWH @phmbressan

ArthurJWH:
> As discussed in the meeting yesterday, it would be nice really to have a Flight.save_rpy and Flight.load_from_rpy method.
> 
> I'm just not entirely sure whether we should add it to this PR or not. If so, we should move the new functions from utilities to the tools module.
> 
> But I was thinking about it and a question has raised: What would happen if we would like to store a multi-stage rocket flight? in this case we would need a composition of multiple Flight objects. Do you have any opinion on this matter, @ArthurJWH @phmbressan

About moving the methods to `tools.py`, I have tried it earlier and I found some circular dependencies. Basically the modules import each other in a circle (e.g. tools import encoder, encoder import flight_plot, flight_plot import plot_helper and plot_helper import tools, or tools import encoder, encoder import function, and function import tools)

There might be some way to do it, but it will definitely take a lot of work

rocketpy/_encoders.py
ArthurJWH: I believe that since it is a `kwargs` parameter, it will be used only once when loading

Gui-FernandesBR:
> > As discussed in the meeting yesterday, it would be nice really to have a Flight.save_rpy and Flight.load_from_rpy method.
> > I'm just not entirely sure whether we should add it to this PR or not. If so, we should move the new functions from utilities to the tools module.
> > But I was thinking about it and a question has raised: What would happen if we would like to store a multi-stage rocket flight? in this case we would need a composition of multiple Flight objects. Do you have any opinion on this matter, @ArthurJWH @phmbressan
> 
> About moving the methods to `tools.py`, I have tried it earlier and I found some circular dependencies. Basically the modules import each other in a circle (e.g. tools import encoder, encoder import flight_plot, flight_plot import plot_helper and plot_helper import tools, or tools import encoder, encoder import function, and function import tools)
> 
> There might be some way to do it, but it will definitely take a lot of work

Nevermind, I think it is ok to keep it where it currently is.
Let's proceed with this PR, so we can test the feature before improve it.

PR #803, opened 2025-04-13
ENH: allow users to provide custom samplers

Lucas-Prates:
## Pull request type

## Checklist

## Current behavior

Currently, the user must use distributions that are specified in numpy.

## New behavior

Allows the user to build upon the abstract class `CustomSampler` which must contain two methods: `sample` and `reset_seed`. Note that the user may implement very complex samplers that use real data, so long as it conforms to returning the data in the format specified.

## Breaking change

- [x] No

## Additional information

- [ ] Add example to Monte Carlo class usage notebook
- [ ] Verify that the seed is being properly reset in parallel Monte Carlo simulations


Lucas-Prates:
@Gui-FernandesBR, @MateusStano and @phmbressan, I would like, if possible, a design check in this feature. I am not sure if this is what you had in mind.

In order to allow the user to specify his own sampler, I thought of providing an abstract class `CustomSampler` and allow him to implement two methods: `sample` and `reset_seed`. These are completely abstracted away from the Stochastic classes, and we only care about the return value. The sample itself might be very complex and even fetch outside data if the user wants to.

I provided a notebook to test the feature design. There, I implemented a MixtureOfGaussian distribution which is not natively implemented in numpy. (Think of a bimodal gaussian distribution if you want, as exemplified below.

image

test_custom_sampler.ipynb
Gui-FernandesBR, line 1: Please delete this jupyter notebook.

I really recommend you to create a section in our docs to briefly describe the feature

Gui-FernandesBR (approved):
Plrase update the changelog

test_custom_sampler.ipynb
Lucas-Prates: Sure, the notebook is only to get a feel if this is the intended design. I thought it was not a good idea to make tests to something that could change drastically.

MateusStano (approved):

MateusStano:
Just missing changelog, then we can merge

Issue #802, opened 2025-04-05
BUG: Flight class with Two Parachute and time_overshoot = False don't define some functions


kevin-alcaniz:
**Describe the bug**

When simulating a `Rocket` with two `Parachute` instances in the `Flight` class and setting the `time_overshoot` attribute to `False`, several functions return undefined values: `stability_margin`, `mach_number`, `free_stream_speed`, `stream_velocity_x`, `wind_velocity_x`, and `z`.

As a result, running a `MonteCarlo` simulation with this setup raises a `ValueError` because these values can't be retrieved.

Interestingly, the issue doesn’t occur when the `Rocket` has only one of those `Parachute` instances — it works fine regardless of which one is used.

Could you take a look?

PR #805, opened 2025-04-18
Bug: Fix StochasticFlight time_overshoot None bug

### LINKED ISSUE #804: BUG: MonteCarlo simulations not working when there are StochasticParachutes
**Describe the bug**

The `MonteCarlo.simulate` method encounters a `ValueError` when evaluating the pressure on the parachutes.

**To Reproduce**

Try to run a Monte Carlo simulation with a `StochasticParachute`. Otherwise, just run the _monte_carlo_class_usage_ notebook.

**Error debugging**

I debugged the code for a bit and the problem seems to be that the `time_overshoot` has a default of `None` in the `StochasticFlight` class, introduced in #785 . A `Flight` object is created with `time_overshoot=None`, which in turn makes the `__simulate` method skip running the `__calculate_and_save_pressure_signals` function, which in turn makes `parachute.clean_pressure_signal` empty and, finally, causes an error in `__transform_pressure_signals_lists_to_functions` since an empty argument is passed to a `Function`.

**Possible fix**

When I set `time_overshoot=True`, that works just fine. 

When I set `time_overshoot=False`, an `ValueError` is encountered once again, this time because several arrays related to contain NaN or Inf when trying to create the cache ? (See the screenshot)

An easy fix is to make `time_overshoot` equals of the base flight when it is None:

```
if time_overshoot is None:
    self.time_overshoot = flight.time_overshoot
else:
    self.time_overshoot = time_overshoot
```

However, that will not fix the problem when `time_overshoot` is `False` for the base flight. This is probably connected to issue #802 . It might be better to take a closer look.

**Screenshots**

Bug encountered in pressure evaluation (`time_overshoot=None`):
Image

Bug encountered (`time_overshoot=False`)

Image

**Additional context**

Shout-out to **MRT014 [LURA]**, a discord member that found the bug and provided me good feedback. Thank you :)

Lucas-Prates:
## Pull request type

## Checklist

## Current behavior

`time_overshoot=None` causes an error in `Flight._simulate`.

## New behavior

Fixes #804 by setting `time_overshoot` to the base flight value when it is `None`. 

## Breaking change

- [x] No

## Additional information

Still need to fix `time_overshoot=False` problem described in #802

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ Attention: The newest changes should be on top -->
 
 ### Added
 
+- BUG: Fix StochasticFlight time_overshoot None bug [#805] (https://github.com/RocketPy-Team/RocketPy/pull/805)
 - ENH: Implement Multivariate Rejection Sampling (MRS) [#738] (https://github.com/RocketPy-Team/RocketPy/pull/738) 
 - ENH: Create a rocketpy file to store flight simulations [#800](https://github.com/RocketPy-Team/RocketPy/pull/800)
 - ENH: Support for the RSE file format has been added to the library [#798](https://github.com/RocketPy-Team/RocketPy/pull/798)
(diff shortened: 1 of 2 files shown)


Gui-FernandesBR (approved):
Please update the CHANGELOG.

LGTM

PR #810, opened 2025-04-22
ENH: support for ND arithmetical operations for Function.

phmbressan:
## Pull request type

## Checklist

## Current behavior

PR #785 introduced the possibility of 2D `Function` multiplication. The intent of this PR is to generalize this to the other arithmetical operations and to ND `Functions` for better scalability (e.g. so as not keep adding if blocks for every dimension, like 3D, 4D etc).

## New behavior

The generalization is done with the following changes:
- Use the `self._domain` and `self._image` for array manipulation, instead of unidimensional `self.x_array` and `self.y_array`;
- Multivariate lambda expressions are generated to allow callable operations:
    - The implementation is less trivial, requiring dynamically building a multivariate `lambda`;
    - This PR implementation worked very well on testing, but, if it does not hold review scrutinity due the use of `eval`, it can easily be converted into an `exception` and disallowing it.
    
Tests were introduced for both existing 2D operations and for the new generalized ones.

Overall the structure of the arithmetic methods of `Function` got a bit simpler and more regular. Another important notice, is that the current multivariable implementation of `mul` from #785 is non commutative when dealing with 1D*2D, which is fixed by this PR.

## Breaking change

- [ ] Yes
- [X] No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,6 @@ Attention: The newest changes should be on top -->
 ### Added
 - ENH: Support for ND arithmetic in Function class. [#810] (https://github.com/RocketPy-Team/RocketPy/pull/810)
 - ENH: allow users to provide custom samplers [#803](https://github.com/RocketPy-Team/RocketPy/pull/803)
-- BUG: Fix StochasticFlight time_overshoot None bug [#805] (https://github.com/RocketPy-Team/RocketPy/pull/805)
 - ENH: Implement Multivariate Rejection Sampling (MRS) [#738] (https://github.com/RocketPy-Team/RocketPy/pull/738) 
 - ENH: Create a rocketpy file to store flight simulations [#800](https://github.com/RocketPy-Team/RocketPy/pull/800)
 - ENH: Support for the RSE file format has been added to the library [#798](https://github.com/RocketPy-Team/RocketPy/pull/798)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 Attention: The newest changes should be on top -->
 
 ### Added
+- ENH: Support for ND arithmetic in Function class. [#810] (https://github.com/RocketPy-Team/RocketPy/pull/810)
 - ENH: allow users to provide custom samplers [#803](https://github.com/RocketPy-Team/RocketPy/pull/803)
+- BUG: Fix StochasticFlight time_overshoot None bug [#805] (https://github.com/RocketPy-Team/RocketPy/pull/805)
 - ENH: Implement Multivariate Rejection Sampling (MRS) [#738] (https://github.com/RocketPy-Team/RocketPy/pull/738) 
 - ENH: Create a rocketpy file to store flight simulations [#800](https://github.com/RocketPy-Team/RocketPy/pull/800)
 - ENH: Support for the RSE file format has been added to the library [#798](https://github.com/RocketPy-Team/RocketPy/pull/798)
--- a/rocketpy/mathutils/function.py
+++ b/rocketpy/mathutils/function.py
@@ -1948,7 +1948,7 @@ class Function:  # pylint: disable=too-many-public-methods
                     self.__outputs__,
                 )
             else:
-                param_names = [f"x{i}" for i in range(len(self.__inputs__))]
+                param_names = [f"x{i}" for i in range(self.__dom_dim__)]
                 param_str = ", ".join(param_names)
                 func_str = f"lambda {param_str}: -func({param_str})"
                 return Function(
@@ -2131,7 +2131,9 @@ class Function:  # pylint: disable=too-many-public-methods
             A Function object which gives the result of self(x)+other(x).
         """
         other_is_func = isinstance(other, Function)
-        other_array = other._source_type is SourceType.ARRAY if other_is_func else False
+        other_is_array = (
+            other._source_type is SourceType.ARRAY if other_is_func else False
+        )
         inputs = self.__inputs__[:]
         interp = self.__interpolation__
         extrap = self.__extrapolation__
@@ -2139,7 +2141,7 @@ class Function:  # pylint: disable=too-many-public-methods
 
         if (
             self._source_type is SourceType.ARRAY
-            and other_array
+            and other_is_array
             and np.array_equal(self._domain, other._domain)
         ):
             source = np.column_stack((self._domain, self._image + other._image))
@@ -2162,7 +2164,7 @@ class Function:  # pylint: disable=too-many-public-methods
         elif callable(other):
             if other_is_func:
                 other_dim = other.__dom_dim__
-                other = other.get_value_opt if other_array else other.source
+                other = other.get_value_opt if other_is_array else other.source
             else:
                 other_dim = len(signature(other).parameters)
 
@@ -2216,7 +2218,9 @@ class Function:  # pylint: disable=too-many-public-methods
             A Function object which gives the result of self(x)-other(x).
         """
         other_is_func = isinstance(other, Function)
-        other_array = other._source_type is SourceType.ARRAY if other_is_func else False
+        other_is_array = (
+            other._source_type is SourceType.ARRAY if other_is_func else False
+        )
         inputs = self.__inputs__[:]
         interp = self.__interpolation__
         extrap = self.__extrapolation__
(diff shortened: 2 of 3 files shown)


MateusStano (approved):

rocketpy/mathutils/function.py
```diff
@@ -2197,10 +2203,53 @@ def __sub__(self, other):
...
+                    self.__make_arith_lambda(
+                        operator.sub, self.get_value_opt, other, dom_dim, other_dim
+                    )
+                )
+            else:
+                raise TypeError(
+                    f"The number of parameters in the function to be subtracted ({other_dim}) "
+                    f"does not match the number of parameters of the Function ({dom_dim})."
+                )
+        raise TypeError("Unsupported type for subtraction")
```
MateusStano, line 2252: Are all these operation methods the same (add, sub, mul...)? They seem really similar if not equal, except for the operation itself. Can we put all these checks and Function creations in a separate function? And then just call this inside each of the operations methods?

rocketpy/mathutils/function.py
```diff
@@ -2176,7 +2182,7 @@ def __radd__(self, other):
         """
         return self + other
 
-    def __sub__(self, other):
+    def __sub__(self, other):  # pylint: disable=too-many-statements
         """Subtracts from a Function object and returns a new Function object
```
MateusStano, line 2186: Do we need to have the __sub__ defined like this? It used to just call __add__ and __mul__

I guess you are avoiding performing some checks more than once?

rocketpy/mathutils/function.py
```diff
@@ -2354,58 +2374,57 @@ def __truediv__(self, other):  # pylint: disable=too-many-statements
...
-            # Check if Function objects source is array or callable
-            # Check if Function objects have the same domain discretization
-            if (
-                isinstance(other.source, np.ndarray)
-                and isinstance(self.source, np.ndarray)
-                and self.__dom_dim__ == other.__dom_dim__
-                and np.array_equal(self.x_array, other.x_array)
-            ):
-                # operate on grid values
+        self_array = isinstance(self.source, np.ndarray)
```
MateusStano, line 2377: Shouldn't we have an attribute that defines which type of Function self is (if it is array defined or callable)? Then you don't need this check everywhere for both self and other

MateusStano (changes requested):
Can you check execution speed of the code? With Monte Carlo sims specially

There are a lot of new

phmbressan:
> Can you check execution speed of the code? With Monte Carlo sims especially
> 
> The operations seems different so I just want to make sure it is close to how it was before

### Performance Benchmarks

#### Function Composition (`f + g`) and Evaluation (`h(x)`)

| Test                         | Before                        | After                         |
|------------------------------|-------------------------------|-------------------------------|
| `f + g` (lambda)             | 16.2 μs ± 1.24 μs             | 17.5 μs ± 0.41 μs             |
| `h(3)` (lambda eval)         | 1.59 μs ± 0.013 μs            | 725 ns ± 5.47 ns              |
| `f + g` (array function)     | 110 μs ± 0.82 μs              | 102 μs ± 1.01 μs              |
| `h(2.5)` (array eval)        | 3.85 μs ± 0.045 μs            | 3.74 μs ± 0.068 μs            |

#### Monte Carlo Simulation

| Metric                   | Before         | After          |
|--------------------------|----------------|----------------|
| Avg Time per Iteration   | 1.337 s        | 1.276 s        |
| Total Simulations        | 250            | 250            |
| Total Wall Time          | 334.3 s        | 319.0 s        |

Overall the results are about the same speed, mostly within run to run variance. The exception is the Function lambda evaluation that is now consistently faster and I don't know why, still checking.

rocketpy/mathutils/function.py
phmbressan: I preferred to define the `sub`, since the "correct" definition would call add and neg (or mul, as you mentioned). This would make only the `sub` operation take about twice the time. Even though that may be a meaningless time save, I believe, for consistency, it makes sense to have all of them take about the same time.

rocketpy/mathutils/function.py
```diff
@@ -1927,13 +1941,22 @@ def __neg__(self):
...
-                self.__extrapolation__,
-            )
+            if self.__dom_dim__ == 1:
+                return Function(
+                    lambda x: -self.source(x),
+                    self.__inputs__,
+                    self.__outputs__,
+                )
+            else:
+                param_names = [f"x{i}" for i in range(len(self.__inputs__))]
```
Gui-FernandesBR, line 1951: Does it work for any N-d function? Not only 2 or 3D?

rocketpy/mathutils/function.py
```diff
@@ -2197,10 +2215,56 @@ def __sub__(self, other):
         result : Function
             A Function object which gives the result of self(x)-other(x).
         """
-        try:
-            return self + (-other)
-        except TypeError:
-            return Function(lambda x: (self.get_value_opt(x) - other(x)))
+        other_is_func = isinstance(other, Function)
+        other_array = other._source_type is SourceType.ARRAY if other_is_func else False
```
Gui-FernandesBR, line 2219: maybe? 

```suggestion
        other_is_array = other._source_type is SourceType.ARRAY if other_is_func else False
```

rocketpy/mathutils/function.py
phmbressan: Correct. For instance:

```bash
>>> a=Function(lambda x,y,z,w,m,n,k,q: x+y)
>>> b = -a
>>> a(1,5,3,4,5,6,7,8)
6
>>> b(1,5,3,4,5,6,7,8)
-6
>>> a
'Function from R8 to R1 : (x, y, z, w, m, n, k, q) → (Scalar)'
>>> b
'Function from R8 to R1 : (x, y, z, w, m, n, k, q) → (Scalar)'
```

rocketpy/mathutils/function.py
phmbressan: Sure! Done in [ae07fcb](https://github.com/RocketPy-Team/RocketPy/pull/810/commits/ae07fcb1cf3083f5c5050ab945c3e42511ecca47).

rocketpy/mathutils/function.py
MateusStano: Why use the Enum? Its just simpler to have two more constants in the top no?

rocketpy/mathutils/function.py
MateusStano: Okay, let it go for now then. We can get back to this if it is ever needed

rocketpy/mathutils/function.py
Gui-FernandesBR: I prefer using the Enum as well, it provides more readability and type safety.
We all should use Enum classes more often within this project.

tests/unit/test_function.py
```diff
@@ -283,7 +283,7 @@ def test_remove_outliers_iqr(x, y, expected_x, expected_y):
 def test_set_get_value_opt():
     """Test the set_value_opt and get_value_opt methods of the Function class."""
     func = Function(lambda x: x**2)
-    func.source = np.array([[1, 1], [2, 4], [3, 9], [4, 16], [5, 25]])
+    func.set_source(np.array([[1, 1], [2, 4], [3, 9], [4, 16], [5, 25]]))
```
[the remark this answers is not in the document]
Gui-FernandesBR: Nice catch!

rocketpy/mathutils/function.py
```diff
@@ -2108,61 +2130,60 @@ def __add__(self, other):  # pylint: disable=too-many-statements
...
-            # Or if it is just a callable
-            elif callable(other):
-                return Function(lambda x: (self.get_value_opt(x) + other(x)))
+                )
+            else:
+                raise TypeError(
+                    f"The number of parameters in the function to be added ({other_dim}) "
+                    f"does not match the number of parameters of the Function ({dom_dim})."
+                )
+        raise TypeError("Unsupported type for addition")
```
Gui-FernandesBR, line 2182: I usually don't request these kind of changes, but I believe in this case it's worth it, since it regards to the Function class.

Codecoverage is reporting that you didn't cover the raise exceptions. Please add more tests or skip the rows using `# pragma: no-cover`

CHANGELOG.md
```diff
@@ -31,7 +31,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 Attention: The newest changes should be on top -->
 
 ### Added
+- ENH: Support for ND arithmetic in Function class. [#810] (https://github.com/RocketPy-Team/RocketPy/pull/810)
 - ENH: allow users to provide custom samplers [#803](https://github.com/RocketPy-Team/RocketPy/pull/803)
+- BUG: Fix StochasticFlight time_overshoot None bug [#805] (https://github.com/RocketPy-Team/RocketPy/pull/805)
```
Gui-FernandesBR, line 36:
```suggestion
```

MateusStano (approved):

PR #811, opened 2025-04-24
BUG: Unecessary Gyroscope Rotation and Wrong Acceleremoter Rotation

MateusStano:
## Pull request type

## Description

Some rotations were wrong in the sensors class:

- The gyroscope receives an angular velocity described in the body frame. There was an extra rotation being made. That was removed since it is not needed
- The accelerometer receives an acceleration vector described in the inertial frame. This vector was then **wrongly**  rotated from body to inertial, when it should be inertial to body. This was fixed by adding a `.transpose`
- Acceleration and velocity attributes described in the body frame were created in the flight class for clarity

I also took the chance to improve/correct some of the docs strings of some Flight attributes

### DIFF
--- a/rocketpy/prints/sensors_prints.py
+++ b/rocketpy/prints/sensors_prints.py
@@ -74,7 +74,7 @@ class _InertialSensorPrints(_SensorPrints):
         self._print_aligned("Orientation:", self.sensor.orientation)
         self._print_aligned("Normal Vector:", self.sensor.normal_vector)
         print("Rotation Matrix:")
-        for row in self.sensor.rotation_matrix:
+        for row in self.sensor.rotation_sensor_to_body:
             value = " ".join(f"{val:.2f}" for val in row)
             value = [float(val) for val in value.split()]
             self._print_aligned("", value)
--- a/rocketpy/simulation/flight.py
+++ b/rocketpy/simulation/flight.py
@@ -104,11 +104,14 @@ class Flight:
     Flight.z : Function
         Rocket's z coordinate (positive up) as a function of time.
     Flight.vx : Function
-        Rocket's X velocity as a function of time.
+        Velocity of the rocket's center of dry mass in the X (East) direction of
+        the inertial frame as a function of time.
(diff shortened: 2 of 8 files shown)


rocketpy/prints/sensors_prints.py
```diff
@@ -74,7 +74,7 @@ def orientation(self):
         self._print_aligned("Orientation:", self.sensor.orientation)
         self._print_aligned("Normal Vector:", self.sensor.normal_vector)
         print("Rotation Matrix:")
-        for row in self.sensor.rotation_matrix:
+        for row in self.sensor.rotation_sensor_to_body:
```
Gui-FernandesBR, line 77: maybe you should emphasize in the print message that you're using the `rotation_sensor_to_body` matrix rather than the other option. What do you think?

tests/unit/test_sensor.py
```diff
@@ -291,7 +293,7 @@ def test_noisy_rotated_accelerometer(noisy_rotated_accelerometer, example_plain_
         euler313_to_quaternions(*np.deg2rad([60, 60, 60]))
     )
     total_rotation = sensor_rotation @ cross_axis_sensitivity
-    rocket_rotation = Matrix.transformation(U[6:10])
+    rocket_rotation = Matrix.transformation(U[6:10]).transpose
```
Gui-FernandesBR, line 296: does the `transformation` method simply transpose the matrix? if so, shouldn't we just use U instead of transformation(U).transpose?

rocketpy/simulation/flight.py
Gui-FernandesBR, line 1: great work refactoring the docstrings!

Gui-FernandesBR (approved):
LGTM,

Nice work solving the BUG so promptly. Thanks, @MateusStano !!

tests/unit/test_sensor.py
MateusStano: The transformation method creates a rotation matrix from the quaternions

Issue #657, opened 2024-08-19
ENH: Use `EnvironmentAnalysis` to initialize the `Flight` class


Gui-FernandesBR:
Hi, I hope this email finds you well.

My name is Gregorio and I am using RocketPy to do trajectory simulation for competition rockets. I really appreciate all your work.

For a competition we have later this year (EuRoC), we need to do a preliminary simulation and we would like to make an Environmental Analysis and use that as our Environment for the simulation. We can make the EnvironmentalAnalysis and it works fine. However, we cannot use it directly as a parameter for the flight.

We would like to do something like this:

"
Ea = EnvironmentAnalysis(
        start_date=date1,
        end_date=date2,
        start_hour=8,
        end_hour=19,
        latitude=lat ,
        longitude=lon,
        surface_data_file="./single_levels.nc",
        pressure_level_data_file="./pressure_file.nc,
        timezone="CET",
        unit_system="metric",
    )

......

flight = Flight(
    environment = Ea,
    rocket = myRocket,
    rail_length = rail_length,
    inclination = inclination,
    heading = direction,
    name=f"Nominal",
)
"

However, that doesn't work, and we are doing it the following (indirect) way:

"
Ea = EnvironmentAnalysis(
        start_date=date1,  
        end_date=date2,  
        start_hour=8,
        end_hour=19,
        latitude=lat ,
        longitude=lon,
        surface_data_file="./single_levels.nc",
        pressure_level_data_file="./pressure_file.nc,
        timezone="CET",
        unit_system="metric",
    )

# Get the mean values from the Environmenatal Analysis
altitude_list = Ea.altitude_list
temperature_profile = Ea.average_temperature_profile + 273.15 # converting from °C to °K
wind_heading_profile = Ea.average_wind_heading_profile
wind_speed_profile = Ea.average_wind_speed_profile
pressure_profile = Ea.average_pressure_profile * 100 # converting from Pa to HPa

wind_u_profile = -1 * wind_speed_profile * np.sin(np.radians(wind_heading_profile))
wind_v_profile = -1 * wind_speed_profile * np.cos(np.radians(wind_heading_profile))

# Convert the data to the form accepted by set_environment_model()
alt_p = list(zip(altitude_list, pressure_profile))
alt_t = list(zip(altitude_list, temperature_profile))
alt_wind_v = list(zip(altitude_list, wind_v_profile))
alt_wind_u = list(zip(altitude_list, wind_u_profile))

[...]

PR #813, opened 2025-05-05
ENH: Environment object from EnvironmentAnalysis

ArthurJWH:
This method allows the user to create a ```Environment``` object from the ```EnvironmentAnalysis``` object

## Pull request type

## Checklist

## Current behavior

Issue #657 

## New behavior

The ```env``` method have been added, which creates a ```Environment``` object from the data retrieved from the ```EnvironmentAnalysis``` class

## Breaking change

- [ ] Yes
- [ ] No

### DIFF
--- a/rocketpy/environment/environment_analysis.py
+++ b/rocketpy/environment/environment_analysis.py
@@ -2885,3 +2885,25 @@ class EnvironmentAnalysis:  # pylint: disable=too-many-public-methods
         file.write(encoded_class)
         file.close()
         print("Your Environment Analysis file was saved, check it out: " + filename)
+
+    def env(
+        self, gravity=None, date=None, datum="SIRGAS2000", max_expected_height=80000.0
+    ):
+        env = Environment(
+            gravity,
+            date,
+            self.latitude,
+            self.longitude,
+            self.converted_elevation,
+            datum,
+            self.preferred_timezone,
+            max_expected_height,
+        )
+        env.set_atmospheric_model(
+            type="custom_atmosphere",
+            pressure=list(zip(self.altitude_list, self.average_pressure_profile)),
+            temperature=list(zip(self.altitude_list, self.average_temperature_profile)),
+            wind_u=list(zip(self.altitude_list, self.average_wind_velocity_x_profile)),
+            wind_v=list(zip(self.altitude_list, self.average_wind_velocity_y_profile)),
+        )
+        return env


rocketpy/environment/environment_analysis.py
```diff
@@ -2885,3 +2885,25 @@ def save(self, filename="env_analysis_dict"):
         file.write(encoded_class)
         file.close()
         print("Your Environment Analysis file was saved, check it out: " + filename)
+
+    def env(
+        self, gravity=None, date=None, datum="SIRGAS2000", max_expected_height=80000.0
+    ):
```
Gui-FernandesBR, line 2891: Two points:

1. The method would benefit a lot from a docstring!
2. The name could be a lot more clearer. A few options should be: `to_rocketpy_environment`, `to_environment_object`, `get_environment_object`, `instantiate_environment_object`. I'm giving some very different options because I'd like you to propose what you think would be clearer to the final user. Anything apart from "env" should already be fine.

Gui-FernandesBR: Please mention [in docs] you're using the average profiles.

ArthurJWH: Yes. I'm just waiting for a response on how to deal with the inversion function before proceeding, because currently there is a problem with the pressure profile. But after the code works, I'll be making the code cleaner (tests, documentation, the name...)

ArthurJWH:
> This is really nice. So simple and effective. Great work @ArthurJWH ! I think you're getting the hang of it...
> 
> Some extra points I'd like to add before :
> 
> * could you please update the CHANGELOG?
> * How familiar are you with  unit tests?
> * There's an open issue in this repo that may be connected with your PR. Could you please link the PR using the "development" section on the "conversation" section here on GitHub?

I will update the CHANGELOG soon.
For tests, I just had one test asserting that the method would return an ```Environment``` object. But there is this one problem with the pressure profile that, if it is not "bijective", then the invert function would just give an error.
The issue #657 should be already linked, if there is other related issue, let me know.

rocketpy/environment/environment_analysis.py
Gui-FernandesBR, line 1: Environment analysis has this option in init method where you can choose whether you wanna work with imperial or metric unit system.
Could you double check that please?

I'm guessing you should convert units only if the unit system is not metric.

Gui-FernandesBR: you`re right, thanks

rocketpy/environment/environment_analysis.py
```diff
@@ -2885,3 +2885,79 @@ def save(self, filename="env_analysis_dict"):
         file.write(encoded_class)
         file.close()
         print("Your Environment Analysis file was saved, check it out: " + filename)
+
+    def get_environment_object(
+        self, gravity=None, date=None, datum="SIRGAS2000", max_expected_height=80000.0
+    ):
```
MateusStano, line 2891: Could you change the name from `get_` to `create_`? 

`get` implies that you are just accessing an attribute or something of the sort, create would make more sense here.

After this change we can merge

Gui-FernandesBR (approved):
If it works on your machine, it Looks Great To Me (LGTM) then.
Ready to merge.

tests/integration/test_environment_analysis.py
```diff
@@ -55,3 +56,9 @@ def test_exports(mock_show, env_analysis):  # pylint: disable=unused-argument
     os.remove("env_analysis_dict")
     os.remove("wind_rose.gif")
     os.remove("export_env_analysis.json")
+
+
+@pytest.mark.slow
+@patch("matplotlib.pyplot.show")
+def test_get_environment_object(mock_show, env_analysis):  # pylint: disable=unused-argument
+    assert isinstance(env_analysis.get_environment_object(), Environment)
```
MateusStano, line 64: @ArthurJWH this still needs the name change

PR #814, opened 2025-05-06
ENH: create a dataset of pre-registered motors.

leogrosa:
## Pull request type

## Checklist

## Current behavior

The user had to add its own .eng file to use the method ``GenericMotor.load_from_eng_file()`` to create a GenericMotor from it, and RocketPy didn't have a dataset of pre-registered motors.

## New behavior

RocketPy now has its own dataset of pre-registered motors, containing 63 .eng files, located in ```rocketpy/datasets/motors/```.
Three functions are now present in ``utilities.py`` to help:
- ``list_motors_dataset()``: returns a list of all available motors names;
-  ``load_motor_from_dataset()``: load a GenericMotor from a motor name from the dataset;
-  ``show_motors_dataset()``: prints the list of available motors, useful for quick inspection in terminals and Jupyter notebooks.

## Breaking change

- [ ] Yes
- [x] No

## Additional information

The user guide has documentation for the three new functions:
Captura de Tela 2025-05-05 às 23 48 11
Captura de Tela 2025-05-05 às 23 49 17
Captura de Tela 2025-05-05 às 23 49 44

- The following test didn't pass locally: 
```console
FAILED tests/integration/test_environment.py::test_rap_atmosphere - ValueError: The chosen launch time '2025-05-05-02: UTC' is not available in the provided file. Please choose a time within the range of the file, which starts at '2025-05-06-01 UTC'.
```

### To be discussed
- **Should this dataset be installed with the default installation of RocketPy using ``pip``?** For now, the size of the dataset isn't big, so the easiest path is to just add it in the repo. However, in the future, it may become big enough to make it optional. Preparing the repo right now could be a good idea -- but it would demand changes in the PyPI pipeline.


leogrosa:
@Gui-FernandesBR, ok, I'll search for a nice way to make it optional and discuss it in the weekly. I know that it will involve another member of the team that has access to the PyPI configuration.

Regarding your concerns:

1 - There are ways, but the team and I discussed in the weekly that creating the datasets directory would be a good choice because:
- It separates what is meant to be used programmatically, as a resource (the ```datasets/``` directory), from what is meant to be used in examples or sandbox exploration (the ```data/``` directory), organizing it in a structured way (in this case, only having .eng files for motors, and nothing else);
- It allows RocketPy to reference the dataset in other parts of the code using ```rocketpy.datasets```. I believe it wouldn't be good to reference the ```data``` folder like that, since it's not inside the ```rocketpy``` directory;
- It is inspired in other well-stablished projects, like the color palettes in **seaborn**, or the datasets from **scikit-learn** (```sklearn.datasets```).

2 - I thought it would fit better in the utilities because there may be future functions that also use the ```datasets``` directory, but are not necessarily related to the ``GenericMotor`` class. However, the fact that I've created the documentation inside the ``GenericMotor`` part in the user guide looks a little bit like an inconsistency haha, even though I thought it would be better for it to be there when thinking about its use case.

MateusStano:
@leogrosa please try to make the dataset optional on the installation. We don't have anything of the sort in the repo right now so you will need to research a bit on how to do it

leogrosa:
@Gui-FernandesBR @MateusStano 

So, I did some research and found that it's not possible to make *part* of this repository optional when installing via pip.

The only way to achieve that would be by hosting the files elsewhere. This could be done in many ways, but I believe the most scalable and maintainable option would be to **create a separate repository under the RocketPy-Team organization**. We could then include it as an optional dependency in RocketPy, using pip’s `extras_require` mechanism. That way, we avoid implementing our own logic for download handling, file hosting, retries, etc — all of which pip already handles very well.

We could create a repository named `rocketpy-datasets`, but that would lead to the same problem we're trying to avoid: increasing size over time. For instance, if a user wants to install only the motors dataset, they'd still have to download the entire dataset repository, including unrelated data (e.g., weather, atmospheric profiles...).

So, my proposal is to create a dedicated repository for each dataset. For example:
- `rocketpy-motor-dataset`
- `rocketpy-weather-dataset`
- ...

This granularity would allow us to keep each dataset truly optional, and help prevent unnecessary package size growth in the long term.

The only downside is having to maintain multiple small repositories. However, since these are intended to store only static data files, setup should be straightforward. I can even prepare a private “dataset repository template” to simplify this process for future datasets.

What do you guys think?

Gui-FernandesBR:
> @Gui-FernandesBR @MateusStano
> 
> So, I did some research and found that it's not possible to make _part_ of this repository optional when installing via pip.
> 
> The only way to achieve that would be by hosting the files elsewhere. This could be done in many ways, but I believe the most scalable and maintainable option would be to **create a separate repository under the RocketPy-Team organization**. We could then include it as an optional dependency in RocketPy, using pip’s `extras_require` mechanism. That way, we avoid implementing our own logic for download handling, file hosting, retries, etc — all of which pip already handles very well.
> 
> We could create a repository named `rocketpy-datasets`, but that would lead to the same problem we're trying to avoid: increasing size over time. For instance, if a user wants to install only the motors dataset, they'd still have to download the entire dataset repository, including unrelated data (e.g., weather, atmospheric profiles...).
> 
> So, my proposal is to create a dedicated repository for each dataset. For example:
> 
> * `rocketpy-motor-dataset`
> * `rocketpy-weather-dataset`
> * ...
> 
> This granularity would allow us to keep each dataset truly optional, and help prevent unnecessary package size growth in the long term.
> 
> The only downside is having to maintain multiple small repositories. However, since these are intended to store only static data files, setup should be straightforward. I can even prepare a private “dataset repository template” to simplify this process for future datasets.
> 
> What do you guys think?

Thanks for the update, Leo.
I need to think a bit more about the problem before answering you.

Issue #801, opened 2025-04-04
ENH: Enable only radial burning


caioessouza:
**Is your feature request related to a problem? Please describe.**

Rocketpy SolidMotor class only simulates radial and axial simoultaneos burn, but the regression rate on hybrid motor are modeled to be a function of the oxidizer mass flux entering the grain, so it only takes into consideration the radial burn.

**Describe the solution you'd like**

The idea is to implement the option to simulate only radial burn for the cases of a  axially inhibited grain or hybrid motors.

**Additional context**

![Image](https://github.com/user-attachments/assets/0fe3216b-4a66-4625-b86c-c5e0c2af44eb)

![Image](https://github.com/user-attachments/assets/4405a837-68e9-4b63-8870-79c87ba18e49)

Heeg, F., Kilzer, L., Seitz, R., & Stoll, E. (05 2020). Design and Test of a Student Hybrid Rocket Engine with an External Carbon Fiber Composite Structure. Aerospace, 7, 57. doi:10.3390/aerospace7050057

Gui-FernandesBR:
does it happen with **every** hybrid motor?

caioessouza:
> does it happen with **every** hybrid motor?

I can't ensure you that it happens to **every** hybrid motor. It's a model, as the combustion happens in the boundary layer between the liquid oxidizer and the solid fuel, the hypothesis is that the reaction is much more prominent in the port, where the burn is axial, because of the stronger oxidizer flow velocity, while the reaction in the bases of the grain are negligible.

This model is based on the source below and is largely cited, including in NASA SP7002-02.
G. A. MARXMAN, C. E. WOOLDRIDGE and R. J. MUZZY. Fundamentals of Hybrid
Boundary-Layer Combustion. In: Heterogeneous Combustion. Vol. 15. Progress in As-
tronautics and Rocketry.

caioessouza:
I runned some sanity check tests, just to see if the physical behaviour is correct, I used the motor Pro75M1670, defined in the getting started file.

![Image](https://github.com/user-attachments/assets/d487c528-2d34-42c0-9289-e84fd2e14cca)

1. I was expecting to the burn rate to grow in the begining, since it's calculated using
$r = - \frac{\dot m}{A_b \rho}$
and it was exactly what I got.

![Image](https://github.com/user-attachments/assets/2fd9234b-7aff-49e9-9dfc-d995ae98beb5)

[...]

PR #815, opened 2025-05-09
ENH: Enable only radial burning

caioessouza:
## Pull request type

## Checklist

## Current behavior

Rocketpy SolidMotor class simulates the grain regression either radially and axially by default.

This PR is related to the following issue:
ENH: Enable only radial burning #801 

## New behavior

Now it's possible to the user to change from the default behavior to a only radial regression by setting the new "only_radial_burning" optional parameter as True in the SolidMotor class.

## Breaking change

- [ ] Yes
- [x] No

## Additional information

Here is an example of the comparison between the default burning and the only radial one applied to the getting_started.ipynb simulation. The pytest tests haven't passed locally because of a windows problem.

GRAFICO1NOVO
GRAFICO2NOVO
GRAFICO3NOVO

### DIFF
--- a/rocketpy/motors/solid_motor.py
+++ b/rocketpy/motors/solid_motor.py
@@ -300,6 +300,11 @@ class SolidMotor(Motor):
             positions specified. Options are "nozzle_to_combustion_chamber" and
             "combustion_chamber_to_nozzle". Default is
             "nozzle_to_combustion_chamber".
+        only_radial_burn : boolean, optional
+            If True, inhibits the grain from burning axially, only computing
+            radial burn. Otherwise, if False, allows the grain to also burn 
+            axially. May be useful for axially inhibited grains or hybrid motors.
+            Default is False.
 
         Returns
         -------
@@ -316,7 +321,6 @@ class SolidMotor(Motor):
             reshape_thrust_curve=reshape_thrust_curve,
             interpolation_method=interpolation_method,
             coordinate_system_orientation=coordinate_system_orientation,
-            only_radial_burn = only_radial_burn,
         )
         # Nozzle parameters
         self.throat_radius = throat_radius
@@ -484,13 +488,7 @@ class SolidMotor(Motor):
             # Compute state vector derivative
             grain_inner_radius, grain_height = y
             if self.only_radial_burn:
-                burn_area = (
-                    2
-                    * np.pi
-                    * (
-                        grain_inner_radius * grain_height
-                    )
-                )
+                burn_area = 2 * np.pi * (grain_inner_radius * grain_height)
             else:
                 burn_area = (
                     2
@@ -516,12 +514,7 @@ class SolidMotor(Motor):
             grain_inner_radius, grain_height = y
             if self.only_radial_burn:
                 factor = volume_diff / (
-                    2
-                    * np.pi
-                    * (
-                        grain_inner_radius * grain_height
-                    )
-                    ** 2
+                    2 * np.pi * (grain_inner_radius * grain_height) ** 2
                 )
 
                 inner_radius_derivative_wrt_inner_radius = factor * (
@@ -537,8 +530,6 @@ class SolidMotor(Motor):
                         inner_radius_derivative_wrt_height,
                     ],
                     [height_derivative_wrt_inner_radius, height_derivative_wrt_height],
-
-
                 ]
 
             else:
@@ -568,10 +559,8 @@ class SolidMotor(Motor):
                         inner_radius_derivative_wrt_height,
                     ],
                     [height_derivative_wrt_inner_radius, height_derivative_wrt_height],
-
-
                 ]
-                
+
         def terminate_burn(t, y):  # pylint: disable=unused-argument
             end_function = (self.grain_outer_radius - y[0]) * y[1]
             return end_function
@@ -625,9 +614,7 @@ class SolidMotor(Motor):
             burn_area = (
                 2
                 * np.pi
-                * (
-                    self.grain_inner_radius * self.grain_height
-                )
+                * (self.grain_inner_radius * self.grain_height)
                 * self.grain_number
             )
         else:
@@ -812,6 +799,7 @@ class SolidMotor(Motor):
                 "grain_initial_height": self.grain_initial_height,
                 "grain_separation": self.grain_separation,
                 "grains_center_of_mass_position": self.grains_center_of_mass_position,
+                "only_radial_burn": self.only_radial_burn,
             }
         )
 
@@ -855,4 +843,5 @@ class SolidMotor(Motor):
             throat_radius=data["throat_radius"],
             interpolation_method=data["interpolate"],
             coordinate_system_orientation=data["coordinate_system_orientation"],
+            only_radial_burn=data.get("only_radial_burn", False),
         )
--- a/rocketpy/motors/solid_motor.py
+++ b/rocketpy/motors/solid_motor.py
@@ -302,7 +302,7 @@ class SolidMotor(Motor):
             "nozzle_to_combustion_chamber".
         only_radial_burn : boolean, optional
             If True, inhibits the grain from burning axially, only computing
-            radial burn. Otherwise, if False, allows the grain to also burn 
+            radial burn. Otherwise, if False, allows the grain to also burn
             axially. May be useful for axially inhibited grains or hybrid motors.
             Default is False.
 
@@ -343,10 +343,11 @@ class SolidMotor(Motor):
         )
         self.grain_initial_mass = self.grain_density * self.grain_initial_volume
 
-        self.evaluate_geometry()
-
         # Burn surface definition
         self.only_radial_burn = only_radial_burn
+
+        self.evaluate_geometry()
+
         # Initialize plots and prints object
         self.prints = _SolidMotorPrints(self)
         self.plots = _SolidMotorPlots(self)
@@ -489,6 +490,10 @@ class SolidMotor(Motor):
             grain_inner_radius, grain_height = y
             if self.only_radial_burn:
                 burn_area = 2 * np.pi * (grain_inner_radius * grain_height)
+
+                grain_inner_radius_derivative = -volume_diff / burn_area
+                grain_height_derivative = 0
+
             else:
                 burn_area = (
                     2
@@ -500,8 +505,8 @@ class SolidMotor(Motor):
                     )
                 )
 
-            grain_inner_radius_derivative = -volume_diff / burn_area
-            grain_height_derivative = -2 * grain_inner_radius_derivative
+                grain_inner_radius_derivative = -volume_diff / burn_area
+                grain_height_derivative = -2 * grain_inner_radius_derivative
 
             return [grain_inner_radius_derivative, grain_height_derivative]
 
@@ -520,7 +525,7 @@ class SolidMotor(Motor):
                 inner_radius_derivative_wrt_inner_radius = factor * (
                     grain_height - 2 * grain_inner_radius
                 )
-                inner_radius_derivative_wrt_height = factor * grain_inner_radius
+                inner_radius_derivative_wrt_height = 0
                 height_derivative_wrt_inner_radius = 0
                 height_derivative_wrt_height = 0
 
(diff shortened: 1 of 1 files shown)


MateusStano (changes requested):
Are we missing something in the HybridMotor class here? If we want to define a a HybridMotor with only radial burning garin how would we do that currently?

Gui-FernandesBR:
> Are we missing something in the HybridMotor class here? If we want to define a a HybridMotor with only radial burning garin how would we do that currently?

I recommend modifyng  the hybrid class in another PR, it's easier to reivew it.

rocketpy/motors/solid_motor.py
```diff
@@ -314,6 +315,11 @@ class Function. Thrust units are Newtons.
             "nozzle_to_combustion_chamber".
         reference_pressure : int, float, optional
             Atmospheric pressure in Pa at which the thrust data was recorded.
+        only_radial_burn : boolean, optional
+            If True, inhibits the grain from burning axially, only computing
+            radial burn. Otherwise, if False, allows the grain to also burn
```
Gui-FernandesBR, line 320: Redundant

```suggestion
            radial burn. Otherwise, allows the grain to also burn
```

or 

```suggestion
            radial burn. If False, allows the grain to also burn
```

Gui-FernandesBR (commented):
Great PR.

I recommend keeping this PR to solidmotor (possibly renaming it).
Later you could extend the feature to also cover hybrid motors.

Also, I believe two other steps are important:
1. Update the Motor page on our documentation to include this new feature
2. Include new tests!!

Gui-FernandesBR:
But if you really want to include hybrid motors here, not a problem

flight.json
Gui-FernandesBR, line 1: @caioessouza I believe the `flight.json` file was added by mistake, could you take a look please?

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@ Attention: The newest changes should be on top -->
 
 ### Added
 
+- ENH: Enable only radial burning [#815](https://github.com/RocketPy-Team/RocketPy/pull/815)
```
Gui-FernandesBR: please move your new row to the `## [Unreleased] - yyyy-mm-dd` section

tests/unit/test_hybridmotor.py
```diff
@@ -170,9 +170,45 @@ def test_hybrid_motor_inertia(hybrid_motor, spherical_oxidizer_tank):
...
+    )
+
+    assert np.isclose(motor.solid.burn_area(0), burn_area_radial, atol=1e-12)
+
+    # Deactivates the radial burning and recalculate the geometry
+    motor.solid.only_radial_burn = False
+    motor.solid.evaluate_geometry()
+    assert motor.solid.only_radial_burn is False
+
+    # In this case the burning area also considers tha bases of the grain
```
Gui-FernandesBR:
```suggestion
    # In this case the burning area also considers the bases of the grain
```

rocketpy/motors/solid_motor.py
```diff
@@ -521,32 +538,55 @@ def geometry_jacobian(t, y):
...
-            return [
-                [
-                    inner_radius_derivative_wrt_inner_radius,
-                    inner_radius_derivative_wrt_height,
-                ],
-                [height_derivative_wrt_inner_radius, height_derivative_wrt_height],
-            ]
+                inner_radius_derivative_wrt_inner_radius = factor * (
+                    grain_height - 2 * grain_inner_radius
+                )
```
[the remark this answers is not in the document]
caioessouza: This comment already exists a bit below.

flight.json
caioessouza: Yeah, It was a mistake, sorry

rocketpy/motors/solid_motor.py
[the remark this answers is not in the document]
caioessouza: Done.

CHANGELOG.md
```diff
@@ -31,6 +31,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 Attention: The newest changes should be on top -->
 
 ### Added
+- ENH: Enable only radial burning [#815](https://github.com/RocketPy-Team/RocketPy/pull/815)
+- ENH: Controller (AirBrakes) and Sensors Encoding [#849] (https://github.com/RocketPy-Team/RocketPy/pull/849)
+- EHN: Addition of ensemble variable to ECMWF dictionaries [#842] (https://github.com/RocketPy-Team/RocketPy/pull/842)
```
phmbressan: There are duplicates here. Try to get the develop CHANGELOG and only add your PR to the top.

rocketpy/motors/hybrid_motor.py
```diff
@@ -216,6 +216,7 @@ def __init__(  # pylint: disable=too-many-arguments
         interpolation_method="linear",
         coordinate_system_orientation="nozzle_to_combustion_chamber",
         reference_pressure=None,
+        only_radial_burn=True,
```
phmbressan, line 223: There can be made some discussion here on the severity of it, but this is likely a breaking change: the default of this attribute should be `False`, otherwise all the current code with hybrid motors would have its behavior changed. Do you agree @Gui-FernandesBR ?

phmbressan, line 223: This parameter needs to be documented in the ``class`` and ``__init__`` docstrings.

tests/fixtures/motor/hybrid_fixtures.py
```diff
@@ -4,7 +4,7 @@
 
 
 @pytest.fixture
-def hybrid_motor(spherical_oxidizer_tank):
+def hybrid_motor(oxidizer_tank):
```
phmbressan, line 7: I don't oppose to it if needed, but any particular reason for changing the fixture name?

tests/unit/motors/test_solidmotor.py
```diff
@@ -279,3 +279,43 @@ def test_reshape_thrust_curve_asserts_resultant_thrust_curve_correct(
...
+    assert (
+        values[0] == motor.grain_initial_inner_radius
+    )  # expected initial inner radius
+    assert (
+        values[-1] <= motor.grain_outer_radius
+    )  # final inner radius should be less or equal than outer radius
+
+    # Evaluates without error
+    val = motor.grain_inner_radius(0.5)  # evaluate at intermediate time
+    assert isinstance(val, float)
```
phmbressan: Similarly to the other tests, docstrings are needed to describe it and its fixture parameters.

tests/unit/motors/test_hybridmotor.py
```diff
@@ -170,9 +170,46 @@ def test_hybrid_motor_inertia(hybrid_motor, spherical_oxidizer_tank):
...
+def test_hybrid_motor_only_radial_burn_behavior(hybrid_motor):
+    """
+    Test if only_radial_burn flag in HybridMotor propagates to its SolidMotor
+    and affects burn_area calculation.
+    """
+    motor = hybrid_motor
+
+    # Activates the radial burning
+    motor.solid.only_radial_burn = True
+    assert motor.solid.only_radial_burn is True
```
phmbressan: This is quite dangerous. I did not test nor exhaustively review the code, but here are some of my comments that I believe are relevant (feel free to correct me on that) if you want to support _setting_ (post instantiation) this attribute:

- The `SolidMotor.evaluate_geometry` method also sets some of the instance attributes, such as `self.grain_burn_out` and `self.grain_inner_radius`;
- Simply toggling the attribute `only_radial_burn` would not change the values of those attributes, therefore their old values would be kept.

On the top of my head, I see two possible solutions here:

1. Supporting this toggling after instantiation in which other values of the class should be adapted is the task of a setter:

```python
class SolidMotor(Motor):
    def __init__(self, ..., only_radial_burn=False):
        ...
        self._only_radial_burn = only_radial_burn

    @property
    def only_radial_burn(self):
        return self._only_radial_burn

    @only_radial_burn.setter
    def only_radial_burn(self, value):
        self._only_radial_burn = value
        self.evaluate_geometry()

class HybridMotor(Motor):
    def __init__(self, ..., only_radial_burn=False):
        ...
        self._only_radial_burn = only_radial_burn

    @property
    def only_radial_burn(self):
        return self._only_radial_burn

    @only_radial_burn.setter
    def only_radial_burn(self, value):
        self.solid.only_radial_burn = value
        reset_funcified_method(self)
```

2. Alternative simpler solution: another possibility is to disallow the toggling of this attribute after instantiation. This is possible by using a property with no setter. In this way, the radial burn can only be selected on instantiation.

```python
class SolidMotor(Motor):
    def __init__(self, ..., only_radial_burn=False):
        ...
        self._only_radial_burn = only_radial_burn

    @property
    def only_radial_burn(self):
        return self._only_radial_burn
```

@caioessouza could you confirm this is the case, i.e., toggling this attribute would cause the defined attributes from `evaluate_geometry` incorrectly keep their old values? If I understood things correctly, the tests for this new parameter did not catch that, which means they have room for improvement.

I, personally, have a slight preference for option 2, since it avoids complicating more this already delicate interaction between the motor types. Do you have any suggestions @MateusStano or @Gui-FernandesBR ?

phmbressan (changes requested):

tests/unit/motors/test_hybridmotor.py
Gui-FernandesBR: I think the best option is to make `only_radial_burn` not mutable. 

This would make life much easier.

If you want to modify the value of only_radial_burn after instanciating the object, you should instantiate another object

CHANGELOG.md
caioessouza: Check. I just replaced CHANGELOG to the develop one and added my changes

rocketpy/motors/hybrid_motor.py
caioessouza: I don't think this change would break anything, but I think it can change the propellant mass evolution with time. Anyway, this behaviour is more similar to reality, so in my conception even if it changes, it's not necessarily a bad thing. Do you agree @Gui-FernandesBR @MateusStano ? If you think it's a problem, I can change.

caioessouza: Check, please review again to make sure everything is ok

tests/fixtures/motor/hybrid_fixtures.py
caioessouza: The spherical_oxidizer_tank fixture is broken. It's giving negative liquid heights. It need to be fixed in another issue.

tests/unit/motors/test_solidmotor.py
caioessouza: Check, please review again

tests/unit/motors/test_hybridmotor.py
caioessouza: I've just implemented option 2. I agree that makes more sense, because it's a characteristic of each motor, so it's not mutable in  reality anyway.

caioessouza: When I added this 

class SolidMotor(Motor):
    def __init__(self, ..., only_radial_burn=False):
        ...
        self._only_radial_burn = only_radial_burn

    @property
    def only_radial_burn(self):
        return self._only_radial_burn
        
It broke a lot of tests, maybe I've done something wrong.
Can you do this part? @phmbressan

rocketpy/motors/hybrid_motor.py
Gui-FernandesBR: I agree it changes results, and I understand it changes for the better.

@phmbressan don't you think we could proceed with this?

tests/fixtures/motor/hybrid_fixtures.py
Gui-FernandesBR: @caioessouza what issue are you referring to?

caioessouza: There's no issue for it yet, but the spherical_oxidizer_tank fixture is returning negative liquid heights for this burning duration.

rocketpy/motors/hybrid_motor.py
Gui-FernandesBR: I have rebased and fixed the PR. This is the only remaining comment. @phmbressan what do you think?

phmbressan: If we are all in agreement, I understand interpreting this change as an improvement of our modeling rather than a breaking change.

PR #818, opened 2025-05-16
ENH: Individual Fins

MateusStano:
## Checklist

## Description

Individual Fins model have been added:

- Usage can be seen in the documentation files. All equations and modelling are also described in the docs file.

- A lot of structural changes were made to **all** fin classes. Please review this new structure with care.

- If you run a simulation with the standard Fins model and another with individual fins, the results will be a little different. This happens because the individual fin model calculates forces for each fin separately, using the airflow at each fin’s exact position. The standard model (Barrowman) treats all the fins as one and uses a single point along the rocket’s center. In theory, this means that individual fins are more accurate then the standard model

- There are a couple of TODOs relating to cant angle control. The center of pressure position relative to CDM is precalculated in the Rocket class and then accessed in the Flight. This causes a problem if the fin’s cant angle changes during the simulation because that changes the center of pressure position, but right now there’s no easy way to update these values in the Rocket while the simulation is running.

## Breaking change

- [ ] Yes
- [x] No


rocketpy/rocket/aero_surface/fins/trapezoidal_fins.py
```diff
@@ -169,66 +165,11 @@ def __init__(
...
-
-    @sweep_length.setter
-    def sweep_length(self, value):
-        self._sweep_length = value
-        self.evaluate_geometrical_parameters()
-        self.evaluate_center_of_pressure()
-        self.evaluate_lift_coefficient()
-        self.evaluate_roll_parameters()
-
     def evaluate_center_of_pressure(self):
```
Gui-FernandesBR, line 173: isnt the `evaluate_center_of_pressure` common for both TrapFin and TrapFins?

docs/technical/aerodynamics/individual_fins.rst
```diff
@@ -0,0 +1,498 @@
...
+
+:Author: Mateus Stano Junqueira
+:Date: March 2025
+
+Introduction
+============
+
+There are currently three ways to model the aerodynamic effects of fins in RocketPy:
+
+1. By use of the :class:`rocketpy.GenericSurface` or :class:`rocketpy.LinearGenericSurface` classes,
```
Gui-FernandesBR, line 15: This way you skip the "skip line" entry

```suggestion
1. By use of the :class:`rocketpy.GenericSurface` or :class:`rocketpy.LinearGenericSurface` classes, \
```

rocketpy/rocket/aero_surface/fins/_free_form_mixin.py
```diff
@@ -0,0 +1,228 @@
...
+                down = True
+            i += 1
+
+        root_chord = abs(shape_points[0][0] - shape_points[-1][0])
+        ys = [point[1] for point in shape_points]
+        span = max(ys) - min(ys)
+        return (
+            root_chord,
+            span,
+        )
```
Gui-FernandesBR, line 32: do you really need the return?

rocketpy/rocket/rocket.py
```diff
@@ -996,11 +1005,17 @@ def __add_single_surface(self, surface, position):
...
         rail buttons case, and position type.
         """
-        position = (
-            Vector([0, 0, position])
-            if not isinstance(position, (Vector, tuple, list))
-            else Vector(position)
-        )
+        if isinstance(surface, (TrapezoidalFin, EllipticalFin, FreeFormFin)):
+            # TODO: this depends on cant angle, so it should somehow be
+            # recalculated whenever the cant angle of the fin changes
```
Gui-FernandesBR, line 1010: maybe what you want to do is to create a setter method to the cant_angle parameter?

rocketpy/rocket/rocket.py
```diff
@@ -1018,14 +1033,20 @@ def add_surfaces(self, surfaces, positions):
         surfaces : list, AeroSurface, NoseCone, TrapezoidalFins, EllipticalFins, Tail, RailButtons
```
Gui-FernandesBR, line 1033:
```suggestion
        surfaces : list[AeroSurface], AeroSurface
```

rocketpy/rocket/aero_surface/fins/_trapezoidal_mixin.py
```diff
@@ -0,0 +1,179 @@
...
+        # Check if sweep angle or sweep length is given
+        if sweep_length is not None and sweep_angle is not None:
+            raise ValueError("Cannot use sweep_length and sweep_angle together")
+        elif sweep_angle is not None:
+            sweep_length = np.tan(sweep_angle * np.pi / 180) * span
+        elif sweep_length is None:
+            sweep_length = root_chord - tip_chord
+        else:  # Sweep length is given
+            pass
+
```
Gui-FernandesBR, line 56: (minor) these checks could be moved to an another method

rocketpy/rocket/aero_surface/fins/fins.py
```diff
@@ -423,7 +308,7 @@ def compute_forces_and_moments(
             * omega[2]
             / 2
         )
-        M3 = M3_forcing - M3_damping
+        M3 = M3_forcing + M3_damping
         return R1, R2, R3, M1, M2, M3
 
     def to_dict(self, include_outputs=False):
```
Gui-FernandesBR, line 314: As both `Fins` and `TrapezoidalFinsMixin` have the `to_dict` method, which method will the `TrapezoidalFins` class use?

rocketpy/rocket/aero_surface/fins/_free_form_mixin.py
```diff
@@ -0,0 +1,228 @@
...
+                or np.isnan(chord_lead[i])
+                or np.isnan(chord_trail[i])
+            ):
+                chord_lead[i] = 0
+                chord_trail[i] = 0
+            if chord_length[i] < 0 or np.isnan(chord_length[i]):
+                chord_length[i] = 0
+            if chord_length[i] > chord_trail[i] - chord_lead[i]:
+                chord_length[i] = chord_trail[i] - chord_lead[i]
+
```
Gui-FernandesBR, line 130: Should be faster (copilot suggestion), and by far more readable:

```python
        # Replace infinities and handle invalid values in chord_lead and chord_trail
        mask_invalid = (
            np.isinf(chord_lead) | np.isinf(chord_trail) | np.isnan(chord_lead) | np.isnan(chord_trail)
        )
        chord_lead[mask_invalid] = 0
        chord_trail[mask_invalid] = 0

        chord_length = np.where((chord_length < 0) | np.isnan(chord_length), 0, chord_length)
        max_chord = chord_trail - chord_lead
        chord_length = np.minimum(chord_length, max_chord)
```

rocketpy/rocket/aero_surface/fins/elliptical_fin.py
```diff
@@ -0,0 +1,212 @@
...
+        """Calculates and returns the center of pressure of the fin set in local
+        coordinates. The center of pressure position is saved and stored as a
+        tuple.
+
+        Returns
+        -------
+        None
+        """
+        # Center of pressure position in local coordinates
+        cpz = 0.288 * self.root_chord
```
Gui-FernandesBR, line 178: 0.288?

Where does it come from?
Any reference?

docs/technical/aerodynamics/individual_fins.rst
```diff
@@ -0,0 +1,498 @@
...
+^^^
+
+.. jupyter-execute::
+  :hide-code:
+  :hide-output:
+
+  example_rocket.aerodynamic_surfaces.pop()
+  example_rocket.aerodynamic_surfaces.pop()
+
+Given oposing canards at 0 degrees and 180 degrees, having opposite cant angles,
```
Gui-FernandesBR:
```suggestion
Given opposing canards at 0 degrees and 180 degrees, having opposite cant angles, 
```

docs/technical/aerodynamics/individual_fins.rst
```diff
@@ -0,0 +1,498 @@
...
+(:math:`M_{z}`) is still missing the damping term, which is given by:
+
+.. math::
+  M_{damp} = \frac{1}{2} \cdot \rho \cdot \|v_{0}\| \cdot A_{r} \cdot L_{r}^2 \cdot C_{ld\omega}(Ma) \cdot \frac{1}{2} \cdot \omega_z 
+
+.. math::
+  M_{z \, \text{final}} = M_{z} + M_{damp}
+
+Where :math:`C_{ld}` is the roll moment damping coefficient, :math:`L_{r}` 
+is the reference length, which is equal to the rocket diamete, and
```
Gui-FernandesBR:
```suggestion
is the reference length, which is equal to the rocket diameter, and
```

Gui-FernandesBR:
> Is this branch will be in production soon? 

Not at all!

> Where can I read updated README?

I didn't understand this question


--------

If you want to, you can fork this branch and continue the implementation on your fork, thus proposing a PR to our repo later

rocketpy/rocket/aero_surface/fins/trapezoidal_fins.py
MateusStano: No, there are very small differences

rocketpy/rocket/rocket.py
MateusStano: The problem is that what need updating is a the value of the position of the fin saved in rocket.aerodynamic_surfaces. So you can´t really use a setter in the Fin class to change the Rocket class. 

This calculation has very minimal effect, so I left it as a TODO because it will require a bigger code structure related change to fix I believe

rocketpy/rocket/aero_surface/fins/fins.py
MateusStano: `TrapezoidalFin.to_dict` calls `_TrapezoidalMixin.to_dict`. Inside of `_TrapezoidalMixin.to_dict` there is a call to the `Fin.to_dict`

rocketpy/rocket/aero_surface/fins/elliptical_fin.py
MateusStano: Just from barrowman

rocketpy/rocket/aero_surface/fins/_base_fin.py
```diff
@@ -0,0 +1,193 @@
...
+
+    @airfoil.setter
+    def airfoil(self, value):
+        self._airfoil = value
+        self.evaluate_geometrical_parameters()
+        self.evaluate_center_of_pressure()
+        self.evaluate_lift_coefficient()
+        self.evaluate_roll_parameters()
+
+    def evaluate_single_fin_lift_coefficient(self):
```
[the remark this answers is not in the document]
MateusStano: Yes, it is useful in both classes

rocketpy/rocket/aero_surface/fins/_base_fin.py
```diff
@@ -0,0 +1,193 @@
...
+            # Defines clalpha2D as the derivative of the lift coefficient curve
+            # for the specific airfoil
+            self.airfoil_cl = Function(
+                self.airfoil[0],
+                title="Airfoil lift coefficient",
+                interpolation="linear",
+            )
+
+            # Differentiating at alpha = 0 to get cl_alpha
+            clalpha2D_incompressible = self.airfoil_cl.differentiate_complex_step(
```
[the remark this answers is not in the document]
MateusStano: This has been here for a long time. But ill change it in this PR sure. There is no point in it being complex step

rocketpy/prints/aero_surface_prints.py
```diff
@@ -179,14 +274,26 @@ class _TrapezoidalFinsPrints(_FinsPrints):
     """Class that contains all trapezoidal fins prints."""
 
 
+class _TrapezoidalFinPrints(_FinPrints):
```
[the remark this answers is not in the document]
MateusStano: We don´t **need** it. But it keeps our default structure and it is future proof.

rocketpy/rocket/components.py
```diff
@@ -196,9 +197,12 @@ def sort_by_position(self, reverse=False):
 
         Returns
         -------
-        None
+        Components
+            A new Components object sorted by component position.
         """
-        self._components.sort(key=lambda x: x.position.z, reverse=reverse)
+        components = deepcopy(self)
```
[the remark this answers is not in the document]
MateusStano: If this didn´t deepcopy it would change the order of the aero surfaces in the rocket attribute `rocket.aerodynamic_surfaces`. So if someone was accessing the first item with `rocket.aerodynamic_surfaces[0]` it could change the object it is referencing

rocketpy/prints/aero_surface_prints.py
Gui-FernandesBR, line 1: TODO for later: the `rocketpy/prints/aero_surface_prints.py` file could be separated into different, smaller files.

I'd appreciate if you could add this TODO at the top of the file

rocketpy/rocket/aero_surface/fins/_geometry.py
Gui-FernandesBR, line 1: `rocketpy/rocket/aero_surface/fins/_geometry.py` could be divided into smaller parts

I would expect different unit tests, covering each geometry method

rocketpy/rocket/aero_surface/fins/elliptical_fin.py
Gui-FernandesBR: please reference in the code

rocketpy/rocket/rocket.py
Gui-FernandesBR: I saw the TODO, but it lacks your full explanation. By reading the comment, a young developer could easily become keen to solve a very hard non-profitable problem

rocketpy/rocket/rocket.py
```diff
@@ -1085,17 +1100,23 @@ def add_surfaces(self, surfaces, positions):
 
         Parameters
         ----------
-        surfaces : list, AeroSurface, NoseCone, TrapezoidalFins, EllipticalFins, Tail, RailButtons
+        surfaces : list[AeroSurface], AeroSurface
```
Gui-FernandesBR, line 1106: could we fix in list[AeroSurface] only?

```python
        surfaces : list[AeroSurface], AeroSurface
```

.pylintrc
```diff
@@ -225,6 +225,11 @@ good-names=FlightPhases,
            center_of_mass_without_motor_to_CDM,
            motor_center_of_dry_mass_to_CDM,
            generic_motor_cesaroni_M1520,
+           R_phi,
```
Gui-FernandesBR, line 228: What is R?

rocketpy/rocket/aero_surface/fins/trapezoidal_fin.py
```diff
@@ -0,0 +1,197 @@
...
+        angular_position,
+        root_chord,
+        tip_chord,
+        span,
+        rocket_radius,
+        cant_angle=0,
+        sweep_length=None,
+        sweep_angle=None,
+        airfoil=None,
+        name="Fins",
```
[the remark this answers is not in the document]
Gui-FernandesBR: @MateusStano ?

rocketpy/plots/aero_surface_plots.py
```diff
@@ -325,10 +378,129 @@ def draw(self, *, filename=None):
...
+        )
+        ax.set_xlim(0, xlim * 1.1)
+        ax.set_ylim(0, self.aero_surface.span * 1.1)
+        ax.set_xlabel("Root chord (m)")
+        ax.set_ylabel("Span (m)")
+        ax.set_title("Trapezoidal Fin Cross Section")
+        ax.legend(bbox_to_anchor=(1.05, 1.0), loc="upper left")
+
+        plt.tight_layout()
+        plt.show()
```
phmbressan: In this method (and a few other ones) is the `filename` being used? Should we use `show_or_save` plot for them?

rocketpy/rocket/aero_surface/fins/_base_fin.py
```diff
@@ -0,0 +1,295 @@
...
+        self.name = name
+        self._rocket_radius = rocket_radius
+        self._root_chord = root_chord
+        self._span = span
+        self._airfoil = airfoil
+        self._cant_angle = cant_angle
+        self._cant_angle_rad = math.radians(cant_angle)
+        self.geometry = None
+
+        self.d = 2 * rocket_radius
```
phmbressan: Generally I don't mind much the abbreviated names. But even though it is clear to most people this is a diameter, this is documented nowhere (I think) and is a bit too short of a name. Perhaps `rocket_diameter` and make a local alias before big formulas?

rocketpy/rocket/aero_surface/fins/_base_fin.py
```diff
@@ -0,0 +1,295 @@
...
+        self.ref_area = np.pi * rocket_radius**2
+
+        super().__init__(name, self.ref_area, self.d)
+
+    def _run_geometry_update_chain(self):
```
phmbressan: Optional, this name is bit too verbose (e.g. I don't think the word `run` really accomplishes anything here and `chain` gives a sense of order, but it is vague). I would strive for simplicity with `_update_geometry`, but it is just a suggestion.

rocketpy/rocket/aero_surface/fins/_geometry.py
```diff
@@ -0,0 +1,519 @@
...
+            )
+        return data
+
+
+class _EllipticalGeometry(_FinGeometry):
+    """Geometry strategy for elliptical fins."""
+
+    def evaluate_geometrical_parameters(self):  # pylint: disable=too-many-statements
+        """Calculates and saves elliptical fin geometric parameters."""
+        owner = self.owner
```
phmbressan, line 209: This is a bit unusual, since the direction of dependence is reversed here, i.e., a rocketpy more internal object depends on an outer one. If I understood correctly:
- `owner` will be the `Fin` (or `Fins`) themselves. This is odd, since the attribute `geometry` now depends on the class it is part of. It is almost circular;
- this attribute `geometry` mutates the outer object, which is also unusual, unless there is an reason for it.

Could you give some hints on why it was not possible to instantiate a `geometry` with purely a few values from the `Fin` and `Fins` classes, but instead, it was necessary to input the full object?

Ideally, the `geometry` class should not mutate the outer object, but instead have methods and properties (e.g. `geometry.Af`, `geometry.roll_interference_factor` etc) that compute and return the values the outer object (i.e. the fin classes) need.

rocketpy/rocket/aero_surface/fins/elliptical_fin.py
```diff
@@ -0,0 +1,197 @@
...
+        self.plots = _EllipticalFinPlots(self)
+
+    def evaluate_center_of_pressure(self):
+        """Calculates and returns the center of pressure of the fin in local
+        coordinates. The center of pressure position is saved and stored as a
+        tuple.
+
+        Returns
+        -------
+        None
```
phmbressan: Here it says it returns, but the methods returns `None`.

rocketpy/rocket/aero_surface/fins/elliptical_fin.py
```diff
@@ -0,0 +1,197 @@
...
+            name,
+        )
+
+        self.geometry = _EllipticalGeometry(self)
+        self._run_geometry_update_chain()
+
+        self.prints = _EllipticalFinPrints(self)
+        self.plots = _EllipticalFinPlots(self)
+
+    def evaluate_center_of_pressure(self):
```
phmbressan, line 167: Are these `evaluate_` methods kept for backwards compatibility? I don't see why it would be useful to an user, and, hence, they should not be public.

Why not have a new `property` (or a `cached_property` or a `funcify`) that is `.center_of_pressure`? You can keep these `evaluate_` methods for backwards compatibility. Maybe I overlooked some condition on why there should be these `evaluates`, so please, correct me if I am wrong.

Note: a similar reasoning applies to the `evaluates_` of the geometry classes:
- one could probably convert most evaluates into individual properties (e.g. `evaluate_center_of_pressure` -> `.center_of_pressure`);
- when a parameter is set in the `Fin`(s) class, one can either:
    1. recreate the `geometry` instance with the new value;
    2. have a `setter` in the geometry that sets the new value and calls the correct re-evaluate for better performance (in this case, the evaluates should be private `geometry._evaluate...`, since they would only be used internaly).

rocketpy/rocket/aero_surface/fins/_base_fin.py
Gui-FernandesBR: good catch, it's even in our documentatiom to not use single letter attributes

rocketpy/rocket/aero_surface/fins/_base_fin.py
```diff
@@ -0,0 +1,295 @@
...
+        """Recompute all dependent aerodynamic properties after a geometry change."""
+        self.evaluate_geometrical_parameters()
+        self.evaluate_center_of_pressure()
+        self.evaluate_lift_coefficient()
+        self.evaluate_roll_parameters()
```
phmbressan: Also, ideally, this method should not be needed, since the update of the parameters would all be done _inside the geometry_, by a setter. The sequence would similar to:

```
User (action fin.root_chord=0.5)
    TrapezoidalFin.root_chord setter (sets fin.geometry.root_chord=0.5)
        _TrapezoidalFinGeometry.root_chord setter (evaluates self._root_chord=0.5 and self.update_geometry())
```

Note: 
- the update geometry would mostly delete any caches of the geometrical attributes of the _TrapezoidalFinGeometry (e.g. center_of_pressure));
- The `getters` of the `TrapezoidalFin` would generally delegate to the `geometry`, e.g.:
        - `def root_chord -> self.geometry.root_chord`

We can discuss more (or go with the current implementation) if this does not seem to tackle our class schema/needs.

rocketpy/rocket/aero_surface/fins/fin.py
```diff
@@ -0,0 +1,457 @@
...
+        )
+
+        # Cant rotation about body Y
+        R_delta = Matrix(
+            [
+                [np.cos(delta), 0, -np.sin(delta)],
+                [0, 1, 0],
+                [np.sin(delta), 0, np.cos(delta)],
+            ]
+        )
```
phmbressan, line 300: If `phi` and `delta` are guaranteed to be a scalar (not an array of angles), I would use the `math` module for those calculations. Also, saving the variables that are used multiple times gives you a slight speed increase:

```
sin_phi = math.sin(self.cant_angle_rad)
...
```

Of course the gains are marginal to say the least, but it is worth commenting, I believe.

rocketpy/rocket/rocket.py
```diff
@@ -1108,10 +1129,18 @@ def add_surfaces(self, surfaces, positions):
         -------
         None
         """
-        try:
+        if isinstance(surfaces, list):
+            if isinstance(positions, list):
+                if len(surfaces) != len(positions):
+                    raise ValueError(
```
phmbressan: Check if they are `Iterable` instead. This would cause an easily solved problem if one gives a tuple or a set, for instance.

tests/integration/test_rocket.py
```diff
@@ -19,7 +19,7 @@ def test_airfoil(
     calisto.add_surfaces(calisto_tail, -1.313)
 
     test_rocket.add_trapezoidal_fins(
-        2,
+        3,
```
phmbressan, line 22: Was there any issue regarding testing with `2` fins?

tests/unit/rocket/aero_surface/test_individual_fins.py
```diff
@@ -0,0 +1,356 @@
...
+        sweep_angle=15.0,
+        name="roundtrip_trapezoidal_fin",
+    )
+    data = original.to_dict()
+
+    # Act
+    reconstructed = TrapezoidalFin.from_dict(data)
+
+    # Assert
+    assert reconstructed.sweep_length == pytest.approx(original.sweep_length)
```
phmbressan: The same way you used `np.testing` module below (for the `all_close`), could you use the same testing module for validating the scalar values here and in a few instances above? I find it has a much more readable output.

.pylintrc
MateusStano: Rotation matrix. This is used internally only and it is the same symbol as in the technical documentation .rst page. Using this makes it easier to compare the documentation with the code

rocketpy/prints/aero_surface_prints.py
MateusStano: https://github.com/RocketPy-Team/RocketPy/pull/818/commits/ee6b4646592913ca87aed592b7066761de69741f

rocketpy/rocket/aero_surface/fins/_geometry.py
MateusStano: https://github.com/RocketPy-Team/RocketPy/pull/818/commits/3929a3bca586ca0c62d6c799c4a30fb77dd170dd

rocketpy/rocket/aero_surface/fins/elliptical_fin.py
MateusStano: https://github.com/RocketPy-Team/RocketPy/pull/818/commits/6639aca240ec3fd9296cb77fd5a6a0921e2f04dc

rocketpy/rocket/rocket.py
MateusStano: https://github.com/RocketPy-Team/RocketPy/pull/818/commits/878a84f25187be4d44edc342ef63efa33cb3689f

rocketpy/plots/aero_surface_plots.py
MateusStano: https://github.com/RocketPy-Team/RocketPy/pull/818/commits/a3ce54e3cd9f3ac0e08b5cf52130b4b9644becd1

rocketpy/rocket/aero_surface/fins/_base_fin.py
MateusStano: https://github.com/RocketPy-Team/RocketPy/pull/818/commits/6bf61dba4147e963c43afd10dcaaee99bb086814

rocketpy/rocket/aero_surface/fins/_base_fin.py
MateusStano: https://github.com/RocketPy-Team/RocketPy/pull/818/commits/31d5311b2d85c79408c24ac5837d624a3d1e968a

rocketpy/rocket/aero_surface/fins/_geometry.py
MateusStano: https://github.com/RocketPy-Team/RocketPy/pull/818/commits/9c0dd139edab3809e094a268bf6e2b0558fb810f

MateusStano: This kind of cyclic dependence happens on some of the Plots and Prints class I believe. I am not sure that parameters are set there, but it is still a more internal object that depends on an outer one. Do you think it that is a problem as well?

phmbressan: No, because, as you said, typically the `Plots` and `Prints` will be read only, i.e., they should not mutate the objects they own. Furthermore, they are a bit more of special case, which was discussed for a while before the implementation.

rocketpy/rocket/aero_surface/fins/elliptical_fin.py
MateusStano: These `evaluate` methods are from the time when we didn´t use cached_property or funcify or any setters. They are also present in all other aero surfaces.

You make really good points, and I think the current AeroSurface structure needs a rework to simplify things a bit

rocketpy/rocket/aero_surface/fins/_base_fin.py
MateusStano: I'll not change it here since these evaluate methods are used in all AeroSurfaces and thats out of scope for this PR

rocketpy/rocket/aero_surface/fins/fin.py
MateusStano: https://github.com/RocketPy-Team/RocketPy/pull/818/commits/c1f51149d345ee4e4eb4a14760b3306471a16ba3

tests/integration/test_rocket.py
MateusStano: The  Barrowman formulas assume a minimum of 3 fins since it requires axisymmetry. So using 2 fins with the standard Fin Set modelling we had was incorrect

tests/unit/rocket/aero_surface/test_individual_fins.py
MateusStano: https://github.com/RocketPy-Team/RocketPy/pull/818/commits/05f2c4ff7d76db5f94b0abf6a0e4774f71c6906f

rocketpy/rocket/rocket.py
MateusStano: https://github.com/RocketPy-Team/RocketPy/pull/818/commits/09657655b7e2724d28aa1a8a0912a7bafd9381ce

PR #827, opened 2025-06-14
ENH: Discretized and No-Pickle Encoding Options

phmbressan:
## Pull request type

## Checklist

- [X] `CHANGELOG.md` has been updated

## Current behavior

The standard `RocketPyEncoder` currently does not support custom options for users/clients that may have the following needs:

- Discrete encoding: as documented, this is useful to have the `callable` defined `Functions` written out in a human readable format. Of course, this has the drawback that the original `callable` cannot be exactly reproduced on decoding;
- No-Pickle encoding: as it is common on I/O libraries/functions (e.g. [NumPy save](https://numpy.org/doc/2.0/reference/generated/numpy.save.html)) having a toggle for pickling is beneficial for allowing the user/client having better control of security and output size (pickled objects might get large). This specially useful if there is no interest in decoding the result.

## New behavior

The added options `discretize` and `allow_pickle` work as described above. The biggest challenge of the PR (which required a bit of hard coding) was setting discretization bounds for some of RocketPy callable `Functions`. 

## Breaking change

- [ ] Yes
- [X] No

### DIFF
--- a/rocketpy/rocket/aero_surface/fins/fins.py
+++ b/rocketpy/rocket/aero_surface/fins/fins.py
@@ -441,7 +441,7 @@ class Fins(AeroSurface):
             cl = self.cl
             if kwargs.get("discretize", False):
                 cl = cl.set_discrete(
-                    (-np.pi / 12, 0), (np.pi / 12, 2), (10, 10), mutate_self=False
+                    (-np.pi / 6, 0), (np.pi / 6, 2), (10, 10), mutate_self=False
                 )
 
             data.update(
--- a/rocketpy/rocket/aero_surface/fins/fins.py
+++ b/rocketpy/rocket/aero_surface/fins/fins.py
@@ -441,7 +441,7 @@ class Fins(AeroSurface):
             cl = self.cl
             if kwargs.get("discretize", False):
                 cl = cl.set_discrete(
-                    (-np.pi / 12, 0), (np.pi / 12, 2), (50, 50), mutate_self=False
+                    (-np.pi / 12, 0), (np.pi / 12, 2), (10, 10), mutate_self=False
                 )
 
             data.update(
--- a/rocketpy/rocket/aero_surface/fins/fins.py
+++ b/rocketpy/rocket/aero_surface/fins/fins.py
@@ -426,7 +426,7 @@ class Fins(AeroSurface):
         M3 = M3_forcing - M3_damping
         return R1, R2, R3, M1, M2, M3
 
-    def to_dict(self, include_outputs=False):
+    def to_dict(self, **kwargs):
         data = {
             "n": self.n,
             "root_chord": self.root_chord,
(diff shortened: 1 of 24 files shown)


rocketpy/rocket/aero_surface/fins/fins.py
```diff
@@ -437,11 +437,17 @@ def to_dict(self, include_outputs=False):
             "name": self.name,
         }
 
-        if include_outputs:
+        if kwargs.get("include_outputs", False):
+            cl = self.cl
+            if kwargs.get("discretize", False):
+                cl = cl.set_discrete(
```
phmbressan, line 443: @MateusStano , could you give some insights if these boundary values make sense (angle of attack from -15 to 15 and mach 0 to 2)?

rocketpy/rocket/rocket.py
```diff
@@ -1916,38 +1925,78 @@ def to_dict(self, include_outputs=False):
...
+            cp_position = self.cp_position
+            stability_margin = self.stability_margin
+            center_of_mass = self.center_of_mass
+            motor_center_of_mass_position = self.motor_center_of_mass_position
+            reduced_mass = self.reduced_mass
+            total_mass = self.total_mass
+            total_mass_flow_rate = self.total_mass_flow_rate
+            center_of_propellant_position = self.center_of_propellant_position
+
+            if discretize:
```
phmbressan, line 1938: I admit this part of the code is not exactly pretty, but it was the way I found that better guarantees the necessary variables will be discrete.

rocketpy/rocket/rocket.py
```diff
@@ -1892,7 +1892,16 @@ def all_info(self):
         self.info()
         self.plots.all()
 
-    def to_dict(self, include_outputs=False):
+    # pylint: disable=too-many-statements
+    def to_dict(self, **kwargs):
```
Gui-FernandesBR, line 1895: I think this way is more correct:

```suggestion
    def to_dict(self, **kwargs):  # pylint: disable=too-many-statements
```

tests/integration/test_encoding.py
Gui-FernandesBR, line 1: should we also test the pickable option?

Gui-FernandesBR (approved):
LGTM

rocketpy/rocket/rocket.py
phmbressan: I really don't have any strong opinions here, I generally put it above the lines so as to not trigger a line wrap (the comments frequently surpasses the 80 char limit). From the pylint docs, it seems [both are supported](https://pylint.readthedocs.io/en/v2.11.1/faq.html?utm_source=chatgpt.com#how-to-disable-a-particular-message). If you prefer the other way, please let me know.

tests/integration/test_encoding.py
phmbressan: Good point, done in [f3e45bb](https://github.com/RocketPy-Team/RocketPy/pull/827/commits/f3e45bbd8d8bf0fb846ae5c450ec5864033d8146).

rocketpy/rocket/aero_surface/fins/fins.py
MateusStano: I'd suggest to do -30 to 30 and 0 to 4

PR #829, opened 2025-06-17
MNT: fix deprecations and warnings

Gui-FernandesBR:


rocketpy/environment/environment.py
```diff
@@ -451,7 +451,7 @@ def __initialize_utm_coordinates(self):
                 self.initial_utm_letter,
                 self.initial_hemisphere,
                 self.initial_ew,
-            ) = self.geodesic_to_utm(
+            ) = geodesic_to_utm_tools(
```
phmbressan: Is the renaming of this import still necessary?

Gui-FernandesBR: No, I think it is no longer necessary

Issue #669, opened 2024-08-19
MNT: Implement a @deprecated decorator


Gui-FernandesBR:
The @deprecated decorator is used to mark functions or methods as deprecated, indicating that they will be removed in future versions. This helps in managing deprecations and informing users of the codebase about outdated functionalities.
This could help us to organize the code and manage deprecations.

Basic Implementation Steps:

- [ ] Create a decorator function named deprecated.
- [ ] The decorator should issue a warning when the decorated function is called, indicating that it is deprecated.
- [ ] Include a message or version information in the warning to provide more context.
- [ ] Search in the code any deprecation messages and use the decorator you created.

OBS.: Please be aware of 2 different status we may have:
- 1: The method is already deprecated.
- 2: The method is set to be deprecated in a future version (this happens when we want to deprecate something, but we are obligated to give users some time ahead).

Gui-FernandesBR:
> Just wanted to inform that this feature has been added to python 3.13. More info here: https://peps.python.org/pep-0702/

Thank you! 

It will take us 4 years from now until we are finally able to drop support for python 3.12, therefore, I see this issue with less priority, but if someone wants to implement a "brute force" version of the deprecator, that would also be welcome.

PR #830, opened 2025-06-17
MNT: deprecated decorator

Gui-FernandesBR:
Solves #669


phmbressan (commented):
Very interesting implementation.

I know that Python 3.13 added a [similar decorator](https://docs.python.org/3.13/library/warnings.html#warnings.deprecated). I don't think it has the same custom functionalities (such as providing an alternative). Nonetheless, it is worth the read.

Since this Python is far from being RocketPy minimum required version, even if we were to use it, a custom version would be needed to support other Python versions (similarly to what happed with `cached_property` a few years ago). Therefore, I agree with this PR implementation. 

Optionally, we could discuss the possibility of changing the parameters so that it is compatible with the Python standard one when/if we were to change to it in the future. Another idea for discussion is having the Python default one if the user has a `>= 3.13` Python version.

PR #833, opened 2025-06-22
Enh/zero inertia tensor check

aZira371:
## Pull request type

## Checklist

## Current behavior

As of now there are no checks on the singularity of input inertia tensor. This can create some issues where the simulation may still run even if the input inertia tensor was singular and give ill defined results. This was pointed by an user on discord but could not be recreated

## New behavior

A simple check has been added to rocket initialisation for checking the input dry inertia.

## Breaking change

- [ ] Yes
- [x] No

## Additional information

For future this check can be extended to dynamically testing the singularity of inertia tensor. Need for this is to be evaluated.

### DIFF
--- a/rocketpy/rocket/rocket.py
+++ b/rocketpy/rocket/rocket.py
@@ -294,6 +294,12 @@ class Rocket:
         self.I_13_without_motor = inertia[4]
         self.I_23_without_motor = inertia[5]
 
+        # Initial Inertia Tensor determinant singularity check
+        if abs(inertia) == 0:
+            raise ValueError(
+                "The rocket inertia tensor is singular (determinant is zero). "
+            )
+
         # Define rocket geometrical parameters in SI units
         self.center_of_mass_without_motor = center_of_mass_without_motor
         self.radius = radius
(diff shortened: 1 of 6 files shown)


rocketpy/rocket/rocket.py
```diff
@@ -292,6 +292,29 @@ def __init__(  # pylint: disable=too-many-statements
...
+                    self.I_13_without_motor,
+                    self.I_23_without_motor,
+                    self.I_33_without_motor,
+                ],
+            ]
+        )  # Initial Inertia Tensor determinant singularity check
+        if abs(inertia_matrix) == 0:
+            raise ValueError(
+                "The rocket inertia tensor is singular (determinant is zero). "
+            )
```
MateusStano, line 317: I think having the dry inertia as zero should not stop the simulation, since someone might only have the inertias on the motor for some reason.

In any case, we need the determinant of the Rocket + Motor Inertia Matrix to be different from zero. Could you move this check so that it is done when a motor is added?

aZira371: So you mean to say someone might want to simulate just a motor and check the trajectory? 

Yes agreed I'll move it to when the motor is added.

MateusStano:
Closed since this fix was not enough for the issue. It probably needs some extra checks in the flight class, and hopefully out of the simulation loop

PR #834, opened 2025-06-23
Enh/acceleration earth to rocket frame

gdilbaz:
## Pull request type

## Checklist

## Current behavior

Currently, there is no functionality in the code that performs the transformation of acceleration from the Earth frame to the Body frame.

## New behavior

This PR adds acceleration tranformation from Earth frame to Body frame.

## Breaking change

- [ ] Yes
- [x] No

## Additional information

I didn't add the feature as "axial_acceleration" as specified in the original issue in order to avoid confusion. Instead, I chose a clearer name "acceleration_earth_to_body_frame" which better represents what was originally in mind.


monte_carlo_test.errors.txt
Gui-FernandesBR, line 1: please remove these `.txt` files

Gui-FernandesBR: @Lucas-Prates , perhaps your latest PR introduced this problem of creating `.txt` when running Monte Carlo tests. Could you double check that please?

rocketpy/simulation/flight.py
Gui-FernandesBR, line 1: In terms of coding, it is quite fine to me already.
I would just say that we need to implement unit tests for this method before merging.

@MateusStano, what do you think in terms of physics/dynamics? Anything to comment on?

Gui-FernandesBR:
I really love when people have the initiative to open up a PR to this repo. It makes our work so valuable and collaborative.

Thanks!!

monte_carlo_test.errors.txt
gdilbaz: I'll look into that didn't realised thoese were included. Thanks for pointing it out!

MateusStano:
Hey @gdilbaz very cool that you did this!

Can you check if the `_stacked_acceleration_body_frame` property of the Flight class is the same thing that you have implemented? It might be the case that this is redundant

gdilbaz:
> Hey @gdilbaz very cool that you did this!
> 
> Can you check if the `_stacked_acceleration_body_frame` property of the Flight class is the same thing that you have implemented? It might be the case that this is redundant

After checking yes definitely. It looks like I missed it while looking through the properties and there is no need for this PR sorry for the confusion.

PR #835, opened 2025-06-26
ENH: Improve parachute geometric parametrization

ArthurJWH:
## Pull request type

## Checklist

## Current behavior

Parachute radius is assumed to be fixed (=1.5m)

## New behavior

Other than adding a radius parameter, there is also additional height and porosity parameters.
The default is still set to the current behavior.

## Breaking change

- [ ] Yes
- [x] No


rocketpy/rocket/parachute.py
```diff
@@ -154,13 +157,27 @@ def __init__(
...
         self.name = name
         self.cd_s = cd_s
         self.trigger = trigger
         self.sampling_rate = sampling_rate
         self.lag = lag
         self.noise = noise
+        self.parachute_radius = parachute_radius
+        if parachute_height is None:
+            parachute_height = parachute_radius
+        self.parachute_height = parachute_height
```
Gui-FernandesBR: cleaner

```suggestion
        self.parachute_radius = parachute_radius
        self.parachute_height = parachute_height or parachute_radius
```

rocketpy/simulation/flight.py
```diff
@@ -1961,22 +1979,27 @@ def u_dot_parachute(self, t, u, post_processing=False):
...
+        porosity = self.parachute_porosity
+
 
         # Get the mass of the rocket
         mp = self.rocket.dry_mass
 
         # Define constants
-        ka = 1  # Added mass coefficient (depends on parachute's porosity)
-        R = 1.5  # Parachute radius
+        ka = 1.068 * (1 - 1.465 * porosity - 0.25975 * porosity**2 + 1.2626 * porosity**3)
```
Gui-FernandesBR: this `ka` property should be saved inside the Parachute class

rocketpy/simulation/flight.py
```diff
@@ -1961,22 +1979,27 @@ def u_dot_parachute(self, t, u, post_processing=False):
 
         # Get Parachute data
         cd_s = self.parachute_cd_s
+        parachute_radius = self.parachute_radius
+        parachute_height = self.parachute_height
+        porosity = self.parachute_porosity
+
```
Gui-FernandesBR: We don't need these definitions...

We can save a few nanoseconds by not defining them here.

You can use the attributes directly: using self.cds instead of cds

rocketpy/simulation/flight.py
```diff
@@ -1987,12 +2010,12 @@ def u_dot_parachute(self, t, u, post_processing=False):
         # Determine drag force
         pseudo_drag = -0.5 * rho * cd_s * free_stream_speed
         # pseudo_drag = pseudo_drag - ka * rho * 4 * np.pi * (R**2) * Rdot
```
Gui-FernandesBR: These comments??? Should we start using it?

```suggestion
        pseudo_drag = -0.5 * rho * cd_s * free_stream_speed
        # pseudo_drag = pseudo_drag - ka * rho * 4 * np.pi * (R**2) * Rdot
```

rocketpy/rocket/parachute.py
```diff
@@ -154,13 +157,27 @@ def __init__(
             The values are used to add noise to the pressure signal which is
             passed to the trigger function. Default value is ``(0, 0, 0)``.
             Units are in Pa.
+        parachute_radius : float, optional
```
Gui-FernandesBR: can you be more spacific on the parachute_radius definition please? I believe it is still not intuitive what the radius is.

rocketpy/rocket/rocket.py
```diff
@@ -1493,16 +1502,36 @@ def add_parachute(
...
+        parachute_radius : float, optional
+            Radius of the inflated parachute. Default value is 1.5.
+            Units are in meters.
+        parachute_height : float, optional
+            Height of the inflated parachute. Default value is the radius parachute.
+            Units are in meters.
+        porosity : float, optional
+            Porosity of the parachute material, which is a measure of how much air can
+            pass through the parachute material.
+            Default value is 0.0432 (for consistency with previous versions).
```
MateusStano: Could you define more precisely what porosity is? Something like: "The ratio of the area of the porous structure to the total area of the canopy" (I don't know if this is the correct definition in our case, just an example)

MateusStano (approved):

README.md
```diff
@@ -261,6 +261,9 @@ main = calisto.add_parachute(
     sampling_rate=105,
     lag=1.5,
     noise=(0, 8.3, 0.5),
+    parachute_radius=1.5,
+    parachute_height=1.5,
+    porosity=0.0432,
```
Gui-FernandesBR, line 267: having both `cds` and `parachute_radius`? do you think we could possible calculate the cd based on these other parameters?

rocketpy/rocket/parachute.py
```diff
@@ -92,6 +92,15 @@ class Parachute:
         Function of noisy_pressure_signal.
     Parachute.clean_pressure_signal_function : Function
         Function of clean_pressure_signal.
+    Parachute.parachute_radius : float
+        Length of the non-unique semi-axis (radius) of the inflated hemispheroid
+        parachute in meters.
+    Parachute.parachute_height : float
```
Gui-FernandesBR:
```suggestion
    Parachute.parachute_height : float, None
```

or


```suggestion
    Parachute.parachute_height : Optional[float]
```

rocketpy/rocket/rocket.py
Gui-FernandesBR, line 1: Question: how should the Monte Carlo simulations be affected by this PR? 
Should we open another PR allowing the user to vary these new arguments within the `StochasticParachute` (https://docs.rocketpy.org/en/latest/reference/classes/monte_carlo/stochastic_models/stochastic_parachute.html) class?

rocketpy/rocket/parachute.py
```diff
@@ -170,6 +193,11 @@ def __init__(
         self.clean_pressure_signal_function = Function(0)
         self.noisy_pressure_signal_function = Function(0)
         self.noise_signal_function = Function(0)
+        self.parachute_radius = parachute_radius
+        if parachute_height is None:
+            parachute_height = parachute_radius
+        self.parachute_height = parachute_height
```
Gui-FernandesBR:
```python
self.parachute_height = parachute_height or parachute_radius
```

Gui-FernandesBR (approved):
Overall code looks cleaner and better, congratulations on this implementation, @ArthurJWH .

I have a few comments that I believe would help you to improve clarity of your code. Please let me know if you have any questions.

rocketpy/rocket/rocket.py
```diff
@@ -1433,7 +1433,16 @@ def add_free_form_fins(
...
-        self, name, cd_s, trigger, sampling_rate=100, lag=0, noise=(0, 0, 0)
+        self,
+        name,
+        cd_s,
+        trigger,
+        sampling_rate=100,
+        lag=0,
+        noise=(0, 0, 0),
+        parachute_radius=1.5,
+        parachute_height=None,
```
Gui-FernandesBR: I just noticed it now... 

Since it is the Parachute class, there's no need to specify this is the parachute radius or the parachute height.

```suggestion
        radius=1.5,
        height=None,
```

ArthurJWH: I was looking in other Rocket class methods such as add_trapezoidal_fins and set_rail_buttons, and both use `radius` already to refer to the fuselage radius. So changing it in the add_parachute method wouldn't break the consistency?

ArthurJWH: Also, since we already use `radius` in some inputs as the fuselage radius, maybe it will be more clear as an input to explicitly mention parachute

Gui-FernandesBR: I would like to hear a third opinion here, maybe @MateusStano or @phmbressan .

Based on my experience, I don't think parachute_radius is a good idea, it sounds verbose to me.
However, I don't know iff we would have any conflict with other classes.

MateusStano: I think the simpler naming is better. If the documentation is there we can rely on that

rocketpy/simulation/flight.py
```diff
@@ -1963,24 +1998,27 @@ def u_dot_parachute(self, t, u, post_processing=False):
...
         # Calculate added mass
-        ma = ka * rho * (4 / 3) * np.pi * R**3
+        ma = (
+            self.ka
+            * rho
+            * (4 / 3)
+            * np.pi
+            * self.parachute_radius**2
+            * self.parachute_height
+        )
```
MateusStano: This is calculating the volume of the whole sphere right? I think it should be divided by two to account for the hemispheroid shape only

rocketpy/rocket/parachute.py
```diff
@@ -154,6 +166,17 @@ def __init__(
...
+            Length of the non-unique semi-axis (radius) of the inflated hemispheroid
+            parachute. Default value is 1.5.
+            Units are in meters.
+        height : float, optional
+            Length of the unique semi-axis (height) of the inflated hemispheroid
+            parachute. Default value is the radius of the parachute.
+            Units are in meters.
+        porosity : float, optional
+            Porosity of the parachute is the ratio of open space in the canopy to total
+            canopy area. Default value is 0.0432 (for consistency with previous versions).
```
MateusStano:
```suggestion
        porosity : float, optional
            Geometric porosity of the canopy (ratio of open area to total canopy area),
            in [0, 1]. Affects only the added-mass scaling during descent; it does
            not change ``cd_s`` (drag). The default, 0.0432, yields an added-mass
            of 1.0 (“neutral” behavior).
```

Gui-FernandesBR:
@ArthurJWH please let me know when you have this PR ready for a re-review

ArthurJWH:
> @ArthurJWH please let me know when you have this PR ready for a re-review

I think it is ready for the review now

rocketpy/rocket/parachute.py
```diff
@@ -170,6 +197,15 @@ def __init__(
...
         self.noise_signal_function = Function(0)
+        self.radius = radius
+        self.height = height or radius
+        self.porosity = porosity
+        self.ka = 1.068 * (
+            1
+            - 1.465 * self.porosity
+            - 0.25975 * self.porosity**2
+            + 1.2626 * self.porosity**3
+        )
```
Gui-FernandesBR: Could you document Parachute.ka in the class docstring please?

Also, what does ka mean? Could we think about a new, more descriptive name for it?

In python, we rather use the explicit version than implicit version of name variables.


self.aoa is not preferred
self.angle_of_attach is preferred

Gui-FernandesBR (approved):
Best PR o 2025 so far, congratulations, @ArthurJWH !!

rocketpy/simulation/flight.py
Gui-FernandesBR: @ArthurJWH

rocketpy/rocket/parachute.py
ArthurJWH: ka is the added mass coefficient. Should I change the variable to  added_mass_coefficient and document it?

MateusStano: Yes, please. To document it, you just have to add some docstrings to class `Attributes` section

Issue #663, opened 2024-08-19
ENH: Allow overshooting controllers


Gui-FernandesBR:
Needs to change the way we deal with overshooting in the Flight class.

PR #843, opened 2025-08-15
ENH: Refactor Flight class to improve time node handling and sensor/controllers

Gui-FernandesBR:
## Pull request type

## Checklist

## Current behavior vs New behavior
1. The `Flight.__simulate()` method is more readable
2. Controllers can now be used with time overshoot option = True (solves https://github.com/RocketPy-Team/RocketPy/issues/663)
3. the same is valid for sensors

Controllers and Sensors Simulations should run faster!!

## Breaking change

- [x] No

## Additional information
- I need help to review this PR carefully, although tests are passing locally.

### DIFF
--- a/rocketpy/simulation/flight.py
+++ b/rocketpy/simulation/flight.py
@@ -687,7 +687,7 @@ class Flight:
             )
 
             # Initialize phase time nodes
-            self.__setup_phase_time_nodes(phase, phase_index)
+            self.__setup_phase_time_nodes(phase)
 
             # Iterate through time nodes
             for node_index, node in self.time_iterator(phase.time_nodes):
@@ -821,15 +821,13 @@ class Flight:
         if verbose:
             print(f"\n>>> Simulation Completed at Time: {self.t:3.4f} s")
 
-    def __setup_phase_time_nodes(self, phase, phase_index):
+    def __setup_phase_time_nodes(self, phase):
         """Set up time nodes for the current phase.
 
         Parameters
         ----------
         phase : FlightPhase
             The current flight phase.
-        phase_index : int
-            The index of the current phase.
         """
         phase.time_nodes = self.TimeNodes()
 
@@ -951,8 +949,11 @@ class Flight:
             ):
                 continue  # Check next parachute
 
-            # Remove parachute from flight parachutes
-            self.parachutes.remove(parachute)
+            # Remove parachute from flight parachutes (if not already removed)
+            if parachute in self.parachutes:
+                self.parachutes.remove(parachute)
+            else:
+                continue  # Parachute already triggered, skip to next
 
             # Create phase for time after detection and before inflation
             # Must only be created if parachute has any lag
@@ -1591,7 +1592,6 @@ class Flight:
         """Initialize controllers and sensors"""
         self._controllers = self.rocket._controllers[:]
         self.sensors = self.rocket.sensors.get_components()
-        # Note: time_overshoot now supports both controllers and sensors
 
         # reset controllable object to initial state (only airbrakes for now)
         for air_brakes in self.rocket.air_brakes:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@ Attention: The newest changes should be on top -->
 
 ### Changed
 
--
+- ENH: Refactor Flight class to improve time node handling and sensor/controllers [#843](https://github.com/RocketPy-Team/RocketPy/pull/843)
 
 ### Fixed
 
--- a/rocketpy/plots/flight_plots.py
+++ b/rocketpy/plots/flight_plots.py
@@ -720,9 +720,16 @@ class _FlightPlots:
         ax2.grid()
 
         ax3 = plt.subplot(413)
+        # Handle both array-based and callable-based Functions
+        thrust_power = self.flight.thrust_power
+        if callable(thrust_power.source):
+            # For callable sources, discretize based on speed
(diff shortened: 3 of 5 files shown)


rocketpy/simulation/flight.py
```diff
@@ -705,81 +683,12 @@ def __simulate(self, verbose):
...
-                        )
-
-                for controller in node._controllers:
-                    controller(
-                        self.t,
-                        self.y_sol,
-                        self.solution,
-                        self.sensors,
-                    )
+                self.__process_sensors_and_controllers_at_current_node(node, phase)
```
phmbressan, line 711: I don't mind much, since this is a private method naming. However, do you think there might be a more scalable / descriptive name here, so as to avoid adding words to it in the future: e.g. `__process_sensors_and_controllers_and...`? Having trouble naming a method may indicate it may do too much.

Actually, my suggestion would be going for a more standardized approach. I have never seen a fixed convention for this, but it is nice to have some internal consistency when naming checkers/process/handlers, my two cents would be:

- `handler`: returns a boolean and operate on solution variables;
- `process`: can operate on solution variables, but does not return;
- `checker`: returns a boolean, but only reads solution variables (e.g. does not setup phases/nodes, I did not see any pure checkers on this PR).

```py
self.__process_sensors(...)
self.__process_controllers(...)

if self.__handle_parachute_triggers(...):
    break

while running:
    if self.__handle_simulation_events(...):
        break

    if self.time_overshoot:
        # I nested the if clauses for readability
        # Both ways are equivalent and should perform the same
        if self.__handle_overshoot(...):
            break
```

rocketpy/simulation/flight.py
```diff
@@ -1053,6 +726,573 @@ def __simulate(self, verbose):
...
+            ) = self.__calculate_and_save_pressure_signals(
+                parachute, node.t, self.y_sol[2]
+            )
+            if not parachute.triggerfunc(
+                noisy_pressure,
+                height_above_ground_level,
+                self.y_sol,
+                self.sensors,
+            ):
+                return False  # Early exit: parachute not deployed
```
phmbressan: Really not sure, since the code is complex, but shouldn't this be a `continue` (i.e. even if one trigger does not go off, there are other parachutes to check)?

rocketpy/simulation/flight.py
```diff
@@ -1053,6 +726,573 @@ def __simulate(self, verbose):
...
+            # Prepare to leave loops and start new flight phase
+            phase.time_nodes.flush_after(node_index)
+            phase.time_nodes.add_node(self.t, [], [], [])
+            phase.solver.status = "finished"
+            self.parachute_events.append([self.t, parachute])
+            return True
+
+        return False
+
+    def __check_simulation_events(self, phase, phase_index, node_index):
```
phmbressan, line 1008: Since the three methods inside it are handlers, I would also call this one a handler.

rocketpy/simulation/flight.py
```diff
@@ -1053,6 +726,573 @@ def __simulate(self, verbose):
...
+                u_dot = phase.derivative(overshootable_node.t, overshootable_node.y_sol)
+                self.__measure_sensors(
+                    overshootable_node._component_sensors,
+                    u_dot,
+                    overshootable_node.t,
+                    overshootable_node.y_sol,
+                )
+        return False
+
+    def __check_overshootable_parachute_triggers(
```
phmbressan, line 1310: This method and the `__handle_parachute_triggers` are essentially identical (https://www.diffchecker.com/FYbfsMQf/) aside from the history rollback. They could be converted intro a single method with a boolean parameter to toggle the rollback.

phmbressan (commented):
Good work! I hope this is the beginning of having a much more readable and developer friendly `Flight` class.

My only concern is making sure the behavior is exactly the same as before, therefore I made a single important comment on the early return of parachute triggers just to make sure we are on the same page. Of course, all the tests being green is already a big step.

Aside from that, my other suggestions are more style/naming choices. I prefer having some more standardized names for some methods, but I would approve the PR in the current state, since it already does loads for readability.

rocketpy/simulation/flight.py
MateusStano: Yeah, if there is more than one parachute here the pressure signals wont be logged correctly

rocketpy/simulation/flight.py
```diff
@@ -1268,16 +1569,11 @@ def __init_controllers(self):
...
-            if self.time_overshoot:  # pragma: no cover
-                self.time_overshoot = False
-                warnings.warn(
-                    "time_overshoot has been set to False due to the presence "
-                    "of controllers or sensors. "
-                )
-            # reset controllable object to initial state (only airbrakes for now)
-            for air_brakes in self.rocket.air_brakes:
-                air_brakes._reset()
+        # Note: time_overshoot now supports both controllers and sensors
```
Gui-FernandesBR:
```suggestion
```

rocketpy/simulation/flight.py
```diff
@@ -1095,6 +799,604 @@ def __simulate(self, verbose):
...
+            self._controllers, self.solution[-2][0], self.t
+        )
+        overshootable_nodes.add_sensors(
+            self.rocket.sensors, self.solution[-2][0], self.t
+        )
+
+        # Add last time node (always skipped)
+        overshootable_nodes.add_node(self.t, [], [], [])
+
+        if len(overshootable_nodes) < 1:
```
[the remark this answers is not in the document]
Gui-FernandesBR: needs to check

PR #849, opened 2025-09-20
ENH: Controller (AirBrakes) and Sensors Encoding

phmbressan:
## Pull request type

## Checklist

- [X] Tests for the changes have been added. Note: could be expanded
- [X] Lint (`make format && make pylint`) has passed locally 

## Current behavior

Controllers are not correctly encoded nor decoded; this is especially regarding rockets with `AirBrakes` which are currently not exportable to an `.rpy` file. Sensors are also not correctly decoded. 

## New behavior

This PR brings the encoding and decoding of `Controller` and `Sensor` classes. The main challenge for it is that the `Controllers` observe a set of `interactable_objects`, which _must_ be the same ones as the ones added to the `Rocket`. 

It is important to note that a naive decoding would create a copy of the `interactable_objects`. Therefore, hashes were added to the signature of rocketpy objects so as to allow linking them after the decoding to the correct controller.

Retro compatibility should be carefully examined, but it was taken into consideration when writing this PR. The test that uses an `v1.8` `.rpy` file is passing normally. 

## Breaking change

- [ ] Yes
- [X] No


rocketpy/rocket/aero_surface/air_brakes.py
```diff
@@ -206,3 +206,25 @@ def all_info(self):
         """
         self.info()
         self.plots.drag_coefficient_curve()
+
+    # pylint: disable=unused-argument
+    def to_dict(self, **kwargs):
```
Gui-FernandesBR: can you move this to inside the function?

Gui-FernandesBR: this = pylint rule

phmbressan: Solved.

phmbressan:
> LGTM,
> 
> great work.
> 
> Can yoiu check if all the modified lines are included in coverage? tks for the improvement

Could you re-review the changes I made to the .git actions. The tkinter and tcl errors were happening in the windows actions for unknown reasons to me (always these libraries). I tried an experimental test with a non-interative matplotlib backend and everything seem to have run smoothly.

Regarding your point about coverage, the seemingly drop in converage is due to the fact I placed the encoding runs of `sensors` and `controllers` as slow tests (since these sims are slightly slower). I might move those to non-slow tests with the expense of some 5s of run time.

PR #867, opened 2025-11-01
ENH: Parachute Inflation Modeling

ArthurJWH:
## Pull request type

## Checklist

## Current behavior

When triggered, the parachute is modeled as fully opened

## New behavior

The parachute inflation will be simulated using a volumetric flow model with simple inviscid flow

## Breaking change

- [ ] Yes
- [ ] No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ Attention: The newest changes should be on top -->
 - ENH: Add thrustcurve api integration to retrieve motor eng data [#870](https://github.com/RocketPy-Team/RocketPy/pull/870)
 - ENH: Compatibility with MERRA-2 atmosphere reanalysis files [#825](https://github.com/RocketPy-Team/RocketPy/pull/825)
 - ENH: Enable only radial burning [#815](https://github.com/RocketPy-Team/RocketPy/pull/815)
+- ENH: Parachute Inflation Modeling [#867](https://github.com/RocketPy-Team/RocketPy/pull/867)
 
 ### Changed
 


rocketpy/simulation/flight.py
```diff
@@ -1960,40 +2001,80 @@ def u_dot_parachute(self, t, u, post_processing=False):
...
+
+        radius = self.parachute_radius
+        height = self.parachute_height
+        inflated_radius = self.parachute_inflated_radius
+        inflated_height = inflated_radius * height / radius
+
+        # Calculate the surface area of the parachute
+        if radius > height:
+            e = math.sqrt(1 - (inflated_height**2) / (inflated_radius**2))
+            surface_area = math.pi * inflated_radius**2 * (1 + (1 - e ^ 2) / e * math.atanh(e))
```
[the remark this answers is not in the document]
Gui-FernandesBR: true

rocketpy/simulation/flight.py
```diff
@@ -761,7 +761,24 @@ def __simulate(self, verbose):
                         callbacks = [
                             lambda self, parachute_cd_s=parachute.cd_s: setattr(
                                 self, "parachute_cd_s", parachute_cd_s
-                            )
+                            ),
+                            lambda self,
+                            parachute_radius=parachute.parachute_radius: setattr(
```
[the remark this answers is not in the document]
Gui-FernandesBR: true

rocketpy/simulation/flight.py
```diff
@@ -765,6 +765,21 @@ def __simulate(self, verbose):
                                 "parachute_added_mass_coefficient",
                                 added_mass_coefficient,
                             ),
+                            lambda self: setattr(
+                                self,
+                                "parachute_volume",
```
Gui-FernandesBR, line 771: @ArthurJWH why do you need to calculate the `parachute_volume` within the flight simulation? Can you calculate and store it in the Parachute class?

rocketpy/simulation/flight.py
```diff
@@ -2702,31 +2732,58 @@ def u_dot_parachute(self, t, u, post_processing=False):
...
-        )
-
         # Calculate freestream speed
         freestream_x = vx - wind_velocity_x
         freestream_y = vy - wind_velocity_y
         freestream_z = vz
         free_stream_speed = (freestream_x**2 + freestream_y**2 + freestream_z**2) ** 0.5
 
+        # Initialize parachute geometrical parameters
+        radius = self.parachute_radius
```
Gui-FernandesBR: Wha is the benefit behind this `radius = self.parachute_radius`?

rocketpy/simulation/flight.py
```diff
@@ -2702,31 +2732,58 @@ def u_dot_parachute(self, t, u, post_processing=False):
...
+        height = self.parachute_height
+        inflated_radius = (
```
Gui-FernandesBR: calculating `inflated_radius` in the Flight class? Do you really need it?

rocketpy/simulation/flight.py
```diff
@@ -2702,31 +2732,58 @@ def u_dot_parachute(self, t, u, post_processing=False):
...
+        volume_flow = (
+            rho
+            * freestream_z  # considering parachute as vertical
+            * (
+                (math.pi * inflated_radius**2)
+                - (self.parachute_porosity * surface_area)
+            )
+        )
+
+        # Getting time step
```
Gui-FernandesBR, line 2766: do you really need to store step.t0? Why?

Should this really be a public attribute?

rocketpy/simulation/flight.py
ArthurJWH: My original idea was to assume the initial radius being the rocket radius (compartment from where it is ejected), but I cannot access the rocket radius within Parachute class. I can add a new parameter "initial radius" for the Parachute class, if you think that is better

rocketpy/simulation/flight.py
ArthurJWH: I thought it would help with the readability of the code, especially when the parameters are used in the convoluted equations. I can remove it

rocketpy/simulation/flight.py
ArthurJWH: `inflated_radius` is the parachute radius in the current Flight stage, it is used to calculate the surface area of the parachute

rocketpy/simulation/flight.py
Gui-FernandesBR: I understand it.
it does not contribute so much to the readability actually.

We can save a few 0.0001 ns by removing this line

rocketpy/simulation/flight.py
ArthurJWH: I used it to have access to the time step between `u_dot_parachute` calls (not ideal). I can change it for a private attribute, or do you have any other suggestion?

rocketpy/simulation/flight.py
```diff
@@ -2702,31 +2732,58 @@ def u_dot_parachute(self, t, u, post_processing=False):
...
+            e = math.sqrt(1 - (inflated_height**2) / (inflated_radius**2))
+            surface_area = (
+                math.pi * inflated_radius**2 * (1 + (1 - e**2) / e * math.atanh(e))
+            )
+        else:
+            e = math.sqrt(1 - (inflated_radius**2) / (inflated_height**2))
+            surface_area = (
+                math.pi
+                * inflated_radius**2
+                * (1 + inflated_height / (e * inflated_radius) * math.asin(e))
```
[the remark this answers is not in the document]
ArthurJWH: Since the inflated radius / inflated height = radius / height, the if else statement should already account for sign of the value under the square root

rocketpy/simulation/flight.py
Gui-FernandesBR: private is better

rocketpy/rocket/parachute.py
```diff
@@ -98,6 +99,9 @@ class Parachute:
     Parachute.height : float, None
         Length of the unique semi-axis (height) of the inflated hemispheroid
         parachute in meters.
+    Parachute.initial_radius : float, None
+        Initial radius of the parachute on deployment in meters. Used to model the
+        parachute inflation.
```
MateusStano, line 104: This should be more descriptive. What exacly is the initial radius? Is it the radius at the moment of deployment?

MateusStano: Also please change all the other docstrings that mention this

rocketpy/simulation/flight.py
```diff
@@ -2702,31 +2723,61 @@ def u_dot_parachute(self, t, u, post_processing=False):
...
+        # Integrating parachute volume
+        max_volume = (2 / 3) * math.pi * self.parachute_radius**2 * self.parachute_height
+        self.parachute_volume = min(self.parachute_volume + volume_flow * dt, max_volume)
+
+        # Dragged air mass
+        ma = self.parachute_volume * rho
+
+        # Moving time step
+        self.__t0 = t1
+
```
MateusStano, line 2780: Do the equations here have a source? If not can you write them in a simple document, just so it is clearer what exactly is being done here. We can then put them in the documentation

rocketpy/simulation/flight.py
```diff
@@ -2702,31 +2723,61 @@ def u_dot_parachute(self, t, u, post_processing=False):
...
+        )
+
+        # Getting time step
+        self.__t0 = getattr(self, "t0", t)
+        t1 = t
+        dt = t1 - self.__t0
+
+        # Integrating parachute volume
+        max_volume = (2 / 3) * math.pi * self.parachute_radius**2 * self.parachute_height
+        self.parachute_volume = min(self.parachute_volume + volume_flow * dt, max_volume)
```
MateusStano, line 2773: The `self.__t0` and `self.parachute_volume` are currently mutating inside the integrator calls. I don't think that's a good idea and can lead to wrong results because ODE solvers will call udot multiple times per step. I think we should maybe have e new u_dot function that has the parachute_volume as a state

Basically I think you should try to create a new u_dot that computes dV/dt = volume_flow(...) in RHS and return u_dot = `[vx, vy, vz, ax, ay, az, ..., alpha3, parachute_volume]`

Issue #882, opened 2025-11-27
ENH: implementation of quasi static attitude variation to improve lateral motion of 3 DOF


aZira371:
**Is your feature request related to a problem? Please describe.**

The 3-DOF flight simulation keeps a fixed attitude throughout the flight. The quaternion derivatives are always zero, meaning the body axis direction never evolves despite changes in the relative wind direction.

**Describe the solution you'd like**

Introduces an evolving unit direction vector for the body axis using a quasi-static weathercocking model. The rocket's attitude gradually aligns toward the relative wind direction.

Changes:

- New weathercock_coeff parameter in Flight.__init__ (default: 1.0)
- Updated u_dot_generalized_3dof to compute quaternion derivatives proportional to misalignment with relative wind
- Angular velocity = weathercock_coeff * sin(misalignment_angle)
- Added example notebook docs/examples/bella_lui_3dof_vs_6dof_comparison.ipynb comparing 6-DOF and 3-DOF simulations

How the Weathercocking Model Works:

- Calculate the current body z-axis direction from quaternions (attitude vector)
- Calculate the desired direction as the opposite of the freestream velocity (normalized)
- Compute the rotation axis as the cross product of current and desired directions
- Apply angular velocity proportional to sin(misalignment_angle) (for quasi-static alignment)
- Compute quaternion derivatives from this angular velocity

**Additional context**

Comparison of expected results with and w/o weathercocking:

<img width="844" height="552" alt="Image" src="https://github.com/user-attachments/assets/62541287-debd-4256-967e-205409036c22" />

<img width="695" height="276" alt="Image" src="https://github.com/user-attachments/assets/f553517c-23ec-400f-8b28-f90b4234c9a8" />

PR #883, opened 2025-11-27
ENH: 3-dof lateral motion improvement

aZira371:
## Pull request type

## Checklist

## Current behavior

The 3-DOF flight simulation keeps a fixed attitude throughout the flight. The quaternion derivatives are always zero, meaning the body axis direction never evolves despite changes in the relative wind direction.
https://github.com/RocketPy-Team/RocketPy/issues/882

## New behavior

Introduces an evolving unit direction vector for the body axis using a quasi-static weathercocking model. The rocket's attitude gradually aligns toward the relative wind direction.

Changes:

- New weathercock_coeff parameter in Flight.init (default: 1.0)
- Updated u_dot_generalized_3dof to compute quaternion derivatives proportional to misalignment with relative wind
- Angular velocity = weathercock_coeff * sin(misalignment_angle)
- Added example notebook docs/examples/bella_lui_3dof_vs_6dof_comparison.ipynb comparing 6-DOF and 3-DOF simulations

## Breaking change

- [x] No

## Additional information

A new jupyter notebook to be added, depicting comparison of 3 dof and 6 dof. It also highlights use of weathercocking coefficient.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ Attention: The newest changes should be on top -->
 
 ### Added
 
+- ENH: 3-dof lateral motion improvement [#883](https://github.com/RocketPy-Team/RocketPy/pull/883)
 - ENH: Add multi-dimensional drag coefficient support (Cd as function of M, Re, α) [#875](https://github.com/RocketPy-Team/RocketPy/pull/875)
 - ENH: Add save functionality to `_MonteCarloPlots.all` method [#848](https://github.com/RocketPy-Team/RocketPy/pull/848)
 - ENH: add animations for motor propellant mass and tank fluid volumes [#894](https://github.com/RocketPy-Team/RocketPy/pull/894)


docs/examples/bella_lui_3dof_vs_6dof_comparison.ipynb
Gui-FernandesBR, line 1: simply adding the file here won't have much effect.

You should search in the `docs` folder to find a googd place for this simulation. Try to read https://docs.rocketpy.org/en/latest/ and imagine where would you put this analysis. I'm afraid the "flight examples" section is not suitable for this.

aZira371: yes noted will shift it to a better place!

aZira371: i think the notebooks section is a good place in my opinion. It contains utilities usage and sensor note books as well. What do you say @Gui-FernandesBR?

Gui-FernandesBR: Although a good option, notebooks docs are legacy now. You should look towards moving to .rst files.


Also, it is important to actually import your file in some of the index.rst file, otherwise it won't be rendered in the documentation page

aZira371: That is also fair to do!

tests/unit/simulation/test_flight_3dof.py
```diff
@@ -137,3 +137,176 @@ def test_invalid_simulation_mode(example_plain_env, calisto):
...
+        rocket=point_mass_rocket,
+        environment=example_plain_env,
+        rail_length=1,
+        simulation_mode="3 DOF",
+        weathercock_coeff=2.5,
+    )
+    assert flight.weathercock_coeff == 2.5
+
+
+def test_weathercock_coeff_default(example_plain_env, point_mass_rocket):
```
Gui-FernandesBR: for the `test_weathercock_coeff_default`, do you really need to redefine a new flight object? Couldn't you use a fixture like calisto_flight (I dont remeber the name) or similar?

tests/unit/simulation/test_flight_3dof.py
```diff
@@ -137,3 +137,176 @@ def test_invalid_simulation_mode(example_plain_env, calisto):
...
+    flight = Flight(
+        rocket=point_mass_rocket,
+        environment=example_plain_env,
+        rail_length=1,
+        simulation_mode="3 DOF",
+    )
+    assert flight.weathercock_coeff == 1.0
+
+
+def test_weathercock_zero_gives_fixed_attitude(example_plain_env, point_mass_rocket):
```
Gui-FernandesBR: for the `test_weathercock_zero_gives_fixed_attitude`, can you re-use another fixture? 

fixtures make tests faster as pylint re-use their valuies instead of creating the object multiple times

tests/integration/simulation/test_flight_3dof.py
Gui-FernandesBR, line 1: unit tests are tests which "isolatly" test a single method. Otherwise the test is most likely integration tests.


Please read your code twice and decide wheter your tests are really unit tests or migrate then to integration.

please read: https://docs.rocketpy.org/en/latest/development/testing.html

docs/user/three_dof_simulation.rst
```diff
@@ -340,6 +340,316 @@ Here's a complete 3-DOF simulation from start to finish:
...
+- ``weathercock_coeff=1.0``: Default moderate alignment rate
+- ``weathercock_coeff>1.0``: Faster alignment (more stable rocket)
+
+Effect of Weathercocking Coefficient
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Higher values of ``weathercock_coeff`` result in faster alignment with the
+relative wind. This affects the lateral motion and impact point:
+
+.. list-table:: Weathercocking Coefficient Effects
```
Gui-FernandesBR, line 439: Out of curiosity, what is the source of these values?

```rst
.. list-table:: Weathercocking Coefficient Effects
```

Gui-FernandesBR (commented):
overall the code looks clean. I see a few things that concerns me, so I listed below.

rocketpy/simulation/flight.py
```diff
@@ -492,6 +492,7 @@ def __init__(  # pylint: disable=too-many-arguments,too-many-statements
         equations_of_motion="standard",
         ode_solver="LSODA",
         simulation_mode="6 DOF",
+        weathercock_coeff=1.0,
```
Gui-FernandesBR: should the default value really be set to 1.0? Why not `None`?

aZira371: Yes correct I should put none as a default to portray true 3 dof behaviour as default.

tests/integration/simulation/test_flight_3dof.py
aZira371: Let me go through the tests once again and see if I misunderstood some and shift to integration as per my understanding.

tests/unit/simulation/test_flight_3dof.py
aZira371: yes but weathercock model only works with 3 dof, so a point mass rocket would be needed to run with 3 dof flight class. Thus recreating new flight objects with point mass inputs.

docs/user/three_dof_simulation.rst
aZira371: The weathercocking coefficient (wc) in this model represents the overall rate at which the rocket’s body axis aligns with the wind direction. It does this without considering aerodynamic surfaces like fins or calculating aerodynamic forces and torques. Normally, weathercocking depends on the rocket’s static margin and normal force coefficient, because these create aerodynamic forces that slowly turn the rocket into the wind.However, a 3-DOF point mass simulation cannot compute such aerodynamic forces or torques. So, this model artificially forces the rocket’s axis to align to the wind independently of aerodynamic details or angular dynamics.You can think of the wc as a value that approximates effects of static margin and normal force coefficient, but it has no direct physical meaning here. Instead, designers find suitable wc values through trial and error that make sense for their rocket design.This approach is useful for early design stages: designers pick a wc that achieves their target flight behavior and later check if the real aerodynamics match this ideal as the design matures.This makes wc a practical tuning parameter rather than a physically derived coefficient
In short these values are not from a particular source, as more simulations are run in 3 dof we will get a sense of what values lead to different kind of results. This implementation just came out of my head so I am not sure where to find such values but I do expect users to comeback with queries or inputs to verify this model a bit more! This makes 3 dof to behave like a 3.5 dof model XD.

tests/unit/simulation/test_flight_3dof.py
Gui-FernandesBR: ok no problem. But please try to use more fixtures so we can speed up tests

docs/user/three_dof_simulation.rst
Gui-FernandesBR: That's alright, I understand it.

But I mean, where did you learn this from? Can you site your sources?

This is important (not mandatory) to users that want to use this new kind of simulation

aZira371: Cool, I'll update the documentation of 3 dof over the weekend to include the above explanation as well sourves available!

aZira371: Cool, I'll update the documentation of 3 dof over the weekend to include the above explanation as well sourves available!

tests/integration/simulation/test_flight_3dof.py
aZira371: I reviewed the RocketPy testing guidelines and moved test_flight_3dof.py to integration/simulation because the file exercises interactions between multiple components (Flight, Rocket, Motor, Environment) and thus matches the integration-test definition. I hope this is a better placement and interpretation on my end!

Gui-FernandesBR, line 1: @aZira371 all the fixtures need to be defined within the `tests/fixtues` folder

docs/user/three_dof_simulation.rst
```diff
@@ -340,6 +340,337 @@ Here's a complete 3-DOF simulation from start to finish:
...
+        latitude=47.213476,
+        longitude=9.003336,
+        elevation=407,
+    )
+    env.set_atmospheric_model(type="standard_atmosphere")
+    env.max_expected_height = 2000
+
+    # Full 6-DOF Motor
+    motor_6dof = SolidMotor(
+        thrust_source="../data/motors/aerotech/AeroTech_K828FJ.eng",
```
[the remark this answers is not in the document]
aZira371: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/RocketPy-Team/RocketPy/pull/883#discussion_r2575358090)

rocketpy/simulation/flight.py
```diff
@@ -1895,9 +1906,100 @@ def u_dot_generalized_3dof(self, t, u, post_processing=False):
...
+                    omega1_wc, omega2_wc, omega3_wc = (
+                        omega_body.x,
+                        omega_body.y,
+                        omega_body.z,
+                    )
+                    e0_dot = 0.5 * (-omega1_wc * e1 - omega2_wc * e2 - omega3_wc * e3)
+                    e1_dot = 0.5 * (omega1_wc * e0 + omega3_wc * e2 - omega2_wc * e3)
+                    e2_dot = 0.5 * (omega2_wc * e0 - omega3_wc * e1 + omega1_wc * e3)
+                    e3_dot = 0.5 * (omega3_wc * e0 + omega2_wc * e1 - omega1_wc * e2)
+                    e_dot = [e0_dot, e1_dot, e2_dot, e3_dot]
```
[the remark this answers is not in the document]
aZira371: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/RocketPy-Team/RocketPy/pull/883#discussion_r2575358120)

PR #888, opened 2025-11-27
ENH: Built-in flight comparison tool (`FlightComparator`) to validate simulations against external data

Monta120:
## Pull request type

## Checklist

## Current behavior
Currently, users who want to compare RocketPy simulations against real flight data (avionics logs) or other simulators (OpenRocket, RAS Aero) face significant obstacles. They must:

1- Manually extract arrays from RocketPy objects.

2- Write custom code to interpolate their external data to match RocketPy's time steps (or vice versa) to calculate errors.

3- Write custom Matplotlib code to visualize the differences.

## New behavior
This PR introduces a built-in Flight Comparison architecture that automates the validation process.

1) New Class FlightComparator: A dedicated class in rocketpy/simulation/flight_comparator.py that:

- Ingests external data (dictionaries of tuples or RocketPy Functions).

- Automatically Interpolates disparate datasets onto a shared time grid (1,000 points) to allow point-by-point comparison.

- Calculates error metrics: RMSE (Root Mean Square Error), MAE (Mean Absolute Error), and Max Deviation.

- Generates comparison tables for Key Events (Apogee, Max Velocity, Impact Velocity) with relative error percentages.

2) Visualization:

- Comparison Plots: Standardized plots showing Simulation vs. External Data (Top) and Residuals/Error (Bottom).

- 2D Trajectories: Visual comparison of flight paths (e.g., X vs Z) to analyze drift.

## Breaking change

- [ ] Yes
- [X] No

## Additional information
Testing: Added exhaustive unit tests (tests/unit/simulation/test_flight_comparator.py) covering initialization, data conversion, metric calculation, and plotting logic.

### DIFF
--- a/docs/user/flight_comparator.rst
+++ b/docs/user/flight_comparator.rst
@@ -1,17 +1,23 @@
 Flight Comparator
 =================
 
-This example demonstrates how to use the RocketPy ``FlightComparator`` class.
-This class is designed to compare a RocketPy simulation against external data sources,
-such as:
+This example demonstrates how to use the RocketPy ``FlightComparator`` class to
+compare a Flight simulation against external data sources.
+
+Users must explicitly create a `FlightComparator` instance.
+
+
+This class is designed to compare a RocketPy Flight simulation against external
+data sources, such as:
 
 - Real flight data (avionics logs, altimeter CSVs)
 - Simulations from other software (OpenRocket, RASAero)
 - Theoretical models or manual calculations
 
-Unlike ``CompareFlights`` (which compares multiple RocketPy simulations), ``FlightComparator``
-specifically handles the challenge of aligning different time steps and calculating
-error metrics (RMSE, MAE, etc.) between a "Reference" simulation and "External" data.
+Unlike ``CompareFlights`` (which compares multiple RocketPy simulations),
+``FlightComparator`` specifically handles the challenge of aligning different
+time steps and calculating error metrics (RMSE, MAE, etc.) between a
+"Reference" simulation and "External" data.
 
 Importing classes
 -----------------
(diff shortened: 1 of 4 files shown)


docs/user/flight_comparator.rst
Gui-FernandesBR, line 1: are you planning to implement this .rst file? 

If you prefer, you can let it to next PR.

rocketpy/simulation/flight_comparator.py
```diff
@@ -0,0 +1,619 @@
...
+        print(f"\nExternal Data Sources: {list(self.data_sources.keys())}")
+
+        # Display key events comparison table
+        try:
+            events_results = self.compare_key_events()
+            print("\n" + self._format_key_events_table(events_results))
+            print(
+                "\nNote: Values marked with * are approximations (e.g., speed from vz only)"
+            )
+        except Exception as e:  # pylint: disable=broad-exception-caught
```
Gui-FernandesBR: This is not acceptable, please capture the specific exception you want to except. Otherwise just remove the try/exception block and let code fails. Bare exceptions make bugs silent

```suggestion
        except Exception as e:
```

rocketpy/simulation/flight_comparator.py
```diff
@@ -0,0 +1,619 @@
...
+                    f"Skipping '{key}' in '{label}'. Format not recognized. "
+                    "Expected RocketPy Function or (time, data) tuple."
+                )
+
+        self.data_sources[label] = processed_data
+        print(
+            f"Added data source '{label}' with variables: {list(processed_data.keys())}"
+        )
+
+    def compare(
```
Gui-FernandesBR: the `compare` method is quite long. Could your break it into smaller auxiliary methods?

rocketpy/simulation/flight_comparator.py
```diff
@@ -0,0 +1,619 @@
...
+            The reference RocketPy Flight object to compare against.
+
+        Returns
+        -------
+        None
+        """
+        self.flight = flight
+        self.data_sources = {}  # The format is {'Source Name': {'variable': Function}}
+
+    def add_data(self, label, data_dict):
```
Gui-FernandesBR: question: What do you thikn about acepting a `Flight` object here?

rocketpy/simulation/flight_comparator.py
```diff
@@ -0,0 +1,619 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+
+from rocketpy.mathutils import Function
+
+from ..plots.plot_helpers import show_or_save_fig
+
+
+# pylint: disable=too-many-public-methods
+# pylint: disable=too-many-statements
```
Gui-FernandesBR: I don't like the idea behind disabling pylint rules for an entire file. Please make it only when strictly necessary. If you still want to disable these rules, try to make it within specific functions, not the entire file.

```suggestion
```

tests/unit/simulation/test_flight_comparator.py
```diff
@@ -0,0 +1,511 @@
+"""Tests for the FlightComparator class.
+
+This module tests the FlightComparator class which compares RocketPy Flight
+simulations against external data sources such as flight logs, OpenRocket
+simulations, and RAS Aero simulations.
```
Gui-FernandesBR:
```suggestion
simulations, and RASAero simulations.
```

tests/unit/simulation/test_flight_comparator.py
```diff
@@ -0,0 +1,511 @@
...
+
+    comparator.trajectories_2d(plane="xz", filename=filename)
+
+    if filename:
+        assert os.path.exists(filename)
+        os.remove(filename)
+
+
+# Integration test
+def test_full_workflow(flight_calisto):
```
Gui-FernandesBR: the integration tests should be positioned in the `tests/integration`  folder. (https://docs.rocketpy.org/en/latest/development/testing.html)

Gui-FernandesBR (changes requested):
Great work, I'm happy with the results already.

Please solve all the commts, update the CHANGELOG file and then we will merge it when it's ready.

I don't think this one will take a long time to be finished or reviewed.
The implementation seems pretty clean.

rocketpy/simulation/flight_comparator.py
Monta120: Flight already imports FlightComparator(for the helper I added), so importing Flight back into the comparator would create a circular dependency. My plan is to extend add_data so it can also take a Flight instance (or Flight‑like object) via duck typing, e.g. add_data("Other Flight", other_flight) or add_data(other_flight), and automatically register its standard attributes (z, vz, x, y, speed, etc.) while still avoiding the circular import.

docs/user/flight_comparator.rst
Monta120: Yes! I just added the docs/user/flight_comparator.rst page with a full Calisto example, synthetic external data, and usage of Flight.compare, summary, compare_key_events, and trajectories_2d

rocketpy/simulation/flight_comparator.py
Monta120: Refactored compare() into smaller helpers: _process_time_range, _build_time_grid, _setup_compare_figure, _plot_reference_series, _plot_external_sources, and _finalize_compare_figure. The public API and plots remain the same, but the main method now mostly orchestrates these helpers and carries a local # pylint: disable=too-many-statements instead of disabling rules at file level.

tests/unit/simulation/test_flight_comparator.py
Monta120: Done

rocketpy/simulation/flight_comparator.py
Gui-FernandesBR: Should we consider removing the convenience wrapper from the Flight class? Alternatively, we could use a lazy import for FlightComparator inside the specific Flight method that uses it.

Monta120: make format keeps rewriting the lazy imports, so any inline # pylint: disable on those lines eventually gets moved or split, and pylint starts flagging import-outside-toplevel and the import cycle again. Since we’re not allowed to use module‑level pylint disables, there’s no stable way to keep those lazy imports and have pylint + formatting both clean. The only robust fix is to drop the Flight import in FlightComparator and treat Flight as a “Flight‑like” object via duck typing (hasattr(..., "t_final") and "z"), and then import FlightComparator normally at the top of flight.py instead of lazily inside. What do you think?

Monta120: Alternatively, we could remove the Flight.compare convenience wrapper and force users to always create a FlightComparator manually, which would avoid this import edge entirely. But the wrapper is a nice, discoverable entry point on Flight and keeps the common “compare this flight against data” workflow to a single call. I'll let you decide.

Gui-FernandesBR: I need to think baout it for a while. Gonna get back to you soon.

usually when these things happen I try to investigate whether there are similar behaviors throughout the project.

Gui-FernandesBR: @Monta120 I carefully reviewed it and I truly believe the best option here is for us to remove the "convinient wrapper" from the Flight class. This approach would remove the circular imports problem, with the only downside of requiring 1-2 extra lines from the user. I believe it is worth the changes.

tests/fixtures/surfaces/surface_fixtures.py
```diff
@@ -1,3 +1,4 @@
+# pylint: disable=cyclic-import
```
Gui-FernandesBR: Do you really, really need it?

```suggestion
```

Gui-FernandesBR (approved):

PR #911, opened 2025-12-07
ENH: acceleration data to trigger parachutes

### LINKED ISSUE #156: ENH: Acceleration data to trigger parachutes
Currently, custom functions used to trigger parachutes (events) only have access to **Pressure** and the **State Vector** (position, velocity, attitude, angular rates).

However, real-world flight computers (avionics) heavily rely on **Accelerometers (IMU)** to detect flight phases (e.g., Liftoff, Motor Burnout, Apogee).

  * *Example:* A user wants to deploy a drogue parachute only if the rocket detects it is under free-fall (approx 0g total acceleration) or has detected a specific deceleration curve.

Because RocketPy does not expose acceleration to the trigger, users cannot simulate these realistic avionics algorithms.

I would like to update the `Flight` class event handling logic to pass acceleration data to the parachute trigger callbacks.

The trigger signature should ideally support:

```python
def my_trigger(p, y, u_dot):
    # p = pressure
    # y = state vector [x, y, z, vx, vy, vz...]
    # u_dot = derivative [vx, vy, vz, ax, ay, az...]
    
    vertical_acceleration = u_dot[5] 
    return vertical_acceleration < -9.0 # Deploy if falling
```

**Implementation Details**

  * **Target File:** `rocketpy/flight/flight.py`
  * **Mechanism:** Inside the event wrapper methods (which interface with `scipy.integrate`), we need to explicitly call `self.u_dot(t, y)` to obtain the current derivatives.
  * **Performance Note:** Calling `u_dot` inside the event checker effectively doubles the physics load for event detection. This is acceptable for the gained accuracy, but we should document it.
  * **Noise:** Ideally, we should allow a way to inject "sensor noise" into this acceleration before passing it to the trigger, simulating a real MEMS accelerometer.

**Acceptance Criteria**

  - [ ] Update `Flight` class to calculate `u_dot` inside the event checking loop.
  - [ ] Pass the acceleration components (or the full `u_dot`) to the user-defined trigger function.
  - [ ] Add a tutorial example showing how to trigger a parachute based on "Motor Burnout" (sudden drop in acceleration).

**Additional Context**

  * Currently, the state vector `y` contains `[x, y, z, vx, vy, vz, ...]`.
  * The acceleration `[ax, ay, az]` is only available after computing the derivative `u_dot`.

ViniciusCMB:
# Pull Request: Add Acceleration-Based Parachute Triggers with IMU Sensor Simulation

## Pull request type

- [x] Code changes (features)
- [x] Code maintenance (refactoring, tests)
- [x] Documentation update

## Checklist

- [x] Tests for the changes have been added / updated
- [x] Docs have been reviewed and updated
- [x] Formatting (`black`) has been run locally
- [x] Lint and tests have been run locally
- [ ] CHANGELOG.md has been updated (if relevant)

## Why this matters

This is a critical feature for advanced users simulating realistic avionics.
Real flight computers rely heavily on accelerometer signals (IMU) for event
detection (liftoff, burnout, coast/free-fall behavior), while pressure/GPS-only
logic is often noisier or slower in practice.

Until now, parachute triggers primarily relied on pressure/height/state-vector
inputs. This made it difficult to reproduce avionics-style event logic that
depends on acceleration profiles.

## What this PR changes

This PR adds first-class support for acceleration-aware trigger callables by
passing state derivatives (`u_dot`) to user-defined trigger functions.

### 1) Trigger callable signatures support `u_dot`

In `rocketpy/rocket/parachute.py`, callable dispatch supports:

- `3 args`: `(pressure, height, state_vector)`
- `4 args`: `(pressure, height, state_vector, u_dot)` or `(…, sensors)`
- `5 args`: `(pressure, height, state_vector, sensors, u_dot)`

Dispatch is deterministic, and trigger wrappers expose metadata
(`_expects_udot`, `_expects_sensors`) used by `Flight`.

### 2) `Flight` computes `u_dot(t, y)` inside trigger evaluation when needed

In `rocketpy/simulation/flight.py`, trigger checks consistently route through
`_evaluate_parachute_trigger(...)`, and `u_dot` is computed only when required
by trigger metadata.

This follows the architecture discussed in Issue #156: acceleration is not in
the solver state vector and must be obtained from derivative evaluation at the
event-check instant.

### 3) Docs now include acceleration-based trigger examples

`docs/user/parachute_triggers.rst` was updated with custom trigger examples for:

- motor burnout detection
- acceleration-guided apogee/coast logic
- free-fall detection
- liftoff detection

Thresholds are intentionally mission-specific and configurable by users.

### 4) Final API scope from review

Per review direction, this PR keeps the API focused on custom callables and
does not introduce acceleration-specific built-in trigger strings.

## Acceptance criteria coverage (Issue #156)

- [x] Update `Flight` to calculate `u_dot` inside event checking flow
- [x] Pass acceleration data (full `u_dot`) to user-defined triggers
- [x] Add tutorial/docs example for burnout-style acceleration triggering

## Files changed

- `rocketpy/rocket/parachute.py`
  - callable signature detection/dispatch and trigger metadata
  - `u_dot`/sensor-aware trigger wrapper behavior
- `rocketpy/simulation/flight.py`
  - consistent trigger evaluation path with optional `u_dot` computation
  - removal of duplicated fallback signature handling
- `docs/user/parachute_triggers.rst`
  - acceleration-based custom trigger examples and usage guidance
- `tests/unit/test_parachute_triggers.py`
  - coverage for signature behavior and `u_dot` delivery
- `tests/unit/test_parachute_trigger_acceleration.py`
  - removed/merged to align with final callable-focused API

## Breaking change

- [ ] Yes
- [x] No

No breaking change in trigger usage: existing numeric, `"apogee"`, and legacy
3-argument callables remain supported.

## Validation performed

- `black` run on modified Python files
- `pytest tests/unit/test_parachute_triggers.py -q` passing
- additional local lint/test checks executed during review updates

## Related issue

- RocketPy#156

### DIFF
--- a/docs/user/parachute_triggers.rst
+++ b/docs/user/parachute_triggers.rst
@@ -1,0 +1,279 @@
+Acceleration-Based Parachute Triggers
+======================================
+
+RocketPy lets parachute trigger functions access the state derivative
+``u_dot`` (which holds the accelerations at indices ``[3:6]``) in addition to
+pressure, height and the state vector. This enables avionics-style deployment
+logic that mimics how real flight computers use accelerometer (IMU) data to
+detect flight phases such as burnout, free-fall or liftoff.
+
+Overview
+--------
+
+Built-in string and numeric triggers rely on altitude and vertical velocity.
+By writing a **custom trigger function**, you can additionally use acceleration
+to implement mission-specific logic, for example:
+
+- Motor burnout detection (sudden drop in acceleration)
+- Apogee detection combining near-zero velocity with downward acceleration
+- Free-fall / ballistic coast detection (low total acceleration)
+- Liftoff detection (high total acceleration)
+
+For realistic noisy measurements, attach an :doc:`Accelerometer sensor
+</reference/classes/sensors/index>` to the rocket and read it inside the
+trigger instead of feeding the ideal ``u_dot`` directly.
+
+Trigger function signatures
+---------------------------
+
+A custom trigger callable may take **3, 4, or 5** arguments. RocketPy detects
+the signature automatically and only computes ``u_dot`` when a trigger asks for
+it (so legacy triggers pay no performance cost):
+
+- ``(pressure, height, state_vector)`` — the classic signature.
+- ``(pressure, height, state_vector, u_dot)`` — name the 4th argument
+  ``u_dot`` (or ``udot``/``acc``/``acceleration``) to receive the derivative;
+  any other name receives the ``sensors`` list instead.
+- ``(pressure, height, state_vector, sensors, u_dot)`` — receive both.
+
+``state_vector`` is ``[x, y, z, vx, vy, vz, e0, e1, e2, e3, w1, w2, w3]`` and
+``u_dot`` is ``[vx, vy, vz, ax, ay, az, ...]``.
+
+Built-in apogee trigger
+-----------------------
+
+Deploys when the rocket starts descending (vertical velocity becomes negative):
+
+.. code-block:: python
+
+    rocket.add_parachute(
+        name="Main",
+        cd_s=10.0,
+        trigger="apogee",
+        sampling_rate=100,
+        lag=0.5,
+    )
+
+Numeric altitude trigger
+------------------------
+
+Pass a number to deploy at a fixed height above ground level while descending:
+
+.. code-block:: python
+
+    rocket.add_parachute(
+        name="Main",
+        cd_s=10.0,
+        trigger=400,  # meters above ground level
+        sampling_rate=100,
+        lag=0.5,
+    )
+
+Custom trigger: motor burnout
+-----------------------------
+
+Burnout is highly mission-dependent, so it is best expressed as a custom
+trigger with user-defined thresholds.
+
+Logic: detect a drop in vertical or total acceleration once the rocket is above
+a minimum height and still ascending.
+
+.. code-block:: python
+
+    def burnout_trigger_factory(
+        min_height=5.0,
+        min_vz=0.5,
+        az_threshold=-8.0,
+        total_acc_threshold=2.0,
+    ):
+        def burnout_trigger(_pressure, height, state_vector, u_dot):
+            ax, ay, az = u_dot[3], u_dot[4], u_dot[5]
+            total_acc = (ax**2 + ay**2 + az**2) ** 0.5
+            vz = state_vector[5]
+            if height < min_height or vz <= min_vz:
+                return False
+            return az < az_threshold or total_acc < total_acc_threshold
+
+        return burnout_trigger
+
+Attach it to a rocket:
+
+.. code-block:: python
+
+    rocket.add_parachute(
+        name="Drogue",
+        cd_s=1.0,
+        trigger=burnout_trigger_factory(
+            min_height=10.0,
+            min_vz=2.0,
+            az_threshold=-10.0,
+            total_acc_threshold=3.0,
+        ),
+        sampling_rate=100,
+        lag=1.5,
+    )
+
+Custom trigger: apogee by acceleration
+--------------------------------------
+
+Logic: near-zero vertical velocity together with downward acceleration.
+
+.. code-block:: python
+
+    def apogee_acc_trigger(_pressure, _height, state_vector, u_dot):
+        vz = state_vector[5]
+        az = u_dot[5]
+        return abs(vz) < 1.0 and az < -0.1
+
+.. code-block:: python
+
+    rocket.add_parachute(
+        name="Main",
+        cd_s=10.0,
+        trigger=apogee_acc_trigger,
+        sampling_rate=100,
+        lag=0.5,
+    )
+
+Custom trigger: free-fall
+-------------------------
+
+Logic: low total acceleration while descending above a small height.
+
+.. code-block:: python
+
+    def freefall_trigger(_pressure, height, state_vector, u_dot):
+        ax, ay, az = u_dot[3], u_dot[4], u_dot[5]
+        total_acc = (ax**2 + ay**2 + az**2) ** 0.5
+        vz = state_vector[5]
+        return height > 5.0 and vz < -0.2 and total_acc < 11.5
+
+.. code-block:: python
+
+    rocket.add_parachute(
+        name="Drogue",
+        cd_s=1.0,
+        trigger=freefall_trigger,
+        sampling_rate=100,
+        lag=1.5,
+    )
+
+Custom trigger: liftoff
+-----------------------
+
+Logic: detect motor ignition by high total acceleration.
+
+.. code-block:: python
+
+    def liftoff_trigger(_pressure, _height, _state_vector, u_dot):
+        ax, ay, az = u_dot[3], u_dot[4], u_dot[5]
+        total_acc = (ax**2 + ay**2 + az**2) ** 0.5
+        return total_acc > 15.0
+
+.. code-block:: python
+
+    rocket.add_parachute(
+        name="Lift",
+        cd_s=0.5,
+        trigger=liftoff_trigger,
+        sampling_rate=100,
+        lag=0.1,
+    )
+
+Custom trigger: using sensor measurements
+-----------------------------------------
+
+A 5-argument trigger receives both the ``sensors`` list and ``u_dot``, so you
+can cross-check a noisy accelerometer reading against the ideal derivative.
+
+.. code-block:: python
+
+    def advanced_trigger(_pressure, _height, _state_vector, sensors, u_dot):
+        if not sensors:
+            return False
+        acc_reading = sensors[0].measurement
+        if acc_reading is None or len(acc_reading) < 3:
+            return False
+        meas_az = acc_reading[2]
+        az = u_dot[5]
+        return az < -5.0 and meas_az < -5.0
+
+.. code-block:: python
+
+    rocket.add_parachute(
+        name="Advanced",
+        cd_s=1.5,
+        trigger=advanced_trigger,
+        sampling_rate=100,
+    )
+
+.. note::
+
+    For realistic IMU behavior, attach a RocketPy sensor with its own noise
+    model and read it inside the trigger via ``sensors``, instead of relying on
+    the ideal ``u_dot``. See the :doc:`Sensor Classes
+    </reference/classes/sensors/index>` for available sensors.
+
+Full example: dual deployment
+-----------------------------
+
+In RocketPy only one parachute is active at a time, so a dual-deploy avionics
+can be reproduced with two custom triggers — a drogue at burnout and a main at
+a lower altitude:
+
+.. code-block:: python
+
+    from rocketpy import Rocket, Flight, Environment
+
+    # Environment and rocket setup
+    env = Environment(latitude=32.99, longitude=-106.97, elevation=1400)
+    env.set_atmospheric_model(type="standard_atmosphere")
+
+    rocket = Rocket(...)  # configure your rocket (motor, fins, etc.)
+
+    # Drogue: deploy shortly after burnout (acceleration drop while ascending)
+    def drogue_burnout_trigger(_pressure, height, state_vector, u_dot):
+        az = u_dot[5]
+        vz = state_vector[5]
+        return height > 10 and vz > 1 and az < -8.0
+
+    rocket.add_parachute(
+        name="Drogue",
+        cd_s=1.0,
+        trigger=drogue_burnout_trigger,
+        sampling_rate=100,
+        lag=1.5,
+        noise=(0, 8.3, 0.5),  # pressure-signal noise
+    )
+
+    # Main: deploy below 800 m while descending
+    def main_deploy_trigger(_pressure, height, state_vector, u_dot):
+        vz = state_vector[5]
+        az = u_dot[5]
+        return height < 800 and vz < -5 and az > -15
+
+    rocket.add_parachute(
+        name="Main",
+        cd_s=10.0,
+        trigger=main_deploy_trigger,
+        sampling_rate=100,
+        lag=0.5,
+        noise=(0, 8.3, 0.5),
+    )
+
+    # Flight simulation
+    flight = Flight(
+        rocket=rocket,
+        environment=env,
+        rail_length=5.2,
+        inclination=85,
+        heading=0,
+    )
+    flight.all_info()
+
+See Also
+--------
+
+- :doc:`Parachute Class Reference </reference/classes/Parachute>`
+- :doc:`Flight Simulation </user/flight>`
+- :doc:`Sensors </notebooks/sensors>`
--- a/rocketpy/rocket/parachute.py
+++ b/rocketpy/rocket/parachute.py
@@ -24,25 +24,34 @@ class Parachute:
         This parameter defines the trigger condition for the parachute ejection
         system. It can be one of the following:
 
-        - A callable function that takes three arguments:
-          1. Freestream pressure in pascals.
-          2. Height in meters above ground level.
-          3. The state vector of the simulation, which is defined as:
+        - A callable function that can take 3, 4, or 5 arguments:
 
-             `[x, y, z, vx, vy, vz, e0, e1, e2, e3, wx, wy, wz]`.
+          **3 arguments**:
+            1. Freestream pressure in pascals.
+            2. Height in meters above ground level.
+            3. The state vector: ``[x, y, z, vx, vy, vz, e0, e1, e2, e3, wx, wy, wz]``
 
-          4. A list of sensors that are attached to the rocket. The most recent
-             measurements of the sensors are provided with the
-             ``sensor.measurement`` attribute. The sensors are listed in the same
-             order as they are added to the rocket.
+          **4 arguments** (sensors OR acceleration):
+            1. Freestream pressure in pascals.
+            2. Height in meters above ground level.
+            3. The state vector.
+            4. Either:
+               - ``sensors``: List of sensor objects attached to the rocket, OR
+               - ``u_dot``: State derivative including accelerations at indices [3:6]
 
-          The function should return ``True`` if the parachute ejection system
-          should be triggered and False otherwise. The function will be called
-          according to the specified sampling rate.
+          **5 arguments** (sensors AND acceleration):
+            1. Freestream pressure in pascals.
+            2. Height in meters above ground level.
+            3. The state vector.
+            4. ``sensors``: List of sensor objects.
+            5. ``u_dot``: State derivative with accelerations ``[vx, vy, vz, ax, ay, az, ...]``
+
+          The function should return ``True`` to trigger deployment, ``False`` otherwise.
+          The function will be called according to the specified sampling rate.
 
         - A float value, representing an absolute height in meters. In this
           case, the parachute will be ejected when the rocket reaches this height
-          above ground level.
+          above ground level while descending.
 
         - The string "apogee" which triggers the parachute at apogee, i.e.,
           when the rocket reaches its highest point and starts descending.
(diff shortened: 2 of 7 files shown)


rocketpy/rocket/parachute.py
Gui-FernandesBR, line 1: do you really need try/except blocks here?

rocketpy/rocket/parachute.py
```diff
@@ -24,41 +164,56 @@ class Parachute:
...
         system. It can be one of the following:
 
-        - A callable function that takes three arguments:
-          1. Freestream pressure in pascals.
-          2. Height in meters above ground level.
-          3. The state vector of the simulation, which is defined as:
+        - A callable function that can take 3, 4, or 5 arguments:
 
-             `[x, y, z, vx, vy, vz, e0, e1, e2, e3, wx, wy, wz]`.
+          **3 arguments** (legacy):
```
MateusStano: This isn't legacy, it is still a valid option
```suggestion
          **3 arguments**:
```

rocketpy/rocket/parachute.py
```diff
@@ -8,6 +8,146 @@
...
+
+        if height < 5.0 or vz <= 0.5:
+            return False
+
+        # Burnout detected when:
+        # 1. Vertical acceleration becomes very negative (end of thrust phase)
+        # 2. OR total acceleration drops below 2.0 m/s² (coasting detected)
+        return az < -8.0 or total_acc < 2.0
+    except (ValueError, TypeError, IndexError):
+        return False
```
MateusStano: I don't know how useful a function to deploy a parachute right after burnout really is, but I still have a few issues with this:

1. There are a few hard-coded values that should be user defined
2. The try except here is doing basically nothing
3. No need to check Nan of Inf, those values should never be returned in either u_dot nor state_vector

In the end I think this isn't really useful to be included in the code, but it would be a cool example to include in the docs somewhere

rocketpy/rocket/parachute.py
```diff
@@ -8,6 +8,146 @@
...
+        if not all(np.isfinite([ax, ay, az])):
+            return False
+
+        total_acc = np.sqrt(ax * ax + ay * ay + az * az)
+        if not np.isfinite(total_acc):
+            return False
+
+        # Additional safety: ignore spurious low-accel readings at t~0 by
+        # requiring the rocket to be above a small altitude and still ascending
+        vz = float(state_vector[5]) if len(state_vector) > 5 else 0
```
MateusStano: state_vector always has the same length. No need to check it

rocketpy/rocket/parachute.py
```diff
@@ -8,6 +8,146 @@
...
+        ay = float(u_dot[4])
+        az = float(u_dot[5])
+
+        # Defensive checks for NaN/Inf
+        if not all(np.isfinite([ax, ay, az])):
+            return False
+
+        total_acc = np.sqrt(ax * ax + ay * ay + az * az)
+        if not np.isfinite(total_acc):
+            return False
```
MateusStano: Acceleration will never be infinite

rocketpy/rocket/parachute.py
```diff
@@ -8,6 +8,146 @@
...
+def detect_motor_burnout(_pressure, height, state_vector, u_dot):
+    """Detect motor burnout by sudden drop in acceleration.
+
+    Returns True when vertical acceleration becomes significantly negative
+    (indicating end of propulsion phase) OR when total acceleration drops below
+    a threshold indicating coasting/free-fall has begun.
+    """
+    try:
+        if u_dot is None or len(u_dot) < 6:
+            return False
```
MateusStano: this will never happen. No need for this check

rocketpy/rocket/parachute.py
```diff
@@ -8,6 +8,146 @@
...
+
+        vz = float(state_vector[5])
+        az = float(u_dot[5])
+        if not all(np.isfinite([vz, az])):
+            return False
+
+        # Slightly more permissive thresholds to avoid spurious misses
+        return abs(vz) < 1.0 and az < -0.1
+    except (ValueError, TypeError, IndexError):
+        return False
```
MateusStano: Again, same comments as the ones I made for `detect_motor_burnout`.

- A lot of unecessary checks
- I believe this is more useful as an example in the docs than as standard in the code

rocketpy/rocket/parachute.py
```diff
@@ -8,6 +8,146 @@
...
+        if not np.isfinite(vz):
+            return False
+
+        if height < 5.0 or vz >= -0.2:
+            return False
+
+        # More sensitive threshold: detect free-fall at lower acceleration
+        return total_acc < 11.5
+    except (ValueError, TypeError, IndexError):
+        return False
```
MateusStano: Again, same comments as the ones I made for `detect_motor_burnout`.

- A lot of unecessary checks
- I believe this is more useful as an example in the docs than as standard in the code

rocketpy/rocket/parachute.py
```diff
@@ -8,6 +8,146 @@
...
+        if not all(np.isfinite([ax, ay, az])):
+            return False
+
+        total_acc = np.sqrt(ax * ax + ay * ay + az * az)
+        if not np.isfinite(total_acc):
+            return False
+
+        return total_acc > 15.0
+    except (ValueError, TypeError, IndexError):
+        return False
```
MateusStano: Again, same comments as the ones I made for `detect_motor_burnout`.

- A lot of unecessary checks
- I believe this is more useful as an example in the docs than as standard in the code

rocketpy/simulation/flight.py
```diff
@@ -96,6 +97,12 @@ class Flight:
         trigger function evaluation points and then interpolation is used to
         calculate them and feed the triggers. Can greatly improve run
         time in some cases.
+    Note
+    ----
+    Calculating the derivative `u_dot` inside parachute trigger checks will
+    cause additional calls to the equations of motion (extra physics
+    evaluations). This increases CPU cost but enables realistic avionics
+    algorithms that rely on accelerometer data.
```
MateusStano: This note makes more sense if it is on the .rst docs instead of the docstrings.

rocketpy/simulation/flight.py
```diff
@@ -503,6 +510,7 @@ def __init__(  # pylint: disable=too-many-arguments,too-many-statements
         name="Flight",
         equations_of_motion="standard",
         ode_solver="LSODA",
+        acceleration_noise_function=None,
         simulation_mode="6 DOF",
```
MateusStano, line 506: What is this parameter supposed to be doing?

If it is important, please move it somewhere else. It should not be in Flight class constructor

Also, if this is only for noise, I'd suggest just removing because the Accelerometer Sensor already handles noise generation

rocketpy/simulation/flight.py
```diff
@@ -1457,6 +1474,80 @@ def __calculate_and_save_pressure_signals(self, parachute, t, z):
...
+                    expects_sensors = any(p.lower() == "sensors" for p in params[3:])
+                except (ValueError, TypeError):
+                    expects_udot = False
+                    expects_sensors = True
+
+        # Compute u_dot only if needed (performance optimization)
+        u_dot = None
+        if expects_udot:
+            try:
+                u_dot = np.array(derivative_func(t, y), dtype=float)
```
MateusStano: no need to convert to array

rocketpy/simulation/flight.py
```diff
@@ -1457,6 +1474,80 @@ def __calculate_and_save_pressure_signals(self, parachute, t, z):
...
+                        noise = np.asarray(self.acceleration_noise_function())
+                        if noise.size == 3:
+                            # u_dot layout: [vx, vy, vz, ax, ay, az, ...]
+                            u_dot[3:6] = u_dot[3:6] + noise
+                    except (ValueError, TypeError):
+                        # ignore noise errors and continue
+                        pass
+            except (ValueError, TypeError, RuntimeError):
+                # If u_dot computation fails, leave as None
+                u_dot = None
```
MateusStano: u_dot computation should never fail, if it does the simulation is broken anyway. Please remove this try except

rocketpy/simulation/flight.py
```diff
@@ -1457,6 +1474,80 @@ def __calculate_and_save_pressure_signals(self, parachute, t, z):
...
+            if callable(trig_original):
+                try:
+                    sig = inspect.signature(trig_original)
+                    params = list(sig.parameters.keys())
+                    acc_names = {"u_dot", "udot", "acc", "acceleration"}
+                    expects_udot = any(p.lower() in acc_names for p in params[3:])
+                    expects_sensors = any(p.lower() == "sensors" for p in params[3:])
+                except (ValueError, TypeError):
+                    expects_udot = False
+                    expects_sensors = True
```
MateusStano: expects_udot and expects sensors should always be defined in parachutes.py. There should be no need to perform extra checks in the Flight class.

rocketpy/simulation/flight.py
```diff
@@ -947,6 +964,7 @@ def __check_and_handle_parachute_triggers(
                 height_above_ground_level,
                 self.y_sol,
                 self.sensors,
+                None,  # u_dot not computed in non-overshoot path
```
MateusStano: Why not? u_dot needs to be here to make the triggerfunc work properly in case someone uses the new udot param

rocketpy/rocket/parachute.py
```diff
@@ -211,59 +366,121 @@ def __init__(
...
+                try:
+                    return fn(p, h, y, sensors, u_dot)
+                except TypeError:
+                    try:
+                        return fn(p, h, y, u_dot)
+                    except TypeError:
+                        try:
+                            return fn(p, h, y, sensors)
+                        except TypeError:
+                            return fn(p, h, y)
```
MateusStano: This is not necessary. We need to be able to detect exactly what the user `fn` function needs in terms of parameters

MateusStano (changes requested):

ViniciusCMB:
@Gui-FernandesBR, @MateusStano, I was really focused on the original issue and ended up implementing the "liftoff", "burnout", and other triggers as built-ins, even though it doesn't make much sense for the parachute (I believe it might be useful for other deployables, though). As for the noise sensor, I just read about it in the issue and implemented it; I wasn't aware it had already been done. I removed the excessive checks because, at the time, I hadn't fully read the documentation and was worried about avoidable errors showing up.

rocketpy/rocket/parachute.py
```diff
@@ -8,6 +8,22 @@
...
+
+    def trigger(_pressure, height, state_vector, _u_dot=None):
+        vz = float(state_vector[5])
+        if require_descent:
+            return (height <= target_altitude) and (vz < 0)
+        return height <= target_altitude
+
+    return trigger
+
+
```
MateusStano:
```suggestion

```

Please remove this function. It is not used currently

rocketpy/rocket/parachute.py
```diff
@@ -24,41 +40,52 @@ class Parachute:
...
 
         - A float value, representing an absolute height in meters. In this
           case, the parachute will be ejected when the rocket reaches this height
-          above ground level.
+          above ground level while descending.
 
-        - The string "apogee" which triggers the parachute at apogee, i.e.,
-          when the rocket reaches its highest point and starts descending.
+        - A string for built-in triggers:
+          - ``"apogee"``: Apogee detection (velocity-based)
```
MateusStano: I prefer the previous version of this description. Can you remove these changes?

rocketpy/rocket/parachute.py
```diff
@@ -24,41 +40,52 @@ class Parachute:
...
+        ground level in meters, the state vector, sensors list, and u_dot.
+        Returns ``True`` if the parachute ejection system should be triggered
         and ``False`` otherwise.
 
-        .. note:
+        .. note::
 
             The function will be called according to the sampling rate specified.
+            For performance, ``u_dot`` is only computed if the trigger signature
+            indicates it is needed.
```
MateusStano, line 71: No need to explain backend code in this part of the docstring. I suggent removing it

```suggestion
        .. note::

            The function will be called according to the sampling rate specified.

```

rocketpy/simulation/flight.py
```diff
@@ -587,6 +587,19 @@ def __init__(  # pylint: disable=too-many-arguments,too-many-statements
...
+            Proportionality coefficient (rate coefficient) for the alignment rate of the rocket's body axis
+            with the relative wind direction in 3-DOF simulations, in rad/s. The actual angular velocity
+            applied to align the rocket is calculated as ``weathercock_coeff * sin(angle)``, where ``angle``
+            is the angle between the rocket's axis and the wind direction. A higher value means faster alignment
+            (quasi-static weathercocking). This parameter is only used when simulation_mode is '3 DOF'.
+            Default is 0.0 to mimic a pure 3-DOF simulation without any weathercocking (fixed attitude).
+            Set to a positive value to enable quasi-static weathercocking behaviour.
+
+
         Returns
```
MateusStano, line 593: Please remove the weathercock_coeff, this seems to be a merge error

```suggestion
        simulation_mode : str, optional
            Simulation mode to use. Can be "6 DOF" for 6 degrees of freedom or
            "3 DOF" for 3 degrees of freedom. Default is "6 DOF".

        Returns
```

rocketpy/rocket/parachute.py
```diff
@@ -273,54 +300,95 @@ def __init_noise(self, noise):
...
-        # Case 3: The parachute is deployed at apogee
-        elif trigger.lower() == "apogee":
-            # The parachute is deployed at apogee
-            def triggerfunc(p, h, y, sensors):
-                # p = pressure considering parachute noise signal
-                # h = height above ground level considering parachute noise signal
-                # y = [x, y, z, vx, vy, vz, e0, e1, e2, e3, w1, w2, w3]
+        # Special case: "apogee" (legacy support)
+        if isinstance(trigger, str) and trigger.lower() == "apogee":
+
```
MateusStano: This is not legacy support, it is a valid feature

```suggestion
        # Special case: "apogee"
        if isinstance(trigger, str) and trigger.lower() == "apogee":

```

tests/unit/test_parachute_triggers.py
```diff
@@ -0,0 +1,110 @@
...
+        pressure=0.0,
+        height=0.0,
+        y=np.zeros(13),
+        sensors=[],
+        derivative_func=derivative_func,
+        t=0.0,
+    )
+
+    assert res is True
+    assert called.get("ok", False) is True
```
MateusStano: Change the name please, these are not "legacy"

docs/user/parachute_triggers.rst
```diff
@@ -0,0 +1,307 @@
...
+        trigger=advanced_trigger,
+        sampling_rate=100
+    )
+
+Sensor Noise
+------------
+
+For realistic IMU behavior, use RocketPy sensors with their own noise models.
+Parachute trigger functions can receive ``sensors`` and use those measurements
+directly instead of injecting noise in the flight solver.
```
MateusStano: Can you make this a `.. note::`? I think it would look cleaner in the page

docs/user/parachute_triggers.rst
```diff
@@ -0,0 +1,307 @@
...
+    main = Parachute(
+        name="Main",
+        cd_s=10.0,
+        trigger="apogee",
+        sampling_rate=100,
+        lag=0.5
+    )
+
+**Detection criteria:**
+- Vertical velocity becomes negative (descent starts)
```
MateusStano: This is missing the case when trigger is a int,float (`trigger=400`)

docs/user/parachute_triggers.rst
MateusStano: Also, please provide a link to the sensors doc pages

docs/user/parachute_triggers.rst
```diff
@@ -0,0 +1,307 @@
...
+- 5 parameters: Both sensors and acceleration provided
+
+Best Practices
+--------------
+
+1. **Choose appropriate sampling rates**: 50-200 Hz is typical for flight computers
+2. **Add realistic noise**: Real IMUs have noise; simulate it for validation
+3. **Test edge cases**: Verify triggers work at low altitudes, high speeds, etc.
+4. **Use robust custom logic**: Add mission-specific guards and thresholds
+5. **Document custom triggers**: Include detection criteria in docstrings
```
MateusStano: I suggest to remove these sections, they do not add anythin of value in terms of usability

docs/user/parachute_triggers.rst
```diff
@@ -0,0 +1,307 @@
...
+        trigger=burnout_trigger_factory(
+            min_height=10.0,
+            min_vz=2.0,
+            az_threshold=-10.0,
+            total_acc_threshold=3.0,
+        ),
+        sampling_rate=100,
+        lag=1.5,
+    )
+
```
MateusStano: Please add some description of what is the logic of the trigger here. For example, "to check if burn out state has been achieved, the accelerations are checked....

docs/user/parachute_triggers.rst
```diff
@@ -0,0 +1,307 @@
...
+Apogee by Acceleration (Custom Example)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: python
+
+    def apogee_acc_trigger(_pressure, _height, state_vector, u_dot):
+        vz = state_vector[5]
+        az = u_dot[5]
+        return abs(vz) < 1.0 and az < -0.1
```
MateusStano: Please add some description of what is the logic of the trigger here.

docs/user/parachute_triggers.rst
```diff
@@ -0,0 +1,307 @@
...
+Free-fall (Custom Example)
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: python
+
+    def freefall_trigger(_pressure, height, state_vector, u_dot):
+        ax, ay, az = u_dot[3], u_dot[4], u_dot[5]
+        total_acc = (ax**2 + ay**2 + az**2) ** 0.5
+        vz = state_vector[5]
+        return height > 5.0 and vz < -0.2 and total_acc < 11.5
```
MateusStano: Please add some description of what is the logic of the trigger here.

docs/user/parachute_triggers.rst
```diff
@@ -0,0 +1,307 @@
...
+        return total_acc < 5.0 and vz < -10.0
+
+    # Use custom trigger
+    parachute = Parachute(
+        name="Custom",
+        cd_s=2.0,
+        trigger=custom_trigger,
+        sampling_rate=100,
+        lag=1.0
+    )
```
MateusStano: There are two trigger functions defined here... One of them seems to be unused

MateusStano: Also, I don´t really understand the what this trigger is doing that is different from the previous ones?

docs/user/parachute_triggers.rst
```diff
@@ -0,0 +1,307 @@
...
+        # Combine sensor and acceleration logic
+        return az < -5.0 and imu_reading > threshold
+
+    parachute = Parachute(
+        name="Advanced",
+        cd_s=1.5,
+        trigger=advanced_trigger,
+        sampling_rate=100
+    )
+
```
MateusStano: Please make an example more in lines with what would be like using a rocketpy Accelerometer object. We don´t have an IMU sensor object
Also the `.measurement` result will be a list/vector, so please make it coherent. Users will definetly copy and paste these examples into their own code, so it should be as closely to working when pasting as possible

docs/user/parachute_triggers.rst
```diff
@@ -0,0 +1,307 @@
...
+    flight = Flight(
+        rocket=my_rocket,
+        environment=env,
+        rail_length=5.2,
+        inclination=85,
+        heading=0,
+    )
+
+    flight.all_info()
+
```
MateusStano: Can you change this section to be about simulation Drogue parachutes? I think it would make more sense, and make this docs page more complete

We simulate drogue chutes by simply considering the disrefeed chute the equivalent to main here, and the refeed chute the drogue, since in rocketpy only one parachute is considered at a time

docs/user/parachute_triggers.rst
ViniciusCMB: > There are two trigger functions defined here... One of them seems to be unused

My bad, I missed that. Fixed it!

Gui-FernandesBR (approved):
LGTM

PR #920, opened 2025-12-10
ENH: Implement 3-DOF Single Rail Button Flight Phase (Tip-off Analysis)

### LINKED ISSUE #28: ENH: Implement 3-DOF Single Rail Button Flight Phase (Tip-off Analysis)
Currently, the `Flight` simulation transitions instantly from **Phase 1** (Two buttons on rail, 1-DOF linear motion) to **Phase 3** (Free flight, 6-DOF).

It skips an intermediate **Phase 2**: The "Single Rail Button" phase.
This occurs when the upper rail button has left the rail guide, but the lower rail button is still attached. During this split-second interval, the rocket is free to rotate (pitch/yaw) around the lower button due to wind or thrust misalignment, while still being constrained translationally along the rail axis.

Ignoring this phase ignores "Rod Whip" or "Tip-off" effects, which can significantly alter the initial conditions ($w_0$, $\theta_0$) of the free flight, affecting dispersion accuracy.

I would like to implement the equations of motion for this specific phase in the `Flight` class.
The structure is already prepared in the code as `udot_rail2`, but the method is currently empty/unimplemented.

**Implementation Details**
* **Target Method:** `Flight.udot_rail2(t, u)` in `rocketpy/flight/flight.py`.
* **Physics Model:**
    * **Constraints:** The rocket is pinned at the lower rail button location (sliding along the rail vector).
    * **Degrees of Freedom (3-DOF):**
        1.  Translation along the rail ($v_{rail}$).
        2.  Rotation around the local Y-axis (Pitch).
        3.  Rotation around the local Z-axis (Yaw).
    * **Forces:** Thrust, Gravity, Aerodynamics (Lift/Drag), and the *Reaction Force* at the lower button (which acts as a pivot).
* **Transition Logic (in `Flight` main loop):**
    * **Start of Phase 2:** When distance traveled > `Flight.effective_1rl`.
    * **End of Phase 2:** When distance traveled > `Flight.effective_2rl`.

**Helpful Properties**
The `Flight` class already has the necessary geometric properties to determine the start and end of this phase:
* `Flight.effective_1rl`: Distance to clear the first button.
* `Flight.effective_2rl`: Distance to clear the second button.

**Acceptance Criteria**
- [ ] Implement the dynamics in `udot_rail2` to return the derivative of the state vector $[v, \omega, ...]$.
- [ ] Update the integration loop in `Flight` to detect when `z > effective_1rl` and switch the integration method from `udot_rail1` to `udot_rail2`.
- [ ] Ensure `udot_rail2` transitions smoothly to `u_dot` (free flight) once `z > effective_2rl`.
- [ ] Verify that wind acting during this phase causes the rocket to "whip" or turn into the wind before fully leaving the rail.

**Additional Context**
* This is often referred to as "Tip-off" analysis.
* *Warning:* The equations of motion for a body constrained to a line at a specific offset point (lower button) involving rotation are more complex than standard free-body dynamics. Lagragian mechanics might be the easiest way to derive these equations.

Rafit345:
## Pull request type

## Checklist

- [x] Lint (`ruff check`, `ruff format --check`, `pylint`) has passed locally

## Current behavior

Refs #28.

The rail phase ends the moment the *upper* rail button reaches the end of the rail,
at `effective_1rl`, and the simulation jumps straight from 1-DOF rail motion to
6-DOF free flight. The interval in which the rocket is still guided by the *lower*
button alone — the **tip-off** phase — is not modelled, so the rocket begins free
flight with no angular rate and with exactly the rail's attitude.

## New behavior

Adds `Flight.udot_rail2`, an intermediate 3-DOF flight phase covering the interval
between the two rail buttons leaving the rail:

```
udot_rail1        for d <  effective_1rl   (both buttons engaged, 1 DOF)
udot_rail2        for d in [effective_1rl, effective_2rl)  (lower button only, 3 DOF)
u_dot_generalized for d >= effective_2rl   (free flight, 6 DOF)
```

The phase models a **constrained rigid body**: the lower button slides along the fixed
inertial rail axis and roll is suppressed, leaving translation along the rail, pitch
and yaw. It reuses `u_dot_generalized` for the free solution and adds a reaction
wrench — a normal force at the button (perpendicular to the rail, 2 DOF) plus a roll
reaction moment (1 DOF) — solved from a **3x3 linear system** built from the three
constraints: the button's perpendicular acceleration vanishes (2) and the roll angular
acceleration vanishes (1). Aerodynamics, thrust, body-frame gravity and the evolving
inertia tensor therefore come along by construction.

The derivation is documented in `docs/technical/tip_off.rst`.

Enabled with `Flight(..., use_udot_rail2=True)`.

## Breaking change

- [ ] Yes
- [x] No

`use_udot_rail2` defaults to `False`, and a disabled run is **bit-for-bit identical**
to `develop` — verified by comparing the full final state against a clean `develop`
checkout, not just the apogee.

## Additional information

**Validation**

- Constraint satisfaction: the button's perpendicular offset from the rail stays at
  ~1e-13 m; roll rate and roll acceleration stay at zero.
- Physics: with no wind the nose pitches down (the center of mass is ahead of the
  pivot); with a crosswind the rocket weathercocks into the wind before fully leaving
  the rail, which is the acceptance criterion in #28. Apogee shifts about -2 m on
  Calisto.
- 1960 unit and simulation-integration tests pass locally; all CI checks pass.

**Also fixed here:** a `Flight` continued from another `Flight` object never set
`t_initial`, which raised `AttributeError` for any continued flight carrying sensors
or controllers. The bug predates this branch and is fixed by merging the two
initial-solution branches (a refactor the review asked for). Covered by a regression
test.

**Known limitation:** the button is modelled on the rocket axis, so the small roll
coupling through its radial standoff is not represented — roll is instead suppressed
explicitly by the reaction moment. Documented as a modelling assumption; a natural
follow-up.

**References:** "Tip-off effect analysis of a vehicle moving along an inclined
guideway by considering dynamic interactions" (Chou et al.) and "Analysis of Missile
Launchers Part Q: Tipoff Effects in Helical Rail Launchers" (Hosken et al.).

### DIFF
--- a/rocketpy/simulation/flight.py
+++ b/rocketpy/simulation/flight.py
@@ -504,7 +504,7 @@ class Flight:
         verbose=False,
         name="Flight",
         equations_of_motion="standard",
-        use_udot_rail2=True,
+        use_udot_rail2=False,
         ode_solver="LSODA",
         simulation_mode="6 DOF",
     ):
@@ -583,11 +583,14 @@ class Flight:
             solid propulsion rockets. Such equations were used in RocketPy v0
             and are kept here for backwards compatibility.
         use_udot_rail2 : bool, optional
-            If True, enable the intermediate 3-DOF rail phase ``udot_rail2``
-            (tip-off analysis). If False, the flight remains in the 1-DOF
-            ``udot_rail1`` phase until the upper rail button exit and then
-            transitions directly to the generalized 6-DOF dynamics, as in
-            previous versions. Default is True.
+            If True, enable the intermediate 3-DOF "tip-off" rail phase
+            ``udot_rail2``: after the upper rail button leaves the rail (at
+            ``effective_1rl``) the rocket pivots about the still-engaged lower
+            button until it too leaves the rail (at ``effective_2rl``), before
+            the generalized 6-DOF free flight. If False, the flight transitions
+            directly from the 1-DOF ``udot_rail1`` phase to the generalized
+            6-DOF dynamics at the upper button exit, as in previous versions.
+            Default is False.
         ode_solver : str, ``scipy.integrate.OdeSolver``, optional
             Integration method to use to solve the equations of motion ODE.
             Available options are: 'RK23', 'RK45', 'DOP853', 'Radau', 'BDF',
@@ -1052,25 +1055,31 @@ class Flight:
         bool
             True if an event occurred and the simulation should break.
         """
-        # Check for the first time the rocket is between the two rail buttons
-        # (tip-off analysis). Optionally inserts the intermediate 3-DOF
-        # `udot_rail2` phase; see __handle_between_rails_event.
-        if len(self.between_rails_state) == 1 and (
-            self.y_sol[0] ** 2
-            + self.y_sol[1] ** 2
-            + (self.y_sol[2] - self.env.elevation) ** 2
-            >= self.effective_2rl**2
-        ):
-            if self.__handle_between_rails_event(phase, phase_index, node_index):
-                return True
-        # Check for first out of rail event
-        elif len(self.out_of_rail_state) == 1 and (
+        # Check for first out of rail event (upper rail button leaving the
+        # rail, at effective_1rl). This starts the 3-DOF tip-off phase when
+        # enabled, otherwise transitions straight to generalized 6-DOF flight.
+        if len(self.out_of_rail_state) == 1 and (
             self.y_sol[0] ** 2
             + self.y_sol[1] ** 2
             + (self.y_sol[2] - self.env.elevation) ** 2
             >= self.effective_1rl**2
         ):
             return self.__handle_out_of_rail_event(phase, phase_index, node_index)
+        # Check for the lower rail button leaving the rail (at effective_2rl),
+        # which ends the tip-off phase and begins generalized 6-DOF flight.
+        # Only relevant while the intermediate udot_rail2 phase is active.
+        elif (
+            self.use_udot_rail2
+            and len(self.between_rails_state) == 1
+            and len(self.out_of_rail_state) != 1
+            and (
+                self.y_sol[0] ** 2
+                + self.y_sol[1] ** 2
+                + (self.y_sol[2] - self.env.elevation) ** 2
+                >= self.effective_2rl**2
+            )
+        ):
+            return self.__handle_between_rails_event(phase, phase_index, node_index)
 
         # Check for apogee event
         # TODO: negative vz doesn't really mean apogee. Improve this.
@@ -1083,34 +1092,37 @@ class Flight:
 
         return False
 
-    def __handle_out_of_rail_event(self, phase, phase_index, node_index):
-        """Handle the out of rail event.
+    def __root_find_rail_exit_time(self, phase, effective_rl):
+        """Root-find the exact time at which the squared distance travelled
+        (from the launch point, ignoring ground elevation) equals
+        ``effective_rl ** 2``, using cubic Hermite interpolation between the two
+        most recent solution points.
 
         Parameters
         ----------
         phase : FlightPhase
-            The current flight phase.
-        phase_index : int
-            The index of the current phase.
-        node_index : int
-            The index of the current node.
+            The current flight phase (provides the solver step size).
+        effective_rl : float
+            Effective rail length whose crossing is being solved for
+            (``effective_1rl`` for the upper button, ``effective_2rl`` for the
+            lower button).
 
         Returns
         -------
-        bool
-            True to indicate the simulation should break.
+        float
+            Absolute simulation time of the crossing.
         """
         # Check exactly when it went out using root finding
         # Disconsider elevation
         self.solution[-2][3] -= self.env.elevation
         self.solution[-1][3] -= self.env.elevation
         # Get points
-        y0 = sum(self.solution[-2][i] ** 2 for i in [1, 2, 3]) - self.effective_1rl**2
+        y0 = sum(self.solution[-2][i] ** 2 for i in [1, 2, 3]) - effective_rl**2
         yp0 = 2 * sum(
             self.solution[-2][i] * self.solution[-2][i + 3] for i in [1, 2, 3]
         )
         t1 = self.solution[-1][0] - self.solution[-2][0]
-        y1 = sum(self.solution[-1][i] ** 2 for i in [1, 2, 3]) - self.effective_1rl**2
+        y1 = sum(self.solution[-1][i] ** 2 for i in [1, 2, 3]) - effective_rl**2
         yp1 = 2 * sum(
             self.solution[-1][i] * self.solution[-1][i + 3] for i in [1, 2, 3]
         )
@@ -1139,32 +1151,16 @@ class Flight:
             raise ValueError("Multiple roots found when solving for rail exit time.")
         if len(valid_t_root) == 0:  # pragma: no cover
             raise ValueError("No valid roots found when solving for rail exit time.")
-        # Determine final state when upper button is going out of rail
-        self.t = valid_t_root[0] + self.solution[-2][0]
-        interpolator = phase.solver.dense_output()
-        self.y_sol = interpolator(self.t)
-        self.solution[-1] = [self.t, *self.y_sol]
-        self.out_of_rail_time = self.t
-        self.out_of_rail_time_index = len(self.solution) - 1
-        self.out_of_rail_state = self.y_sol
-        # Create new flight phase
-        self.flight_phases.add_phase(
-            self.t,
-            self.u_dot_generalized,
-            index=phase_index + 1,
-        )
-        # Prepare to leave loops and start new flight phase
-        phase.time_nodes.flush_after(node_index)
-        phase.time_nodes.add_node(self.t, [], [], [])
-        phase.solver.status = "finished"
-        return True
+        return valid_t_root[0] + self.solution[-2][0]
 
-    def __handle_between_rails_event(self, phase, phase_index, node_index):
-        """Handle the intermediate rail phase (tip-off analysis).
+    def __handle_out_of_rail_event(self, phase, phase_index, node_index):
+        """Handle the out of rail event (upper rail button leaving the rail, at
+        ``effective_1rl``).
 
-        Records the state at which the rocket first reaches ``effective_2rl``
-        and, when ``use_udot_rail2`` is enabled, inserts the 3-DOF
-        ``udot_rail2`` flight phase.
+        Records the out-of-rail state and inserts the next flight phase: the
+        3-DOF tip-off phase ``udot_rail2`` when it is enabled and there is a
+        nonzero single-button window (``effective_2rl > effective_1rl``),
+        otherwise the generalized 6-DOF dynamics.
 
         Parameters
         ----------
@@ -1178,27 +1174,74 @@ class Flight:
         Returns
         -------
         bool
-            True if a new flight phase was inserted (simulation should break),
-            False otherwise.
+            True to indicate the simulation should break.
         """
+        # Determine final state when upper button is going out of rail
+        self.t = self.__root_find_rail_exit_time(phase, self.effective_1rl)
+        interpolator = phase.solver.dense_output()
+        self.y_sol = interpolator(self.t)
+        self.solution[-1] = [self.t, *self.y_sol]
+        self.out_of_rail_time = self.t
+        self.out_of_rail_time_index = len(self.solution) - 1
+        self.out_of_rail_state = self.y_sol
+        # Create new flight phase: the intermediate 3-DOF tip-off phase if
+        # enabled and the two buttons are distinct, otherwise straight to the
+        # generalized 6-DOF dynamics (previous behavior).
+        if self.use_udot_rail2 and self.effective_2rl > self.effective_1rl:
+            next_derivative = self.udot_rail2
+        else:
+            next_derivative = self.u_dot_generalized
+        self.flight_phases.add_phase(
+            self.t,
+            next_derivative,
+            index=phase_index + 1,
+        )
+        # Prepare to leave loops and start new flight phase
+        phase.time_nodes.flush_after(node_index)
+        phase.time_nodes.add_node(self.t, [], [], [])
+        phase.solver.status = "finished"
+        return True
+
+    def __handle_between_rails_event(self, phase, phase_index, node_index):
+        """Handle the end of the 3-DOF tip-off phase (lower rail button leaving
+        the rail, at ``effective_2rl``).
+
+        Records the between-rails state at the exact ``effective_2rl`` crossing
+        and inserts the generalized 6-DOF free-flight phase.
+
+        Parameters
+        ----------
+        phase : FlightPhase
+            The current flight phase.
+        phase_index : int
+            The index of the current phase.
+        node_index : int
+            The index of the current node.
+
+        Returns
+        -------
+        bool
+            True to indicate the simulation should break.
+        """
+        # Determine final state when lower button is going out of rail
+        self.t = self.__root_find_rail_exit_time(phase, self.effective_2rl)
+        interpolator = phase.solver.dense_output()
+        self.y_sol = interpolator(self.t)
+        self.solution[-1] = [self.t, *self.y_sol]
         self.between_rails_time = self.t
         self.between_rails_time_index = len(self.solution) - 1
         self.between_rails_state = self.y_sol
-        # Optionally insert the udot_rail2 3-DOF phase. If disabled, the solver
-        # remains in the current phase and will transition to generalized
-        # dynamics at the upper button exit as before.
-        if self.use_udot_rail2:
-            self.flight_phases.add_phase(
-                self.t,
-                self.udot_rail2,
-                index=phase_index + 1,
-            )
-            # Prepare to leave loops and start new flight phase
-            phase.time_nodes.flush_after(node_index)
-            phase.time_nodes.add_node(self.t, [], [], [])
-            phase.solver.status = "finished"
-            return True
-        return False
+        # Create the generalized 6-DOF free-flight phase
+        self.flight_phases.add_phase(
+            self.t,
+            self.u_dot_generalized,
+            index=phase_index + 1,
+        )
+        # Prepare to leave loops and start new flight phase
+        phase.time_nodes.flush_after(node_index)
+        phase.time_nodes.add_node(self.t, [], [], [])
+        phase.solver.status = "finished"
+        return True
 
     def __handle_apogee_event(self, phase, phase_index, node_index):
         """Handle the apogee event.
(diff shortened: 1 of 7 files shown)


rocketpy/simulation/flight.py
```diff
@@ -797,6 +810,192 @@ def __simulate(self, verbose):
                     if verbose:
                         print(f"Current Simulation Time: {self.t:3.4f} s", end="\r")
 
+                    # check for the first time the rocket is between the two rail buttons for tip-off analysis
+                    if len(self.between_rails_state) == 1 and (
```
aZira371: The condition len(self.betweenrailsstate) == 1 checks whether the state has been set once, not whether the rocket is currently between the buttons. Once betweenrailsstate is populated, subsequent integration steps will not re-enter this block, so the phase is only added at the exact instant the lower button clears, not maintained throughout the tip-off interval. This means the udot_rail2 phase may execute for only a single time step or not propagate correctly.

rocketpy/simulation/flight.py
```diff
@@ -797,6 +810,192 @@ def __simulate(self, verbose):
...
+                        valid_t_root = [
+                            t_root.real
+                            for t_root in t_roots
+                            if 0 < t_root.real < t1 and abs(t_root.imag) < 0.001
+                        ]
+                        if len(valid_t_root) > 1:  # pragma: no cover
+                            raise ValueError(
+                                "Multiple roots found when solving for rail exit time."
+                            )
+                        if len(valid_t_root) == 0:  # pragma: no cover
```
aZira371: this fallback uses an arbitrary midpoint guess when the Hermite interpolation fails to find a valid root. This fallback has no physical justification, if no root exists in the bracket, taking the midpoint may place the transition at a point where the rocket is not at effective1rl, violating the state's physical meaning. The warning is issued, but the simulation proceeds with potentially incorrect rail-exit kinematics, and downstream analysis (e.g., launch angle) will be corrupted.​
Instead of a fallback, the code should either raise an exception (forcing the user to investigate convergence issues) or attempt a refined bracket search with tighter tolerances.

rocketpy/simulation/flight.py
```diff
@@ -1527,6 +1741,10 @@ def __init_flight_state(self):
             self.out_of_rail_state = self.initial_solution[1:]
             self.out_of_rail_time = self.initial_solution[0]
             self.out_of_rail_time_index = 0
+            # save out of rail 2 state and time with the same data as out of rail
+            self.between_rails_state = self.initial_solution[1:]
```
aZira371: In __init__, the code initializes self.betweenrailsstate and self.betweenrailstime in multiple branches (3 times once in if elif and else each). These three branches set identical values for betweenrailsstate and betweenrailstime. This is redundant. The initialization should be factored out after the branching logic.

rocketpy/simulation/flight.py
```diff
@@ -1874,7 +2096,7 @@ def udot_rail1(self, t, u, post_processing=False):
         return [vx, vy, vz, ax, ay, az, 0, 0, 0, 0, 0, 0, 0]
 
     def udot_rail2(self, t, u, post_processing=False):  # pragma: no cover
-        """[Still not implemented] Calculates derivative of u state vector with
+        """[WIP] Calculates derivative of u state vector with
```
aZira371: Suggestion for WIP:

True 3-DOF tip-off on a rail with one button requires tracking the normal contact force and the reaction moment. The state vector for udot_rail2 appears to reuse the 13-state format from udot_generalized (x, y, z, vx, vy, vz, e0, e1, e2, e3, ω1, ω2, ω3). However, single-button dynamics are inherently unstable without proper constraint enforcement. The angular velocities ω1 and ω2 will grow due to aerodynamic and thrust misalignment torques, but the code does not implement any constraint forces to stabilize the solution. Without these, the integrator may produce non-physical angular accelerations.

aZira371: The existing udot_rail1 (1-DOF rail phase) computes thrust, drag, and rail reaction forces. The new udot_rail2 must account for:

Off-axis aerodynamic forces and moments (due to angle of attack during tip-off)

Thrust misalignment effects

Gravity components in the body frame

Inertia tensor evolution as the rocket rotates

None of these appear in the provided code stub.

tests/unit/simulation/test_udot_rail2_feature.py
```diff
@@ -0,0 +1,212 @@
...
+    velocity = Vector(u_between[3:6])
+    projected = rail * (velocity @ rail)
+
+    diff = r_dot - projected
+    assert float(abs(diff)) < 1e-8, (
+        f"r_dot not a projection onto rail (err={float(abs(diff))})"
+    )
+
+
+def test_udot_rail2_csv_comparison_generation(
```
aZira371: test_udot_rail2_csv_comparison_generation generates CSV files comparing enabled/disabled runs and computes pitch/yaw deltas, but does not assert any expected behavior—it only checks that files exist and contain headers. This is a structural test, not a validation test

aZira371 (changes requested):
I have the following comments for the preliminary implementation. In general looks to be going in a good direction as of now. Will wait for u_dotrail2 implementation to be fully complete in order to review the tests and involved physics.

Gui-FernandesBR:
@aZira371 @Rafit345 this PR is ready for another look. Since the last review the
`udot_rail2` physics was rewritten, the phase transition was fixed, and the branch
was rebased onto current `develop` (no conflicts). All checks are green.

### The two blocking problems

**1. The phase transition was inverted.** The thresholds were swapped: `between_rails`
keyed on `effective_2rl` as the leading `if`, `out_of_rail` on the smaller
`effective_1rl` as the `elif`. So `out_of_rail` fired first and `udot_rail2` was
inserted *after* free flight, where it never exited. The observed phase order was
`rail1 -> u_dot_generalized -> udot_rail2`. It is now
`rail1 -> udot_rail2` (at `effective_1rl`, upper button out) `-> u_dot_generalized`
(at `effective_2rl`, lower button out), which is also what broke ~32 existing flight
tests with `ValueError: No valid roots found when solving for rail exit time`.

**2. The dynamics had no reaction force**, which is the substance of your review
comment: it reused free-flight dynamics plus an ad-hoc roll-zeroing and a projection
of the velocity onto the rail. That is now replaced by proper constrained dynamics,
along the lines you asked for:

- The lower button slides along the fixed inertial rail axis; roll is suppressed.
- The unknowns are a **normal reaction force** at the button (perpendicular to the
  rail, 2 DOF) and a **roll reaction moment** (1 DOF).
- They are solved from a **3x3 linear system** built from the three constraints: the
  button's acceleration perpendicular to the rail vanishes (2) and the roll angular
  acceleration vanishes (1). `6 - 3 = 3` DOF remain: slide along the rail, pitch, yaw.
- It builds on `u_dot_generalized`, so the off-axis aerodynamics, thrust,
  body-frame gravity and the evolving inertia tensor you listed are all included by
  construction rather than re-derived. The reaction wrench is added to the same
  `T20`/`T21` totals and the same solve is applied.
- The velocity is **not** projected onto the rail any more. The constraint acts at
  the acceleration level on the *button*; the CDM legitimately acquires a small
  perpendicular velocity as the rocket pivots.

The derivation is now in the technical documentation, `docs/technical/tip_off.rst`,
next to the other equations of motion.

### Your other comments

- **Midpoint fallback with no physical justification** - removed. The root finder was
  factored into one shared helper (`__root_find_rail_exit_time`) used by both events,
  and it raises rather than guessing.
- **`len(between_rails_state) == 1` does not mean "currently between the buttons"** -
  correct, and that is the intent: it is a "has this event fired yet" guard, the same
  idiom as the pre-existing `len(out_of_rail_state) == 1`. What keeps the phase active
  is the phase itself, not this check. It now also requires `out_of_rail` to have
  already fired, so it can only trigger inside the tip-off window.
- **Redundant `between_rails` initialization across three branches** - fixed. The two
  initial-solution branches are now one. Worth flagging: they were not identical - the
  `Flight`-object branch also failed to set `t_initial`, which raised
  `AttributeError: 'Flight' object has no attribute 't_initial'` for any continued
  flight carrying sensors or controllers. That bug predates this branch; merging the
  branches fixes it, and there is a regression test.
- **CSV test asserts nothing** - agreed, it was structural. The test file was rewritten
  around behaviour: phase ordering, the opt-in default, the on-rail constraint, zero
  roll, the gravity tip-off direction, and the rail axis being defined when an
  `initial_solution` skips the rail. No CSV artifacts.
- **Copilot's points** - `use_udot_rail2` now defaults to `False`, the docstring entry
  moved to match the signature order, the leftover commented-out code and unused
  imports/variables are gone, the `pragma: no cover` on `udot_rail2` was dropped, and
  the test names follow `test_methodname_expectedbehaviour`.

Two Copilot comments I deliberately left alone: the `a += 1e-5  # TODO: why??` in the
root finder and the `negative vz doesn't really mean apogee` TODO. Both are
pre-existing on `develop` and unrelated to tip-off; fixing them here would change
results for every flight and belongs in its own PR.

### Validation

- `use_udot_rail2=False` is **bit-for-bit identical to `develop`** - verified by
  running the same scenario against a clean `develop` worktree and comparing the full
  final state, not just the apogee. This is why the feature is opt-in for now.
- Constraint satisfaction: the button's perpendicular offset from the rail stays at
  ~1e-13 m, and the roll rate and roll acceleration stay at zero.
- Physics: with no wind the nose pitches **down** (the center of mass is ahead of the
  pivot), and the rocket weathercocks into a crosswind - the acceptance criterion in
  #28. Apogee shifts by about -2 m on Calisto.
- Suites: 1960 unit + simulation-integration tests pass locally; all 10 CI checks pass.

### Known limitation

The button is modelled on the rocket axis, so the small roll coupling through its
radial standoff is not represented; roll is instead suppressed explicitly by the
reaction moment. This is documented as a modelling assumption and is a natural
follow-up.

rocketpy/simulation/flight.py
```diff
@@ -504,6 +504,7 @@ def __init__(  # pylint: disable=too-many-arguments,too-many-statements
         verbose=False,
         name="Flight",
         equations_of_motion="standard",
+        use_udot_rail2=False,
```
Gui-FernandesBR, line 507: adding "use_udot_rail2" argument at this position will be a break change. We should rather append it to the end of the function

Gui-FernandesBR:
Status check on this one: it went conflicting with `develop`. @Rafit345 could you merge `develop` in and push? All 19 checks were green before the conflict appeared, so this should just be a base update.

@aZira371 the `CHANGES_REQUESTED` review here dates from 2025-12-13 and a lot of commits have landed since then. Once the conflicts are resolved, could you take another look? That stale review is currently the only thing blocking this PR.

aZira371:
> Status check on this one: it went conflicting with `develop`. @Rafit345 could you merge `develop` in and push? All 19 checks were green before the conflict appeared, so this should just be a base update.
> 
> @aZira371 the `CHANGES_REQUESTED` review here dates from 2025-12-13 and a lot of commits have landed since then. Once the conflicts are resolved, could you take another look? That stale review is currently the only thing blocking this PR.
> 

I'll review this soon!

rocketpy/simulation/flight.py
aZira371: agreed gui. use_udot_rail2=False is inserted mid-signature, not at the end (rocketpy/simulation/flight.py, Flight.__init__) — it sits between equations_of_motion and ode_solver/simulation_mode. Any caller passing ode_solver/simulation_mode positionally silently gets the wrong value. This is a breaking change, should move it after simulation_mode.

rocketpy/simulation/flight.py
```diff
@@ -1890,9 +2009,27 @@ def udot_rail1(self, t, u, post_processing=False):
 
         return [vx, vy, vz, ax, ay, az, 0, 0, 0, 0, 0, 0, 0]
 
-    def udot_rail2(self, t, u, post_processing=False):  # pragma: no cover
-        """[Still not implemented] Calculates derivative of u state vector with
-        respect to time when rocket is flying in 3 DOF motion in the rail.
+    def udot_rail2(self, t, u, post_processing=False):
```
aZira371, line 2012: udot_rail2 recomputes total_mass and get_inertia_tensor_at_time(t) (plus the parallel-axis H/I_CM correction) from scratch right after u_dot_generalized already computed the same things internally. Since the ODE solver evaluates the RHS many times per step, this doubles interpolation cost for every call during the tip-off window. Worth exposing those intermediates from u_dot_generalized rather than recomputing.

rocketpy/simulation/flight.py
```diff
@@ -4224,6 +4455,7 @@ def to_dict(self, **kwargs):
                     "time": self.time,
                     "out_of_rail_velocity": self.out_of_rail_velocity,
                     "out_of_rail_state": self.out_of_rail_state,
+                    "between_rails_state": self.between_rails_state,
```
aZira371, line 4458: between_rails_time/between_rails_state are tracked and included in to_dict(), but flight_prints.py/flight_plots.py were never updated with a "Between Rails Time/Velocity" alongside the existing "Rail Departure" fields. Since the entire point of this feature is studying tip-off for dispersion, users who opt in have no way to see the window in .info() or standard plots except manually.

rocketpy/simulation/flight.py
```diff
@@ -1911,8 +2048,102 @@ def udot_rail2(self, t, u, post_processing=False):  # pragma: no cover
...
+        # The generalized EOM store r_CM / r_NOZ as (point -> CDM) vectors, i.e.
+        # the negative of the true-frame position; hence the sign flips below.
+        csys = self.rocket._csys
+        cdm = self.rocket.center_of_dry_mass_position
+        r_CM = Vector([0, 0, -self.rocket.com_to_cdm_function.get_value_opt(t)])
+        lower_button_z = self.rocket.rail_buttons[0].position.z
+        r_button = Vector([0, 0, (lower_button_z - cdm) * csys])
+
+        # Inertia about the instantaneous center of mass (sign-independent).
+        inertia_tensor = self.rocket.get_inertia_tensor_at_time(t)
```
aZira371, line 2078: udot_rail2 always builds its constraint Jacobian from the full 3×3 inertia-tensor path, but if equations_of_motion="solid_propulsion" or simulation_mode="3 DOF" is also selected, the "free" solution it patches comes from a different reduced axisymmetric-inertia formulation (u_dot/u_dot_generalized_3dof). Mixing the two isn't tested, warned against, or guarded. Recommend either a guard/warning or a test proving the combination is still physically consistent.

aZira371 (changes requested):

Gui-FernandesBR (approved):

rocketpy/simulation/flight.py
Gui-FernandesBR: Fixed in f66bc588. `use_udot_rail2` now sits at the end of `Flight.__init__`, after `post_step_callback`, and its docstring entry moved with it.

Worth noting the window was wider than when you flagged it: `develop` has since added `post_step_callback` in that same tail region, so a positional caller had three arguments to land on, not two.

rocketpy/simulation/flight.py
Gui-FernandesBR: Fixed in f66bc588. `u_dot_generalized` now publishes the total mass and the CM inertia it just computed, and `udot_rail2` reuses them when the cached `t` matches instead of interpolating both again — which it did on every RHS evaluation inside the window.

One detail that made this safe: the two call sites build `r_CM` with opposite sign conventions (the generalized EOM store point-to-CDM vectors, the tip-off code uses the true body frame). Since `H = (r_CM.cross_matrix @ -r_CM.cross_matrix) * total_mass` is quadratic in `r_CM`, `I_CM` is identical either way, so the cached value is directly reusable.

Verified rather than assumed: I ran the same flight with the cache in place and with it forced off through a descriptor, and compared the full solution arrays. Max absolute difference was exactly `0.0`.

rocketpy/simulation/flight.py
Gui-FernandesBR: Fixed in f66bc588 — you were right that the feature was effectively invisible to the people it is for.

`Flight` now exposes `between_rails_velocity` and `tip_off_duration` alongside the existing `between_rails_time`/`between_rails_state`. `flight_prints.py` prints a **Tip-Off State** block (duration, lower-button departure time and velocity, angle of attack) right after the Rail Departure fields, and `flight_plots.py` shades the window across all four attitude subplots, which is where tip-off actually shows up.

All of it is inert when the phase is off: the print returns early and the shading helper is a no-op, so existing output is byte-identical. There are tests for both the enabled and the disabled path.

rocketpy/simulation/flight.py
Gui-FernandesBR: Fixed in f66bc588, and the concern was real — I checked before writing the guard.

The phase is selected at the rail-departure event, independently of the mode resolution in `__init_equations_of_motion`. And `udot_rail2` starts by calling `self.u_dot_generalized`, which is a *rebound* attribute: under `simulation_mode="3 DOF"` it points at `u_dot_generalized_3dof`, which returns `[vx, vy, vz, ax, ay, az, 0, 0, 0, 0, 0, 0, 0]` — no attitude at all. So the constraint patch would have been correcting angular accelerations of a solution that does not model rotation. Same class of problem with `equations_of_motion="solid_propulsion"`, which binds the reduced axisymmetric `u_dot`.

Both combinations now raise `ValueError` at construction with a message naming the offending argument. This also covers point-mass motors, since those force `simulation_mode` to `"3 DOF"` on their own — a user could have hit this without ever typing `"3 DOF"`. Three tests cover the two rejections and the fact that the guard stays quiet when the phase is not requested.

rocketpy/simulation/flight.py
Gui-FernandesBR: This one is resolved in the current state of the branch — the arbitrary midpoint fallback you objected to is gone.

What is there now is narrow and, I think, defensible: the only fallback left is on `np.linalg.LinAlgError` when the 3x3 constraint system is singular, which warns and returns the unconstrained derivative for that step. No guessed root, no silently corrupted rail-exit kinematics.

rocketpy/simulation/flight.py
Gui-FernandesBR: Resolved on the current branch. The triple initialization is gone: `between_rails_state`/`between_rails_time`/`between_rails_time_index` are set once in `__init_solution_monitors`, and the only other assignment is in the `initial_solution` path, where the values are genuinely different (the supplied starting state, since both rail phases are skipped). The third site is the event handler recording the actual departure.

rocketpy/simulation/flight.py
Gui-FernandesBR: This was the substantive objection, and the implementation changed to meet it — the phase no longer leaves the single-button dynamics unconstrained.

`udot_rail2` now solves for an explicit reaction wrench: a normal force at the lower button (two components perpendicular to the rail) plus a roll reaction moment, three unknowns recovered from a 3x3 constraint Jacobian built by probing the response to unit reactions. The three constraints are the button's acceleration perpendicular to the rail being zero (2) and zero roll angular acceleration (1). The free solution from `u_dot_generalized` is then patched with that wrench.

So the stabilizing constraint forces you said were missing are exactly what is being solved for. The derivation is written up in `docs/technical/tip_off.rst`.

It is also tested rather than asserted: the lower button stays on the rail line to under 1e-6 m across the window, and roll acceleration and roll rate stay at 1e-12 and 1e-9.

Gui-FernandesBR: Addressed by the change in approach since December. `udot_rail2` no longer reimplements the force model — it calls `u_dot_generalized` for the free solution and overrides only the linear and angular accelerations with the constrained values.

That means everything on your list comes in through the generalized equations, which already compute it: per-surface aerodynamic forces and moments via `surfaces_cp_to_cdm` (so off-axis effects at angle of attack are included), net thrust including pressure thrust, `weight_in_body_frame` for gravity, and the inertia tensor together with its time derivative `I_dot`. Nothing is recomputed in a reduced form inside the tip-off phase.

tests/unit/simulation/test_udot_rail2_feature.py
Gui-FernandesBR: Agreed with the original criticism, and the CSV test it was about no longer exists. The current `test_udot_rail2_feature.py` asserts behaviour:

- the lower button stays on the rail line, max perpendicular offset under 1e-6 m;
- roll acceleration under 1e-12 and roll rate under 1e-9 across the window;
- with no wind the nose pitches *down* under gravity, and by a physically plausible amount (under one degree) rather than just "some number";
- phase ordering rail1 -> rail2 -> generalized, and the opt-in default.

f66bc588 adds eight more covering the mode guard, the reporting, and objects restored without `__init__`.

PR #924, opened 2026-01-28
Feature/cluster motor structure

ayoubdsp:
## Pull request type

* [x] Code changes (bugfix, features)
* [x] Code maintenance (refactoring, formatting, tests)

## Checklist

* [x] Tests for the changes have been added (if needed)
* [ ] Docs have been reviewed and added / updated
* [x] Lint (`black rocketpy/ tests/`) has passed locally
* [x] All tests (`pytest tests/integration/motors/test_cluster_motor.py`) have passed locally
* [ ] `CHANGELOG.md` has been updated (if relevant)

## Current behavior

Currently, RocketPy does not have native support for clustered motor configurations.
Users attempting to simulate a cluster (e.g., 3x or 4x motors) must manually create a "custom" single motor with scaled thrust curves and manually calculate the complex inertia changes caused by the off-axis mass distribution using external tools. This process is error-prone and does not dynamically update the inertia tensor correctly as propellant is consumed.

## New behavior

This PR introduces a new `ClusterMotor` class in `rocketpy.motors`.

* **Wrapper Architecture:** It wraps an existing `SolidMotor` object, preserving all its internal ballistics.
* **Automatic Scaling:** Thrust, Dry Mass, and Propellant Mass are automatically scaled by the number of motors ().
* **Physics-Based Inertia:** The class implements the **Parallel Axis Theorem (Huygens-Steiner)** to correctly calculate the global inertia tensor of the cluster. It accounts for the radial distance () of the motors from the central axis, dynamically updating  (Pitch/Yaw) and  (Roll) as the propellant mass decreases during the burn.

## Breaking change

* [x] No

## To do
Once implementation has been validated, we will still need to improve the draw() function to enable the drawing of grouped projects. For now, I have only implemented a rear view, so that the placement of the engines is visible.

This feature was developed by the French rocket science association IPSA SPACE SYSTEMS. Come say hello to us on Instagram! 
https://www.instagram.com/ipsa_space_systems/

### DIFF
--- a/rocketpy/motors/cluster_motor.py
+++ b/rocketpy/motors/cluster_motor.py
@@ -1,0 +1,228 @@
+import matplotlib.pyplot as plt
+import numpy as np
+from rocketpy import Function
+from rocketpy.motors import Motor
+
+
+class ClusterMotor(Motor):
+    """
+    A class representing a cluster of N identical motors arranged symmetrically.
+
+    This class aggregates the physical properties (thrust, mass, inertia) of
+    multiple motors using the Parallel Axis Theorem (Huygens-Steiner theorem).
+
+    Attributes
+    ----------
+    motor : SolidMotor
+        The single motor instance used in the cluster.
+    number : int
+        The number of motors in the cluster.
+    radius : float
+        The radial distance from the rocket's central axis to the center of each motor.
+    """
+
+    def __init__(self, motor, number, radius):
+        """
+        Initialize the ClusterMotor.
+
+        Parameters
+        ----------
+        motor : SolidMotor
+            The base motor to be clustered.
+        number : int
+            Number of motors. Must be >= 2.
+        radius : float
+            Distance from center of rocket to center of motor (m).
+        """
+        self.motor = motor
+        self.number = number
+        self.radius = radius
+        dry_inertia_cluster = self._calculate_dry_inertia()
+
+        super().__init__(
+            thrust_source=motor.thrust_source,
+            nozzle_radius=motor.nozzle_radius,
+            burn_time=motor.burn_time,
+            dry_mass=motor.dry_mass * number,
+            dry_inertia=dry_inertia_cluster,
+            center_of_dry_mass_position=motor.center_of_dry_mass_position,
+            coordinate_system_orientation=motor.coordinate_system_orientation,
+            interpolation_method="linear",
+        )
+
+        self.throat_radius = motor.throat_radius
+        self.grain_number = motor.grain_number
+        self.grain_density = motor.grain_density
+        self.grain_outer_radius = motor.grain_outer_radius
+        self.grain_initial_inner_radius = motor.grain_initial_inner_radius
+        self.grain_initial_height = motor.grain_initial_height
+        self.grains_center_of_mass_position = motor.grains_center_of_mass_position
+        self._thrust = self.motor.thrust * self.number
+        self._propellant_mass = self.motor.propellant_mass * self.number
+        self._propellant_initial_mass = self.number * self.motor.propellant_initial_mass
+        self._center_of_propellant_mass = self.motor.center_of_propellant_mass
+        Ixx_term1 = self.motor.propellant_I_11 * self.number
+        Ixx_term2 = self.motor.propellant_mass * (0.5 * self.number * self.radius**2)
+        self._propellant_I_11 = Ixx_term1 + Ixx_term2
+        self._propellant_I_22 = self._propellant_I_11
+
+        Izz_term1 = self.motor.propellant_I_33 * self.number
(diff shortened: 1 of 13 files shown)


rocketpy/motors/cluster_motor.py
```diff
@@ -0,0 +1,263 @@
...
+
+    Attributes
+    ----------
+    motor : SolidMotor
+        The single motor instance used in the cluster.
+    number : int
+        The number of motors in the cluster.
+    radius : float
+        The radial distance from the rocket's central axis to the center of each motor.
+    """
```
MateusStano: This is a good description of the class. One question, this is only modelling cases where the motors are constructed radially? A symmetric case considering a motor in the middle (aligned with upper frame central axis) is not modelled here right?

Can you clarify this aspect in the docstrings, please?

Also, I would suggest changing the name of the class to `RingClusterMotor` or `AnnularClusterMotor` (or something similar to your preference).

rocketpy/motors/cluster_motor.py
```diff
@@ -0,0 +1,263 @@
...
+        Izz_loc = self.motor.dry_I_33
+        m_dry = self.motor.dry_mass
+
+        Izz_cluster = self.number * Izz_loc + self.number * m_dry * (self.radius**2)
+        Ixx_cluster = self.number * Ixx_loc + (self.number / 2) * m_dry * (
+            self.radius**2
+        )
+        Iyy_cluster = self.number * Iyy_loc + (self.number / 2) * m_dry * (
+            self.radius**2
+        )
```
MateusStano: Where do the terms `(self.number/2)` come from here? Are you sure this works in cases where `self.number=2`? I believe if there are two motors, the values of Ixx and Iyy should be different

rocketpy/motors/cluster_motor.py
```diff
@@ -0,0 +1,263 @@
...
+    def _evaluate_propellant_inertia(self):
+        """Calculates the dynamic inertia of the propellant using Steiner's theorem."""
+        Ixx_term1 = self.motor.propellant_I_11 * self.number
+        Ixx_term2 = self.motor.propellant_mass * (0.5 * self.number * self.radius**2)
+        self._propellant_I_11 = Ixx_term1 + Ixx_term2
+        self._propellant_I_22 = self._propellant_I_11
+
+        Izz_term1 = self.motor.propellant_I_33 * self.number
+        Izz_term2 = self.motor.propellant_mass * (self.number * self.radius**2)
+        self._propellant_I_33 = Izz_term1 + Izz_term2
```
MateusStano: these inertias calculations probably also have issues in the case of `self.number=2`

MateusStano (changes requested):
This is really good! Just requested a few changes and it should be good to merge

ayoubdsp:
Hi @MateusStano, thanks for the great feedback! You were spot on about the inertia calculations.I've renamed the class to RingClusterMotor and updated the docstrings to make it clear that this specifically models an annular configuration without a central motor.To fix the issue with the $N=2$ case, I completely removed the N/2 approximation. Instead, the code now calculates the exact geometric summation using the true $x$ and $y$ coordinates of each motor. This ensures the transverse inertias are perfectly accurate and naturally asymmetric when there are only two motors, while staying mathematically consistent for larger clusters.Finally, I updated the test suite to validate this exact math and fixed that np.isclose ValueError we were getting by making sure the Function objects are properly evaluated at a given time $t$. All tests are now passing smoothly!Let me know if everything looks good to merge !

Gui-FernandesBR:
Implemented one pending merge-readiness item on this PR:\n\n- updated CHANGELOG.md with an Unreleased entry for RingClusterMotor (#924).\n\nCommit: d10a942\n\n@MateusStano could you please take a final look so we can move this to merge?

Gui-FernandesBR:
Update: I rebased eature/cluster-motor-structure onto current develop and resolved conflicts to keep the PR merge-ready.\n\nAdditional adjustments included in this update:\n- added CHANGELOG.md Unreleased entry for RingClusterMotor (#924)\n- fixed post-rebase plotting imports/type dispatch in 
ocketpy/plots/rocket_plots.py\n\nValidation run locally:\n- pytest tests/integration/motors/test_ring_cluster_motor.py -q\n- pytest tests/unit/test_plots.py -q\n- 
uff check rocketpy/plots/rocket_plots.py rocketpy/motors/ring_cluster_motor.py tests/integration/motors/test_ring_cluster_motor.py\n\nLatest branch head: 378ec7ce.\n\n@MateusStano could you please re-review for merge?

Gui-FernandesBR (approved):
gotta update documentation and also make a few tests to ensure it's working.

Gui-FernandesBR:
Sorry for the long wait @ayoubdsp . Thank you so much for your great work, it still impresses me a lot. Awesome!

PR #958, opened 2026-04-29
MNT: refactor parachute implementation

### LINKED ISSUE #31: Recovery options and drag coeefficients?
Hey, love this project! I'm the founder of open research rocketry and i'd love to chat to you about various aspects of this and some of our burgeoning projects over in that community. Was wondering if you had/would consider adding different recovery system design options (we have some parachute design calculators and scripts over on our webpage and gitlab repo). We are using either a designed diameter or a reference surface area and a drag coefficient and I wondered if it could be built into your project? (Full disclosure... I am much more mechanical so I might be way off the mark!) If you are on riot/matrix we would love to have you drop in the chat! https://riot.im/app/#/room/#OpenResearchRocketry:matrix.org

Lucas-Prates:
## Pull request type

## Checklist

- [ ] Tests for the changes have been added **(modified)**
- [X] Docs have been reviewed and added / updated **(partially)**

## Current behavior

The current implementation of parachutes in RocketPy considers a simplified drag based round parachute, henceforth called Hemispherical Parachute. The `Parachute` class stores basic parachute information (drag, triggers, etc...) and all the physics (equations of motion) is handled by the `Flight` class inside the `u_dot_parachute` function. The problem is: different parachute types and models require different physics, which would create several "u_dot_parachute"s functions (or a really big messy one). Moreover, different parachutes require different information (some have lift coefficient, others have yaw coefficient, others have bank angle, angle of attack, and so on), so perhaps we should have different parachute classes, not just a big one.

## New behavior

The `Parachute` class is now an abstract class which cannot be instantiated. All parachute models which will be created from now on should have it as a parent class. Our current implementation of the parachute is in the `HemisphericalParachute` class. 

Each parachute class will have only the attributes it uses for its modelling and **it contains the u_dot function!** The idea is that the `Parachute` class has an abstract method called `u_dot_parachute` which each child must implement! Hence, the design proposal is to allow each parachute model to handle the physics! The SciPy solver is still called in the `Flight` class: an `u_dot_parachute_wrapper` is constructed in the `Flight` class to allow the parachute to use flight information and still maintain its arguments conforming to the SciPy solver requirement.

The `Rocket` method `add_parachute` was also changed. Since the idea is to have several parachute models, we now construct the parachute and then pass it as and object to that method. Hence, we now have a syntax such as

```
Main = HemisphericalParachute(
    "Main",
    cd_s=10.0,
    trigger=800,
    sampling_rate=105,
    lag=1.5,
    noise=(0, 8.3, 0.5),
)

Drogue = HemisphericalParachute(
    "Drogue",
    cd_s=1.0,
    trigger="apogee",
    sampling_rate=105,
    lag=1.5,
    noise=(0, 8.3, 0.5),
)

calisto.add_parachute(Main)
calisto.add_parachute(Drogue)
```

Hence, `add_parachute` also changed (compare to how it was done before).

## Breaking change

- [X] Will be (currently still support legacy parachute use)

## Additional information

I look forward for your opinions on design choice and modifications/improvements!

I modified the tests and notebooks to work properly. However, I did not change much of the documentation, so it probably teaches how to add a Parachute incorrectly. Moreover, I did a quick dirty job on the Monte Carlo modification, it needs to be rewritten.

TODO LIST:

[ ] Verify and modify the documentation
[ ] Modify the Stochastic version of parachute

:parachute:  **I'm baking up a low-fidelity 3DoF Parafoil model which motivated this hard refactor :)** :parachute:


rocketpy/rocket/rocket.py
```diff
@@ -1648,26 +1653,49 @@ def add_parachute(
...
-        )
-        self.parachutes.append(parachute)
-        return self.parachutes[-1]
+        if parachute is not None:
+            self.parachutes.append(parachute)
+        else:
+            # For backwards compatibility
+            deprecation_message = (
+                "Passing parachute parameters directly to 'add_parachute' method is "
+                + "deprecated and will be removed in a future release. Please create "
```
Gui-FernandesBR: "will be removed in a future release" is usually not a good idea

Can you already set a version for this feature to be removed? If we are 1.x.y, try 1.(x+2).0, for example

2 versiosn is tipically acceptable

docs/notebooks/trajectory.kml
Gui-FernandesBR, line 1: do you really need to add this file? `docs/notebooks/trajectory.kml`

Lucas-Prates: Nope, artifact of doing "git add docs/*". Thank you for catching it! :)

rocketpy/rocket/rocket.py
Lucas-Prates: Thanks for the observation! Will set "1.12" in the message but will keep in mind that it may change since this probably will take a while.

Gui-FernandesBR:
Quick update on #958 (latest head: `ad0777bc`):

- rebased branch on current `develop` and resolved conflicts
- added missing `CHANGELOG.md` Unreleased entry for the parachute refactor
- aligned `Parachute` trigger docstrings with runtime behavior (callable may receive sensors as 4th argument)
- set concrete deprecation removal target for legacy `Rocket.add_parachute(...)` usage to `1.14.0`

Local checks run for this update:
- `ruff check` on touched files
- `pytest tests/unit/rocket/test_parachute.py -q`
- `pytest tests/integration/simulation/test_flight.py -q`

Gui-FernandesBR (approved):

Gui-FernandesBR (approved):

Gui-FernandesBR (approved):

Issue #959, opened 2026-05-08
BUG: Controllers called twice per time node in Flight.__simulate


Krissal1234:
In `Flight.__simulate`, controllers are called twice per time node. 
This is a leftover loop from the refactor introduced in #843.

In commit 45a891eb, `__process_sensors_and_controllers_at_current_node` 
was introduced for sensor and controller processing. However, 
the original controller loop was not removed from `__simulate`.

Any flight simulation with a controller (e.g. air brakes) will trigger 
this bug. You can verify it by adding a print statement to your controller:

def air_brakes_controller(t, y_sol, solution, sensors):
    print(t)

You will see each timestamp printed twice, confirming the controller 
is being called twice per time node.

A controller that depends on the time elapsed since the last call (e.g. a PID 
controller) will compute incorrect values since it is called twice at the same timestamp.

karlhillx:
I've reproduced this and can confirm the duplicate call. It's in `TimeNodes.merge()` — `_controllers` is appended twice. I'll open a PR to fix it.

Gui-FernandesBR:
Please refer to https://github.com/RocketPy-Team/RocketPy/pull/949

Next release will already incorporate these changes. Before that, you can install rocketpy directly from the develop branch.

I'm closing this issue as solved, thanks!

PR #960, opened 2026-05-08
BUG: Remove duplicate controller call in Flight.__simulate

Krissal1234:
## Pull request type

## Checklist

## Current behavior
Controllers are called twice per time node in `Flight.__simulate`. See #959.

## New behavior
Controllers are called exactly once per time node, inside 
`__process_sensors_and_controllers_at_current_node` as intended.

## Breaking change
- [x] No

## Additional information
The duplicate loop was introduced in #843 when 
`__process_sensors_and_controllers_at_current_node` was added 
for sensor and controller processing, but the original 
controller loop in `__simulate` was not removed.


Gui-FernandesBR:
seems like this problem has already been solved. We just have to locate the original PR and link it here

MateusStano:
Closing because this was already solved in https://github.com/RocketPy-Team/RocketPy/pull/949

PR #964, opened 2026-06-02
Fix controller duplication in TimeNodes.merge()

### LINKED ISSUE #959: BUG: Controllers called twice per time node in Flight.__simulate

karlhillx:
Closes #959

Removed duplicate _controllers += node._controllers in
TimeNodes.merge() that caused each controller to be called
twice per time node during simulation.

Added unit test verifying merge accumulation behavior.

### DIFF
--- a/tests/integration/simulation/test_flight.py
+++ b/tests/integration/simulation/test_flight.py
@@ -14,7 +14,7 @@ plt.rcParams.update({"figure.max_open_warning": 0})
 )
 @patch("matplotlib.pyplot.show")
 # pylint: disable=unused-argument
-def test_all_info(mock_show, request, flight_fixture):
+def test_all_info(request, flight_fixture):
     """Test that the flight class is working as intended. This basically calls
     the all_info() method and checks if it returns None. It is not testing if
     the values are correct, but whether the method is working without errors.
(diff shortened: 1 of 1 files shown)


Gui-FernandesBR:
the only modified file is a test file. The PR description is not aligned with the PR changes.

I'm closing this PR since it doesn't really solve a problem or improve anything.

Please feel free to improve yout branch and propose a new PR. 

Remind to open PRs always heading to the `develop` branch please

image

PR #965, opened 2026-06-07
FEAT: new actuator class for roll, throttle, and thrust vector control

zuorenchen:
## Pull request type

- [X] Code changes (features)

## Checklist

- [X] Tests for the changes have been added (if needed)
- [ ] Docs have been reviewed and added / updated
- [ ] Lint (`black rocketpy/ tests/`) has passed locally 
- [ ] All tests (`pytest tests -m slow --runslow`) have passed locally
- [ ] `CHANGELOG.md` has been updated (if relevant)

## Current behavior
- In https://github.com/RocketPy-Team/RocketPy/issues/154, TVC feature is requested.
- Only `air_brake` is implemented as a controller

## New behavior
- A new `Actuator` base class is added.
  - This class follows the implementation of `air_brake`
  - The actuator range can be clamped to limit
  - The actuator will be saturated to the rate limit (if the actuator `demand_rate` (a.k.a. `sampling_rate`) is discrete)
  - The actuator dynamics is modelled as a 1st-order low pass IIR filter  (if the actuator `demand_rate` (a.k.a. `sampling_rate`) is discrete)
- Inheriting the  `Actuator` base class are
  - `throttle`: an actuator that applies a throttle gain value on top of motor thrust. 
  - `roll`: an actuator that applies a magic / hand-of-god roll torque on the rocket
  - `thrust vector`: actuators that gimbal the motor thrust vector on x and y axis

- New files:
  - actuator classes in new `/actuator` folder
  - prints for actuators
  - test_actuator (generated by copilot)
  - example applied on `halcyon_flight_sim`
- Modified files:
  - `rocket.py`: implement add_actuator methods
  - `flight.py`: apply actuators on the ODEs

__known issue__
- Throttling doesn't change the propellant mass flow. It is essentially throttling the ISP of motor.

## Breaking change

- [ ] Yes
- [X] No

## Additional information

co-authored-by: RickyRicato <rickywang44@gmail.com> @chichunwang

PR #968, opened 2026-06-15
ENH: Events Class and Flight Rework

MateusStano:
## Summary

The addition of custom events was long overdue. In this PR the Event class was added. It was basically impossible to do this without changing a lot of the flight class, so I took the opportunity to rework it and solve all TODOs there.

Several upcoming features will build on top of this (parafoil, multistage, more controllers...), so I want this branch to act as a hub for a few related changes:

- [ ] Make `class Event` serializable
  - [ ] Fix broken tests (broken due to fails on controller serialization)
- [ ] Rework controllers (minor) so they are better designed for use with specific classes, like what is done in #965.
- [ ] Make `flight.solution` an instance of a new `Solution` class. It should behave like the current array, but also allow a variable state length. This is important for adding new parachute models such as parafoil, and for letting us define more complex derivative functions.
- [ ] Aerodynamic classes refactor to make `GenericSurface` the mother class of all other aero surfaces.

Here is a summary of all the changes:

## `flight.py`

Some flight class methods were reworked, some were moved:

- The flight now moves forward using **events**, not a fixed list of phases.
- `flight_derivatives.py`: the motion functions (`u_dot`, `u_dot_generalized`,
  `u_dot_generalized_3dof`, rail and parachute), moved out of `flight.py`.
- `flight_phase.py`: now has the `_FlightPhases` and `_TimeNodes` classes.
- `event_calling.py`: utilities for callling the events in the flight class correctly
- `event_commands.py`: applies the commands an event returns back to the flight.
- a propper logging feature was added

## New event code (`rocketpy/simulation/events/`)

- `event.py` (`Event`): An event has a trigger and a callback. The trigger is checked each step; when it is true, the callback runs. Callbacks can read the state, save data, log results, and send commands that change the simulation. It comes with ready-made presets like `apogee` and `burnout`. It also only computes expensive values when an event actually needs them.
- `commands.py` (`Commands`): what a callback can ask for. An event can start a
  new phase, swap the motion function, turn other events on or off, add or
  remove controllers, undo a step (rollback), or stop the flight.
- `event_builders.py`: ready-made events that recreate the normal flight steps
  (like leaving the rail).
- `exact_time_solvers.py`: finds the exact time an event happens inside a step
  (using a few math methods).

Parachutes, controllers and sensors are no longer special cases in the flight loop. Each one now has a `to_event()` method that wraps it into an `Event`. The built-in flight milestones (apogee, out of rail, impact) are now events too. 

Parachute triggers and Controllers controller_function now take `**kwargs` only and read what they need by name (`pressure`, `height_agl`, `state`, `sensors`, etc.); the old positional parachute and controller signatures still work but raise a `DeprecationWarning`.

`Flight.__init_events()` just collects all events (core events, then sensors, parachutes, controllers, and user custom events) and runs them uniformly, and commands an event returns (like swapping the derivative or starting a phase) are applied back to the flight by `event_commands.py`.

Parachute noise was also removed. The `noise` parameter on `Parachute` is now deprecated and has no effect (removal in v1.13). To model a noisy trigger, a `Barometer` should be used and accesses the noisy measurement in the trigger via `kwargs['sensors_by_name']`. This fits the new model, where sensors are events and the trigger reads their measurements.

## Plots and prints

- `flight_plots.py` and `flight_prints.py` now show sensor data too.
- `compare_flights.py` updated.
- New plots to all_info, and now they show case events:

image

## Docs

- New guides: `docs/user/event_usage.rst` and `docs/user/sensors_usage.rst`.
- New technical page: `docs/technical/simulation_loop.rst`.
- Update all places that used old definition of trigger functions (with positional arguments) to now be used with **kwargs

## Notes for reviewers

This will be a difficult one to review, specially because the diff tracking in flight.py is not that helpful given I changed the order of a few methods. So I suggest that the best way to understand how events work is to read the user guide: [`docs/user/event_usage.rst`. It walks through the API with runnable examples.](https://docs.rocketpy.org/en/enh-events/user/event_usage.html)

## Breaking change

- [x] No (I hope)


rocketpy/rocket/aero_surface/generic_surface.py
```diff
@@ -260,17 +262,32 @@ def compute_forces_and_moments(
...
-            Tuple containing angular velocities around the x, y, z axes.
+        density : Function
+            Atmospheric density as a function of altitude. Used to compute the
+            Reynolds number at the surface altitude.
+        dynamic_viscosity : Function
+            Atmospheric dynamic viscosity as a function of altitude. Used to
+            compute the Reynolds number at the surface altitude.
+        z : float
+            Altitude of the surface, used to evaluate ``density`` and
+            ``dynamic_viscosity``.
```
Gui-FernandesBR, line 273: let it clear if its above sea level ou ground level please 

```python
        z : float
            Altitude of the surface, used to evaluate ``density`` and
            ``dynamic_viscosity``.
```

docs/technical/simulation_loop.rst
phmbressan, line 1: <details>
<summary>
A diagram like below would probably help this page a lot.
</summary>

```mermaid
flowchart TD
    A["Initialize Flight"] --> B["Build event set<br/>and flight phases"]
    B --> C["Start phase<br/>create solver and time-node schedule"]
    C --> D["Advance solver<br/>and store state"]
    D --> E["Evaluate events<br/>scheduled, continuous, or interpolated"]
    E --> F["Run triggered callbacks"]
    F --> G["Apply Commands<br/>update events, dynamics, phases, or termination"]

    G --> H{"Next action"}
    H -- Continue phase --> D
    H -- Roll back --> I["Restore event state<br/>and restart solver"]
    I --> D
    H -- Start new phase --> C
    H -- Flight complete --> J["Finalize and cache results"]

    classDef main fill:#17324d,color:#fff,stroke:#17324d;
    classDef process fill:#e8f1f8,color:#17324d,stroke:#5b84a6;
    classDef event fill:#fff3d6,color:#5b4100,stroke:#d39b24;
    classDef decision fill:#fff,color:#333,stroke:#666;

    class A,B,J main;
    class C,D,I process;
    class E,F,G event;
    class H decision;
```

</details>

<details>
<summary>
Or a more complex/comprehensive:
</summary>

```mermaid
flowchart TD
    A["Flight.__init__"] --> B["Initialize events, equations of motion,<br/>solution history, and solver settings"]

    B --> C["Classify events by priority"]
    C --> C1["Non-overshootable<br/>scheduled nodes or continuous checks"]
    C --> C2["Sampled + overshootable<br/>checked through interpolation"]

    C1 --> D["Flight.__simulate"]
    C2 --> D

    D --> E["_FlightPhases<br/>initial phase + max-time sentinel"]
    E --> F{"Next flight phase?"}

    F -- Yes --> G["Create phase solver<br/>derivative, initial state, time bound"]
    G --> H["_TimeNodes builds phase schedule<br/>phase bounds + sampled events"]
    H --> I{"Next time node?"}

    I -- Yes --> J["Set solver bound to next node"]
    J --> K["Evaluate events scheduled<br/>at current node"]

    K --> L["Event pipeline<br/>trigger → optional exact-time refinement<br/>→ callback → Commands"]
    L --> M["Apply commands"]

    M --> M1["Enable, disable, or add events<br/>update _TimeNodes"]
    M --> M2["Set derivative or start phase<br/>insert _FlightPhase and finish current solver"]
    M --> M3["Terminate flight<br/>truncate phases and finish solver"]

    M1 --> N["Re-sync solver bound if<br/>node schedule changed"]
    N --> O
    M2 --> O
    M3 --> O

    O{"Solver still running?"}
    O -- Yes --> P["solver.step"]
    P --> Q["Store time, state, and<br/>function-evaluation count"]

    Q --> R["Build sampled event nodes between<br/>previous and current solver times"]
    R --> S["Use dense output to evaluate<br/>interpolated states"]

    S --> T{"Overshoot event<br/>requires rollback?"}
    T -- Yes --> U["Restore interpolated<br/>time and state"]
    T -- No --> V
    U --> V["Combine triggered overshoot events<br/>with continuous events"]

    V --> W["Evaluate by priority through<br/>the Event pipeline"]
    W --> X{"Rolled back, but phase<br/>and derivative unchanged?"}

    X -- Yes --> Y["Rebuild solver from<br/>rolled-back state"]
    X -- No --> Z["Continue with current solver state"]
    Y --> AA
    Z --> AA

    AA["Derivative post-processing<br/>when dynamics-changing events exist"]
    AA --> O

    O -- No --> I
    I -- No --> F
    F -- No --> AB["Finalize simulation<br/>cache outputs and sensor data"]

    classDef flight fill:#17324d,color:#fff,stroke:#17324d;
    classDef structure fill:#e8f1f8,color:#17324d,stroke:#5b84a6;
    classDef event fill:#fff3d6,color:#5b4100,stroke:#d39b24;
    classDef command fill:#f5e8f7,color:#53245c,stroke:#9b59a5;
    classDef decision fill:#fff,color:#333,stroke:#666;

    class A,B,D,AB flight;
    class E,G,H,J,P,Q,R,S,Y,AA structure;
    class C,C1,C2,K,L,V,W event;
    class M,M1,M2,M3,N,U command;
    class F,I,O,T,X decision;
```

</details>

docs/user/event_usage.rst
```diff
@@ -0,0 +1,973 @@
...
+      
+      print(f"Flight time = {simple_flight.t:.4f} s\n")
+      
+      print(f"Callback log: {simple_event.callback_log}\n")
+      
+      print(simple_event)
+
+**trigger** (optional)
+  A callable that returns ``True`` when the event should fire. If ``None``, the
+  event acts as a passive hook and always triggers when called.
```
phmbressan, line 203: I think one could elaborate more on the meaning of `always` here. I would phrase it like this (if I understood correctly): 

    If `None`, the `Event` `callback` is executed unconditionally when 
    sampled by the simulation (see `sampling_rate` below). Therefore
    this `Event` acts as a passive hook.

docs/user/event_usage.rst
```diff
@@ -0,0 +1,973 @@
...
+  Automatically enable a disabled event based on a condition. Uses the same
+  formats as ``disable_on``:
+
+  - String preset: ``"apogee"`` or ``"burnout"``
+  - Simulation time threshold
+  - Callable predicate
+
+  .. tip::
+    The times when the event is enabled through ``enable_on`` are recorded in
+    the list ``event.enabled_times`` for later inspection.
```
phmbressan, line 448: Which has precedence if both `disable_on` and `enable_on` simultaneously trigger?

docs/user/sensors_usage.rst
```diff
@@ -0,0 +1,569 @@
...
+    calisto.draw(plane="yz")
+
+Using Sensors With a Controller (Air Brakes)
+--------------------------------------------
+
+The main motivation for simulating sensors is to feed their measurements into a
+controller, such as an air brakes system, exactly as the real avionics would.
+Inside a controller function, every sensor attached to the rocket is available
+through the ``sensors`` keyword argument (a list, in the order they were added)
+and through ``sensors_by_name`` (a dictionary keyed by the sensor ``name``).
```
phmbressan, line 225: Why both? If it really eases the use, I understand, but it seems a bit much, since the two parameters are so similar, if I got it correctly: `sensors == sensors_by_name.values()`. Personally, I would keep only the `dict` version, but maybe I am out of touch here.

docs/user/sensors_usage.rst
```diff
@@ -0,0 +1,569 @@
...
+
+``flight.sensor_data`` is a dictionary keyed by the sensor object. Each value is
+a list of ``(time, *measurement)`` tuples (or a list of such lists if the same
+sensor object was added to the rocket in more than one position):
+
+.. jupyter-execute::
+
+    data = test_flight.sensor_data[accelerometer]
+    print("Number of accelerometer samples:", len(data))
+    print("First sample (t, ax, ay, az):", data[0])
```
phmbressan, line 335: If (big if) this is the only place the actual `Sensor` object is used for keying values, I would probably consider adding a method to the `Flight` class (even though I would generally advise against it so as to avoid bloat). Something akin to:

```python
def get_sensor_values(self, sensor: Sensor):
    """get values/can be reused across reruns"""
    try:
        return self.sensor_data[sensor]
    except KeyError as e:
        raise ... no sensor ... found ... from e
```

This is, since naming/parameter names is becoming a pretty common key across a few uses, it wouldn't surprise me one attempting:

```python
my_flight.sensor_data["my_accelerometer"]
```

Non blocking, since the correct way is already documented.

rocketpy/control/controller.py
```diff
@@ -1,216 +1,382 @@
...
+            Condition that automatically re-enables a disabled controller,
+            using the same formats as ``disable_on``. When the condition is met
+            while the controller is disabled, it re-enables before the next
+            trigger evaluation. Defaults to ``None`` (no automatic enabling).
+        controller_needs : list or frozenset of str or None, optional
+            Declares which expensive simulation values the controller function
+            accesses. Valid keys: ``'state_dot'``, ``'pressure'``,
+            ``'state_history'``. When ``None`` (default), all values are
+            computed on every call. Pass an explicit list or frozenset to skip
+            computing values the controller does not use.
```
phmbressan, line 138: This works opposite to the `Event`, right? Passing arguments saves simulation time here, instead of adding as in `Event`. Was this made to avoid a braking change due to the earlier default parameters of the `Controller` (i.e. it required `state_history`)? I don't think the earlier `Controller` required `pressure` and `state_dot` though, so could the default here be `self.controller_needs = controller_needs or ["state_history"]`?

rocketpy/control/controller.py
```diff
@@ -1,216 +1,382 @@
...
+            self.event.callback_log = value
+
+    @property
+    def return_log(self):
+        """Alias for :attr:`log`."""
+        return self.log
+
+    @return_log.setter
+    def return_log(self, value):
+        self.log = value
```
phmbressan, line 308: I believe this is an unnecessary duplicate.

rocketpy/rocket/parachute.py
```diff
@@ -128,6 +128,7 @@ def __init__(
         height=None,
         porosity=0.0432,
         drag_coefficient=1.4,
+        trigger_needs=None,
```
phmbressan, line 131: Consider putting a `, *,` parameter before it if you want it to always be last and not positionally usable.

This might apply to other classes also.

rocketpy/rocket/parachute.py
```diff
@@ -258,61 +260,101 @@ def __init_noise(self, noise):
...
+        if noise[0] != 0 or noise[1] != 0 or noise[2] != 0:
+            warnings.warn(
+                "The `noise` parameter on Parachute is deprecated and has no "
+                "effect; it will be removed in v1.13. Use a Sensor (e.g. a "
+                "Barometer) with built-in noise instead, and read the noisy "
+                "measurement via `kwargs['sensors_by_name']` in your trigger "
+                "function.",
+                DeprecationWarning,
+                stacklevel=3,
+            )
```
phmbressan, line 272: I believe the `noise` parameter is here for too long to remove it that quickly and also in a class that has no `_` prefix.

rocketpy/simulation/events/exact_time_solvers.py
```diff
@@ -0,0 +1,282 @@
...
+
+    if y0 == y1:
+        raise ValueError(no_root_error_message)
+
+    try:
+        event_time = brentq(f, t0, t1, xtol=xtol, rtol=rtol, maxiter=maxiter)
+    except Exception as e:
+        raise ValueError(no_root_error_message) from e
+
+    return {"event_time": event_time, "event_state": interpolator(event_time)}
```
phmbressan, line 194: Could there be any benefits here on making use of different [`scipy` root finders](https://docs.scipy.org/doc/scipy-1.16.1/reference/optimize.html#scalar-functions) via the [`root_scalar`](https://docs.scipy.org/doc/scipy-1.16.1/reference/generated/scipy.optimize.root_scalar.html#scipy.optimize.root_scalar)?

rocketpy/simulation/flight.py
```diff
@@ -569,16 +584,23 @@ def __init__(  # pylint: disable=too-many-arguments,too-many-statements
...
         name : str, optional
             Name of the flight. Default is "Flight".
         equations_of_motion : str, optional
             Type of equations of motion to use. Can be "standard" or
             "solid_propulsion". Default is "standard". Solid propulsion is a
             more restricted set of equations of motion that only works for
-            solid propulsion rockets. Such equations were used in RocketPy v0
-            and are kept here for backwards compatibility.
+            solid propulsion rockets. Solid motors are automatically mapped to
+            "solid_propulsion".
```
phmbressan, line 603: Haven't we attempted this automatic mapping before and there were marginal differences in output? Also, two comments:

- Correct me if I am wrong, but isn't the largest difference between the two sets of equation the inertias and their derivatives? Can't we allow the usage of the simpler set to any rocket, with the warning for other kind of motors that it is less accurate?
- Can't we merge the two sets and simply put an if clause on the current `u_dot_generalized` to replicate the older `u_dot` (e.g. skip products of inertia computation)?

phmbressan:
Rebasing this branch with `develop` will really be something else. Once this PR is done, I would even consider squashing and cherry-picking on top of `develop`.

Overall, the new `Event` API is outstanding, the possibilities of new/complex simulations scenarios are rather interesting.

As it was mentioned in the PR description, a `Solution` (or analogous) class, with quickly accessible attributes would ease the use of indexing quite a bit. If the class has a closed-set of attributes (which I think it is likely), I recommend `__slots__` to make it slightly faster and memory efficient.

docs/user/event_usage.rst
MateusStano: Done

docs/user/event_usage.rst
MateusStano: enable_on is checked before the trigger and disable_on after the callback, so if both are true on the same check the event is enabled, fires, and is disabled again. Now documented under enable_on and covered by a unit test.

docs/user/sensors_usage.rst
MateusStano: I think keeping both makes sense. Sensors is a list in the order the sensors were added, which is what the positional-argument controller path and the parachute trigger receive, and sensors_by_name is the same objects keyed by name for lookups. Both are just attributes of the Flight, so I think it costs nothing to expose them

rocketpy/control/controller.py
MateusStano: `needs` were removed with most recent commits

rocketpy/control/controller.py
MateusStano: controller.py will get its refactor on another PR. Ill leave this here

rocketpy/simulation/events/exact_time_solvers.py
MateusStano: Maybe, but I dont think it is needed. It would be easy to add if we ever feel the need though

rocketpy/simulation/flight.py
MateusStano: The main difference isn't the inertia products, it's the mass model. u_dot assumes the propellant CM stays at a fixed position in the body, and this is only valid for solids.
I think the automatic mapping has been in the code for a while also. For solids the difference in results is very very small. Also I think you can use the simpler model for any rocket, just have to set it mannualy

docs/reference/classes/Solution.rst
```diff
@@ -0,0 +1,57 @@
...
+----------------
+
+These classes hold the state history produced by a :class:`rocketpy.Flight`
+simulation. They are not part of RocketPy's top-level namespace; a flight's
+solution is reached through :attr:`Flight.solution <rocketpy.Flight.solution>`.
+
+The ``Solution`` holds every row of the flight in one list, in the order they
+were flown. The flight phases sit alongside it, each saying where its own rows
+begin. So a row is read from the solution, and a phase tells you what that row
+means::
```
phmbressan, line 11:
```suggestion
means:
```

docs/reference/classes/Solution.rst
```diff
@@ -0,0 +1,57 @@
...
+
+.. autoclass:: rocketpy.simulation.solution.Solution
+   :members:
+
+Post-process variables
+~~~~~~~~~~~~~~~~~~~~~~
+
+On its way to each state derivative, a flight phase works out quantities it
+never integrates: the accelerations, the aerodynamic forces and moments, and
+the net thrust. Read them through ``flight.solution.post``::
```
phmbressan, line 29:
```suggestion
the net thrust. Read them through ``flight.solution.post``:
```

rocketpy/simulation/solution.py
```diff
@@ -0,0 +1,1532 @@
...
+
+    It also reads like a plain list of rows: ``len(solution)``, iteration,
+    ``solution[-1]``, slicing and ``numpy.array(solution)`` all give 14-value
+    canonical rows ``[t, x, y, z, vx, vy, vz, e0, e1, e2, e3, w1, w2, w3]``,
+    whatever the phase they came from actually integrated.
+
+    Writing is done by the simulation alone, through the private methods
+    below. They take the raw states of the phase being flown, since that is
+    what the integrator produces.
+    """
```
phmbressan, line 301: Do you think having the solution variables as static attributes would be a major performance hit or maintainability concern?

In my head I underestimated a bit the amount of operations this `Solution` class actually performs, but I was conjecturing something more akin to a `dataclass`, e.g. `solution.vy` should I want the `vy`, instead of using string keyed dicts, since it is better suited to static code completion.

I get that this might be complicated here by the presence of the phase, those would probably be still string keyed. But maybe having a `State` dataclass would be a not too difficult compromise:

```py
solution.phases[-1].state.vy

event_context.state.vy
# As opposed to event_context["state"][4]
```

The ease of use needs also depends on how much the user will interact with the `Solution` class, so perhaps it is not important here. 

This is also my question to the event context that is frequently used by _user defined_ triggers.

If the string keyed approach is the most reasonable way, I understand, but we should at least consider the usability difference, between someone having to count array indexes or read the documentation (or a `.available_names`) for the exact string name of the attribute, as opposed to simply typing `context.` and having a list of names by the code editor.

rocketpy/simulation/solution.py
```diff
@@ -0,0 +1,1532 @@
...
+
+        Returns
+        -------
+        dict
+            A format marker, the flight's rows, and one entry per phase as
+            produced by :meth:`_PhaseSolution.to_dict`. Pass it to
+            :meth:`from_dict` to rebuild the solution.
+        """
+        return {
+            "format": "rocketpy/solution",
```
phmbressan, line 1179: Where is this being used? It seems similar to the `signature` parameter that is automatically added by the `RocketPyEncoder` to rocketpy classes.

phmbressan (commented):
Is the intention of the PR to allow the action test runners to run? If yes, I would already remove the draft.

PR #970, opened 2026-06-16
ENH: Add custom exceptions and unstable rocket warning (#285)

### LINKED ISSUE #285: Custom Exception errors and messages

After years using RocketPy, we already know what are common errors and mistakes that usually new users tends to do. Of course we will always try to improve our docs based on that experience, but I think that we could think on how to avoid the frustraction of not having a clue of what is wrong with your simulation

Create dedicated exception and warning messages for common problems as I mentioned above. An example would be the static margin with negative values, we should alert the user right after he/she creates the rocket object.

**Additional context**

TBD further researches on this topic.

Matanski:
Closes #285

## Changes
- Added `rocketpy/exceptions.py` with three new types:
  - `InvalidParameterError` – raised when `Rocket` is constructed with non-positive `radius` or `mass`
  - `InvalidInertiaError` – raised when the `inertia` tuple has wrong length (not 3 or 6)
  - `UnstableRocketWarning` – issued automatically when static margin is negative at motor ignition
- Added input validation in `Rocket.__init__`
- Added stability check in `evaluate_static_margin` 
- Exported all three names from the top-level `rocketpy` package

## Testing
All existing unit tests pass (`106 rocket + 70 motor tests`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ Attention: The newest changes should be on top -->
 
 ### Added
 
+- ENH: Add custom exceptions and unstable rocket warning [#970](https://github.com/RocketPy-Team/RocketPy/pull/970)
 - ENH: MNT: Remove redundant wind_heading/wind_direction functions from EnvironmentAnalysis [#1041](https://github.com/RocketPy-Team/RocketPy/pull/1041)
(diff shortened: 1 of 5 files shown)


Gui-FernandesBR:
@Matanski a previous PR alraedy tried to solve the same problme but got rejected. Do you think your implementation is different (or better) than #764

Gui-FernandesBR (commented):
only pending CHANGELOG.md update

PR #947, opened 2026-03-31
ENH: Monte Carlo Formatting Options (#945)

khushalkottaru:
<!-- You are awesome! Your contribution to RocketPy is fundamental in our endeavour to create the next generation solution for rocketry trajectory simulation! -->

<!-- You may use this template to describe your Pull Request. But if you believe there is a better way to express yourself, don't hesitate! -->

## Pull request type
<!-- Remove unchecked box items. -->

- [x] Code changes (bugfix, features)

## Checklist
<!-- Remove irrelevant items to this PR. -->

- [x] Tests for the changes have been added (if needed)
- [ ] Docs have been reviewed and added / updated
- [x] Lint (`black rocketpy/ tests/`) has passed locally 
- [x] All tests (`pytest tests -m slow --runslow`) have passed locally
- [ ] `CHANGELOG.md` has been updated (if relevant)

## Current behavior
<!-- Describe current behavior or link to an issue. -->

The MonteCarlo class only reads and writes .txt files (JSONL format, one JSON object per line) for simulation inputs, outputs, and errors. Users who want to analyze results in Excel or other tools must manually convert these files to CSV or JSON themselves.

## New behavior
<!-- Describe changes introduced by this PR. -->

The MonteCarlo class can now directly export simulation results to CSV and JSON formats, and import previously exported CSV/JSON files back in. New methods:                                                                                                                                                                                                                                                                    
   
  - export_outputs_to_csv(filename) / export_outputs_to_json(filename)                                                                                                                                                                                                                                                                                                                                                            
  - export_inputs_to_csv(filename, flatten=False) / export_inputs_to_json(filename)
  - export_errors_to_csv(filename, flatten=False) / export_errors_to_json(filename)                                                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                                                                                  
  The existing .txt (JSONL) format remains the default for simulation output since it supports streaming writes. CSV/JSON are post-processing export targets. The import_outputs, import_inputs, and import_errors methods also now accept .csv and .json files in addition to .txt.

## Breaking change
<!-- If this introduces a breaking change, please describe the impact and migration path for existing applications below. Remove the unchecked box item. -->

- [ ] Yes
- [x] No

## Additional information
<!-- Include any relevant details or screenshots. If none, remove this section. -->
Related to issue #945.

PR #971, opened 2026-06-16
ENH: Add export_results_to_csv and export_results_to_json to MonteCar…

Matanski:
## Description

Closes #945

Adds two new public methods to the `MonteCarlo` class so users can export
simulation results directly to CSV or JSON without manually converting the
existing `.txt` output files.

### New methods

- `export_results_to_csv(filename=None)` — writes one column per variable
  and one row per simulation run. Defaults to `<filename>.results.csv`.
- `export_results_to_json(filename=None)` — writes the `results` dict as a
  pretty-printed JSON object. Defaults to `<filename>.results.json`.

Both methods work on results from a freshly run simulation or from previously
imported output files, and accept an optional custom output path.

### Example usage

```python
mc = MonteCarlo(filename="my_sim", ...)
mc.simulate(100)

mc.export_results_to_csv()          # → my_sim.results.csv
mc.export_results_to_json()         # → my_sim.results.json
mc.export_results_to_csv("out.csv") # → custom path

### Changes
rocketpy/simulation/monte_carlo.py: added export_results_to_csv and
export_results_to_json methods; added import csv at the top.
### Testing
All 114 existing unit tests pass. The new methods were manually verified to
produce correct CSV headers/rows and valid JSON output.

### DIFF
--- a/rocketpy/simulation/monte_carlo.py
+++ b/rocketpy/simulation/monte_carlo.py
@@ -13,6 +13,7 @@ change in future versions. Users are encouraged to check for updates and read th
 latest documentation.
 """
 
+import csv
 import json
 import os
 import traceback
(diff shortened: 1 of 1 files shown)


Gui-FernandesBR:
duplicated with #947, which has just been merged

PR #973, opened 2026-06-19
ENH: Add python loggers

### LINKED ISSUE #450: ENH: use of built-in python loggers instead of prints

YuriCastroDev:
## Pull request type

## Checklist

## Current behavior

RocketPy uses bare `print()` calls scattered across the codebase to report simulation progress, warnings, and errors. This approach has no severity levels, cannot be redirected to a file, and cannot be silenced or filtered by the end user. Closes #450.

## New behavior

Internal runtime events now use Python's built-in `logging` module following best practices for libraries, while user-requested output (`.info()`, `.all_info()`) continues to use `print()` unchanged.

**What changed:**

- `rocketpy/__init__.py` registers a `NullHandler`, so the library is **silent by default** with no output unless the user configures a handler.
- 27 execution modules now expose `logger = logging.getLogger(__name__)`, creating a hierarchy (`rocketpy.simulation.flight`,  `rocketpy.environment.environment`, etc.) that users can filter individually.
- Log levels assigned according to severity:
  - `DEBUG`: high-frequency solver ticks, internal detail
  - `INFO`: operation confirmations (simulation completed, files saved)
  - `WARNING`: unexpected but recoverable situations (missing motor, geometry auto-corrected, feature silently skipped)
  - `ERROR`: optional dependency unavailable, feature broken
- `rocketpy/prints/` folder kept as `print()` methods like `flight.info()`
  and `rocket.all_info()` continue to print directly to the terminal as before.
- `_SimMonitor.reprint()` in `MonteCarlo` kept for backwards compatibility and now delegates to `logger.info()`.
- New `rocketpy/utils.py` with `enable_logging()` helper function so users can activate logging with a single line:

```python
import rocketpy
rocketpy.utils.enable_logging(level="INFO")
```

Note on verbose parameter: Flight(verbose=True) retains its original real-time terminal progress ticker. The same information is also emitted via logger.debug(), so users who configure a logging handler will capture it as well.

Tests updated: 3 existing tests migrated from capsys to caplog where the output source changed from print() to logger.*. 6 new unit tests added for rocketpy.utils.enable_logging().

## Breaking change

- [x] No

## Additional information

- 31 files modified (27 source + 3 tests + 1 new utils.py).


Gui-FernandesBR (commented):
does it change current behavior?

If I open the getting start notebook, does it work just like as it should?

YuriCastroDev:
@Gui-FernandesBR yes, it does change current behavior. The getting started notebook would not work the same because .info() / .all_info() calls are now silent due to the nullhandler. To get the same output, the user would need to configure a handler before running simulation.

Gui-FernandesBR:
> @Gui-FernandesBR yes, it does change current behavior. The getting started notebook would not work the same because .info() / .all_info() calls are now silent due to the nullhandler. To get the same output, the user would need to configure a handler before running simulation.

yes that's my concern... Is there a way where we keep the same behavior for current code, but also benefit from built-in loggers without duplicating code? thta's the real challenge

YuriCastroDev:
Hm... makes sense. Keeping print() for explicitly user-requested output 
like flight.info() and reserving logger for internal runtime events 
(simulation completed, file saved, motor overwritten, etc.) follows 
the Python logging guide's own distinction. 

What do u think?

Gui-FernandesBR:
> Hm... makes sense. Keeping print() for explicitly user-requested output like flight.info() and reserving logger for internal runtime events (simulation completed, file saved, motor overwritten, etc.) follows the Python logging guide's own distinction.
> 
> What do u think?

yes but that would imply in us duplicating code verytime we want to print a message, right?

I wonder how other simulator libraries deal with the same situation.

If you could bring us some references so we can base our opinion on different examples... I'd appreciate it a lot.

YuriCastroDev:
I'll look into other libraries for references, once I have the answer I'll come back, tks for review

Gui-FernandesBR:
@YuriCastroDev you let me know when this PR is ready for review, ok? Thx a lot for your contribution!

YuriCastroDev:
@Gui-FernandesBR The PR is ready for your review! Thanks for your support! :)

rocketpy/environment/environment.py
```diff
@@ -2973,6 +2977,6 @@ def from_dict(cls, data):  # pylint: disable=too-many-statements
 
     results = doctest.testmod()
     if results.failed < 1:
-        print(f"All the {results.attempted} tests passed!")
+        logger.info("All the %d tests passed!", results.attempted)
```
Gui-FernandesBR:
```suggestion
        logger.debug("All the %d tests passed!", results.attempted)
```

rocketpy/environment/environment.py
```diff
@@ -2973,6 +2977,6 @@ def from_dict(cls, data):  # pylint: disable=too-many-statements
...
     else:
-        print(f"{results.failed} out of {results.attempted} tests failed.")
+        logger.warning("%d out of %d tests failed.", results.failed, results.attempted)
```
Gui-FernandesBR:
```suggestion
        logger.error("%d out of %d tests failed.", results.failed, results.attempted)
```

rocketpy/environment/tools.py
```diff
@@ -850,6 +854,6 @@ def utm_to_geodesic(  # pylint: disable=too-many-locals,too-many-statements
 
     results = doctest.testmod()
     if results.failed < 1:
-        print(f"All the {results.attempted} tests passed!")
+        logger.info("All the %d tests passed!", results.attempted)
     else:
-        print(f"{results.failed} out of {results.attempted} tests failed.")
+        logger.warning("%d out of %d tests failed.", results.failed, results.attempted)
```
Gui-FernandesBR:
```suggestion
        logger.error("%d out of %d tests failed.", results.failed, results.attempted)
```

rocketpy/mathutils/function.py
```diff
@@ -4405,6 +4409,6 @@ def reset_funcified_methods(instance):
 
     results = doctest.testmod()
     if results.failed < 1:
-        print(f"All the {results.attempted} tests passed!")
+        logger.info("All the %d tests passed!", results.attempted)
     else:
-        print(f"{results.failed} out of {results.attempted} tests failed.")
+        logger.warning("%d out of %d tests failed.", results.failed, results.attempted)
```
Gui-FernandesBR:
```suggestion
        logger.error("%d out of %d tests failed.", results.failed, results.attempted)
```

rocketpy/mathutils/vector_matrix.py
```diff
@@ -1108,6 +1111,6 @@ def from_dict(cls, data):
 
     results = doctest.testmod()
     if results.failed < 1:
-        print(f"All the {results.attempted} tests passed!")
+        logger.info("All the %d tests passed!", results.attempted)
     else:
-        print(f"{results.failed} out of {results.attempted} tests failed.")
+        logger.warning("%d out of %d tests failed.", results.failed, results.attempted)
```
Gui-FernandesBR:
```suggestion
        logger.error("%d out of %d tests failed.", results.failed, results.attempted)
```

rocketpy/simulation/monte_carlo.py
```diff
@@ -418,9 +421,9 @@ def __sim_producer(self, seed, sim_monitor, mutex, error_event):  # pylint: disa
                 try:
                     mutex.acquire()
                     if error_event.is_set():
-                        sim_monitor.reprint(
-                            "Simulation Interrupt, files from simulation "
-                            f"{sim_idx} saved."
+                        logger.warning(
+                            "Simulation interrupt. Files from simulation %d saved.",
+                            sim_idx,
```
phmbressan: @YuriCastroDev 

Have you checked if these changes behave well regarding parallel monte carlo?

The sim_monitor was in use to have a shared instance that makes sure concurrent messaging will not end up mangled in the shell output. Perhaps the logging API can do this out of the box, I am not sure.

One way to check would be either:

1. Running the slow test that specifically has a parallel monte carlo run with a `-s` flag to capture the output and keep an eye on the logged messages. E.g.:

```bash
pytest -m slow -s --runslow tests/integration/simulation/test_monte_carlo.py
```

2. Or, run the notebook in `docs/notebooks/monte_carlo_analysis/monte_carlo_class_usage.ipynb`, which has a parallel monte carlo run.

From a very quick search, here is what the [Python docs](https://docs.python.org/3/howto/logging-cookbook.html#logging-to-a-single-file-from-multiple-processes) has to say about it. It seems that multiprocessing logging might not be straightforward.

YuriCastroDev: @phmbressan 

I'll check this and get back to you. Any other edge case you'd like me to verify besides the parallel monte carlo run, or is this the main concern?

rocketpy/plots/aero_surface_plots.py
Gui-FernandesBR, line 1: Do you agree that these message will stop making sense here?

The prints existed to separated each plot from the next one.
If we turn these into info logs, the user may not seeing them (unless they implement a handler).

On the other hand,  if a user consumes the log without seeing the plot, it doesn't really make sense at all (i.e. reading "airfoil lift curve" doesn't make as much sense as reading "plotting airfoil lift curve")

In other words, we need to rethink what we really want to display in each message within the plots module.

rocketpy/rocket/aero_surface/nose_cone.py
```diff
@@ -445,9 +449,10 @@ def final_shape(x):
...
             self._length = nosecone_x[-1]
-            print(
-                "Due to the chosen bluffness ratio, the nose "
-                f"cone length was reduced to {self.length} m."
+            logger.warning(
+                "Due to the chosen bluffness ratio, the nose cone length was "
+                "reduced to %.4f m.",
+                self.length,
             )
         self.fineness_ratio = self.length / (2 * self.base_radius)
```
Gui-FernandesBR, line 458: we are suprassing somehting really important here, no?

rocketpy/tools.py
```diff
@@ -1469,6 +1473,6 @@ def find_obj_from_hash(obj, hash_, depth_limit=None):
 
     res = doctest.testmod()
     if res.failed < 1:
-        print(f"All the {res.attempted} tests passed!")
+        logger.info("All the %d tests passed!", res.attempted)
     else:
-        print(f"{res.failed} out of {res.attempted} tests failed.")
+        logger.warning("%d out of %d tests failed.", res.failed, res.attempted)
```
Gui-FernandesBR: please search and replace in the entire  codebase...

```python
        logger.error("%d out of %d tests failed.", res.failed, res.attempted)
```

rocketpy/utils.py
Gui-FernandesBR, line 1: could you kindly add a documentation within docs/ folder teaching how to use it?

Gui-FernandesBR: the docstring already helps a lot, but beginners may benefit from a more detailed instruction, perhaps adding it to the getting started notebook could help!

Gui-FernandesBR (commented):
I already reviewd 23/31 files. there are some changes I already see as important.
But overall the Pr looks great!

rocketpy/simulation/monte_carlo.py
phmbressan: I believe the other comments already address anything else I had in mind. Good work on using "%" strings rather than f-strings, it should be better for performance.

Gui-FernandesBR:
@YuriCastroDev please let me know when you fix all the comments. 

I will mark the PR as "draft' once again

rocketpy/rocket/aero_surface/nose_cone.py
YuriCastroDev: Indeed, since this is a warning about an automatic change, I used UserWarning "the code worked, but something happened that you should know about".
Commit: https://github.com/RocketPy-Team/RocketPy/pull/973/changes/a29f05fcaaea1458a729c8d6373a5a9dd3ccc982

rocketpy/tools.py
YuriCastroDev: Ok, I found only one occurrence.
Commit: https://github.com/RocketPy-Team/RocketPy/pull/973/changes/a29f05fcaaea1458a729c8d6373a5a9dd3ccc982

rocketpy/utils.py
YuriCastroDev: Sure, it'll be a pleasure! After addressing all the remaining comments, I'll take care of it.

rocketpy/utils.py
```diff
@@ -0,0 +1,51 @@
+import logging
+
+
+def enable_logging(level="WARNING"):
+    """Enable RocketPy logging output to the console.
+
+    Attaches a StreamHandler to RocketPy's root logger so that internal
+    runtime events (simulation progress, warnings, errors) are printed to
+    the terminal. By default, only WARNING and above are shown.
```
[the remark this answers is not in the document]
YuriCastroDev: Commit: https://github.com/RocketPy-Team/RocketPy/pull/973/changes/21b4f7d1fc5b2f4f16e10c38d6f66ee42af7e66c

rocketpy/simulation/flight.py
```diff
@@ -777,6 +781,7 @@ def __simulate(self, verbose):
                     self.y_sol = phase.solver.y
                     if verbose:
                         print(f"Current Simulation Time: {self.t:3.4f} s", end="\r")
+                    logger.debug("Current Simulation Time: %3.4f s", self.t)
```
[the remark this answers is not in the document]
YuriCastroDev: Commit: https://github.com/RocketPy-Team/RocketPy/pull/973/changes/21b4f7d1fc5b2f4f16e10c38d6f66ee42af7e66c

rocketpy/simulation/flight_comparator.py
```diff
@@ -739,38 +745,35 @@ def summary(self):  # pylint: disable=too-many-statements
...
         None
         """
-        print("\n" + "=" * 60)
-        print("FLIGHT COMPARISON SUMMARY")
-        print("=" * 60)
-
-        print("\nRocketPy Simulation:")
-        print(
-            f"  - Apogee: {self.flight.apogee:.2f} m at t={self.flight.apogee_time:.2f} s"
+        logger.info("FLIGHT COMPARISON SUMMARY")
```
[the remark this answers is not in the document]
YuriCastroDev: Commit: https://github.com/RocketPy-Team/RocketPy/pull/973/changes/21b4f7d1fc5b2f4f16e10c38d6f66ee42af7e66c

rocketpy/simulation/monte_carlo.py
YuriCastroDev: @phmbressan 

I ran the verification as you suggested with:

`pytest -m slow -s --runslow tests/integration/simulation/test_monte_carlo.py`

The progress ticker was indeed silent because _SimMonitor runs in a separate process and logging doesn't work in that context. I fixed it by keeping print() in those methods, the output looks correct and all 4 tests passed.

Commit: https://github.com/RocketPy-Team/RocketPy/pull/973/changes/2fb3fe375dcb570096877c57ecd853ec8b418f3a

rocketpy/utils.py
YuriCastroDev: @Gui-FernandesBR 

I was thinking about adding something to the PR template, but haven't fully thought it out yet, my idea would be to remind reviewers to always check print() vs logger.* usage. What do u think? any ideas?

Gui-FernandesBR (approved):

PR #974, opened 2026-06-20
BUG: fix wind heading and direction wraparound interpolation

### LINKED ISSUE #253: BUG: Wind Heading Profile Plots are not that good
**Describe the bug**

Currently there's no way to prevent wrong interpolation of wind heading values when creating wind heading profile function. This causes plots to present weird behavior of a high variance of wind heading.
image

**To Reproduce**

Any time when the wind heading (or direction) varies more than 180ºfrom one pressure level to the other. Those case might present a wrong interpolation of wind heading values over the altitude, specially when plotting it.

**Expected behavior**

See the followinf link for reference.
https://stackoverflow.com/questions/16058096/conditionally-draw-connecting-lines-in-matplotlib
image

**Additional context**

Suggestion: Maybe a polar plot might help on the solution!

Gui-FernandesBR:
This fixes the issue where linear interpolation of wind heading and direction values over discrete altitudes caused unexpected spikes/wiggles around the 360/0 wraparound boundary.

Closes #253

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ Attention: The newest changes should be on top -->
 
 ### Fixed
 
+- BUG: fix wind heading and direction wraparound interpolation [#974](https://github.com/RocketPy-Team/RocketPy/pull/974)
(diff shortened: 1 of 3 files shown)


Gui-FernandesBR:
wind_wraparound

PR #1045, opened 2026-07-04
CI: create a CI for testing docs updates + solve different docs issues

Gui-FernandesBR:


.github/workflows/docs.yml
```diff
@@ -0,0 +1,104 @@
+name: Documentation
+
+on:
+  # Only PRs targeting master (base branch = master) and pushes to master.
+  # This is the gate: a PR cannot be merged into master without this check.
```
Gui-FernandesBR:
```suggestion
```

docs/reference/classes/aero_surfaces/_BaseFin.rst
```diff
@@ -1,3 +1,5 @@
+:orphan:
+
 _BaseFin Class
 --------------
```
Gui-FernandesBR, line 4: question: why do we need this here?

Gui-FernandesBR: :orphan: marks this page as intentionally not part of any toctree. _BaseFin is a private base class, so we don't want it in the nav (it's excluded from index.rst), but we keep the page so the Bases: _BaseFin cross-references from Fins/Fin resolve. Since the new docs CI builds with sphinx-build -W, the "document isn't included in any toctree" warning would otherwise fail the build — :orphan: silences exactly that.

PR #1048, opened 2026-07-04
Rel: v1.13.0

Gui-FernandesBR:
One of the biggest releases of RocketPy so far!!


rocketpy/utils.py
Gui-FernandesBR, line 1: Now I have notice that we are creating a new module named "utils.py", while we should use the already existing tools.py module

rocketpy/prints/aero_surface_prints.py
```diff
@@ -1,6 +1,7 @@
 from abc import ABC, abstractmethod
 
 
+# TODO: the rocketpy/prints/aero_surface_prints.py file could be separated into different, smaller files.
```
Gui-FernandesBR, line 4: I think we can solve this TODO before releasing v1.13, it is worth doing it

```python
# TODO: the rocketpy/prints/aero_surface_prints.py file could be separated into different, smaller files.
```

.github/renovate.json
Gui-FernandesBR, line 2: we need to check whether this file will conflict with the renovate file already present in the master branch

rocketpy/exceptions.py
Gui-FernandesBR, line 1: I could add more exceptions to this file and start using the customexceptions even more

docs/user/airbrakes.rst
```diff
@@ -372,6 +375,41 @@ controller function. If you want to disable this feature, set ``clamp`` to
     For more information on the :class:`rocketpy.AirBrakes` class
     initialization, see  :class:`rocketpy.AirBrakes.__init__` section.
 
+.. _discrete-vs-continuous-controllers:
+
+Discrete vs. Continuous Controllers
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Gui-FernandesBR: Maybe controllers and airbrakes should be splitted in the documentation pages

phmbressan (changes requested):
My local --runslow tests gave:

```bash
FAILED tests/integration/environment/test_environment.py::test_wyoming_sounding_atmosphere - AssertionError: assert np.float64(7725.0) < 1e-08
```

I cannot debug the changes right now, but it seems significant. Maybe related to unit conversions when reading `csv` or `API` responses.

rocketpy/prints/aero_surface_prints.py
MateusStano: I´d leave it for now, there will be bigger changes comming to aero_surfaces soon

rocketpy/utils.py
MateusStano: better to move this to utilities maybe

Gui-FernandesBR:
There are a lot of files... I read the most of them already, but it's safer if we spend a few days reviewing and testing around. I still plan to open the getting started notebook and run each cell just to verify everything is ok.

Gui-FernandesBR:
A quick checklist:
1. Run and fix slow tests (including warnings)
2. Build the documentation
3. Update CHANGELOG.mg one last time
4. Use Claude Fable to investigate possible bugs or errors

rocketpy/rocket/parachutes/hemispherical_parachute.py
```diff
@@ -0,0 +1,399 @@
...
+        cd_s,
+        trigger,
+        sampling_rate,
+        lag=0,
+        noise=(0, 0, 0),
+        radius=None,
+        height=None,
+        porosity=0.0432,
+        drag_coefficient=1.4,
+    ):
```
MateusStano, line 128: Since we are adding the class, we should remove the `noise` parameter, we have a much more robust and useful implementation of noise in the Sensors now and the way noise is implemented in the flight class is sub optimal (some lists get appended to at every step of the simulation for very little gain)

rocketpy/rocket/parachutes/hemispherical_parachute.py
```diff
@@ -0,0 +1,399 @@
...
+        name,
+        cd_s,
+        trigger,
+        sampling_rate,
+        lag=0,
+        noise=(0, 0, 0),
+        radius=None,
+        height=None,
+        porosity=0.0432,
+        drag_coefficient=1.4,
```
MateusStano, line 127: Shouldnt drag_coefficient already be set for the current geometry (Hemispherical)?

Gui-FernandesBR (commented):
LGTM!

rocketpy/rocket/parachutes/hemispherical_parachute.py
```diff
@@ -0,0 +1,399 @@
+import numpy as np
+
+from .parachute import Parachute
+
+
+class HemisphericalParachute(Parachute):
+    """Implements a hemispherical parachute.
+
+    Attributes
+    ----------
```
MateusStano, line 10: This is the only thing that worries me about this release. I am okay with the way the deprecation of `Parachute` class was handled, but I don´t currently see any benefit for this new class 

It makes sense to have classes for specific parachute geometries only if we have more than one geometry AND a more generic parachute case

When we had only the Parachute class, all users, independent of parachute geometry would use that one, and that was a good enough simplification. Now users will feel limited to the hemispherical geometry. An easy fix for this would be to make the Parachute class a generic parachute (that receives Volume instead of radius/height for example) instead of making it abstract. This would also reduce the level of breaking change

Gui-FernandesBR: Maybe a GenericParachute class?

PR #1052, opened 2026-07-06
ENH: seed sensor measurement noise per instance

### LINKED ISSUE #1042: Sensor noise uses the global NumPy RNG and can't be seeded (not reproducible)
I ran into this while driving the IMU/GNSS sensors in a seeded run, and wanted to check whether it's a known limitation before proposing anything.

### Current behaviour

The measurement noise for the sensors (white noise and random-walk bias drift in `Sensor`/`InertialSensor`/`ScalarSensor`, and the position/altitude accuracy in `GnssReceiver`) is drawn from the process-global NumPy RNG: `rocketpy/sensors/sensor.py` lines 556 and 561 (`InertialSensor.apply_noise`) and 766/774 (`ScalarSensor.apply_noise`), and `rocketpy/sensors/gnss_receiver.py` lines 93-95 (`GnssReceiver.measure`).

None of the sensor constructors take a `seed`, so there's no way to make the noise reproducible without reseeding the *global* `numpy.random` state by hand. That has two consequences. First, a simulation that's otherwise reproducible still produces different sensor measurements on every run; the rest of the stochastic stack already seeds explicitly (`StochasticModel` builds `np.random.default_rng(seed)` and `MonteCarlo` derives its streams from a `SeedSequence`), so the sensors are the one place that escapes it. Second, under `multiprocess` the child workers inherit a copy of the global RNG state, so sensor noise can come out correlated (or seed-independent) across parallel runs.

### Minimal example

```python
import numpy as np
from rocketpy.mathutils.vector_matrix import Vector
from rocketpy import Accelerometer

acc = Accelerometer(sampling_rate=100, noise_density=0.1, noise_variance=1.0)
print(acc.apply_noise(Vector([0, 0, 0])))  # changes every run; no `seed` to pin it
```

The only way to make this repeatable today is `np.random.seed(...)` on the global state, which also perturbs every other consumer of the global RNG and isn't safe across processes.

### Desired behaviour

Each sensor draws its noise from its own `numpy.random.Generator`, seeded deterministically, so the noise is reproducible for a given seed and independent of the global RNG. This matches what `StochasticModel` and `MonteCarlo` already do.

### Proposed solution

Add a `seed=None` argument to the sensor constructors, store `self._rng = np.random.default_rng(seed)`, and replace the `np.random.normal(...)` calls with `self._rng.normal(...)`. `seed=None` keeps the current "random by default" behaviour, but per instance, so it no longer reaches into the global RNG. Reproducibility then lives at construction, since a freshly-seeded sensor replays the same sequence. That matches how `StochasticModel` and `MonteCarlo` are already seeded (seed once, draw sequentially).

I've prototyped this with unit tests (reproducible from a seed, decorrelated across sensors, independent of the global RNG, and unchanged for the zero-noise ideal case) and would be glad to open a PR if this direction sounds reasonable. Happy to discuss the API first, for example whether the seed should live on the sensor or be threaded down from `add_sensor`/the rocket.

thc1006:
## Pull request type

## Checklist

- [x] Lint passes locally (`ruff check` and `ruff format --check`; pylint 10/10 on the touched files)
- [x] Sensor unit + integration tests pass locally (50 passing); relying on CI for the full slow suite
- [x] `CHANGELOG.md` updated

## Current behavior

Sensor measurement noise is drawn from the process-global NumPy RNG (`np.random.normal`), so there's no way to seed it. Noisy-sensor runs aren't reproducible, and the noise is fragile under parallel or forked execution where workers share or reset the global RNG. This is #1042.

## New behavior

Adds an optional `seed` argument to the sensors, threaded through `Sensor`, `InertialSensor`, `ScalarSensor` and the four concrete sensors. Each sensor builds its own `numpy.random.Generator` from `np.random.default_rng(seed)` and draws its white noise and random walk from that instead of `np.random`.

`seed=None` (the default) keeps the noise random but per-instance, so existing behavior is unchanged unless you pass a seed. With a seed, the noise is reproducible and independent of the global RNG state.

`_reset` is left alone on purpose (no re-seed): re-seeding on every reset would make each run in a repeated or Monte Carlo loop draw identical noise, which isn't the goal.

`tests/unit/sensors/test_sensor_seeding.py` covers same-seed reproducibility, decorrelation across seeds, independence from the global RNG (and not consuming it, which is the regression guard for the original bug), and the GnssReceiver path.

## Breaking change

- [x] No

With `seed=None` the output is identical to before; the argument is opt-in.

## Additional information

I kept `seed` out of the docstrings and `to_dict`/`from_dict` to keep the change minimal and consistent with the existing style (a couple of the sensor `__init__`s don't document every parameter either). Can add either if you'd rather.

Closes #1042

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ Attention: The newest changes should be on top -->
 
 - ENH: ENH: Refactor flight.py latitude/longitude to use inverted_haversine [#1055](https://github.com/RocketPy-Team/RocketPy/pull/1055)
 - ENH: TST: Add unit tests for evaluate_reduced_mass and remove TODO [#1051](https://github.com/RocketPy-Team/RocketPy/pull/1051)
+- ENH: seed sensor measurement noise per instance [#1052](https://github.com/RocketPy-Team/RocketPy/pull/1052)
 - ENH: FIX: pre release hardening [#1047](https://github.com/RocketPy-Team/RocketPy/pull/1047)
 - ENH: DEV: Claude Code ruff hooks (auto-format on edit + pre-push lint guard) [#1046](https://github.com/RocketPy-Team/RocketPy/pull/1046)
 - ENH: CI: create a CI for testing docs updates + solve different docs issues [#1045](https://github.com/RocketPy-Team/RocketPy/pull/1045)
--- a/rocketpy/sensors/accelerometer.py
+++ b/rocketpy/sensors/accelerometer.py
@@ -78,6 +78,7 @@ class Accelerometer(InertialSensor):
         cross_axis_sensitivity=0,
         consider_gravity=False,
         name="Accelerometer",
+        seed=None,
     ):
         """
         Initialize the accelerometer sensor
@@ -193,6 +194,7 @@ class Accelerometer(InertialSensor):
             temperature_scale_factor=temperature_scale_factor,
             cross_axis_sensitivity=cross_axis_sensitivity,
             name=name,
+            seed=seed,
         )
         self.consider_gravity = consider_gravity
         self.prints = _InertialSensorPrints(self)
--- a/rocketpy/sensors/accelerometer.py
+++ b/rocketpy/sensors/accelerometer.py
@@ -170,6 +170,11 @@ class Accelerometer(InertialSensor):
             acceleration. Default is False.
         name : str, optional
             The name of the sensor. Default is "Accelerometer".
+        seed : int, optional
+            Seed for the random number generator that draws the measurement
+            noise. If given, the noise becomes reproducible and independent of
+            the process-global NumPy RNG. Default is None, meaning the noise is
+            seeded from fresh entropy per instance.
 
         Returns
         -------
@@ -301,4 +306,5 @@ class Accelerometer(InertialSensor):
             cross_axis_sensitivity=data["cross_axis_sensitivity"],
             consider_gravity=data["consider_gravity"],
             name=data["name"],
+            seed=data.get("seed"),
         )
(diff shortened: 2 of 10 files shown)


MateusStano (changes requested):
> ##
> 
> I kept `seed` out of the docstrings and `to_dict`/`from_dict` to keep the change minimal and consistent with the existing style (a couple of the sensor `__init__`s don't document every parameter either). Can add either if you'd rather.

Please do add it to the docstrings! If there are any parameters from `__init__`  missing in its docstrings than there is something wrong there!

Also add `seed`  to `to_dict` and `from_dict`, please.

Otherwise this is a great PR!

thc1006:
Thanks, done. I added the `seed` docstring to every sensor `__init__` (and while I was in `Gyroscope` I filled in the `name` entry that was missing there too), and threaded `seed` through `to_dict`/`from_dict`, including `GnssReceiver`'s custom `to_dict`. `from_dict` reads it with `data.get("seed")` so dicts saved before this change still load, defaulting to `None`.

Also added tests: the seed round-trips through the JSON encoder for every sensor type, plus a case for the missing-key default. I went through the encoder rather than a bare `from_dict(to_dict())` because the inertial sensors' vectorized noise fields come back as lists that way, which is how the library actually saves and loads them.

Gui-FernandesBR (approved):
LGTM

thc1006:
Hi @MateusStano, a nudge on this one. I addressed both of your asks in `300e96e2`: `seed` is now documented in every sensor `__init__` (I also filled in the `name` entry that was missing from `Gyroscope` while I was there), and it's threaded through `to_dict` and `from_dict`, including `GnssReceiver`'s custom `to_dict`. `from_dict` reads it with `data.get("seed")` so dicts saved before this change still load.

Gui approved afterward and rebased the branch onto develop today, so the change request from your first pass is the last thing holding it. Whenever you get a chance to take another look, I'd appreciate it.

MateusStano (approved):

PR #1054, opened 2026-07-07
ENH: reproducible Monte Carlo via per-simulation-index seeding

### LINKED ISSUE #1053: Monte Carlo results aren't reproducible across serial and parallel runs (seeding is per-worker, not per-simulation)
**Describe the bug**

`MonteCarlo.simulate()` seeds the stochastic models per worker in parallel mode, and effectively once (at model construction) in serial mode. Because of that, the random inputs a given simulation index gets depend on how the run was executed rather than on the simulation index itself. Serial and parallel runs produce different input samples for the same setup, and parallel runs aren't reproducible run to run at all.

Two things are going on in `rocketpy/simulation/monte_carlo.py` (on `develop`):

1. Parallel mode spawns one seed per worker from a fresh, unseeded `SeedSequence`, so there's no fixed seed to reproduce against:

   ```python
   # __run_in_parallel, line 342
   seeds = np.random.SeedSequence().spawn(n_workers)
   ```

   Each worker then reseeds every stochastic model from its own per-worker seed and draws sequentially for whatever simulations it happens to pull off the shared counter:

   ```python
   # __sim_producer, lines 406-408
   self.environment._set_stochastic(seed)
   self.rocket._set_stochastic(seed)
   self.flight._set_stochastic(seed)
   ```

   ```python
   # __sim_producer, line 411
   sim_idx = sim_monitor.increment() - 1
   ```

   So simulation `i`'s inputs depend on which worker won it and in what order, which is nondeterministic across runs and changes with `n_workers`.

2. Serial mode (`__run_in_serial`) never reseeds during `simulate()`; it just draws one long sequential stream from each `Stochastic*` model's generator (seeded once at construction via `default_rng(seed)`). That stream is unrelated to the per-worker streams parallel mode produces.

The upshot: `simulate(n)` serial, `simulate(n, parallel=True, n_workers=2)`, and `simulate(n, parallel=True, n_workers=4)` all sample different inputs for the same stochastic setup, and the parallel ones differ every time you run them.

**To Reproduce**

Conceptually, with the same stochastic `environment`, `rocket`, and `flight`:

```python
mc_serial = MonteCarlo("out_serial", environment, rocket, flight)
mc_serial.simulate(10)

mc_par2 = MonteCarlo("out_par2", environment, rocket, flight)
mc_par2.simulate(10, parallel=True, n_workers=2)

mc_par4 = MonteCarlo("out_par4", environment, rocket, flight)
mc_par4.simulate(10, parallel=True, n_workers=4)
```

Compare the sampled inputs in the three `.inputs.txt` files: simulation `i` does not get the same inputs across the three, and re-running either parallel case gives a different set again.

**Expected behavior**

Simulation index `i` should get the same random inputs regardless of whether the run is serial or parallel and regardless of `n_workers`. A Monte Carlo run should be reproducible from a single seed, and `append` should extend the same stream rather than restart it.

**Additional context**

The reason the current scheme can't give that is that the seed is tied to the worker, not the simulation. Seeding per simulation index fixes it. Derive one root `SeedSequence` for the whole run and spawn one child per simulation, then reseed the stochastic models from `child_seeds[i]` right before running simulation `i`:

```python
# once, at the start of simulate()
root = np.random.SeedSequence(random_seed)      # root seed for the run
child_seeds = root.spawn(number_of_simulations)

# before running simulation i (same call in serial and inside each worker)
self.environment._set_stochastic(child_seeds[i])
self.rocket._set_stochastic(child_seeds[i])
self.flight._set_stochastic(child_seeds[i])
```

`SeedSequence.spawn` is prefix stable, so child `i` is the same object whether you spawned 10 or 100 children and whether you're in serial or in any worker. That makes the sampled inputs identical across serial, parallel(2), and parallel(N), and it makes `append=True` deterministic since later indices keep extending the same prefix.

One open design point: `MonteCarlo` / `simulate()` don't currently take a `random_seed` (today the only seed lives on the `Stochastic*` constructors and only reaches the serial path). The fix needs a root seed entry point somewhere, either a new `random_seed` argument on `simulate()`/`MonteCarlo` or deriving the root from the stochastic models' existing seeds. I can put up a PR once we agree on where the root seed should live.

**Breaking change**

Yes. This changes the exact numbers a given seed produces, so any stored Monte Carlo baseline regenerates, including `tests/integration/simulation/test_monte_carlo.py::test_monte_carlo_simulate`. The migration is a one-time baseline refresh; there's no API break for users who don't pin exact samples.

### LINKED ISSUE #1087: Sensor accepts a SeedSequence seed but cannot serialize it
### Summary

`Sensor.__init__` accepts anything `numpy.random.default_rng` accepts, keeps it, and hands it straight back from `to_dict()`:

```python
self._seed = seed
self._rng = np.random.default_rng(seed)
...
"seed": self._seed,
```

`default_rng` accepts a `SeedSequence`, so constructing a sensor with one succeeds. `RocketPyEncoder` has no case for `SeedSequence`, so serializing that sensor raises `TypeError`, and it raises at whatever point the object graph is written out rather than where the value came in.

### Reproduction

```python
import json
import numpy as np
from rocketpy.sensors import Accelerometer
from rocketpy._encoders import RocketPyEncoder

seed = np.random.SeedSequence(0).spawn(1)[0]
sensor = Accelerometer(sampling_rate=100, seed=seed)   # accepted
json.dumps(sensor.to_dict(), cls=RocketPyEncoder)      # TypeError
```

### Why it is worth closing

Spawning a `SeedSequence` per component is the idiom NumPy documents for decorrelating sub-streams, so it is a natural thing to reach for when seeding several sensors from one run seed. It works right up until something serializes the flight.

We hit exactly that downstream: a simulation seeded its three sensors from spawned children, every run worked, and packing the result to JSON failed for every user. The fix there was to convert each child to an int before it reaches the sensor, which works but throws away the child's `spawn_key`, so the resulting stream is not the one the child would have produced.

### Two ways to close it

**Accept it.** Document the seed type as `int | Sequence[int] | SeedSequence | None`, serialize a tagged form of `SeedSequence.state`, and rebuild with `SeedSequence(**state)` on load. That also lets a caller keep the child's exact stream through a round trip, which converting to an int cannot.

**Reject it.** Validate in the constructor that the seed is JSON representable and raise a clear `TypeError` naming the type. Narrow the docstring and type hints to match.

Either is fine. What is worth avoiding is the current state, where construction succeeds and the failure surfaces much later in an unrelated call.

### A related note on width

If the tagged-state route is taken, it also sidesteps a second problem. Converting a child to a full-width int gives a 128-bit value, which is exact in Python but larger than a JavaScript `Number` can represent. A submission we produce carries three such integers, and any consumer parsing that JSON with IEEE-754 doubles would round them.

Happy to send a PR for whichever direction you prefer.

thc1006:
Rewritten from current `develop`. The old head is not replayed: it is kept at `thc1006/RocketPy@backup/pr-1054-cf99e872`, and what came out of it went up separately as #1169, #1170, #1181 and #1182.

What is left here is the thing #1053 asked for and nothing else.

```
                  production files   production   new functions
old head cf99e872               15   +1544/-1091             49
this                             1      +120/-25             5
```

The other three files are the two test files and a paragraph in `docs/user/stochastic.rst`.

**Stacked on #1181.** This branch carries that commit, so the diff shown here includes it. Happy to rebase once it lands.

## Pull request type

## Checklist

- `CHANGELOG.md`: one line, written by hand. The workflow that would have written it has not run since #1112, which is #1173.

## Current behavior

`simulate()` has no seed at all. The two run paths seed differently, and one of them does not seed:

```
monte_carlo.py:472   seeds = np.random.SeedSequence().spawn(n_workers)   # fresh entropy, one per worker
__run_in_serial                                                          # never reseeds anything
```

So a study gives different results run to run, different results in the two modes, and different results again when the worker count changes. That is #1053.

Underneath it, the two paths do not even agree on what a simulation is called. Three simulations, measured:

```
serial   indices written: [1, 2, 3]
parallel indices written: [0, 1, 2]
```

An index therefore meant nothing on its own, and no index-keyed guarantee was possible before that was settled. It is the first commit here, separately revertible.

## New behavior

`simulate(..., random_seed=None)`, keyword-only. Every simulation takes the child of that root belonging to its index.

The child is derived directly rather than by spawning the ones before it. `spawn` appends `n_children_spawned + i` to the parent key, so rebuilding that one child reproduces it bit for bit and a worker reaches any index from four picklable values instead of a list a million long. There is a test comparing it against `spawn(n)[i]`, for every index, for each of `int`, `None`, a sequence, and a `SeedSequence`.

Measured over real flights, four simulations:

```
serial(42)  == serial(42)               True
serial(42)  == parallel(2 workers, 42)  True
serial(42)  == parallel(4 workers, 42)  True
serial(42)  == serial(7)                False
```

The last line is the control. Without it the first three pass on a run that ignores the seed.

The per-worker seed is gone rather than kept alongside. A worker decides nothing about sampling now, so `__sim_producer` no longer takes a seed and `__run_in_parallel` no longer spawns one per worker. `pylint` finding the argument unused is what made that obvious.

An index is claimed in one operation. `keep_simulating()` and `increment()` were two calls on a manager proxy, so two workers could both see the last slot free and then claim one index each, the second past the end. `claim_next_index()` does both under a lock in the manager's own process, and the two methods it replaces are gone, so the pair cannot be written again by accident. The index is also written after the data collectors now, so one named `index` cannot take the number the row's seed was derived from.

Addresses #1053. Not `Closes`, because the default branch is `master` and a closing keyword only fires into it.

It also addresses the spawn half of #1076. `test_an_index_keeps_its_inputs_when_the_workers_are_spawned` runs a real `simulate()` on two and four workers under `spawn` and compares the sampled inputs against serial. The forkserver half is still only the derivation helpers, so that issue stays open.

## Breaking change

- [x] Yes

Two, both worth stating plainly.

Fixed-seed results change, because there were none: nothing in `simulate()` was reproducible before, so nothing that was reproducible stops being so. Every study is now sampled from a different place than it would have been.

Serial results are numbered one lower. A three-simulation serial run wrote indices 1, 2, 3 and now writes 0, 1, 2. Anything reading those indices back sees the shift. Append already assumed zero-based, since `num_of_loaded_sims` counts rows and a two-row checkpoint resumes at index 2, which the serial path never used.

## What this does not do

Appending continues the same stream when the same seed is given again, since an index maps to a seed and to nothing else. Nothing here records the seed, so nothing here can tell you whether a later append was given the same one. Persisting the root so that can be checked is #1075, and the docstring says so rather than implying more.

Not in here, and each already has its own place: the nominal lifetime (#1169), component streams (#1170), worker failure reporting (#1182), checkpoint and manifest work (#1075). The old branch had all of them at once, which is what made it unreviewable.

## Additional information

```
pytest tests/unit                     2202 passed
pytest rocketpy --doctest-modules       48 passed
pytest tests/acceptance                 18 passed
ruff check . / ruff format --check .   clean
pylint rocketpy/ tests/ docs/          10.00/10, exit 0
```

Seven `tests/unit` failures on my machine are optional dependencies it does not have: four in `test_sensitivity.py` for `statsmodels`, three in `test_environment_analysis.py` for `windrose` and `ipywidgets`. All seven are `ImportError` before any assertion.

Each mechanism is pinned by a mutation, and each leaves a control standing:

|---|---|---|
| serial numbering back to one-based | three | the parallel case, which was already right |
| the per-index seeding, in both loops | three: reproducibility, split independence, append | the twelve derivation tests |
| every index reuses the root's own key | all six derivation tests | the rest |
| the caller's `SeedSequence` used as given | the consumption test | the rest |
| the entropy copy dropped | the aliasing test | the consumption test |
| the off-by-one put back in the failure message | the serial failure test | the numbering tests |
| the worker stops seeding the index it claimed | the worker-loop test, and the split-independence test with it | the two single-run seed tests |
| the index written before the collectors instead of after | the collector test | the numbering tests |
| the claim split back into a check and an increment | nothing | everything, which is the second finding below |

The last two rows are new. The patch had two lines uncovered, both reachable only when something goes wrong or when the worker loop runs in another process, so each now has a test that reaches it here: the failure message names the index the parallel path would, and the loop a worker runs, driven in this process, produces the run serial produces.

**One mutation survived, and finding that was the point.** I had written the entropy copy twice, once in `_root_seed_sequence` and once in `_root_state_of`. Removing either alone changed nothing, because the other still did the job; only removing both turned a test red. So one of them could not be distinguished from its own absence. Kept the one at the boundary where the value stops being the caller's, deleted the other.

**A second mutation survives, and I would rather say so than imply a test I do not have.** Putting the two-call claim back leaves everything green. A thread-level test does not reproduce the race even at `setswitchinterval(1e-9)` with eight claimers over five thousand slots, and the existing parallel test stayed green over five runs with the old loop restored. The window is one IPC round trip and two workers have to arrive inside it. The two tests I added pin the claim contract and both die to a boundary mutation, but the fix rests on construction rather than on a failing test.

The copy is not decoration. `SeedSequence` keeps a sequence entropy by reference:

```
caller passes [1, 2, 3], then sets caller[0] = 999
without the copy   the children move
with it            they do not
```

Merged with #1169, #1170, #1182 and #1171's follow-up into a throwaway tree on `develop` and run there. That found two things no branch shows on its own: `pylint` R0915 on `__sim_producer`, which is under the limit on each branch and over it once both grew, and ten of #1182's tests calling the producer with the seed argument this branch removes, which git merged without a conflict marker. Both are noted for whichever lands second.

### DIFF
--- a/rocketpy/simulation/monte_carlo.py
+++ b/rocketpy/simulation/monte_carlo.py
@@ -16,6 +16,7 @@ latest documentation.
 import csv
 import json
 import os
+import threading
 import traceback
 import warnings
 from copy import deepcopy
@@ -489,11 +490,10 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
         )
         sim_idx = sim_monitor.count
         try:
-            while sim_monitor.keep_simulating():
-                # Counted from zero, as the parallel path already does. The two
-                # named the same simulation differently: three of them wrote
-                # 1, 2, 3 here and 0, 1, 2 there.
-                sim_idx = sim_monitor.increment() - 1
+            # Counted from zero, as the parallel path already does: the two
+            # used to name the same simulation 1, 2, 3 and 0, 1, 2.
+            while (claimed := sim_monitor.claim_next_index()) is not None:
+                sim_idx = claimed
                 inputs_json, outputs_json = "", ""
 
                 self.__seed_this_simulation(sim_idx)
@@ -610,8 +610,7 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
             Event signaling an error occurred during the simulation.
         """
         try:
-            while sim_monitor.keep_simulating():
-                sim_idx = sim_monitor.increment() - 1
+            while (sim_idx := sim_monitor.claim_next_index()) is not None:
                 inputs_json, outputs_json = "", ""
 
                 self.__seed_this_simulation(sim_idx)
@@ -911,8 +910,6 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
             export_item: getattr(flight, export_item)
             for export_item in self.export_list
         }
-        outputs_dict["index"] = sim_idx
-
         if self.data_collector is not None:
             additional_exports = {}
             for key, callback in self.data_collector.items():
@@ -923,6 +920,9 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
                         f"An error was encountered running 'data_collector' callback {key}. "
                     ) from e
             outputs_dict = outputs_dict | additional_exports
+
+        # After the collectors, so one cannot take the row's own number.
+        outputs_dict["index"] = sim_idx
 
         return (
             json.dumps(outputs_dict, cls=RocketPyEncoder, **self._export_config) + "\n"
@@ -1899,13 +1899,20 @@ class _SimMonitor:
         self.n_simulations = n_simulations
         self.start_time = start_time
         self.completed_count = 0
+        self._claim = threading.Lock()  # proxy calls run in the manager
 
-    def keep_simulating(self):
-        return self.count < self.n_simulations
+    def claim_next_index(self):
+        """The next index to run, or None once every one has been claimed.
 
-    def increment(self):
-        self.count += 1
-        return self.count
+        One call, because a separate check and increment let two workers both
+        see the last slot free and then claim an index each past the end.
+        """
+        with self._claim:
+            if self.count >= self.n_simulations:
+                return None
+            claimed = self.count
+            self.count += 1
+            return claimed
 
     def print_update_status(self):
         """Prints a message on the same line as the previous one and replaces
--- a/rocketpy/simulation/monte_carlo.py
+++ b/rocketpy/simulation/monte_carlo.py
@@ -413,14 +413,18 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
             n_simulations=self.number_of_simulations,
             start_time=time(),
         )
+        sim_idx = sim_monitor.count
         try:
             while sim_monitor.keep_simulating():
-                sim_monitor.increment()
+                # Counted from zero, as the parallel path already does. The two
+                # named the same simulation differently: three of them wrote
+                # 1, 2, 3 here and 0, 1, 2 there.
+                sim_idx = sim_monitor.increment() - 1
                 inputs_json, outputs_json = "", ""
 
                 flight = self.__run_single_simulation()
-                inputs_json = self.__evaluate_flight_inputs(sim_monitor.count)
-                outputs_json = self.__evaluate_flight_outputs(flight, sim_monitor.count)
+                inputs_json = self.__evaluate_flight_inputs(sim_idx)
+                outputs_json = self.__evaluate_flight_outputs(flight, sim_idx)
 
                 self._append_simulation_record(inputs_json, outputs_json)
 
@@ -434,7 +438,7 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
                 f.write(inputs_json)
 
         except Exception as error:
-            print(f"Error on iteration {sim_monitor.count}: {error}")
+            print(f"Error on iteration {sim_idx}: {error}")
             with open(self._error_file, "a", encoding="utf-8") as f:
                 f.write(inputs_json)
             raise error
--- a/rocketpy/simulation/monte_carlo.py
+++ b/rocketpy/simulation/monte_carlo.py
@@ -339,6 +339,20 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
             number of workers will be equal to the number of CPUs available.
             A minimum of 2 workers is required for parallel mode.
             Default is None.
+        random_seed : int, sequence of int or numpy.random.SeedSequence, optional
+            Fixes what every simulation draws. Simulation ``i`` takes the same
+            inputs whichever way the run was split up, so serial and parallel
+            results agree and the number of workers does not reach the
+            sampling. Keyword-only. Default is None, which draws fresh entropy
+            and reproduces nothing.
+
+            Appending continues the same stream when the same seed is given
+            again, since an index maps to a seed and to nothing else. Nothing
+            here records the seed, so nothing here can tell you that a later
+            append was given the same one; that is #1075.
+
+            A ``Generator`` or ``BitGenerator`` is refused rather than read.
+            Pass the seed it was built from.
         kwargs : dict
             Custom arguments for simulation export of the ``inputs`` file. Options
             are:
--- a/rocketpy/simulation/monte_carlo.py
+++ b/rocketpy/simulation/monte_carlo.py
@@ -18,6 +18,7 @@ import json
 import os
 import traceback
 import warnings
+from copy import deepcopy
 from numbers import Real
 from pathlib import Path
 from time import time
@@ -31,6 +32,7 @@ from rocketpy.plots.monte_carlo_plots import _MonteCarloPlots
 from rocketpy.prints.monte_carlo_prints import _MonteCarloPrints
 from rocketpy.simulation.flight import Flight
 from rocketpy.tools import (
+    _seed_sequence_to_int,
     generate_monte_carlo_ellipses,
     generate_monte_carlo_ellipses_coordinates,
     import_optional_dependency,
@@ -42,6 +44,57 @@ from rocketpy.tools import (
 # simulate() writes one JSON object per line and reads that same shape back, so
 # this is the only format it can both resume from and overwrite safely.
 _SIMULATION_LOG_SUFFIX = ".txt"
+
+
+def _root_seed_sequence(random_seed):
+    """The immutable root a run derives every simulation's seed from.
+
+    A ``SeedSequence`` is rebuilt from its full state rather than used as
+    given, since ``spawn`` advances a counter the caller still holds. A
+    ``Generator`` is refused rather than read, because using a consume-on-use
+    object as an immutable seed cannot mean what it says.
+    """
+    if isinstance(random_seed, np.random.SeedSequence):
+        return np.random.SeedSequence(**random_seed.state)
+    if isinstance(random_seed, (np.random.Generator, np.random.BitGenerator)):
+        raise TypeError(
+            f"random_seed must be an int, a sequence of non-negative integers, "
+            f"or a numpy.random.SeedSequence, not a "
+            f"{type(random_seed).__name__}. Pass the seed the generator was "
+            f"built from."
+        )
+    return np.random.SeedSequence(random_seed)
+
+
+def _root_state_of(root):
+    """A root as the four picklable values a worker can rebuild it from.
+
+    Sent to each worker instead of the object, and instead of the list of
+    children, so a run of a million simulations costs four values. The entropy
+    is copied because a sequence one is kept by reference all the way from the
+    caller, who could otherwise still move every child by editing their list.
+    """
+    return (
+        deepcopy(root.entropy),
+        tuple(root.spawn_key),
+        root.pool_size,
+        root.n_children_spawned,
+    )
+
+
+def _seed_of_simulation(root_state, sim_idx):
+    """The seed for one simulation index, without spawning the ones before it.
+
+    ``spawn`` derives child ``i`` by appending ``n_children_spawned + i`` to
+    the parent spawn key, so rebuilding that one child directly reproduces it
+    and any index can be reached from the four values above alone.
+    """
+    entropy, spawn_key, pool_size, base = root_state
+    return np.random.SeedSequence(
+        entropy=entropy,
+        spawn_key=(*spawn_key, base + sim_idx),
+        pool_size=pool_size,
+    )
 
 
 def _refuse_logs_this_run_cannot_write(
@@ -265,6 +318,8 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
         append=False,
         parallel=False,
         n_workers=None,
+        *,
+        random_seed=None,
         **kwargs,
     ):
         """
@@ -317,6 +372,11 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
         self._export_config = kwargs
         self.number_of_simulations = number_of_simulations
         self._initial_sim_idx = self.num_of_loaded_sims if append else 0
+        # Validated here, before __setup_files truncates anything, so an
+        # unusable seed cannot cost a previous run its results. Kept as four
+        # picklable values rather than as the object, since a worker rebuilds
+        # any index from them.
+        self.__root_state = _root_state_of(_root_seed_sequence(random_seed))
 
         # Before anything is opened: __setup_files truncates for append=False.
         _refuse_logs_this_run_cannot_write(
@@ -422,6 +482,7 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
                 sim_idx = sim_monitor.increment() - 1
                 inputs_json, outputs_json = "", ""
 
+                self.__seed_this_simulation(sim_idx)
                 flight = self.__run_single_simulation()
                 inputs_json = self.__evaluate_flight_inputs(sim_idx)
                 outputs_json = self.__evaluate_flight_outputs(flight, sim_idx)
@@ -473,13 +534,14 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
             )
 
             processes = []
-            seeds = np.random.SeedSequence().spawn(n_workers)
 
-            for seed in seeds:
+            # No seed per worker any more: every simulation takes its own from
+            # its index, so the workers are interchangeable and how many there
+            # are does not reach the sampling.
+            for _ in range(n_workers):
                 sim_producer = multiprocess.Process(
                     target=self.__sim_producer,
                     args=(
-                        seed,
                         sim_monitor,
                         mutex,
                         simulation_error_event,
@@ -521,13 +583,11 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
             raise ValueError("Number of workers must be at least 2 for parallel mode.")
         return n_workers
 
-    def __sim_producer(self, seed, sim_monitor, mutex, error_event):  # pylint: disable=too-many-statements
+    def __sim_producer(self, sim_monitor, mutex, error_event):
         """Simulation producer to be used in parallel by multiprocessing.
 
         Parameters
         ----------
-        seed : int
-            The seed to set the random number generator.
         sim_monitor : _SimMonitor
             The simulation monitor object to keep track of the simulations.
         mutex : multiprocess.Lock
@@ -536,15 +596,11 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
             Event signaling an error occurred during the simulation.
         """
         try:
-            # Ensure Processes generate different random numbers
-            self.environment._set_stochastic(seed)
-            self.rocket._set_stochastic(seed)
-            self.flight._set_stochastic(seed)
-
             while sim_monitor.keep_simulating():
                 sim_idx = sim_monitor.increment() - 1
                 inputs_json, outputs_json = "", ""
 
+                self.__seed_this_simulation(sim_idx)
                 flight = self.__run_single_simulation()
                 inputs_json = self.__evaluate_flight_inputs(sim_idx)
                 outputs_json = self.__evaluate_flight_outputs(flight, sim_idx)
@@ -583,6 +639,20 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
             )
             error_event.set()
             mutex.release()
+
+    def __seed_this_simulation(self, sim_idx):
+        """Reseed the three models from this index's own child of the root.
+
+        Per index rather than per worker, which is what makes a simulation's
+        inputs the same however the run was split up. The child is split three
+        ways so the environment, rocket and flight draw independently instead
+        of sharing one stream.
+        """
+        child = _seed_of_simulation(self.__root_state, sim_idx)
+        environment, rocket, flight = child.spawn(3)
+        self.environment._set_stochastic(_seed_sequence_to_int(environment))
+        self.rocket._set_stochastic(_seed_sequence_to_int(rocket))
+        self.flight._set_stochastic(_seed_sequence_to_int(flight))
 
     def __run_single_simulation(self):
         """Runs a single simulation and returns the inputs and outputs.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ Attention: The newest changes should be on top -->
 
 ### Added
 
+- ENH: Reproducible Monte Carlo through per-simulation-index seeding [#1054](https://github.com/RocketPy-Team/RocketPy/pull/1054) [#1053](https://github.com/RocketPy-Team/RocketPy/issues/1053)
 - ENH: Support fixed-time parachute deployment triggers [#1133](https://github.com/RocketPy-Team/RocketPy/pull/1133) [#437](https://github.com/RocketPy-Team/RocketPy/issues/437)
 - DOC: Add SIL parachute ejection integration example [#1131](https://github.com/RocketPy-Team/RocketPy/pull/1131) [#524](https://github.com/RocketPy-Team/RocketPy/issues/524)
 - ENH: List NOAA atmosphere datasets and fetch latest [#1136](https://github.com/RocketPy-Team/RocketPy/pull/1136) [#660](https://github.com/RocketPy-Team/RocketPy/issues/660)
--- a/docs/user/stochastic.rst
+++ b/docs/user/stochastic.rst
@@ -297,3 +297,7 @@ better reflecting the inherent uncertainties in rocketry.
     or split over any number of workers, and appending with the same seed
     carries the same stream on. Without it a run draws fresh entropy and
     reproduces nothing.
+
+    What an index fixes is the inputs it draws. A stochastic input reads its
+    nominal from the object it wraps, and a run moves that as it goes, so the
+    flight itself can still differ by which worker took the index.
--- a/docs/user/stochastic.rst
+++ b/docs/user/stochastic.rst
@@ -289,3 +289,11 @@ better reflecting the inherent uncertainties in rocketry.
 .. note::
     See the ``MonteCarlo`` class documentation for more information on how to run \
     Monte Carlo simulations with stochastic objects.
+
+.. note::
+    A whole run is fixed by ``MonteCarlo.simulate(random_seed=...)`` rather than
+    by seeding these models yourself. Each simulation takes its seed from its
+    own index, so simulation 7 draws the same inputs whether the run was serial
+    or split over any number of workers, and appending with the same seed
+    carries the same stream on. Without it a run draws fresh entropy and
+    reproduces nothing.
--- a/rocketpy/stochastic/stochastic_model.py
+++ b/rocketpy/stochastic/stochastic_model.py
@@ -8,7 +8,7 @@ import numpy as np
 from rocketpy.mathutils.function import Function
 from rocketpy.stochastic.custom_sampler import CustomSampler
 
-from ..tools import get_distribution
+from ..tools import _seed_sequence_to_int, get_distribution
 
 
 def _names_as_spawn_key(input_names):
@@ -41,6 +41,18 @@ def _format_number(value):
     return f"array of shape {np.shape(value)}"
 
 
+def _seed_as_entropy(seed):
+    """A seed as something ``SeedSequence`` will take as entropy.
+
+    A parallel run is handed a ``SeedSequence``, which it will not take. Any
+    other seed goes through untouched, so the stream an int reaches stays where
+    it was.
+    """
+    if not isinstance(seed, np.random.SeedSequence):
+        return seed
+    return _seed_sequence_to_int(seed)
+
+
 def _sampler_seed(seed, input_names):
     """Derive a seed for one sampler, or for one group that shares a generator.
 
@@ -54,10 +66,10 @@ def _sampler_seed(seed, input_names):
     # Sorted here rather than trusting the caller, so a future call site cannot
     # give one group two different seeds by listing its members another way.
     root = np.random.SeedSequence(
-        entropy=seed, spawn_key=_names_as_spawn_key(tuple(sorted(input_names)))
+        entropy=_seed_as_entropy(seed),
+        spawn_key=_names_as_spawn_key(tuple(sorted(input_names))),
     )
-    words = root.generate_state(4, dtype=np.uint32)
-    return sum(int(word) << (32 * position) for position, word in enumerate(words))
+    return _seed_sequence_to_int(root)
 
 
 # TODO: Stop using assert in production code. Use exceptions instead.
--- a/rocketpy/tools.py
+++ b/rocketpy/tools.py
@@ -1377,6 +1377,17 @@ def euler313_to_quaternions(phi, theta, psi):
     return e0, e1, e2, e3
 
 
+def _seed_sequence_to_int(seed_sequence):
+    """Returns a ``SeedSequence`` as the 128-bit ``int`` it can be rebuilt from.
+
+    Folded through ``generate_state`` rather than read off ``entropy``, since
+    the children of one root differ only by ``spawn_key``, and combined by
+    value so it does not depend on byte order.
+    """
+    words = seed_sequence.generate_state(4, dtype=np.uint32)
+    return sum(int(word) << (32 * position) for position, word in enumerate(words))
+
+
 def get_matplotlib_supported_file_endings():
     """Gets the file endings supported by matplotlib.
 
--- a/tests/unit/simulation/test_monte_carlo_parallel_runs.py
+++ b/tests/unit/simulation/test_monte_carlo_parallel_runs.py
@@ -1,0 +1,32 @@
+import pytest
+
+from rocketpy.simulation.monte_carlo import MonteCarlo
+
+
+@pytest.mark.parametrize("parallel", [False, True])
+def test_a_monte_carlo_run_finishes(
+    stochastic_environment, stochastic_calisto, stochastic_flight, tmp_path, parallel
+):
+    # The parallel path hands each worker a SeedSequence rather than an int, and
+    # nothing else in the suite exercises that. A worker that dies on it is not
+    # reported, so this reads as a hang rather than as a failure.
+    #
+    # Built here rather than taken from the monte_carlo_calisto fixture, whose
+    # own filename is fixed, since `filename` is a plain attribute and the three
+    # working paths are settled when the object is constructed.
+    analysis = MonteCarlo(
+        filename=str(tmp_path / "study"),
+        environment=stochastic_environment,
+        rocket=stochastic_calisto,
+        flight=stochastic_flight,
+    )
+
+    analysis.simulate(
+        number_of_simulations=2,
+        append=False,
+        parallel=parallel,
+        n_workers=2 if parallel else None,
+    )
+
+    assert analysis.num_of_loaded_sims == 2
+    assert str(tmp_path) in str(analysis.output_file)
--- a/tests/unit/simulation/test_monte_carlo_seeding.py
+++ b/tests/unit/simulation/test_monte_carlo_seeding.py
@@ -12,10 +12,30 @@ from rocketpy.simulation.monte_carlo import (
 )
 
 
+def _without_object_identity(value):
+    """The same record with every ``hash`` dropped, at any depth.
+
+    ``RocketPyEncoder`` records ``hash(obj)`` beside a serialized ``Function``,
+    and that is the object's identity in the process that wrote it, not
+    anything that was drawn. It agrees between two runs that share memory and
+    differs between two that do not, which is a property of the start method
+    rather than of the seed.
+    """
+    if isinstance(value, dict):
+        return {
+            key: _without_object_identity(item)
+            for key, item in value.items()
+            if key != "hash"
+        }
+    if isinstance(value, list):
+        return [_without_object_identity(item) for item in value]
+    return value
+
+
 def _sampled_inputs(analysis):
     with open(analysis.input_file, "r", encoding="utf-8") as written:
         rows = [json.loads(line) for line in written if line.strip()]
-    return {row["index"]: row for row in rows}
+    return {row["index"]: _without_object_identity(row) for row in rows}
 
 
 def _a_run(tmp_path, stem, models, *, parallel=False, workers=None, seed=None, count=4):
--- a/tests/unit/simulation/test_monte_carlo_seeding.py
+++ b/tests/unit/simulation/test_monte_carlo_seeding.py
@@ -1,6 +1,7 @@
 import json
 import os
 
+import multiprocess
 import numpy as np
 import pytest
 
@@ -12,30 +13,32 @@ from rocketpy.simulation.monte_carlo import (
 )
 
 
-def _without_object_identity(value):
-    """The same record with every ``hash`` dropped, at any depth.
+def _what_was_drawn(value):
+    """The same record without the parts that say which process wrote it.
 
-    ``RocketPyEncoder`` records ``hash(obj)`` beside a serialized ``Function``,
-    and that is the object's identity in the process that wrote it, not
-    anything that was drawn. It agrees between two runs that share memory and
-    differs between two that do not, which is a property of the start method
-    rather than of the seed.
+    Two of the fields a serialized ``Function`` carries are properties of the
+    writer rather than of the draw. ``hash`` is the object's identity in that
+    process. A callable ``source`` is its pickle, and the same callable pickles
+    to different bytes under ``spawn``: measured on one drag curve, the parent
+    and a forked child agree and a spawned child does not, for a value the
+    fixture does not vary at all. Everything a run actually draws is numeric
+    and stays.
     """
     if isinstance(value, dict):
         return {
-            key: _without_object_identity(item)
+            key: _what_was_drawn(item)
             for key, item in value.items()
-            if key != "hash"
+            if key != "hash" and not (key == "source" and isinstance(item, str))
         }
     if isinstance(value, list):
-        return [_without_object_identity(item) for item in value]
+        return [_what_was_drawn(item) for item in value]
     return value
 
 
 def _sampled_inputs(analysis):
     with open(analysis.input_file, "r", encoding="utf-8") as written:
         rows = [json.loads(line) for line in written if line.strip()]
-    return {row["index"]: _without_object_identity(row) for row in rows}
+    return {row["index"]: _what_was_drawn(row) for row in rows}
 
 
 def _a_run(tmp_path, stem, models, *, parallel=False, workers=None, seed=None, count=4):
@@ -175,3 +178,30 @@ def test_an_appended_run_carries_the_same_stream_on(models, tmp_path):
     part.simulate(4, append=True, random_seed=42)
 
     assert _sampled_inputs(part) == _sampled_inputs(whole)
+
+
+@pytest.fixture(name="spawned_workers")
+def _spawned_workers():
+    """Start workers the way Windows does, wherever the test happens to run."""
+    was = multiprocess.get_start_method()
+    multiprocess.set_start_method("spawn", force=True)
+    yield
+    multiprocess.set_start_method(was, force=True)
+
+
+@pytest.mark.usefixtures("spawned_workers")
+@pytest.mark.skipif(os.cpu_count() < 4, reason="needs four workers to be four")
+def test_an_index_keeps_its_inputs_when_the_workers_are_spawned(models, tmp_path):
+    """The same guarantee on the start method Windows uses.
+
+    A spawned child rebuilds the models by unpickling rather than inheriting
+    them, so this is a different question from the one above and was worth
+    asking separately: the first version of these tests compared serialized
+    callables, which differ between processes for a value nothing varies.
+    """
+    serial = _a_run(tmp_path, "serial", models, seed=42)
+    two = _a_run(tmp_path, "two", models, parallel=True, workers=2, seed=42)
+    four = _a_run(tmp_path, "four", models, parallel=True, workers=4, seed=42)
+
+    assert _sampled_inputs(serial) == _sampled_inputs(two)
+    assert _sampled_inputs(serial) == _sampled_inputs(four)
--- a/tests/unit/simulation/test_monte_carlo_seeding.py
+++ b/tests/unit/simulation/test_monte_carlo_seeding.py
@@ -1,5 +1,6 @@
 import json
 import os
+from time import time
 
 import multiprocess
 import numpy as np
@@ -10,6 +11,7 @@ from rocketpy.simulation.monte_carlo import (
     _root_seed_sequence,
     _root_state_of,
     _seed_of_simulation,
+    _SimMonitor,
 )
 
 
@@ -205,3 +207,46 @@ def test_an_index_keeps_its_inputs_when_the_workers_are_spawned(models, tmp_path
 
     assert _sampled_inputs(serial) == _sampled_inputs(two)
     assert _sampled_inputs(serial) == _sampled_inputs(four)
+
+
+class _ALockNobodyContends:
+    """The manager's mutex, with only this process to hand it to."""
+
+    def acquire(self):
+        pass
+
+    def release(self):
+        pass
+
+
+class _AnEventNobodySet:
+    """The failure event, which nothing in a clean run reaches for."""
+
+    def is_set(self):
+        return False
+
+    def set(self):
+        raise AssertionError("the producer reported a failure")
+
+
+def test_a_worker_loop_draws_the_study_serial_draws(models, tmp_path):
+    """The loop a worker runs, driven here, produces the run serial produces."""
+    # Every other check of this starts a child process, where the same code
+    # runs and nothing in the test can watch it.
+    serial = _a_run(tmp_path, "serial", models, seed=42, count=4)
+    environment, rocket, flight = models
+    worker = MonteCarlo(
+        filename=str(tmp_path / "worker"),
+        environment=environment,
+        rocket=rocket,
+        flight=flight,
+    )
+    worker.simulate(number_of_simulations=0, append=False, random_seed=42)
+
+    worker._MonteCarlo__sim_producer(
+        _SimMonitor(initial_count=0, n_simulations=4, start_time=time()),
+        _ALockNobodyContends(),
+        _AnEventNobodySet(),
+    )
+
+    assert _sampled_inputs(worker) == _sampled_inputs(serial)
--- a/tests/unit/simulation/test_monte_carlo_seeding.py
+++ b/tests/unit/simulation/test_monte_carlo_seeding.py
@@ -1,0 +1,157 @@
+import json
+import os
+
+import numpy as np
+import pytest
+
+from rocketpy.simulation.monte_carlo import (
+    MonteCarlo,
+    _root_seed_sequence,
+    _root_state_of,
+    _seed_of_simulation,
+)
+
+
+def _sampled_inputs(analysis):
+    with open(analysis.input_file, "r", encoding="utf-8") as written:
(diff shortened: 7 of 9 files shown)


tests/integration/simulation/test_monte_carlo_determinism.py
```diff
@@ -0,0 +1,177 @@
...
+):
+    """Run a Monte Carlo with a stubbed ``Flight`` and return inputs by index."""
+    monkeypatch.setattr(mc_module, "Flight", _StubFlight)
+    montecarlo = MonteCarlo(
+        filename=str(tmp_path / tag),
+        environment=environment,
+        rocket=rocket,
+        flight=flight,
+    )
+    montecarlo.simulate(**simulate_kwargs)
```
phmbressan: Could you double check if after running this test, no files are left dangling on the test file system. If that is the case, using a `try-finally` block to remove those (as done in a few other `MonteCarlo` tests) is an option.

phmbressan (approved):
The implementation is very clear and throughout, nice work.

The explanation on the concepts behind per index seeding (both in the issue and PR description) were rather helpful. I agree having reproducible results was an issue with the parallel per worker seeding.

Regarding the decisions on parameter naming, I agree with most of the decisions taken here. Moreover, the `rng` attribute is well docstringed, so it shouldn't be a matter of confusion to the user.

@MateusStano could you give your two cents on the changes here before we proceed with a merge?

tests/integration/simulation/test_monte_carlo_determinism.py
thc1006: Thanks for the review. On the dangling files: I checked, and nothing gets left behind. The test passes `filename=tmp_path / tag` into `MonteCarlo`, so the `.inputs.txt`, `.outputs.txt` and `.errors.txt` all land under pytest's per-test `tmp_path` and get cleaned up with it. I reran it and confirmed the repo and working directory stay clean, so there's nothing here for a `try/finally` to remove.

The other MonteCarlo tests need that cleanup because their fixture uses a fixed `filename="monte_carlo_test"`, which writes into the cwd. This one keeps everything inside `tmp_path` instead. I'm happy to match the try/finally style if you'd prefer consistency across the file, but the tmp_path route seemed cleaner to me. Your call.

rocketpy/simulation/monte_carlo.py
```diff
@@ -402,15 +464,11 @@ def __sim_producer(self, seed, sim_monitor, mutex, error_event):  # pylint: disa
...
-            self.flight._set_stochastic(seed)
-
             while sim_monitor.keep_simulating():
                 sim_idx = sim_monitor.increment() - 1
                 inputs_json, outputs_json = "", ""
 
+                self.__seed_simulation(child_seeds[sim_idx])
                 flight = self.__run_single_simulation()
                 inputs_json = self.__evaluate_flight_inputs(sim_idx)
                 outputs_json = self.__evaluate_flight_outputs(flight, sim_idx)
```
MateusStano: I am not sure about this so @thc1006 could you verify, please?

But I believe adding this `child_seeds[sim_idx]` can break at the very end of the simulation. If you are running 8 simulations, and we are at step 7. We could have one worker check 7<8 -> yes, and a second worker also check 7<8 -> yes, because keep_simulating() and increment() are two separate steps and nothing locks them together. Then both workers increment: the first gets `sim_idx = 7` (ok), but the second gets `sim_idx = 8`. Since child_seeds only has 8 item, `child_seeds[8]` does not exist

I think this race was already there before, but it didnt crash the simulation. Could you check and fix, please?

rocketpy/simulation/monte_carlo.py
```diff
@@ -267,10 +278,46 @@ def __setup_files(self, append):
...
+        The per-index child seed is split three ways so the environment,
+        rocket and flight draw from independent streams instead of sharing
+        one. Seeding per simulation index (not per worker) is what makes the
+        sampled inputs invariant to the execution mode and to the number of
+        workers.
+        """
+        env_seed, rocket_seed, flight_seed = child_seed.spawn(3)
+        self.environment._set_stochastic(env_seed)
+        self.rocket._set_stochastic(rocket_seed)
+        self.flight._set_stochastic(flight_seed)
```
MateusStano: This .spawn() call seems to make the case of using SeedSequence/Generator not repeatable

From what I understand if a user builds a SeedSequence (or Generator) once and passes the same object to simulate() twice, the first will spawns a set of children and the second run will spawn another set

MateusStano (changes requested):

rocketpy/simulation/monte_carlo.py
thc1006: You're right, this is a real race. `keep_simulating()` and `increment()` are separate manager calls, so near the end multiple workers can pass the `count < n` check before any of them increments and then claim `sim_idx == n`, which the new `child_seeds[sim_idx]` lookup turned into an `IndexError` (before, it just wrote an extra record). Fixed in 9c020b6: the claim now goes through a small `_claim_next_index` helper that holds the shared mutex across the check and the increment, so each index is handed out once and the counter never runs past `number_of_simulations`. There's a deterministic unit test for it (a barrier plus a widened check-to-increment window) that over-claims and fails if the lock is removed. Thanks for catching it.

rocketpy/simulation/monte_carlo.py
thc1006: Agreed, the current code mutated the caller's seed lineage. `SeedSequence.spawn()` advances `n_children_spawned`, and the helper returned the original object, so a second `simulate()` with the same object spawned a different set. Fixed in 9c020b6 with two changes:

1. A supplied `SeedSequence` is now copied from its full `state` (entropy, spawn_key, pool_size, n_children_spawned) before spawning, so the caller's object is untouched and repeated calls reproduce the same children. Copying only `entropy` would drop `spawn_key`, which matters for a spawned child.
2. I dropped `Generator`/`BitGenerator` from the accepted types. A generator is a stateful RNG, not a seed: its `bit_generator.seed_seq` is only the original lineage, so a generator advanced by a million draws would still seed the run identically to a fresh one. SPEC 7's `rng` semantics would instead consume it, the opposite of a reproducible seed. `random_seed` now takes an int, a sequence of ints, or a `SeedSequence`; a generator user can pass `rng.bit_generator.seed_seq`.

Tests cover the same-object reproducibility, `n_children_spawned` staying put, a spawned child with a non-empty `spawn_key`, and the generator rejection.

thc1006:
@MateusStano friendly ping when you have a moment. Both points from your last pass are addressed: the parallel index claim now holds the shared mutex across the check-and-increment (with a deterministic test that goes red if the lock is removed), and a supplied `SeedSequence` is copied from its full state before spawning, so repeated calls reproduce and the caller is left untouched. I replied inline on both threads. The test matrix and lint pass on the current head; the only red is the soft codecov patch check, which is the parallel-only lines I covered in the thread above. Whenever you get a chance to take another look, I would appreciate it.

thc1006:
Heads up that this changed enough since the last look to be worth a fresh pass rather than merging on the earlier approval. @MateusStano @phmbressan when you have a moment.

What is new since the review:

- Per-index child seeds are now derived in O(1) by extending the root `spawn_key` (bit-identical to `spawn(n)[i]`) instead of pre-spawning the whole list, so nothing O(N) is pickled to each worker.
- The samplers get a plain 128-bit `int` rather than a `SeedSequence`, so a custom sampler's int-typed `reset_seed` keeps working (a `SeedSequence` raises `TypeError` in `random.Random` since 3.11). The int combines all four words by value, so it is byte-order independent.
- List-valued stochastic attributes now draw from the model's seeded generator, so `random_seed` governs them too. That closes the gap the previous description called out as a known limitation.
- Added a start-method-invariance test that runs under fork, spawn and forkserver in ordinary CI, since 3.14 moved the POSIX default to forkserver.

Both earlier concerns are still handled: the parallel claim holds the mutex across the check and the increment, and a supplied `SeedSequence` is copied from its full state. #1071 opened for the same issue in the meantime; the description notes how this relates and what it borrows. A re-review whenever you get the chance would be appreciated.

thc1006:
@MateusStano @phmbressan a follow-up pass turned up a few more things worth fixing, so I have pushed them and would appreciate another look when you have time.

Since your reviews:

- Component seeds: `StochasticRocket._set_stochastic` handed the same seed to the body and to every surface, motor, rail button and parachute, so a main and a drogue parachute drew the same cd_s and lag quantiles. Each component now gets its own child of the run's seed.
- List sampling: `dict_generator` and `_randomize_position` sampled list-valued attributes (component positions among them) with the stdlib `random.choice`, which `random_seed` did not control. Both now draw through the model's seeded generator.
- File safety: `simulate()` truncated the output files before the seed was validated, so a rejected seed destroyed a previous run's results. Validation runs first now.
- Docs: corrected the seed helper's note about `RandomState`, since a 128-bit int does not fit its 32-bit seed.

I also marked the earlier threads resolved. The race and the SeedSequence copy are both fixed in the current code, and the dangling-files question checked out: the run writes only under `tmp_path`.

A few larger items from the same review are better as their own issues, so I opened #1075 (append continuation), #1076 (a full parallel test under spawn and forkserver) and #1077 (a seed for `simulate_convergence`), and linked them from the description. Thanks for the careful reviews.

wuisabel-gif:
I opened #1071 for the same issue before spotting this one, and have closed it in favor of this PR, yours is the more complete solution (per-component child seeds, list-valued attribute sampling, int seeds for custom `reset_seed`, and the fork/spawn/forkserver invariance test all go beyond what mine did). Sorry for the drive-by overlap. If it'd help lighten the load, I'm happy to pick up one of your follow-ups #1077 (seed for `simulate_convergence`) or #1078 (the flaky VTK animation tests).

thc1006:
> I opened #1071 for the same issue before spotting this one, and have closed it in favor of this PR, yours is the more complete solution (per-component child seeds, list-valued attribute sampling, int seeds for custom `reset_seed`, and the fork/spawn/forkserver invariance test all go beyond what mine did). Sorry for the drive-by overlap. If it'd help lighten the load, I'm happy to pick up one of your follow-ups #1077 (seed for `simulate_convergence`) or #1078 (the flaky VTK animation tests).

for sure go ahead thx!!!

thc1006:
@MateusStano a summary of where this got to, since I pushed the last of it without saying so and there is no reason you would know it is ready.

**TLDR: both things you raised are fixed and the threads are resolved. Whenever you have time.**

Your two points, and what they turned into:

**The race between `keep_simulating()` and `increment()`.** You were right that indexing `child_seeds` by a counter read separately from the check meant two workers could land on the same index near the limit. The seed is now derived from the simulation index rather than taken from a shared list, so there is no shared position to race on. It is also O(1) per index instead of spawning a list up front, which was the other reason to change it.

**The `.spawn()` call making a caller's `SeedSequence` non repeatable.** Also right. `spawn()` advances `n_children_spawned` on the object it is called on, so passing the same `SeedSequence` twice gave different runs. It derives without mutating the caller's object now, and there is a test that runs the same `SeedSequence` twice and compares.

Both threads are marked resolved above.

Since your review I also found three more while writing tests for it, and they are in the same branch:

- list valued stochastic attributes were sampled from the global RNG, so a seeded run was not fully seeded
- rocket components shared a seed with each other, so two parachutes drew the same values
- an invalid seed was caught after the output file had been truncated, which lost data on a typo

Ten checks green. `phmbressan` approved on 10 July and asked for your read before merging.

There is no hurry from my side. Flagging it because the last push was 27 July with nothing said, so it has been sitting looking like it still needs work when it does not.

wuisabel-gif (commented):
I reviewed the latest changes and ran the relevant Monte Carlo unit and integration tests locally; they pass. I did not find any blocking issues with the seed derivation, worker index claiming, repeated SeedSequence handling, list sampling, or component seed separation.

One small documentation suggestion: it may be worth clarifying that parallel runs are reproducible by simulation index, while the physical order of records in the output file may still depend on worker completion order. Other than that, the implementation and test coverage look good to me.

thc1006:
Pushed four commits. Review turned up six problems with the seeding change and they are all fixed, plus four more I found going back over my own fix.

The six:

1. **Baseline drift.** `_set_stochastic` re-validates every kwarg, and validation read the nominal back off `self.obj`. `StochasticEnvironment.create_object` writes the sampled value onto that same object on purpose, so a factor like `wind_velocity_x_factor` compounded from one simulation to the next. Serial and parallel then disagreed, because the drift depends on how many simulations a worker happened to run before that index. The nominal is captured once now, when the model is built.
2. **Air brakes were never reseeded.** They are in `create_object` and were not in the reseed loop, so they drew from wherever the generator had been left. Every seeding test passed because no fixture had an air brake. The collections are declared in one place now, and a test scans `create_object`'s source so one added later cannot quietly miss the reseed.
3. **CP and thrust eccentricity** were validated once at add time, against the generator as it stood then. Reseeding replaced the generator but not those values.
4. **A failure in the index claim raised `UnboundLocalError` inside the worker's own error handler**, so nothing was written, nothing was printed, and the run ended with no record of what went wrong.
5. **`join()` returns None however a child ended**, so the shared event was the only signal the parent had, and a worker can leave without setting it. Exit codes are checked now, and after the run every index it claimed has to have left exactly one input row and one output row. Neither file shows that on its own: the rows look well formed, and reading them back keyed by index hides a duplicate behind the row that overwrote it.
6. **Worker invariance was only exercised under `fork`.**

Going back over that, four more:

* Setting the shared event is the first thing the worker's handler does, and it can raise on its own once the manager has gone. That is the same masking bug as 4, one layer up. Reporting is best effort now and never replaces the failure it is reporting.
* The source scan in 2 only walked `for` loops. The same loop written as a comprehension would have gone straight past it.
* A row cut off mid-write is exactly what the check in 5 exists to diagnose, and parsing strictly turned it into a `JSONDecodeError` that points nowhere. It is reported as the index that went missing instead.
* The check in 5 was a regression for Ctrl-C. Both run paths catch it, print that the files are saved and return, so the check then called the run a failure one line after saying the opposite. Stopped runs are exempt.

Worth flagging about the fixtures, since four of these were invisible to a green test suite: the shared `stochastic_environment` has zero wind at every altitude, and zero times any factor is zero, so 1 could not show up in it at all. A bare `StochasticAirBrakes` gives every parameter a standard deviation of zero. Both are set up so the property can actually fail, and the comparison asserts the eccentricities and the air brake are among the fields it compares, since stripping object identity could otherwise empty it.

Locally: full suite green apart from four pre-existing `statsmodels` import failures, the start-method gate passes on fork, spawn and forkserver, and `pylint rocketpy/ tests/ docs/` reports nothing on the files I touched.

@MateusStano the three threads from your review are resolved. Happy to split any of this into a separate PR if that is easier to look at.

thc1006:
@MateusStano this has moved a long way since your review at `761c092`. The three threads you opened are resolved, and two later rounds are in: a per-component seed split, air brakes and eccentricity brought into the reseed, and a pass over the worker and parent failure paths so a run that went wrong cannot report itself as a success. All eight CI jobs are green on `ef6088ea`.

Details are in the PR body rather than repeated here. Two items from the last round are filed as #1090 and #1091 instead of growing this PR, and #1053 stays open until those land, so the closing line is now "partially addresses".

thc1006:
Thanks for this. Four of the findings are real and were mine, and one of them breaks the property this branch exists to establish. Fixed in `96f6d1a` and `a4ca0b5`. The rest I have not taken, and I want to be explicit about why rather than quietly skipping them.

## Fixed

**The base generator drew from the component collections.** This is the serious one. `dict_generator` walks the whole instance, so `parachutes` and `air_brakes` were sampled as ordinary lists, and since this branch moved list choices onto the model's own generator, that draw shifted every later one. `StochasticRocket.dict_generator` then discards it.

```
parachutes=0   mass=14.387755684095282
parachutes=1   mass=14.387755684095282
parachutes=2   mass=14.263716760201630
```

Attaching a drogue changed the sampled mass under a fixed seed. One correction to your write-up: `motors`, `aerodynamic_surfaces` and `rail_buttons` are `Components`, not `list`, so they never reached that branch. Only the two plain lists did. I excluded all five anyway, since the override discards all five.

Also worth recording, because it nearly cost me the test: one component does not reproduce it. `integers(1)` has a single outcome and NumPy returns it without consuming state, so my first attempt passed while the bug was still there. The test now covers 0, 1 and 2.

I did not take the `__stochastic_dict` route. It drops things that are sampled today and arrive outside `kwargs`, so it is a wider behaviour change than the bug needs. A class-level tuple of collections the subclass rebuilds does the same job without touching what anything else samples.

**The parent could wait forever on a worker that was killed outright.** Confirmed with a real manager lock rather than a stub, which is the reason the existing crash test missed it: its `is_alive()` is always `False`, so it only covers the case where everything has already stopped.

```
holder acquires the manager lock, then os._exit(7)
waiter blocks in lock.acquire()
parent: returns, holder=7, waiter=None
parent without the fix: still waiting when the probe was killed at 40 s
```

The wait now also ends on a non-zero exit code, but only when it is unbounded. You are right that the grace period must not be cut short by a crash it already knows about, and it has its own timeout, so keying on that rather than adding a flag.

**`append` below the checkpoint.** Correct, and the consequence is what you said: it ran nothing, reported success, and left more simulations on disk than the caller asked for. It raises now, and there is a control test that appending to the size it already has is still a valid no-op.

**NumPy integers.** Correct, and the `# noqa: E721` was me silencing the symptom. I had used `type()` deliberately, to keep `True` out, and did not notice it took every NumPy integer with it. Both are checked explicitly now.

Local run: ruff, `ruff format --check`, `pylint rocketpy/ tests/ docs/` exit 0, 1941 unit passed, 114 integration passed. Each of the four is pinned by a mutation that undoes it.

## Not taken, and why

These are about files this PR does not touch. It changes six production files: `monte_carlo.py`, `stochastic_model.py`, `stochastic_rocket.py`, `tools.py`, `requirements.txt`, `CHANGELOG.md`.

- **CustomSampler instances sharing a seed.** The problem is real. `_validate_custom_sampler(name, value, seed)` and its call sites predate this branch and are not in its diff, so the correlation you describe is the behaviour on `develop` too. It deserves its own issue and a `seed_group` design discussion, which is roughly the size of this PR again.
- **MonteCarlo bypassing `StochasticFlight.create_object`.** `stochastic_flight.py` is untouched here. This is #1090, already filed.
- **The parachute being rebuilt without its sampled geometry**, and **`isinstance(member, (str, int, float) or callable(member))`.** Both in `stochastic_parachute.py`, untouched. The second one is a nice catch and I would not have found it.
- **`custom_sampler.rst`.** I fixed the fixture, not the document. Fair, and small enough that I will do it if you would rather it came with this.

I would rather file these than grow a PR that is already large. If you disagree on any one of them, say so and I will pull it in.

## Not convinced

**Dependency floors.** `numpy>=1.21.3` and `scipy>=1.8.0` are probably right for the package, but this PR touches `requirements.txt` by one line and the floors are not something it changed. That is a separate decision with its own compatibility question.

**CHANGELOG.** Fair that one line understates it. I will expand it, though the breaking part was already flagged on the PR itself.

thc1006:
Two commits on top of your rebase, @Gui-FernandesBR. Both are defects in this branch's own code that I found while going back over it, so they belong here rather than in a follow-up.

**`27589f6` the fast suite was not gating what its docstring claimed.** It said taking the platform default gated `spawn` on macOS and `forkserver` on 3.14's POSIX default. Neither is true. `multiprocess` hard-codes `fork` on every POSIX platform, macOS included, with a `#FIXME: spawn` still sitting beside the darwin branch:

```python
if sys.platform == 'darwin':
    _default_context = DefaultContext(_concrete_contexts['fork'])  # FIXME: spawn
else:
    _default_context = DefaultContext(_concrete_contexts['fork'])
```

There is no version check in that file, so 3.14 does not change it either. The shipped parallel path was therefore gated on `fork` everywhere except Windows, and the failure message named the stdlib start method rather than the one that made the workers.

The thorough test already covers all three through the real path, and it costs 26 s against 89 s for the rest of that directory, so it is unmarked now and the default-taking one is deleted rather than corrected. Its assertions were a strict subset.

**`047517c` the captured root seed was not a snapshot.** `SeedSequence` keeps a sequence entropy by reference and the capture stored that reference, so a caller who passed a list and later reused it changed the child seeds of a run that had already read the seed:

```
entropy = [1, 2, 3]
mc.simulate(2, random_seed=entropy)
entropy[0] = 999999      # moved every index of that run
```

An int seed was never exposed to this, which is why it went unnoticed. Deep-copied on the way in now.

Each fix has a test that fails when the fix is removed, and the mutation only takes the new test with it. Local: ruff clean, `pylint rocketpy/ tests/ docs/` exit 0, 1970 unit passed, 77 integration passed.

On the rebase itself: I had rebased locally onto the same `develop` before I saw yours, so I checked the trees matched before putting anything on top rather than force-pushing over you. They did, byte for byte, so these two are the only new content.

@MateusStano, your review is from 10 July and the branch has been through a rebase and a fair amount of rework since. Whenever you have time, a fresh look at the current head would be welcome.

thc1006:
Three things from this round, one of which was mine and wrong.

**The renumber advice, `1b291dc`.** The refusal for a legacy one-based checkpoint said "Re-run the study, or renumber the file down by one", while the comment two lines above it said the fix is to re-baseline rather than retry. The comment was right and the message was contradicting it.

Renumbering lines the indices up and leaves the seeds behind. Those rows came from the old sequential scheme, so a renumbered file carries rows 0..n-1 that this release's per-index derivation would never have produced for those indices, and appending onto it joins two different seedings without saying so. The message now says that, and a test fails if the old advice comes back.

**Hole recovery.** I went looking for the same kind of over-promise there and did not find one. The docstring already says what it does:

> A file with a hole in it is refused rather than repaired. Filling holes needs the workers to claim from a plan instead of counting on from the end, which is #1075; until then, refusing loudly beats resuming in the wrong place quietly.

Leaving that as it is.

**The #1102 interaction.** I said earlier that the two conflict in one line of imports. That was true when #1102 was `bffa1fac` and is not now: it has since grown a `seed_group` contract, a grouped reset pass and signature changes to `_validate_custom_sampler` and `_validate_factors`, so it is a lifecycle change rather than an import clash.

The current heads happen to merge with no textual conflict at all, which is the part worth not trusting, so I built the merge and ran it rather than reading it:

```
_validate_factors signature        (self, input_name, input_value)
_validate_custom_sampler signature (self, input_name, sampler)
_reset_custom_samplers  present
component_collections   present
_nominal                present
_random_choice          present

tests/unit/stochastic + tests/unit/simulation   291 passed, 4 skipped
```

Both sides survive and the call sites line up. Whichever lands second should still be rebased and rerun rather than merged on that evidence, but there is nothing to resolve by hand today.

Local on this head: ruff clean, `pylint rocketpy/ tests/ docs/` exit 0, 1971 unit passed.

thc1006:
Pushed `ba029ca5`, which is the serial error path you flagged.

A failure after sampling wrote the inputs to `.errors.txt` with no traceback, while a worker writes `{index, ...inputs, error: traceback}`. So the file the run tells you to read named which inputs failed and not why. Both paths now build the row through one helper.

Three smaller things on the same path, since they only show up once the record is worth reading:

- `inputs_json` is cleared once the pair is on disk. A failure inside `print_update_status()` was reporting an already committed row as one that never finished.
- `sim_idx` is bound before the loop, so a failure on the first iteration has an index to record rather than none.
- the re-raise is bare. `raise error` re-enters the handler frame and it appears twice in the traceback.

`KeyboardInterrupt` keeps its own handler. An interrupt is not a failure with a traceback worth writing to the error file, and it still logs the inputs that did not finish.

On the frame point, my first test for it asserted the failing frame survives, which is true of both spellings, so it passed with the fix removed. The assertion is now that the handler frame appears once:

```
bare raise    ['<module>', 'bare', 'raiser']
raise error   ['<module>', 'rethrow', 'rethrow', 'raiser']
```

Three tests, each pinned by reverting the line it covers and watching it fail.

Also rewrote the `append` docstring, which said only that results are appended. It now says `number_of_simulations` is the target total rather than a number to add, that a lower value is refused, and that the root seed is not stored in the files so the same `random_seed` has to be passed again. Both are covered by tests already in the branch.

One claim from the review I could not stand behind: that `.csv` and `.json` cannot be appended to. The resume index comes from `num_of_loaded_sims`, which is parsed records, not the raw line count I had been looking at, so I have no evidence for it and left it out.

Local: ruff clean, pylint 10.00/10 exit 0, 2159 passed and 45 skipped across unit and integration.

thc1006:
Pushed `47241087`. Both failure-path findings were real, and checking them turned up two more of the same kind, so all four are in this commit.

**Worker committed payload.** Confirmed at the line you pointed at. `_record_simulation()` commits the pair, `print_update_status()` raises, and the handler builds the error record from an `inputs_json` that still holds the committed payload. Index 0 then appears in `.inputs.txt`, `.outputs.txt` and `.errors.txt` with its full sampled inputs. The serial path got this last round; this is its other half.

**Serial reporting replacing the failure.** Also confirmed. The worker guards its own error write and the serial path did not, so an unwritable error file raised `OSError` in place of the exception it was called to record. Reporting is now best effort with a `RuntimeWarning`, not silence, and the traceback is captured before the write. The inputs kept on Ctrl-C are guarded the same way: an interrupt should not become a crash because a file would not open.

Two more found while checking those:

- `_bring_the_fleet_down` says in its docstring that it will not raise over the failure being handled, but only `error_event.set()` was guarded. The two waits under it were not. My own text over-promising, so it is fixed rather than reworded.
- the parallel parent used `raise error`, which you flagged. Bare now, matching serial.

Four tests, each pinned by reverting the line it covers:

```
worker payload not cleared    -> test_a_worker_progress_failure_does_not_repeat_committed_inputs
serial write guard removed    -> test_a_serial_reporting_failure_does_not_replace_the_original
fleet wait unguarded          -> test_a_shutdown_failure_does_not_replace_the_error_it_is_handling
parent back to `raise error`  -> test_the_parent_does_not_repeat_its_own_frame_in_the_traceback
```

**Changelog.** Rewritten and moved to `Changed`. I checked the zero-based claim rather than repeating it: on `develop` the serial path uses the post-increment `sim_monitor.count` while parallel already used `increment() - 1`, so serial was 1-based and parallel 0-based. This branch makes both 0-based, which shifts the `index` field in serial logs. The entry now carries that, the fixed-seed sample change, and that old checkpoints cannot be resumed.

On `_wait_for_workers` polling `N x 0.1s` per round: confirmed by reading, and it is a real ceiling on how fast the parent notices the event with a large fleet. I have left it out of this commit. It is a responsiveness change with its own timing test surface, and this branch is already carrying the seeding change plus four failure-path fixes. Happy to do it here if you would rather it not wait, otherwise I will file it.

Local: ruff clean, pylint 10.00/10 exit 0, 2163 passed and 45 skipped across unit and integration.

thc1006:
Done here rather than as a follow-up issue, `d5ad9ed2`. Tracing it first changed my mind about where it belonged.

`_wait_for_workers` does not exist on `develop`. Nor do `_bring_the_fleet_down`, `_stop_any_worker_still_running` or `_workers_that_crashed`. All four arrive with this branch, so an issue would have pointed at line numbers nobody could find upstream.

What `develop` does today:

```python
for sim_producer in processes:
    sim_producer.join()              # unbounded, in start order
if simulation_error_event.is_set():  # only after every one has joined
    raise RuntimeError(...)
```

The event is not read until the whole fleet has finished, so the run never leaves early at all. The polling you flagged is already a large improvement on that, which is why it was not a regression. It was still worth closing, and it is a smaller change than I had assumed when I said it had its own timing surface.

Measured before, against stubs that never finish, with the event set part way through a round:

```
  1 worker     0.10 s
  8 workers    0.80 s
 32 workers    3.20 s
100 workers   10.01 s
```

One sleep per round now instead of a blocking join per worker. `is_alive()` at the top of the loop already reaps, and the join sweep at the end still runs, so nothing is left unreaped.

One existing test had to change. `test_the_shutdown_window_is_not_cut_short_by_a_worker_already_known_dead` asserted how many times a worker had been joined, which was the mechanism rather than the behaviour, so it failed on a change that kept the behaviour intact. It now measures how long the window lasted. The wait can only overshoot its deadline, never undershoot it, so a floor well under the timeout holds on a coarse clock.

The new test runs at two fleet sizes against the same ceiling, since the claim is that neither depends on the count. Reverting the loop fails the 24-worker case and leaves the single-worker one passing, which is the control.

Local: ruff clean, pylint 10.00/10 exit 0, 2165 passed and 45 skipped.

thc1006:
Both are real and both are in `2f338797`.

**Warning under a warnings-as-errors filter.** Reproduced outside the codebase first, since the claim is about the warning machinery rather than about this file:

```
default filter   caller sees Boom: injected
-W error         caller sees RuntimeWarning: could not be written: OSError(...)
```

So the guard was best effort only under the default filter, and the helper's name and docstring promised more than that. Both reporters now go through one helper that overrides the filter for its own warning and swallows whatever the machinery raises.

**Raw cleanup in `finally`.** Confirmed. The handler's cleanup was already best effort, but `finally` calls the same thing again on the way out and did so raw, so a failure there landed on the caller instead of the exception being re-raised.

One thing I decided differently from your sketch, and want to flag rather than bury. `finally` also runs after a clean run, where nothing is in flight and raising would have been fine and arguably better. I could have suppressed only when an exception is in flight, but warning in both cases is the simpler rule and a cleanup failure is still reported either way. Happy to make it conditional if you would rather keep the clean path loud.

Two tests, each pinned by reverting the line it covers.

On the polling item: that one is already in, at `d5ad9ed2`, one commit before the head your review was taken against. Worth saying what tracing it turned up, because it changed the framing. `_wait_for_workers`, `_bring_the_fleet_down`, `_stop_any_worker_still_running` and `_workers_that_crashed` do not exist on `develop` at all, so this was never a regression against anything. `develop` does:

```python
for sim_producer in processes:
    sim_producer.join()              # unbounded, in start order
if simulation_error_event.is_set():  # only after every one has joined
```

The event is not read until the whole fleet has finished, so today the run never leaves early. Measured before the change, with the event set part way through a round: 0.10 s at one worker, 3.20 s at 32, 10.01 s at 100. It is one sleep per round now, so the delay does not grow with the fleet.

On the transactional input/output pair: agreed, and agreed it should not expand this branch. I will file it rather than leave it in a thread.

On the #1102 interaction: agreed on the shape, and `_validate_custom_sampler` does not reseed on that branch, so the exactly-once contract survives the merge. I have been testing the two together on a throwaway merge into `develop` after each push rather than trusting the textual merge; the last run was ruff clean, pylint exit 0, 2204 passed and 45 skipped, with the mutations that undo each branch's key fix still failing their own tests. I will rerun it on this head.

Local: ruff clean, pylint 10.00/10 exit 0, 2167 passed and 45 skipped.

thc1006:
Both are in, `146785eb`, along with the two documentation points.

**Manager lock release.** Reproduced first, with a stub that releases cleanly once and then loses the connection:

```
before   BrokenPipeError: [Errno 32] Broken pipe
after    Boom: the real simulation failure
```

One context manager owns the lifecycle at all three sites now. Release is best effort only while another exception is on its way out. With nothing in flight a failed release still raises, so a run does not carry on against a dead manager, and there is a test for that direction too.

Worth recording how the first probe misled me, since anyone checking this will hit the same thing. A mutex that fails on every release cannot reach the bug: the first clean claim fails on its own release, before there is anything to mask. The stub has to die partway through.

**Fleet close.** The wait and the forced stop are best effort now. `_fail_if_a_worker_did_not_finish` stays raw, since it is the verdict on the run rather than housekeeping. The event query is a helper that treats an unreachable proxy as a reported failure and names it beside the crash list:

```
before   ConnectionResetError: manager is gone        ("exited with 7" lost)
after    RuntimeError: ... exited with 7 ... the worker error event became unavailable
```

Five tests, four mutations, each killing the test that covers it.

**Docs.** The Notes said an interrupted run can be continued. Not every one can, and a parallel stop is exactly the case that leaves a hole, so that now points at #1075. And `random_seed` said the sampled inputs are identical across modes: what is identical is the index to inputs mapping. Rows land in completion order because a worker takes the log lock once its simulation is done, so a byte for byte diff of two runs would read as reproducibility being broken. No test for that second one on purpose, since the suite already reads the logs into a dict keyed by index, and a test asserting the order does differ would pass or fail on luck.

On the transactional pair, agreed, and filed as #1110 rather than grown into this branch.

**On the #1102 integration.** You flagged that this branch still passes `seed` into `_validate_custom_sampler`, and it does. I merged the two into a throwaway tree on `develop` to see what that costs, and git resolves it correctly on its own: `_reset_custom_samplers` is called once, `_validate_custom_sampler` takes no seed, and this branch's `_nominal`, `__nominal_values` and seeded `_random_choice` all survive. The reason nothing is lost is that `_validate_factors` only handed the seed onward to the reseed that #1102 moved out and does better.

The combined regression you suggested cannot be written yet. `seed_group` exists only on #1102 and `__child_seed` and `random_seed` only here, so no single branch can host it. It belongs in the rebase, and I will add it there rather than leave it implied.

Combined tree on `develop` with #1102, #1103 and #1108: ruff and format clean, pylint exit 0, 2211 passed and 45 skipped, and the mutations that undo each branch's key fix still fail their own tests. This branch alone: 2172 passed, CI 9/9.

thc1006:
All three in, head is `9ed65ae9`, CI 10/10.

**The data collector.** Reproduced before fixing, because the permutation is the part that matters and a constant would not have shown it:

```
sim_idx 0  ->  index 1
sim_idx 1  ->  index 0

indices {0, 1}, each once, against a run of 2   ->  every check passes
```

Three places, since one is not enough. `index` is a reserved key and collector keys have to be strings. `simulate()` validates again, before `__setup_files` opens the logs `w+`, so a rejected run leaves the previous one intact. And the run's own index is written after the custom fields rather than before.

On the string requirement, the reason turned out to be stronger than style. Keys go through JSON, so a non-string one comes back changed, and it can collide:

```python
{1: "int", "1": "str", True: "bool"}  ->  {"1": "bool", "1": "str"}
```

Three keys, one column after the round trip. That is in the test's docstring now.

Four tests. One of them exists only to record why the first matters: a collector returning a permutation passes every completeness check, so a test asserting the row is malformed would not have caught this.

**The develop commit.** Rebased, and it had a consequence worth mentioning. `cbafd728` is the cherry-pick of my own `23be0bab`, taken from the head before I moved two imports to the top of the file, so `pylint` exits 16 on `develop` right now with C0415 on `test_custom_sampler.py`. This branch inherited that on rebase and its lint went red. Fixed here because it is what turns this branch red; #1111 does the same as part of the wider assert change.

I checked the integration you asked about rather than assuming the merge got it right: `__nominal_values`, `_nominal()` and the seeded `_random_choice` are all still here, alongside `_reset_custom_samplers`, `seed_group` and the explicit `raise`. `_validate_custom_sampler` no longer reseeds.

**The two boundaries.** Both added to the `random_seed` notes. The Function metadata one is measured rather than asserted: the cross-start-method test filters `signature` and `.source` and its docstring records that six fields differ across the boundary and all six are those.

I have not brought the other thirteen asserts in `stochastic_model.py` into this branch. They are #1111, which covers 21 across five files and which Gui already agreed to; doing them here as well would only conflict.

Local: ruff and format clean, pylint exit 0, 2223 passed and 45 skipped.

ting-hong-shieh:
I checked the current head, `7053db255f3687c8434eac7524b5cdb38c38a906`, without changing the branch.

- The six focused unit suites for Monte Carlo determinism, log integrity, worker exits/failures, stochastic-model sampling, and rocket component seeding pass: **126 passed**.
- GitHub currently shows all ten Actions checks green, including the six OS/Python test jobs, lint, and docs.
- GitHub reports the pull request as `CONFLICTING` / `DIRTY` against `develop`. The current `develop` SHA is `cb6106a717207dd8fc2dfe1446d80ff75022f21b`.

The remaining blocker appears to be repository state rather than an observed test failure. @thc1006, could you rebase or resolve the conflict against the current `develop`? The earlier Changes Requested review will still need a fresh pass after the resulting head changes.

Local command:

```text
pytest tests/unit/simulation/test_monte_carlo_determinism.py tests/unit/simulation/test_monte_carlo_log_integrity.py tests/unit/simulation/test_monte_carlo_worker_exit.py tests/unit/simulation/test_monte_carlo_worker_failures.py tests/unit/stochastic/test_stochastic_model.py tests/unit/stochastic/test_stochastic_rocket_seeding.py -q
```

Environment: RocketPy source at the head above; Python 3.12.6; NumPy 2.5.2; pytest 9.1.1; macOS 26.5.2 arm64.

rocketpy/simulation/monte_carlo.py
```diff
@@ -180,15 +188,54 @@ def simulate(
...
         parallel : bool, optional
             If True, the simulations will be run in parallel. Default is False.
         n_workers : int, optional
             Number of workers to be used if ``parallel=True``. If None, the
             number of workers will be equal to the number of CPUs available.
             A minimum of 2 workers is required for parallel mode.
             Default is None.
+        random_seed : int, numpy integer, sequence of ints, or SeedSequence, optional
+            Root seed for the run. When provided, the mapping from simulation
+            index to sampled inputs is reproducible and identical across serial
```
wuisabel-gif: This looks good to me overall. One small wording point: the seed tree guarantees reproducibility of the sampled inputs, but not necessarily the complete flight trajectory because some process-global randomness remains outside it. Otherwise, the architecture and test coverage look solid.

Gui-FernandesBR:
This is a quite large PR... I wonder if it would be easy and possible to split it into smaller PRs... If if becomes a too hard task, no problem, just give us more time so we can properly review it

thc1006:
**TL;DR: splitting it. I am putting this back to draft now so nobody spends review time on the current head. Nothing gets thrown away, and I will link the series here before I rewrite the branch.**

Thanks for saying it, and thanks for offering the alternative of just taking more time. Splitting is the better answer. You were right to ask, and I should have stopped adding to this head sooner.

One piece of good news about the size: a fair amount of what accumulated here has landed separately while this sat. The seeded list choice, the per-sampler `CustomSampler` streams, and the eccentricity declarations from #1167 and #1168 are all on `develop` now. I checked each against `develop` rather than assuming, so the split is smaller than 45 commits makes it look.

What I plan to lift out, each based on current `develop`, each green and revertible on its own:

- a stable nominal baseline across reseeding
- independent `StochasticRocket` component streams
- worker shutdown and failure propagation
- checkpoint and log integrity
- persisted append lineage, under #1075

#1054 then keeps only the `random_seed` API, the per-index child derivation, and the tests that cover those directly. I will leave #1126 where it is rather than duplicate it.

@ting-hong-shieh thanks for checking the head and reporting the conflict against `develop`; that saved me a step, and the rebase falls out of the split anyway.

I will ask for review on one at a time rather than sending five at once. No action needed from either of you until the first one is up.

thc1006:
First one is up: #1169, a stable nominal across reseeding. 179 lines over three files, ten checks green, and it stands on its own with nothing else from here.

The series, which I will fill in as each goes up:

- #1169 a stable nominal across reseeding
- independent `StochasticRocket` component streams
- `MonteCarlo` worker shutdown and failure propagation
- checkpoint and log integrity
- persisted append lineage, under #1075

Only #1169 is asking for review right now. The head being split from is kept at `thc1006/RocketPy@backup/pr-1054-cf99e872`, so nothing here goes missing when this branch is rewritten.

thc1006:
Tidying up after myself here, and stating where this actually stands.

**TL;DR:** the branch was rewritten as promised. Head is `83666408`, four commits, +484/-19 against `develop`, still draft. The old 45-commit head is kept at `thc1006/RocketPy@backup/pr-1054-cf99e872` and is not coming back.

What is left in this PR is the thing #1053 asked for and nothing else:

- `simulate(..., random_seed=...)`, keyword-only.
- An immutable root, kept as four picklable values rather than the object.
- Each simulation's child derived from its index in O(1), bit-identical to `spawn(n)[i]`.
- Serial numbering brought in line with parallel, which used to differ by one.

Measured over real flights, four simulations: `serial(42)` equals `parallel(2 workers, 42)` equals `parallel(4 workers, 42)`, and differs from `serial(7)`.

Where the rest of the old branch went, all separately reviewable:

| | |
|---|---|
| #1169 | stable nominal across reseeding |
| #1170 | independent `StochasticRocket` component streams |
| #1181 | the seed type a parallel worker is actually handed |
| #1182 | worker failure reporting and run completeness |
| #1075 | persisted append lineage, still open |

**Correcting something I wrote here earlier.** I said the split-independence guarantee held under `fork` and not under `spawn`, which is what Windows uses. That was my test, not the code. Two fields a serialized `Function` carries belong to the process that wrote the row rather than to the draw: `hash` is the object's identity there, and a callable `source` is its pickle, and the same callable pickles to different bytes in a spawned child. Both drag curves involved are declared `None` on the fixture, so nothing varies them and neither field ever carried a draw.

Measured with the models seeded by hand and no per-index seeding in the way: the parent, a forked child and a spawned child build the same rocket, mass to the last digit. There is a test on the spawn start method now rather than an assumption, and taking the per-index seeding away still turns three of these red, so dropping those two fields has not made the comparison vacuous.

Still true and worth saying: appending continues the same stream only if you pass the same seed again, since nothing here records it. That check is #1075. This stays draft until #1181 lands, which is the only thing under it.

I have also deleted nine of my own older comments on this PR. They were progress notes naming commits that no longer exist in the branch, so following them led nowhere. Everything that answered one of you is untouched.

Reporting every push here was the wrong habit, and I have stopped.

thc1006:
Pushed 3aada2cd, which restores the atomic claim from the earlier review.

`keep_simulating()` and `increment()` were two calls on a manager proxy, so two workers could both see the last slot free and then claim one index each, the second past the end. `claim_next_index()` does both under a lock in the manager's own process, and the two methods it replaces are gone, so the pair cannot be written again by accident.

I want to be straight about the test, because I could not build one that observes the race. A thread-level version does not reproduce it even at `setswitchinterval(1e-9)` with eight claimers over five thousand slots, and putting the two-call loop back leaves the existing parallel test green over five runs. The window is one IPC round trip and two workers have to arrive inside it. The two tests I did add pin the claim contract and both die to a boundary mutation, but the fix rests on construction rather than on a failing test, and I would rather say that than imply otherwise.

The other half of the commit: the index is written after the data collectors, so one named `index` cannot take the number the row's seed was derived from. #1182 rejects such a collector at construction; this makes it impossible in the row as well, which seemed worth having on a branch where that number decides the seed.

thc1006:
@MateusStano thanks for your patience on this one, and for the race in particular. You were right, and it is fixed in 3aada2cd.

Your reading was exactly it. `keep_simulating()` and `increment()` are two calls on a manager proxy with nothing holding them together, so two workers can both see the last slot free and then claim one index each, the second past the end. `claim_next_index()` does the check and the increment under a lock in the manager's own process now, and the two methods it replaces are gone, so the pair cannot be written again by accident.

One thing I want to be straight about: I could not build a test that observes the race. A thread-level version does not reproduce it even at `setswitchinterval(1e-9)` with eight claimers over five thousand slots, and putting the two-call loop back leaves the existing parallel test green over five runs. The window is one IPC round trip and both workers have to arrive inside it, so your reading that it was already there without ever crashing a run matches what I measured. The two tests I added pin the claim contract and both die to a boundary mutation, but the fix rests on construction rather than on a failing test, and I would rather say that than imply otherwise.

Your second point, the `spawn()` call leaving a caller's `SeedSequence` unrepeatable across two runs, is handled by the rewrite: the root is read rather than consumed, and `test_a_caller_seed_sequence_is_not_consumed` passes the same object to `simulate()` twice and compares what came out.

The branch is out of draft now and all nine checks are green. It is still stacked on #1181, so the diff shown here carries that commit until it lands.

thc1006:
Pushed 97f3b45c: a narrowing of the note, and one test stub.

I checked the guarantee this branch documents against an input whose nominal moves. The wording said simulation 7 draws the same inputs whichever way the run was split, which is true, and I had been reading it as meaning the simulation is the same, which it is not. Four simulations with a wind factor, serial against two workers:

| index | serial apogee | 2 workers | same |
|---|---|---|---|
| 0 | 3055.1136 | 3055.1136 | yes |
| 1 | 3065.0436 | 3046.6257 | no |
| 2 | 4148.0313 | 4109.5307 | no |
| 3 | 3885.0862 | 3929.9086 | no |

The recorded inputs match for every index. Three of the four apogees do not. A stochastic input reads its nominal from the object it wraps and a run moves that as it goes, so an index seeded identically can still fly differently depending on which worker took it. That is what #1169 fixes, and it is a prerequisite for reading this note the way it invites.

Rather than deepen the stack I narrowed the note to what this branch delivers on its own. The combined regression belongs on whichever of the two lands second.

Gui-FernandesBR:
Closing this in favour of #1187 — not because anything here was rejected, but because that PR already contains this work in full and carrying both would mean reviewing the same code twice.

Concretely, #1187 is a strict superset of this branch:

- `tests/unit/simulation/test_monte_carlo_seeding.py` and `tests/unit/simulation/test_monte_carlo_simulation_index.py` are byte-identical in both.
- `rocketpy/simulation/monte_carlo.py` goes from +120/-25 here to +326/-27 there, and every line of this diff is present except five comment lines that #1187 rewrote — they describe the append/lineage behaviour it goes on to implement.
- The shared base (`rocketpy/tools.py`, `rocketpy/stochastic/stochastic_model.py`, `tests/unit/stochastic/test_seed_types.py`) is identical again in #1181, which is small enough to land on its own and will shrink #1187 when it does.

On the review state here: the `changes requested` is stale. @MateusStano's two substantive points were both fixed and resolved — the end-of-run race where two workers could pass the `keep_simulating()` check before either incremented and then claim `sim_idx == n`, and `SeedSequence.spawn()` advancing the caller's own lineage so a second `simulate()` with the same object drew a different tree. The one thread still open is @wuisabel-gif's wording point, that the seed tree fixes the sampled inputs rather than the whole trajectory; #1187 carries that into its `docs/user/stochastic.rst` note ("What an index fixes is the inputs it draws").

Thanks @thc1006 for the work and the iterations, and @phmbressan for the review — none of it is lost, it just continues on #1187. Reopen if you think something here does not actually survive in that branch.

Issue #932, opened 2026-03-05
BUG: Incorrect vector rotation in GenericSurface.compute_forces_and_moments


zuorenchen:
**Describe the bug**

Incorrect vector rotation in GenericSurface.compute_forces_and_moments(). 
Here, we want to represent the black `L`, `D`, `Q` vectors in the blue body frame.

The M_BA matrix should be direction cosine matrix (DCM), however it is given in vector rotation matrices. Vector rotation matrices rotate the vector without changing/rotating reference frame. 

Current implementation rotates the force vectors to match body axis, but results are however presented in aero frame.

If it is intended to use rotation matrix instead of DCM, the rotation should be:
- Start with the `Q`, `L`, `D` vectors matching `+X`, `-Y`, `-Z` in body frame (i.e. Vector([side, -lift, -drag]) )
- Rotate the vector by `-alpha` on X axis and `beta` on Y axis.
- The results will match the black  `Q`, `L`, `D` vectors and presented in body frame

Another POV: 
- Only focus on blue Zb axis and body frame.
- Rotate the Zb axis by `-alpha` on X axis and `beta` on Y axis.
- the result will match the black V vector

<img width="786" height="542" alt="Image" src="https://github.com/user-attachments/assets/4424cab2-d766-4bad-a739-b3de85a65c7a" />

<img width="889" height="1052" alt="Image" src="https://github.com/user-attachments/assets/88e1e5df-5692-443c-85c5-5592db56f5b1" />

_____
**To Reproduce & Expected behavior & Screenshots**

<img width="1506" height="308" alt="Image" src="https://github.com/user-attachments/assets/b0fe4ce8-a31c-4b61-aa41-5c07b14b8c90" />

**Additional context**

ting-hong-shieh:
I reduced this to three unit force vectors at `alpha = 0.2 rad` and `beta = 0.1 rad`. Current `develop` is `cb6106a717207dd8fc2dfe1446d80ff75022f21b`.

For the aerodynamic vector `[Q, -L, -D]`, the current code applies the positive-alpha/negative-beta rotation. Applying the direction-cosine mapping `Rx(-alpha) @ Ry(+beta)` gives:

| Unit coefficient | Current body result | DCM body result |
| --- | --- | --- |
| side (`Q=1`) | `[0.995004165, -0.019833838, 0.097843395]` | `[0.995004165, -0.019833838, -0.097843395]` |
[...]

PR #1064, opened 2026-07-11
ENH: Refactor Aero Surface, Make `GenericSurface` the Mother Class

MateusStano:
TBD


docs/user/rocket/generic_surface.rst
```diff
@@ -273,12 +355,22 @@ An example of a ``.csv`` file is shown below:
...
 
-After the definition of the ``GenericSurface`` object, it must be added to the
-rocket's configuration:
+.. note::
+   The ``reynolds`` axis is by default based on the reference length (the
+   rocket diameter). Published rocket data often bases the Reynolds number on
+   the **body length** instead, which for a slender rocket is much larger. If
+   your table uses a different length, pass it as ``reynolds_length`` when
+   creating the surface so the Reynolds number the simulation feeds your table
+   matches the one it was built against.
```
phmbressan, line 364: This phrase is a bit odd, especially the ending "so the Reynolds number the simulation...".

What if the user table is based on the reference length and `reynolds_length` is given? Good docstrings here is important to avoid confusion.

docs/user/rocket/generic_surface.rst
```diff
@@ -435,3 +568,209 @@ shown below:
...
+   base_drag = GenericSurface(
+      reference_area=rocket.area,
+      reference_length=2 * rocket.radius,
+      coefficients={"cA": 0.4},  # axial (drag) coefficient
+      active_during="power_off",
+   )
+
+This is also how a full-vehicle model captures the powered/coasting drag
+difference: build one ``"power_on"`` and one ``"power_off"`` surface and add them
+together (see :ref:`fullbodyaerodynamics`).
```
phmbressan, line 707: Wouldn't it be more fitting to accept (also) an `Event` here? Under the hood, are the default options `power_on` and `power_off` converted to a simulation `Event`?

docs/user/rocket/generic_surface.rst
```diff
@@ -435,3 +568,209 @@ shown below:
...
+      coefficients={...},
+   )
+   rocket.add_full_body_aerodynamics(surface)
+
+Because it is just another aerodynamic surface, a full-vehicle model can be
+**mixed** with modelled add-on surfaces (e.g. a measured body plus modelled
+canards). Pass ``overwrite=True`` to make it the rocket's **only** aerodynamics:
+every existing aerodynamic surface is removed and both built-in drag curves are
+cleared, so the supplied surface(s) provide the complete force set.
+
```
phmbressan, line 741: Maybe a small section or a few more words to clarify exactly when which drag is used could be beneficial. If I understood it correctly:

- `Rocket.power_on_drag` and `Rocket.power_off_drag`: are the default and are used when `GenericSurfaces` are present (e.g. fins, tail, nosecones etc). The other aerodynamic parameters (e.g. lift) do come from a composition of the `GenericSurfaces`, i.e. only drag is ignored (except if the `active_during` is given, which adds to the rocket drag);
- `Rocket.add_full_body_aerodynamics` with `overwrite=False`: as a normal `GenericSurface`, the drag still is the `Rocket.power_on_drag` and `Rocket.power_off_drag`. It could be briefly clarified what is the mathematical/modelling difference of this surface to the non-full-body ones;
- `Rocket.add_full_body_aerodynamics` with `overwrite=True`: the `Rocket.power_on_drag` and `Rocket.power_off_drag` are ignored in favor of the surface. 

Maybe I am mixing things up, but if I got it right, maybe renaming overwrite to `overwrite_rocket_drag` could help here.

docs/user/rocket/generic_surface.rst
```diff
@@ -435,3 +568,209 @@ shown below:
...
+A rocket's drag differs between powered and coasting flight (the motor plume
+lowers the base drag). To capture this, build two surfaces, set each one's
+``active_during`` to ``"power_on"`` and ``"power_off"``, and pass them as a list;
+each then produces force only during its phase.
+
+Extracting a rocket's coefficients
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can also collapses an assembled rocket into a single
```
phmbressan, line 750:
```suggestion
You can also collapse an assembled rocket into a single
```

docs/user/rocket/generic_surface.rst
```diff
@@ -435,3 +568,209 @@ shown below:
...
+for a surface that only exists (or only matters) during a phase. It is accepted
+by both :class:`rocketpy.GenericSurface` and
+:class:`rocketpy.LinearGenericSurface`, and accepts:
+
+- ``"always"`` (default): the surface always contributes force.
+- ``"power_on"``: only while the motor is burning (up to burnout).
+- ``"power_off"``: only after the motor has burned out.
+- a callable ``active_during(t, flight)`` returning ``True`` when the surface is
+  active at time ``t`` (in seconds) of the given :class:`rocketpy.Flight`, for
+  any custom window.
```
phmbressan, line 693: I understand that this is rather intuitive for 99% of rockets and users. However I would like to bring to discussion whether the naming/division `power_on` and `power_off` is adequate model flexibility and future proofing. I ask this, since it has a few assumptions:

- Assumes a single engine rocket; 
- Assumes the effect of the engine on drag is always the same, no matter the throttling or TVC.

Maybe there are more. Overall it feels a bit restrained in terms of design (I understand having this as default/pre-built/easy to use options), but having the option to tweak the parameters (not only toggle) based on the any `flight` parameter or on an `Event` might be interesting.

docs/user/rocket/generic_surface.rst
```diff
@@ -435,3 +568,209 @@ shown below:
...
+.. tip::
+   Tabulated coefficients default to ``extrapolation="constant"`` so they never
+   run to non-physical values past the tabulated envelope. Override it only when
+   you have a specific reason (e.g. ``"zero"`` to make a contribution vanish
+   outside the modeled range).
+
+.. seealso::
+   These arguments are forwarded to each :class:`rocketpy.Function`; see
+   :meth:`rocketpy.Function.set_interpolation` and
+   :meth:`rocketpy.Function.set_extrapolation` for the full list of methods.
```
phmbressan, line 674: Overall, should the recommendation of `"constant"` extrapolation be valid for both `LinearGenericSurface` (based on derivatives) and `GenericSurface`? I ask this, since maybe it is a matter of keeping consistency: if a coefficient has a constant extrapolation, its derivative extrapolation should likely be zero.

I don't know what is more appropriate in this case, but in the `Motor` class, for instance, the `Motor.propellant_mass` has constant extrapolation, whilst the `Motor.net_mass_flow_rate` has zero extrapolation exactly for this reason. I don't know, maybe the most logical for this surfaces case would be constant extrapolation for derivative coefficients and (consequentially) linear extrapolation for the normal coefficients.

rocketpy/rocket/aero_surface/linear_generic_surface.py
```diff
@@ -164,13 +242,30 @@ def __init__(
             coefficients=coefficients,
             center_of_pressure=center_of_pressure,
             name=name,
+            reynolds_length=reynolds_length,
+            extrapolation=extrapolation,
+            interpolation=interpolation,
```
phmbressan, line 247: I would probably make these two `**kwargs` or after a `,*,`. I believe their use here is more justified, but the same parameter is present as positional in the `Motor` class and I feel it is very underused/not useful there.

rocketpy/rocket/aero_surface/fins/_base_fin.py
```diff
@@ -47,20 +56,46 @@ def __init__(
...
+        computed by the first ``_update_geometry_chain`` call."""
+        LinearGenericSurface.__init__(
+            self,
+            reference_area=self.reference_area,
+            reference_length=self.reference_length,
+            coefficients={},
+            center_of_pressure=(self.cpx, self.cpy, self.cpz),
+            name=self.name,
+        )
+        self._barrowman_initialized = True
```
phmbressan, line 98: The architecture here is a bit convoluted, I understand the reasoning, but it is hard to follow due to the amount of methods that need to be called at specific places in the code (e.g. before barrowman/post barrowman) and that have similar names (`update_coefficient`/`calculate_all_coefficients` etc).

Another factor `_update_geometry_chain()` sometimes constructs the superclass and sometimes updates it, which also involves a fair bit of care.

ting-hong-shieh:
I checked two behaviors from the upstream worklist on head `84d3b6a9f0b4a8074d764130dc3d189bdc402036`.

**Scalar coefficients**

`GenericSurface(1, 1, {"cN": 0.75, "cA": 0.2, "cm": -0.1})` constructs successfully. The three values evaluate unchanged and `cN.depends_on == ()`. The focused initialization/constant selection reports **13 passed, 23 deselected**.

**Wind-to-body force mapping (#932)**

At `alpha = 0.2 rad`, `beta = 0.1 rad`, the head produces these body vectors for unit wind-frame coefficients:

```text
side: [ 0.995004165, -0.019833838, -0.097843395]
lift: [ 0.000000000, -0.980066578,  0.198669331]
drag: [-0.099833417, -0.197676812, -0.975170327]
```

Those match `Rx(-alpha) @ Ry(+beta)` applied to `[Q, -L, -D]`; current `develop` has the opposite signs in the affected components. I posted the complete before/after table on #932.

Both pieces appear covered by this draft, but the branch also includes a substantially larger aerodynamic redesign. If a smaller review unit would help, the scalar `AeroCoefficient` support and the #932 force-frame correction are independently testable candidates for extraction. I have not opened overlapping pull requests.

Environment: Python 3.12.6; NumPy 2.5.2; pytest 9.1.1; macOS 26.5.2 arm64. Current `develop`: `cb6106a717207dd8fc2dfe1446d80ff75022f21b`.

Gui-FernandesBR:
This is a huge PR, yet not finished tho

PR #1071, opened 2026-07-19
BUG: seed Monte Carlo per simulation index, not per worker (#1053)

### LINKED ISSUE #1053: Monte Carlo results aren't reproducible across serial and parallel runs (seeding is per-worker, not per-simulation)

wuisabel-gif:
## Pull request type

## Checklist

- [x] Lint passes locally (`ruff check` and `ruff format --check`; pylint 10/10 on the touched files)
- [x] Affected unit suites pass locally (`test_monte_carlo.py`, `tests/unit/stochastic`); relying on CI for the full slow suite
- [x] `CHANGELOG.md` updated

## Current behavior

`MonteCarlo.simulate()` seeds the stochastic models per worker in parallel mode and effectively once at construction in serial mode, so the inputs a given simulation index receives depend on how the run was executed rather than on the index. Parallel spawns one seed per worker from a fresh unseeded `SeedSequence`, and each worker draws for whatever sims it pulls off the shared counter; serial draws one long sequential stream. Result: serial, `n_workers=2`, and `n_workers=4` all sample different inputs, and parallel runs aren't reproducible run to run. This is #1053.

## New behavior

Each simulation's seed is derived from `(master seed, simulation index)` via `SeedSequence`, reseeding the models before every simulation in both paths. Adds an optional `seed` to `MonteCarlo`; the master entropy is captured once, so `seed=None` stays random but consistent across execution modes within a run. A given index now samples identical inputs regardless of serial/parallel or `n_workers`, and a fixed seed reproduces a whole run. Integer child seeds (not `SeedSequence` objects) keep custom samplers' `reset_seed` working.

## Breaking change

- [x] No

With `seed=None` the sampling is unchanged in spirit (still random), but numeric samples for existing unseeded runs will differ, since seeding is now per-index rather than one sequential stream. The `seed` argument is opt-in.

## Additional information

Adds `test_seed_makes_inputs_reproducible_by_index` and `test_unseeded_monte_carlo_is_index_consistent` in `tests/unit/simulation/test_monte_carlo.py`.

Closes #1053

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,8 @@ Attention: The newest changes should be on top -->
 
 ### Fixed
 
+- BUG: seed Monte Carlo per simulation index, not per worker [#1071](https://github.com/RocketPy-Team/RocketPy/pull/1071)
+
 ## [v1.13.0] - 2026-07-04
 
 ### Added
--- a/rocketpy/simulation/monte_carlo.py
+++ b/rocketpy/simulation/monte_carlo.py
@@ -96,6 +96,7 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
         flight,
         export_list=None,
         data_collector=None,
+        seed=None,
     ):
         """
         Initialize a MonteCarlo object.
@@ -132,6 +133,14 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
                     "date": lambda flight: flight.env.date,
                 }
 
+        seed : int, optional
+            Master seed for the whole analysis. Per-simulation seeds are
+            derived from it by simulation index, so a given index draws the
+            same inputs whether the run is serial or parallel and regardless
+            of the number of workers. If None (default), a random master seed
+            is drawn once; results are then random but still consistent across
+            execution modes within that run.
+
         Returns
         -------
         None
@@ -159,6 +168,9 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
         self.export_list = self.__check_export_list(export_list)
         self._check_data_collector(data_collector)
         self.data_collector = data_collector
+
+        self._seed = seed
+        self._base_entropy = np.random.SeedSequence(seed).entropy
 
         self.import_inputs(self.filename.with_suffix(".inputs.txt"))
         self.import_outputs(self.filename.with_suffix(".outputs.txt"))
@@ -285,6 +297,7 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
                 sim_monitor.increment()
                 inputs_json, outputs_json = "", ""
 
+                self.__seed_stochastic_models(sim_monitor.count - 1)
                 flight = self.__run_single_simulation()
                 inputs_json = self.__evaluate_flight_inputs(sim_monitor.count)
                 outputs_json = self.__evaluate_flight_outputs(flight, sim_monitor.count)
@@ -339,13 +352,11 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
             )
 
             processes = []
-            seeds = np.random.SeedSequence().spawn(n_workers)
 
-            for seed in seeds:
+            for _ in range(n_workers):
                 sim_producer = multiprocess.Process(
                     target=self.__sim_producer,
                     args=(
-                        seed,
                         sim_monitor,
                         mutex,
                         simulation_error_event,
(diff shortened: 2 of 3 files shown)


Gui-FernandesBR:
duplicated with #1053

wuisabel-gif:
Closing as a duplicate of #1054 by @thc1006 (the original issue author), which predates this and is already in review. Apologies for the overlap — I should have checked open PRs first. Left a note on #1054.

PR #1085, opened 2026-07-22
BUG: rocket with a late-starting thrust curve never leaves the rail (#411)

### LINKED ISSUE #411: BUG: the rocket does not take off the rail
**Describe the bug**

- I did everything correctly when creating my Flight object, but for some reason the rocket doesn't fly.
- I guess this is happening because my thrust curve doesn't start at time == 0 seconds
- Therefore the SciPy's integrator adopts really large time steps, which make the thrust curve to be ignored.

**To Reproduce**

Try to follow the code block below. You're gonna see that the rocket never leaves the position (0, 0, 0).

```python
liquid_motor = LiquidMotor(
        thrust_source="data/SEBLM/test124_Thrust_Curve.csv",
        burn_time=(8, 20), # this is important to reproduce the error!
        dry_mass=10,
        dry_inertia=(5, 5, 0.2),
        center_of_dry_mass_position=0,
        nozzle_position=-1.364,
        nozzle_radius=0.069 / 2,
    )
liquid_motor.add_tank(pressurant_tank, position=2.007)
liquid_motor.add_tank(fuel_tank, position=-1.048)
liquid_motor.add_tank(oxidizer_tank, position=0.711)

calisto_liquid_modded= Rocket(
        radius=0.0635,
        mass=14.426,
        inertia=(6.321, 6.321, 0.034),
        power_off_drag="data/calisto/powerOffDragCurve.csv",
        power_on_drag="data/calisto/powerOnDragCurve.csv",
        center_of_mass_without_motor=0,
        coordinate_system_orientation="tail_to_nose",
    )
calisto_liquid_modded.add_motor(liquid_motor, position=-1.373)

test_flight = Flight(
        rocket=calisto_liquid_modded,
        environment=Environment(),
        rail_length=5,
        inclination=85,
        heading=0,
        max_time_step=np.inf # this is important to reproduce the error!
)

test_flight.all_info()
```

**Expected behavior**

A description of what you expected to happen.

**Screenshots**

**Additional context**

```bash
tests/test_flight.py::test_liquid_motor_flight
  RocketPy\rocketpy\plots\flight_plots.py:107: UserWarning: Attempting to set identical low and high zlims makes transformation singular; automatically expanding.
    ax1.set_zlim3d([0, max_z])

tests/test_flight.py::test_liquid_motor_flight
  RocketPy\rocketpy\plots\flight_plots.py:108: UserWarning: Attempting to set identical low and high ylims makes transformation singular; automatically expanding.
    ax1.set_ylim3d([min_xy, max_xy])

tests/test_flight.py::test_liquid_motor_flight
  RocketPy\rocketpy\plots\flight_plots.py:109: UserWarning: Attempting to set identical low and high xlims makes transformation singular; automatically expanding.
    ax1.set_xlim3d([min_xy, max_xy])
```

**Possible solutions:**

While discussing this internally, we came up with some suggestions:
- We could set the `max_time_step` as 0.1 * `burn_duration`, for example, to ensure that any simulation will have at least 10 steps.
- The default value of `max_time_step` could now be set to 0.1 instead of `np.inf`
- We could stop initializing the flight simulations at t=0s, instead, starting it at t= burn_time[0]

wuisabel-gif:
## Pull request type

## Checklist

- [x] Tests for the changes have been added (regression test fails without the fix, passes with it)
- [x] Lint passes locally (`ruff check`, `ruff format --check`; pylint 10/10 on the touched file)
- [x] Affected flight tests pass locally (`tests/unit/simulation/test_flight.py`, 56 passed)
- [x] `CHANGELOG.md` updated

## Current behavior

Fixes #411. A motor whose thrust curve starts at `t > 0` (e.g. `burn_time=(8, 20)`) never lifts the rocket off the rail — the flight stays at `(0, 0, 0)` with `out_of_rail_time == 0`.

**Root cause** (as the issue originally guessed): the rail phase's only time nodes are `[t=0, max_time]`. With no thrust at `t=0` the rocket is stationary, so LSODA — whose `max_step` defaults to `inf` — takes one huge step that jumps clean over the burn. The thrust is never sampled and the rocket never accelerates. A normal motor accelerates at `t=0`, forcing small steps, so it only bites for late-starting curves.

Minimal repro (same rocket and total impulse, only the start time differs):

```
[thrust starts t=0] out_of_rail_t=0.329  apogee=10329 m
[thrust starts t=8] out_of_rail_t=0.000  apogee=0 m   <- never leaves the pad
```

## New behavior

`__setup_phase_time_nodes` now forces a solver time node at the motor's ignition and burn-out whenever `burn_start_time > 0`, so the burn is always sampled. With the fix the delayed case flies identically to the `t=0` case, just shifted:

```
[thrust starts t=8] out_of_rail_t=8.329  apogee=10329 m
```

The change is **guarded to `burn_start_time > 0`**, so ordinary motors (which ignite at `t=0`) hit none of the new code and are byte-for-byte unaffected — confirmed by the unchanged `t=0` result above and the existing flight tests staying green.

## Breaking change

- [x] No

## Additional information

Adds `test_flight_with_delayed_burn_leaves_rail` in `tests/integration/simulation/test_flight.py`; it asserts a `burn_time=(8, 20)` rocket leaves the rail after ignition and reaches altitude. It fails on `develop` and passes with this change.

Closes #411


wuisabel-gif:
@RocketPy-Team Could a maintainer please approve the pending GitHub Actions workflows for this fork-based PR? The runs are waiting for approval before they can start. Thank you!

Gui-FernandesBR (approved):
Approving. I reproduced both the bug and the fix, and checked the thing that actually matters for a change inside `__setup_phase_time_nodes` — whether ordinary motors move.

**The bug is real and total.** I reverted only `rocketpy/simulation/flight.py` to `develop`, kept your test, and it fails exactly as you describe:

```
assert flight.out_of_rail_time > 8.0
E  assert 0 > 8.0
```

**Ordinary motors are byte-for-byte unaffected.** Two rockets with the same total impulse, differing only in start time, printed at full precision:

| | develop | with the fix |
|---|---|---|
| `t=0` `out_of_rail_time` | `0.3289354306801905` | `0.3289354306801905` |
| `t=0` `apogee` | `12904.559694188314` | `12904.559694188314` |
| `t=0` `function_evaluations` | `2368463` | `2368463` |
| `t=8` `out_of_rail_time` | `0` | `8.329156434166551` |
| `t=8` `apogee` | `0` | `12904.497507159389` |
| `t=8` `function_evaluations` | `9` | `1953325` |

The `t=0` column is identical down to the solver's function-evaluation count, which is the strongest evidence I could get that the `burn_start > 0` guard means ordinary motors take exactly the same path. And `function_evaluations = 9` on `develop` for the delayed case is a nice confirmation of the diagnosis: the solver really does take one enormous step and stop.

The delayed flight also lands on the same apogee as the `t=0` one to within 0.06 m, which is what you would expect from the same impulse applied 8 s later.

Also checked: `EmptyMotor` is constructed with `burn_time=1`, so its `burn_start_time` is 0 and it falls into the guard rather than the new branch. The `getattr(motor, "burn_start_time", 0) or 0` is defensive but harmless.

Local runs: `tests/unit` 1848 passed / 16 skipped; `tests/unit/simulation/test_flight.py` + `tests/integration/simulation/test_flight.py` 79 passed / 9 skipped; `ruff check` and `ruff format --check` clean; `pylint rocketpy/simulation/flight.py` 10.00/10.

**On the Codecov rename** — I was ready to ask you to drop it as unrelated scope, and then I looked into it. `.github/workflows/codecov.yml` is not a workflow at all; it is Codecov's `coverage:` config sitting in the workflows directory, so Actions has been trying to execute it and failing on **every single run** since #500 introduced it. Moving it to `.codecov.yml` is where Codecov documents it and makes that noise go away. Confirmed the rename is content-identical (`diff` is empty), so keep it — it is a real fix, and worth calling out in the PR body rather than leaving it looking incidental.

Last note: this now has its first green run — all six matrix legs plus Codecov — after #1084 landed and I updated the branch. Every prior run on this PR died in the un-isolated integration step, which is why it never had one.

PR #1050, opened 2026-07-05
Add opening shock force estimation to Parachute class

Matanski:
## Summary
Addresses #161 — estimates the peak transient "opening shock" force
experienced by recovery hardware during parachute inflation.

## Changes
- Added `opening_shock_coefficient` parameter to `Parachute.__init__`
  (default 1.5, per Knacke's typical range of 1.2–2.0).
- Added `Parachute.calculate_opening_shock_force(air_density, velocity)`,
  implementing F₀ = Cx · Cd·S · q from Knacke's "Parachute Recovery
  Systems Design Manual" (1992, Section 5.5).
- Included the new coefficient in `to_dict`/`from_dict` with backward-
  compatible defaulting for old serialized data.
- Added unit tests covering defaults, formula correctness, velocity
  scaling, and serialization round-trip.

## Test plan
- `pytest tests/unit/rocket/test_parachute.py` — 19 passed
- Full unit suite: `pytest tests/unit` — 1546 passed, 12 skipped

PR #1092, opened 2026-08-05
ENH: Add simplified opening shock force estimation

ArthurJWH:
Moving forward with PR #1050 : Add opening shock force estimation to Parachute class


Gui-FernandesBR (approved):
Approved. Moving this out of `Parachute` into `utilities` is the right call, and the migration is complete: `grep -n "shock" rocketpy/rocket/parachute.py` comes back empty, so nothing was left behind from #1050.

What I checked locally:

- `pytest tests/unit/test_utilities.py` — 39 passed, 2 skipped
- `ruff check` clean, `ruff format --check` clean, `pylint rocketpy/utilities.py` — 10.00/10
- The formula matches Knacke's `Cx * Cd*S * q`. Spot-checking a realistic case (`cd_s=10`, `rho=1.225`, `V=30 m/s`, `Cx=1.5`) gives ~8.3 kN, which is the right order of magnitude for sizing a shock cord.
- The function is pure and nothing in the flight path calls it, so this is a pure addition with no regression surface.
- It picks up the API reference automatically, since `docs/reference/classes/utils/utilities.rst` uses `.. automodule:: rocketpy.utilities :members:`.

One nit, not blocking — in the `opening_shock_coefficient` docstring:

```
        :meth:`calculate_opening_shock_force`. Typical values range from
```

That target is left over from the `Parachute`-method version in #1050 and no longer exists anywhere in the repo (this line is its only occurrence). Two small consequences: it is a dangling cross-reference, and `docs.yml` builds with `-W --keep-going` — that workflow is scoped to PRs targeting `master`, so it won't go red here, but it could surface on the next `develop` -> `master` PR. It is also self-referential, since the parameter is described as feeding the very function it is documented in.

A one-liner fixes it, e.g.:

```
    opening_shock_coefficient : float, optional
        Empirical coefficient (commonly noted Cx) used to estimate the
        peak transient force experienced during parachute inflation.
        Typical values range from 1.2 to 2.0 depending on the deployment
        method and canopy type. Default value is 1.5.
```

Happy either way — feel free to fold it in here, or we can take it in a follow-up.

Two things I looked at that need no action: a negative velocity (descent, `vz < 0`) returns the same force as the positive case, which is correct given the `V**2`; and the function isn't re-exported in `rocketpy/__init__.py`, but no utilities function is (`fin_flutter_analysis`, `calculate_equilibrium_altitude` and friends aren't either), so it matches the existing convention.

Issue #1095, opened 2026-08-08
StochasticParachute rejects the callable triggers its docstring accepts


thc1006:
**Describe the bug**

`StochasticParachute._validate_trigger` rejects the callable triggers its own docstring says it accepts:

```python
assert isinstance(trigger, list) and all(
    isinstance(member, (str, int, float) or callable(member))
    for member in trigger
), "`trigger` must be a list of callables, string 'apogee' or ints/floats"
```

The `or` sits inside the `isinstance` call rather than beside it. `(str, int, float)` is a non-empty tuple and therefore truthy, so `X or callable(member)` short-circuits to `X` and `callable(member)` is never evaluated. The check reduces to `isinstance(member, (str, int, float))`, and a callable is none of those.

`Parachute` accepts a callable trigger, `Flight` calls it, and the docstring three lines above says "a list of callables, string 'apogee' or ints/floats". Only the stochastic wrapper refuses one.

**To Reproduce**

```python
def at_apogee(pressure, height, state):
    return state[5] < 0

base = Parachute("c", 10.0, "apogee", 105, 1.5)
StochasticParachute(base, trigger=[at_apogee])
```

```
[callable]        REJECTED
["apogee"]        accepted
[800]             accepted
[callable, 800]   REJECTED
```

The two documented non-callable forms pass, so the argument reaches the check and the check is what refuses it.

**Expected behavior**

A list of callables is accepted, matching the docstring and `Parachute`. Moving the `or` out of the call is the whole fix:

```python
assert isinstance(trigger, list) and all(
    isinstance(member, (str, int, float)) or callable(member)
    for member in trigger
), "`trigger` must be a list of callables, string 'apogee' or ints/floats"
```

Worth a test with a callable in the list, since the current expression passes every test that only uses `"apogee"` or a number.

**Additional context**

Not a regression, and not touched by #1054. Found while reviewing that PR.

Related: #670 (migrate parachute triggers into controllers), #1086 (triggers evaluated twice per time node).

Verified on `develop` at `1691119`, Python 3.12.13.

Signed-off-by: thc1006 <84045975+thc1006@users.noreply.github.com>

Gui-FernandesBR:
Fixed on `develop` in 6224be7e (#1103). The `or` inside `isinstance` is gone; validation now goes through `_is_a_trigger`, which accepts a callable, `"apogee"` case-insensitively, or a real number.

[...]

Issue #1094, opened 2026-08-08
Sampled parachute geometry never reaches the rocket StochasticRocket builds


thc1006:
**Describe the bug**

`StochasticParachute.create_object()` builds a `Parachute` from a complete draw:

```python
generated_dict = next(self.dict_generator())
return Parachute(**generated_dict)
```

`StochasticRocket.create_object()` then throws that object away and builds a second one from six of its ten fields:

```python
parachute = self._create_parachute(parachute)
rocket.add_parachute(
    name=parachute.name,
    cd_s=parachute.cd_s,
    trigger=parachute.trigger,
    sampling_rate=parachute.sampling_rate,
    lag=parachute.lag,
    noise=parachute.noise,
)
```

`radius`, `height`, `porosity` and `drag_coefficient` are not passed. `Parachute.__init__` accepts all four and `Rocket.add_parachute` forwards all four, so the omission is silent: the second parachute re-derives `radius` from `cd_s` and the *default* drag coefficient, and `height` falls back to that radius.

A study that randomizes canopy geometry therefore flies the nominal geometry, while `last_rnd_dict` records the sampled values it did not use. The inputs log and the simulation disagree, and nothing reports it.

**To Reproduce**

```python
stochastic_rocket.add_parachute(
    StochasticParachute(
        chute, radius=(2.0, 0.3), height=(1.5, 0.2),
        porosity=(0.05, 0.01), drag_coefficient=(1.4, 0.2),
    )
)
stochastic_rocket._set_stochastic(42)
built = stochastic_rocket.create_object()
logged = stochastic_rocket.last_rnd_dict["parachutes"][0]
```

Comparing what was drawn against what reached the rocket:

```
                 sampled                on rocket
cd_s             10.0                   10.0
radius            2.0914151239263292     1.5078600877302688
height            1.6500902391612915     1.5078600877302688
porosity          0.05940564716391214    0.0432
drag_coefficient  1.1920031787519008     1.4
```

`radius` and `height` come out equal because both were re-derived. `sqrt(cd_s / 1.4 / pi)` is 1.5078600877303, which is the value in both rows: the derivation used the default drag coefficient, not the sampled one.

**Expected behavior**

The sampled parachute reaches the rocket. `_create_parachute` has already built exactly the right object, so attaching it rather than rebuilding it would be enough:

```python
rocket.parachutes.append(self._create_parachute(parachute))
```

[...]

Issue #1093, opened 2026-08-08
Every CustomSampler in a stochastic model is reset with the same seed


thc1006:
**Describe the bug**

Every `CustomSampler` attached to one stochastic model is reset with the same seed. `StochasticModel._validate_custom_sampler` ends in

```python
sampler.reset_seed(seed)
```

and `seed` is the model's seed, passed unchanged to each sampler in turn. A sampler that follows the documented pattern and builds `np.random.default_rng(seed)` therefore starts from the same state as every other sampler in that model. Two parameters that share a distribution family draw the same value, and two that do not still draw the same underlying deviate.

Quoting the NumPy documentation on parallel random number generation:

> NumPy allows you to spawn new (with very high probability) independent `BitGenerator` and `Generator` instances via their `spawn()` method.

and, in the section on sequences of integer seeds, on deriving several seeds from one root:

> It is quite likely that multiple invocations of the program with different seeds will get overlapping sets of worker seeds [...] subsets of the workers will return identical results, causing a bias in the overall ensemble of results.

Handing one integer to several independent generators is the case that guidance exists to steer away from. Here it is not merely likely to overlap, it is the same integer every time.

**To Reproduce**

Two samplers, different means and different standard deviations, on one rocket:

```python
class Gaussian(CustomSampler):
    def __init__(self, mean, sd):
        self.mean, self.sd = mean, sd

    def sample(self, n_samples=1):
        return list(self.rng.normal(self.mean, self.sd, n_samples))

    def reset_seed(self, seed=None):
        self.rng = np.random.default_rng(seed)


rocket = StochasticRocket(
    rocket=calisto, mass=Gaussian(14.426, 0.5), radius=Gaussian(0.0635, 0.001)
)
rocket._set_stochastic(4242)
drawn = next(rocket.dict_generator())
```

Recovering the standard normal deviate behind each draw:

```
mass     14.659110385288670   ->  0.466220770577340
radius   0.063966220770577    ->  0.466220770577341
```

The two agree to fifteen significant figures. The last digit is the scaling, not the draw. Every later pair agrees the same way, so the correlation is exactly 1 rather than merely high, and a study that varies both parameters is really varying one.

**Expected behavior**

Separate samplers are separate parameters and should be independent, with the model's seed deciding all of them together. Spawning one `SeedSequence` child per sampler would do that and stay reproducible.

[...]

PR #1097, opened 2026-08-08
DOC: make the CustomSampler examples answer to their seed

### LINKED ISSUE #1096: CustomSampler documentation teaches a reset_seed that does nothing
**Describe the bug**

Both `CustomSampler` examples in `docs/user/custom_sampler.rst` build a generator and drop it on the floor:

```python
def reset_seed(self, seed=None):
    np.random.default_rng(seed)
```

The return value is not assigned, so the call has no effect. `sample` then draws from the process-global RNG instead:

```python
mixture_id_list = np.random.binomial(1, self.prob_tuple[0], n_samples)
samples_list[i] = np.random.normal(self.means_tuple[0], self.sd_tuple[0])
```

A user who copies the documented sampler gets one whose `random_seed` does nothing, and the failure is quiet: the study runs, the results look reasonable, and only a second run with the same seed shows they were never reproducible. The page is the reference for writing a sampler, so this is the shape most user samplers will have.

Lines 55, 91, 219 and 241 all have it.

**To Reproduce**

A sampler with the shape the page teaches, reset to the same seed twice:

```python
class AsDocumented(CustomSampler):
    def sample(self, n_samples=1):
        return list(np.random.normal(14.426, 0.5, n_samples))

    def reset_seed(self, seed=None):
        np.random.default_rng(seed)


sampler = AsDocumented()
sampler.reset_seed(4242)
first = sampler.sample(3)
sampler.reset_seed(4242)
second = sampler.sample(3)
```

```
run 1  [14.901167, 14.230133, 14.249566]
run 2  [14.340185, 14.369796, 13.612586]
```

Same seed, different samples. Through `MonteCarlo.simulate(random_seed=...)` the effect is the same, since that is where `reset_seed` is called from.

**Expected behavior**

`reset_seed` keeps the generator and `sample` draws from it:

```python
def reset_seed(self, seed=None):
    self.rng = np.random.default_rng(seed)

def sample(self, n_samples=1):
    mixture_id_list = self.rng.binomial(1, self.prob_tuple[0], n_samples)
    ...
```

The bivariate example needs one more thing. It caches a block of samples, so `reset_seed` has to discard the cache as well, otherwise the first draws after a reseed come from the generator that was replaced.

**Additional context**

The same mistake was in `tests/fixtures/monte_carlo/custom_sampler_fixtures.py` and is fixed there in #1054, which does not touch the documentation. Filing this so the two do not drift apart.

Happy to send the doc patch. It is small and I have the corrected form already.

Verified on `develop` at `1691119`.

thc1006:
Closes #1096.

## Pull request type

## Checklist

The other three items are removed rather than ticked. Nothing under `rocketpy/` or `tests/` changes, so the suite and the linters have nothing to say about this branch and I did not run them on it. What I did run is the documentation build and the code the page contains, both below.

## Current behavior

Both `CustomSampler` examples build a generator and discard it:

```python
def reset_seed(self, seed=None):
    np.random.default_rng(seed)
```

Nothing is assigned, so the call has no effect, and `sample` goes on drawing from the process-global `np.random`. A sampler written by following this page ignores `random_seed` entirely. The failure is quiet: the study runs, the numbers look reasonable, and only a second run with the same seed shows they were never reproducible. Nothing in RocketPy seeds the global RNG, so there is no accident that would rescue it.

The bivariate generator has two further defects, found while fixing the first.

It caches 1000 samples up front. Even with `reset_seed` corrected, those samples came from the generator that was replaced, so the first 1000 draws after a reseed would still ignore the new seed.

Its refill test is

```python
if self.samples_generated < self.used_samples_x:
    self.generate_samples(n_samples)
```

which only becomes true once the cache has already run out, and the slice that follows is not bounded either. A study longer than the cache silently receives short lists:

```
draw 1..1000   sample(n_samples=1) -> 1 value
draw 1001      sample(n_samples=1) -> []
               dict_generator does sample(...)[0] -> IndexError
```

## New behavior

`reset_seed` keeps the generator, `sample` draws from it, and `__init__` routes through `reset_seed` so construction and reseeding agree. For the bivariate generator, `reset_seed` also drops the cache, and the refill covers the shortfall rather than reacting after the fact.

Executing the code blocks straight out of the page, before and after:

```
                                          before        after
mixture, same seed twice                  differs       same
bivariate, same seed twice                differs       same
1501 draws of one sample, short returns   501           0
wind X against wind Y                     IndexError    0.6986
```

The last row is the property the shared generator exists for. The measured correlation matches 0.171/sqrt(0.2*0.3) = 0.6981, so splitting the seeding did not split the two wind components.

I ran those checks by parsing the `jupyter-execute` blocks out of the `.rst` and executing them, rather than retyping the classes, so what passed is what the page says.

The added `.. warning::` is there because the mistake is easy to repeat and produces no error when repeated.

## Breaking change

- [x] No

Documentation only. No module in `rocketpy/` is touched.

## Additional information

The same mistake was in `tests/fixtures/monte_carlo/custom_sampler_fixtures.py` and is fixed in #1054. This uses the same corrected shape so the two do not drift apart. It is independent of that PR and can go in first or second.

Sphinx built clean with the CI settings (`-W --keep-going`, `DOCS_SKIP_EXECUTE=1`): no warning from `custom_sampler.rst`, and the rendered page contains the corrected code with no global draws left.

Related: #1093, #1094, #1095, filed from the same review.

### DIFF
--- a/docs/user/custom_sampler.rst
+++ b/docs/user/custom_sampler.rst
@@ -52,7 +52,7 @@ distributions.
                 2-Tuple that contains the probability of each normal distribution 
                 of the mixture. Its entries should be non-negative and sum up to 1.
             """
-            np.random.default_rng(seed)
+            self.reset_seed(seed)
             self.means_tuple = means_tuple
             self.sd_tuple = sd_tuple
             self.prob_tuple = prob_tuple
@@ -71,12 +71,12 @@ distributions.
                 List containing n_samples samples
             """
             samples_list = [0] * n_samples
-            mixture_id_list = np.random.binomial(1, self.prob_tuple[0], n_samples)
+            mixture_id_list = self.rng.binomial(1, self.prob_tuple[0], n_samples)
             for i, mixture_id in enumerate(mixture_id_list):
                 if mixture_id:
-                    samples_list[i] = np.random.normal(self.means_tuple[0], self.sd_tuple[0])
+                    samples_list[i] = self.rng.normal(self.means_tuple[0], self.sd_tuple[0])
                 else:
-                    samples_list[i] = np.random.normal(self.means_tuple[1], self.sd_tuple[1])
+                    samples_list[i] = self.rng.normal(self.means_tuple[1], self.sd_tuple[1])
 
             return samples_list
 
@@ -88,7 +88,14 @@ distributions.
             seed : int, optional
                 Seed for the random number generator.
             """
-            np.random.default_rng(seed)
+            self.rng = np.random.default_rng(seed)
+
+.. warning::
+    Keep the generator on the instance and draw from it in *sample*. Calling
+    ``np.random.default_rng(seed)`` and discarding the result is a no-op, and
+    *sample* then draws from the process-global ``np.random`` instead, which
+    ``random_seed`` does not reach. The study still runs; it is simply not
+    reproducible, and nothing says so.
 
 This is an example of a distribution that is not implemented in numpy. Note that it is
 a general distribution, so we can use it for many different variables.
@@ -216,14 +223,9 @@ below implements an example of such a generator
             seed : int, optional
                 Number to seed random generator, by default None
             """
-            np.random.default_rng(seed)
-            self.samples_list = []
-            self.samples_generated = 0
-            self.used_samples_x = 0
-            self.used_samples_y = 0
             self.mean = mean
             self.cov = cov
-            self.generate_samples(1000)
+            self.reset_seed(seed)
 
         def generate_samples(self, n_samples = 1):
             """Generate samples from bivariate Gaussian and append to sample list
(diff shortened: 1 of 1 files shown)


thc1006:
Worth knowing when reviewing this one: the Documentation workflow is scoped to `branches: [master]`, so it does not run here. A docs change merged into `develop` gets no build check until it reaches `master`, which means my local run is the only thing standing behind the reStructuredText.

For the record, that run was `sphinx-build -b html -W --keep-going` with `DOCS_SKIP_EXECUTE=1` and `MPLBACKEND=Agg`, the same settings as the workflow. No warning from `custom_sampler.rst`, the page rendered, and the rendered HTML contains the corrected code with no global draws left in it.

Happy to open a separate issue about the branch filter if you think docs PRs onto `develop` should be building. It is a one-line change to the workflow and not something I would fold into this PR.

Gui-FernandesBR (approved):

Issue #1091, opened 2026-08-04
Parachute pressure noise is outside the Monte Carlo seed tree


thc1006:
**Describe the bug**

`Parachute` takes its pressure noise from the process-global NumPy RNG:

```python
# rocketpy/rocket/parachute.py
self.noise_signal = [[-1e-6, np.random.normal(noise[0], noise[1])]]   # line 270
...
+ beta * np.random.normal(noise[0], noise[1])                          # line 285
```

`__init_noise` draws once when the object is built, and `noise_function` keeps drawing from that same global state every time `Flight` samples the trigger. Nothing in `StochasticParachute` reaches it. The stochastic model seeds `cd_s`, `lag`, the sampling rate and the other constructor parameters, but the noise generator is not part of that.

`Flight` adds this noise to the pressure it hands the trigger, and the trigger height is computed from the noisy pressure, so it can move the deployment time and with it the descent, the impact point and the flight time.

There is a second copy of the problem in `StochasticRocket.create_object()`. It builds one `Parachute` through `stochastic_parachute.create_object()` (line 749) to read the parameters off, then builds another through `rocket.add_parachute(...)` (line 841). Both constructions draw from the global RNG and the first object is thrown away.

**To Reproduce**

The shared fixtures already use non-zero noise, so nothing special is needed to reach this:

```python
# tests/fixtures/parachutes/parachute_fixtures.py
noise=(0, 8.3, 0.5)
```

Run the same seed twice in one process, or the same simulation index under different worker counts, and compare parachute events or trajectory outputs rather than `.inputs.txt`. Under `fork` the workers inherit one global state, so their first simulations can share a noise stream. Under `spawn` and `forkserver` each process starts from unrelated entropy, so the same index can get different noise depending on which worker picked it up.

**Expected behavior**

A run identified by a root seed and a simulation index should fly the same trajectory, not only record the same sampled parameters.

Give `Parachute` its own generator and take the noise from it:

```python
class Parachute:
    def __init__(self, ..., rng=None):
        self.rng = np.random.default_rng() if rng is None else rng
```

then derive a child seed for the runtime noise alongside the existing parameter seed, and make sure the `Parachute` that ends up on the `Rocket` is the one holding it. Removing the double construction would be worth doing at the same time, since it is what makes "which object got which stream" ambiguous.

**Additional context**

[...]

PR #1098, opened 2026-08-08
BUG: fly the parachute the simulation sampled

### LINKED ISSUE #1078: Flaky `test_flight_animation_export_gif`: intermittent bus error in VTK off-screen rendering

### LINKED ISSUE #1094: Sampled parachute geometry never reaches the rocket StochasticRocket builds

thc1006:
Closes #1094.

## Pull request type

## Checklist

`pylint` exits 0, and `pytest tests/unit tests/integration` is 2033 passed, 44 skipped. I did not run `-m slow --runslow`. No code under `rocketpy/simulation/` changes, but the values this patch corrects are read by the flight path, so that is a statement about the diff and not about the blast radius. The breaking-change section below has the numbers.

## Current behavior

`StochasticParachute.create_object()` builds a `Parachute` from a complete draw:

```python
generated_dict = next(self.dict_generator())
return Parachute(**generated_dict)
```

`StochasticRocket.create_object()` then discards that object and builds a second one from six of its ten fields:

```python
parachute = self._create_parachute(parachute)
rocket.add_parachute(
    name=parachute.name,
    cd_s=parachute.cd_s,
    trigger=parachute.trigger,
    sampling_rate=parachute.sampling_rate,
    lag=parachute.lag,
    noise=parachute.noise,
)
```

`radius`, `height`, `porosity` and `drag_coefficient` are not passed. `Parachute.__init__` accepts all four and `Rocket.add_parachute` forwards all four, so the omission is silent: the second parachute re-derives `radius` from `cd_s` and the *default* drag coefficient, and `height` falls back to that radius.

Randomizing the geometry is not required to hit this, which is the part I had wrong when I filed the issue. A parachute configured with an explicit radius and no randomization at all:

```
                 configured   flown
radius           2.0          1.5193034332796236
height           1.5          1.5193034332796236
porosity         0.05         0.0432
drag_coefficient 1.4          1.4
```

`radius` and `height` come back equal because both were re-derived. So the affected set is every study whose parachute carries geometry, not only those that sample it, and `last_rnd_dict` records values the flight never used.

## New behavior

The object `_create_parachute` already built is the one that gets attached:

```python
rocket.parachutes.append(self._create_parachute(parachute))
```

`Rocket.add_parachute` does three things: construct, append, return. Nothing else is skipped by attaching directly. Same table after the change:

```
radius           2.0          2.0
height           1.5          1.5
porosity         0.05         0.05
```

This also stops `Parachute.__init__` running twice per parachute per simulation. The second run drew the initial pressure noise from the global NumPy RNG again, which is #1091: fixing the seed tree without fixing this would leave the second initialization undoing the first.

## Tests

Four in `tests/unit/stochastic/test_stochastic_rocket.py`, and each of the two failure modes is pinned by a mutation:

| mutation | fails |
|---|---|
| rebuild through `add_parachute`, as today | 3 |
| build the parachute but never attach it | 2 |

The overlap is deliberate. `test_the_parachute_is_attached_exactly_once` is the control: it survives the first mutation, because the old code did attach parachutes, and it catches the second, which would otherwise let a passing suite fly every Monte Carlo rocket with no parachutes at all.

## Breaking change

- [x] Yes

`radius` and `height` feed the added mass term in `flight.py`, so this is descent dynamics and not just bookkeeping:

```python
ma = self.parachute_added_mass_coefficient * rho * (2 / 3) * np.pi \
     * self.parachute_radius**2 * self.parachute_height
```

For a parachute configured with `radius=2.0, height=1.5` against the 1.5193 it was flying, added mass moves by about 71%. Landing points move with it. That is the bug being fixed, so the old numbers were not the right ones, but anyone with a baselined study will see them change.

One fewer draw from the global NumPy RNG per parachute per simulation, so anything downstream of that stream shifts too. #1091 is the reason that stream should not be there at all.

## One case I would like your call on

There is a third group, and I got it wrong in the issue before measuring it.

If a parachute is built with no explicit geometry, `Parachute.__init__` derives `radius` from `cd_s`, and `StochasticParachute` captures that derived value. The old code then re-derived it downstream from the *sampled* `cd_s`, so a study randomizing `cd_s` had its canopy grow and shrink with it. Attaching the sampled object stops that: `radius` now stays at the value derived from the nominal `cd_s`.

```
cd_s nominal 10.0, sampled 10.609

radius flown, before   1.5531   (from the sampled cd_s)
radius flown, after    1.5079   (from the nominal cd_s)
added mass                      -8.5%
```

Both are defensible. The old one is arguably more physical, since a larger drag area ought to mean a larger canopy. The new one is at least self-consistent, since the object now flies what it recorded, whereas before it recorded 1.5079 and flew 1.5531.

I did not try to keep the old coupling, because deriving geometry from a sampled `cd_s` is a modelling decision rather than a bug fix, and it belongs in `StochasticParachute` rather than here. If you would rather it were preserved, say so and I will follow up with it, either in this PR or a separate one.

## Additional information

Found while reviewing #1054. The two overlap in `rocketpy/stochastic/stochastic_rocket.py`: #1054 changes the base model's list sampling and adds a `component_collections` skip, this one changes the parachute loop below it. They merge cleanly, and I ran the combined tree to check rather than assuming: `tests/unit/stochastic` and `tests/unit/simulation` are 275 passed, 4 skipped on the merge of the two heads. Whichever lands second still wants a rebase.

Related: #1091 (parachute pressure noise outside the seed tree), and #1093, #1095, #1096 from the same review.


thc1006:
CI here is red on #1078 and not on anything in this branch. Three runs so far, all of them dying the same way:

```
Pytest (macos-latest, 3.14)   exit 139, in pyvista's plotter.render during
                              test_flight_animation_export_gif
everything else               cancelled by fail-fast, having found nothing
```

In the run that got furthest, the unit suite was 1855 passed and the doctests 46 passed before the integration job crashed. Locally the whole thing is green: `pytest tests/unit tests/integration` is 2033 passed, 44 skipped, and `pylint rocketpy/ tests/ docs/` exits 0.

I have added the numbers to #1078. The short version is that it is a 3.14 fault rather than a platform one: nine failures on macOS 3.14 and five on ubuntu 3.14 across the last forty runs, the same stack on both, and nothing on 3.10 or on Windows.

Not asking anyone to chase that here. Flagging it so the red is not read as this branch.

thc1006:
Two corrections to the description above, both mine and both now edited in.

I wrote that this touches no file #1054 changes. That is wrong: both change `rocketpy/stochastic/stochastic_rocket.py`. They merge cleanly and I ran the combined tree rather than assuming, `tests/unit/stochastic` and `tests/unit/simulation` are 275 passed and 4 skipped on the merge of the two heads, but the claim as written was false.

I also wrote that nothing here touches the flight path, in the same breath as a breaking-change section explaining that `radius` and `height` feed the added mass term. No code under `rocketpy/simulation/` changes, which is what I meant, but that is a statement about the diff and not about what the change does.

Gui-FernandesBR (approved):

thc1006:
Thanks @Gui-FernandesBR. One check before this goes in, because the description has a question in it that an approval does not obviously answer, and I would rather ask twice than have it merge unnoticed.

The section is "One case I would like your call on". A parachute built without an explicit radius has one derived from `cd_s`. The old double construction re-derived it downstream from the *sampled* `cd_s`, so the canopy tracked the drag area. Attaching the sampled object stops that, and the radius stays at the value derived from the nominal `cd_s`. Measured on one parachute, that is about 8.5% on added mass, in the opposite direction to the 71% this PR fixes for explicitly configured geometry.

Both behaviours are defensible and I did not want to pick. If you are happy with the frozen-derived version, nothing needs to change and I will note it in the issue. If you would rather the derived case kept tracking `cd_s`, that is a change to `StochasticParachute` rather than to this loop, and I would send it separately rather than growing this one.

Rebased onto current `develop` a few minutes ago, so the run on it is post-#1084.

Gui-FernandesBR:
Take the frozen-derived version — nothing needs to change here.

The reason is that the geometry only feeds `ma`. Terminal velocity comes from `cd_s` alone (`flight.py:2770`), so the old coupling moved the opening transient, not the steady descent that sets the landing point. Running your numbers with `mp ~ 16 kg`: added mass +9.3%, but it is ~33% of `mp + ma`, so acceleration moves ~3% — against the ~43-71% error on explicitly configured geometry.

Your own argument settles it anyway: the object recorded 1.5079 and flew 1.5531. A run whose `last_rnd_dict` does not describe the flight cannot be audited. And if we ever want the coupling back, it belongs in `StochasticParachute.create_object` as a visible decision, not as a side effect of double construction. Anyone who wants it today can already correlate `radius`/`height` with `cd_s` via a custom sampler.

Please note it in the issue as you offered. Thanks for asking twice, and for the two corrections to the description — the `stochastic_rocket.py` overlap with #1054 was worth catching.

Issue #1101, opened 2026-08-08
Populate Changelog fails for every pull request from a fork


thc1006:
**Describe the bug**

`Populate Changelog` fails for every pull request opened from a fork, and only for those.

```yaml
on:
  pull_request:
    types: [closed]
    branches:
      - develop
```

GitHub does not pass secrets to a `pull_request` workflow when the head is a fork, so `secrets.RELEASE_TOKEN` is empty and the first step stops:

```
##[error]Input required and not supplied: token
```

The job goes red after five seconds, and since nothing else refers to the changelog, the merge completes and the entry is simply absent.

**To Reproduce**

Recent merges into `develop`, by where the branch lived:

```
#1097  fork (thc1006)      failure
#1070  fork (zuorenchen)   failure
#1052  fork (thc1006)      failure

#1079  RocketPy-Team       success
#1082  RocketPy-Team       success
#1056  RocketPy-Team       success
```

The split is clean in both directions across every run I can still see.

**Expected behavior**

The entry is written whoever the contributor is.

This is not losing history, and I want to be accurate about the size of it. Both older cases were caught by hand:

- #1052 merged on 07-11, entry added on 07-12 in `ba9d130`
- #1070 merged on 07-19, entry added the same day in `d31822d` under "changelog consolidation" while preparing v1.13.0

So what it actually costs is somebody noticing, once per external contribution, and often not until a release is being assembled. #1097 merged this morning and is missing now, which is what sent me looking.

`pull_request_target` would fix it, and unusually for that trigger it is safe here. The job never touches the contributor's branch: it checks out `RocketPy-Team/RocketPy` at `ref: develop` and runs `.github/scripts/update_changelog.py` from that same trusted tree. There is no `head.sha` or `head.ref` anywhere in the file, so nothing from the pull request is executed.

```yaml
on:
  pull_request_target:
    types: [closed]
    branches:
      - develop
```

`github.event.pull_request` keeps the same shape, so `PR_NUMBER` and the `merged == true` guard carry over unchanged.

If you would rather not use that trigger at all, `on: push: branches: [develop]` also gets secrets, at the cost of working out which pull request the push came from.

**Additional context**

Worth saying that this lands on outside contributions specifically, which are the ones where a changelog line does the most good and where the contributor is least likely to know it went missing.

`7f85c072` from April is titled "DEV: correct auto-changelog access token", so the token has had attention before. This looks like a different failure to that one, but I have not gone back to check what that commit was fixing.

[...]

PR #1102, opened 2026-08-08
BUG: give each CustomSampler its own stream instead of the model's seed

thc1006:
Addresses #1093. Not `Closes`, because the keyword only fires when a pull request targets the default branch and this targets `develop`.

## Pull request type

## Checklist

`pylint` exits 0, `pytest tests/unit tests/integration` is 2049 passed, 44 skipped.

## Current behavior

`_validate_custom_sampler` ended in `sampler.reset_seed(seed)`, and `seed` was the model's own, handed unchanged to every sampler on it. A sampler written the way the documentation teaches builds `np.random.default_rng(seed)`, so two of them started from identical state and drew identical values. Not nearly identical, the same to every digit:

```
mass    14.659110385288670  ->  0.466220770577340
radius   0.063966220770577  ->  0.466220770577341
```

Those are the standard normal deviates behind each draw; the last digit is the scaling. A study varying both parameters was varying one, and any correlation it reported between them was an artefact of the seeding.

## New behavior

Each sampler, or each group of samplers that share a generator, gets a child derived from the model's seed and the input names:

```python
def _sampler_seed(seed, input_names):
    root = np.random.SeedSequence(
        entropy=seed, spawn_key=_names_as_spawn_key(tuple(sorted(input_names)))
    )
    words = root.generate_state(4, dtype=np.uint32)
    return sum(int(word) << (32 * position) for position, word in enumerate(words))
```

Keyed by name rather than position, so declaring another parameter does not move the streams already there. The name is length-prefixed into spawn-key words, which no two names share, rather than hashed: a collision would put two samplers back on one stream, which is the bug this exists to prevent. `wd4s4xka50` and `p56cjcee10` are both valid identifiers whose CRC32 is 1560575156, and an earlier version of this gave them the same seed.

The child keeps its full 128 bits, matching the width the Monte Carlo seeding uses rather than being cut to 64.

## Samplers that share a generator

`docs/user/custom_sampler.rst` documents two wrappers over one bivariate generator, so that wind X and wind Y stay correlated. `CustomSampler` gains a `seed_group` property, `self` by default, and those wrappers return the generator they share. A group is seeded once between its members, from a seed derived from all their names.

Resetting each member in turn discarded every seed but the last and left the group's stream decided by whichever member went last, so adding a third wrapper moved the first two. The group is reset directly when it knows how, rather than through one member, since the member cannot be assumed to reset identically or to hold nothing of its own.

The correlation is untouched by any of this. It comes from sharing one `samples_list`, not from sharing a seed:

```
theoretical, 0.171/sqrt(0.2*0.3)   0.6981
measured                           0.6986
```

That is the documented pattern executed out of the `.rst` rather than retyped.

The example also filled a 1000-pair cache inside `reset_seed`. With per-index seeding that reset happens once per simulation, so it built a thousand pairs to use one, every time. The cache starts empty now and `top_up` fills the shortfall on first use.

## Tests

Sixteen in `tests/unit/stochastic/test_custom_sampler.py`. Each fix has a mutation that fails a named test and takes nothing else:

| mutation | fails |
|---|---|
| one seed for every sampler | `two_samplers_do_not_draw_the_same_deviate` |
| fresh entropy per reseed | the reproducibility and ordering tests |
| name key back to CRC32 | `two_names_that_a_hash_would_collide` |
| seed cut to 64 bits | `the_sampler_seed_keeps_the_full_width` |
| seeding in declaration order | `a_shared_generator_lands_on_the_same_seed` |
| group keyed by sampler | `a_shared_group_is_seeded_once` |
| reset through a member | `a_group_that_can_reset_itself_is_reset_directly` |
| helper stops sorting | `a_group_key_does_not_depend_on_the_order` |
| `RuntimeError` only | the legacy `RandomState` test |

The fresh-entropy one is the one worth having. Independence is easy to get by throwing the seed away, and that passes the first test while losing the property the class exists for.

## Breaking change

- [x] Yes

Sampled values change for any study using a `CustomSampler`, since each now receives a different seed. Anything leaning on the accidental correlation will see the parameters move apart, which is the fix rather than a regression, but a baselined study will notice.

A sampler built on the legacy `numpy.random.RandomState` will now fail: it refuses seeds above 2**32-1 and these are 128 bits. The error names the input it belongs to and keeps the original as its cause, rather than surfacing as a bare `ValueError`.

## Additional information

Found while reviewing #1054. The two overlap in `rocketpy/stochastic/stochastic_model.py`: that PR changes the base model's list sampling and nominal snapshot, this one changes the sampler lifecycle. They merge cleanly today and I ran the merged tree rather than assuming, 291 passed and 4 skipped, but whichever lands second wants a rebase rather than a trust in that.

Related: #1096, whose eager cache this takes a bite out of.

### DIFF
--- a/rocketpy/stochastic/stochastic_model.py
+++ b/rocketpy/stochastic/stochastic_model.py
@@ -4,6 +4,7 @@ Stochastic classes.
 """
 
 from random import choice
+from zlib import crc32
 
 import numpy as np
 
@@ -11,6 +12,20 @@ from rocketpy.mathutils.function import Function
 from rocketpy.stochastic.custom_sampler import CustomSampler
 
 from ..tools import get_distribution
+
+
+def _sampler_seed(seed, input_name):
+    """Derive one sampler's seed from the model's, so it gets its own stream.
+
+    Keyed by the input's name rather than its position, so declaring another
+    parameter does not move the stream of the ones already there. ``crc32``
+    because it is stable across processes, which ``hash`` is not.
+    """
+    root = np.random.SeedSequence(
+        entropy=seed, spawn_key=(crc32(input_name.encode("utf-8")),)
+    )
+    return int(root.generate_state(1, dtype=np.uint64)[0])
+
 
 # TODO: Stop using assert in production code. Use exceptions instead.
 # TODO: Each validation method should have a test case.
@@ -467,7 +482,7 @@ class StochasticModel:
             If the input is not in a valid format.
         """
         try:
-            sampler.reset_seed(seed)
+            sampler.reset_seed(_sampler_seed(seed, input_name))
         except RuntimeError as e:
             raise RuntimeError(
                 f"An error occurred in the 'reset_seed' method of {input_name} CustomSampler"
--- a/rocketpy/stochastic/stochastic_model.py
+++ b/rocketpy/stochastic/stochastic_model.py
@@ -4,7 +4,6 @@ Stochastic classes.
 """
 
 from random import choice
-from zlib import crc32
 
 import numpy as np
 
@@ -14,17 +13,33 @@ from rocketpy.stochastic.custom_sampler import CustomSampler
 from ..tools import get_distribution
 
 
+def _name_as_spawn_key(input_name):
+    """Encode a name into spawn-key words with no two names sharing an encoding.
+
+    A hash would be shorter, but a collision puts two samplers back on one
+    stream, which is the bug this keying exists to prevent. The length prefix
+    is what makes it injective.
+    """
+    encoded = input_name.encode("utf-8")
+    payload = len(encoded).to_bytes(4, "little") + encoded
+    payload += b"\0" * (-len(payload) % 4)
+    return tuple(
+        int.from_bytes(payload[at : at + 4], "little")
+        for at in range(0, len(payload), 4)
+    )
+
+
 def _sampler_seed(seed, input_name):
     """Derive one sampler's seed from the model's, so it gets its own stream.
 
     Keyed by the input's name rather than its position, so declaring another
-    parameter does not move the stream of the ones already there. ``crc32``
-    because it is stable across processes, which ``hash`` is not.
+    parameter does not move the stream of the ones already there.
     """
     root = np.random.SeedSequence(
-        entropy=seed, spawn_key=(crc32(input_name.encode("utf-8")),)
+        entropy=seed, spawn_key=_name_as_spawn_key(input_name)
     )
-    return int(root.generate_state(1, dtype=np.uint64)[0])
+    words = root.generate_state(4, dtype=np.uint32)
+    return sum(int(word) << (32 * position) for position, word in enumerate(words))
 
 
 # TODO: Stop using assert in production code. Use exceptions instead.
@@ -97,6 +112,8 @@ class StochasticModel:
         self.__random_number_generator = np.random.default_rng(seed)
         self.last_rnd_dict = {}
 
+        self._reset_custom_samplers(seed)
+
         # TODO: This code block is too complex. Refactor it.
         # TODO: Resetting a instance should not require re-validation.
         for input_name, input_value in self.__stochastic_dict.items():
@@ -104,9 +121,7 @@ class StochasticModel:
                 attr_value = None
                 if input_value is not None:
                     if "factor" in input_name:
-                        attr_value = self._validate_factors(
-                            input_name, input_value, seed
-                        )
+                        attr_value = self._validate_factors(input_name, input_value)
                     elif input_name not in self.exception_list:
                         if isinstance(input_value, tuple):
                             attr_value = self._validate_tuple(input_name, input_value)
@@ -116,7 +131,7 @@ class StochasticModel:
                             attr_value = self._validate_scalar(input_name, input_value)
                         elif isinstance(input_value, CustomSampler):
                             attr_value = self._validate_custom_sampler(
-                                input_name, input_value, seed
+                                input_name, input_value
                             )
                         else:
                             raise AssertionError(
@@ -303,7 +318,7 @@ class StochasticModel:
             get_distribution("normal", self.__random_number_generator),
         )
 
-    def _validate_factors(self, input_name, input_value, seed):
+    def _validate_factors(self, input_name, input_value):
         """
         Validate factor arguments.
 
@@ -332,7 +347,7 @@ class StochasticModel:
         elif isinstance(input_value, list):
             return self._validate_list_factor(input_name, input_value)
         elif isinstance(input_value, CustomSampler):
-            return self._validate_custom_sampler(input_name, input_value, seed)
+            return self._validate_custom_sampler(input_name, input_value)
         else:
             raise AssertionError(
                 f"`{input_name}`: must be either a tuple or listor a custom sampler"
@@ -463,9 +478,34 @@ class StochasticModel:
                 isinstance(member, int) and member >= 0 for member in input_value
             ), f"`{input_name}` must be a list of positive integers"
 
-    def _validate_custom_sampler(self, input_name, sampler, seed=None):
+    def _reset_custom_samplers(self, seed):
+        """Give every sampler its own stream, in an order that does not move.
+
+        Sorted rather than declaration order, because two samplers can share
+        one generator on purpose, as the documented wind pair does, and then
+        whichever is reset last decides the stream. Its own pass rather than
+        the loop below, whose order sets ``__dict__`` and so the order every
+        other input is drawn in.
+        """
+        for input_name in sorted(self.__stochastic_dict):
+            sampler = self.__stochastic_dict[input_name]
+            if not isinstance(sampler, CustomSampler):
+                continue
+            try:
+                sampler.reset_seed(_sampler_seed(seed, input_name))
+            except RuntimeError as error:
+                raise RuntimeError(
+                    f"An error occurred in the 'reset_seed' method of "
+                    f"{input_name} CustomSampler"
+                ) from error
+
+    def _validate_custom_sampler(self, input_name, sampler):
         """
         Validate a custom sampler.
+
+        Seeding is not done here. It happens in ``_reset_custom_samplers``,
+        which runs in a fixed order because two samplers can share one
+        generator and whichever is reset last decides the stream.
 
         Parameters
         ----------
@@ -473,21 +513,15 @@ class StochasticModel:
             Name of the input argument.
         sampler : CustomSampler object
             Custom sampler provided by the user
-        seed : int, optional
-            Seed for the random number generator. The default is None
 
         Raises
         ------
         AssertionError
             If the input is not in a valid format.
         """
-        try:
-            sampler.reset_seed(_sampler_seed(seed, input_name))
-        except RuntimeError as e:
-            raise RuntimeError(
-                f"An error occurred in the 'reset_seed' method of {input_name} CustomSampler"
-            ) from e
-
+        assert isinstance(sampler, CustomSampler), (
+            f"`{input_name}` must be a CustomSampler, not {type(sampler).__name__}"
+        )
         return sampler
 
     def _validate_airfoil(self, airfoil):
--- a/rocketpy/stochastic/stochastic_model.py
+++ b/rocketpy/stochastic/stochastic_model.py
@@ -545,9 +545,13 @@ class StochasticModel:
         AssertionError
             If the input is not in a valid format.
         """
-        assert isinstance(sampler, CustomSampler), (
-            f"`{input_name}` must be a CustomSampler, not {type(sampler).__name__}"
-        )
+        # Raised rather than asserted, the same way #1103 handles it: `python -O`
+        # strips an assert, and the documented AssertionError is kept so callers
+        # that already catch it still do.
+        if not isinstance(sampler, CustomSampler):
+            raise AssertionError(
+                f"`{input_name}` must be a CustomSampler, not {type(sampler).__name__}"
+            )
         return sampler
 
     def _validate_airfoil(self, airfoil):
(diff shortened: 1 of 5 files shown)


thc1006:
Both of these were right and both are fixed in `4a1b6ac`.

**CRC32 collides.** 32 bits is not enough for a stream domain, and a collision puts two samplers back on one stream, which is exactly what this PR exists to prevent. The reproducer holds:

```
crc32("wd4s4xka50") == crc32("p56cjcee10") == 1560575156
_sampler_seed(4242, ...) == 9598975395077455406   for both
```

My "107 built-in names, no collisions" was the wrong test. It says nothing about a subclass, a future field, or a name a user picks. The name is length-prefixed into spawn-key words now, which no two names can share, and the child keeps its full 128 bits instead of being cut to 64. That also lines it up with the width #1054 uses rather than having the two disagree.

**The shared generator was order-coupled, and that one I introduced.** Before this branch both wrappers got the same seed, so order did not matter. One seed per name made whichever wrapper reset last decide the stream:

```
declared x then y   (0.63119306, 0.61655487)
declared y then x   (0.61655487, 0.63119306)
```

Seeding is its own pass over sorted names now, so the shared generator lands on the same seed whichever way the model was written.

## Where I did not follow the suggestion

You proposed a `seed_group` property on `CustomSampler`, with grouping and a per-group reset. I did the sorted pass instead, which is two lines against about forty and needs no change to the ABC or the documented wrappers.

It fixes what I broke. What it leaves is that two wrappers reading one generator take successive values, so swapping the declaration swaps which wrapper gets which number. That is inherent to sharing a generator rather than a seeding property, it is true on `develop` today, and the seed contract holds either way. If you would rather have the group abstraction anyway, say so and I will send it, but I would rather not add the API on my own initiative for a case that already behaves.

## One thing worth flagging

My first attempt at the order fix was to sort the main validation loop, which looked like a one-word change. It is not: that loop's order sets `__dict__`, and `dict_generator` walks `__dict__`, so sorting it moved the draw order of every tuple in every model. A three-input model went from `zulu, alpha, mike` to alphabetical, with different values. The seeding pass is separate for that reason, and there is a test pinning it.

Local: ruff clean, `pylint rocketpy/ tests/ docs/` exit 0, 2041 passed and 44 skipped. Three mutations, one per fix, each taking only its own test.

thc1006:
You were right and I was wrong to argue it away. `1defe34` adds the group semantics.

My case for the sorted pass was that it fixed what I had broken and the rest was inherent. It did fix the order coupling, but it left something I had also introduced, and I did not go looking for it:

```
shared group of wind_x, wind_y      x=0.63119306  y=0.61655487   generator reset 4x
add wind_z to the same generator    x=-0.6761184  y=0.49262482   generator reset 6x
```

Before this branch every sampler took the model's seed, so adding a member to a shared group changed nothing for the others. Keying by name protects independent samplers from exactly that, and left the shared group unprotected, because its stream was whatever the last-sorted member's seed made it. So the inconsistency was mine, not inherent.

`CustomSampler` now has a `seed_group` property, `self` by default, and a group is seeded once between its members with a seed derived from all their names. The documented wind wrappers return their shared generator. The generator is reset once rather than once per wrapper, which also stops the 1000-sample cache in that example being rebuilt and thrown away, so it takes a bite out of #1096 as well.

```
shared group                     reset 1x
independent sampler added        group unmoved
X.seed_group is Y.seed_group     True
correlation                      0.6986 against 0.6981
```

Adding a member to a group still moves that group's stream. That one does look inherent to me: the group is keyed by its members, and a different set of members is a different group. Say if you disagree.

## One thing I did and then undid

Grouping keys on `id`, which is unique only among live objects, so I went looking for whether a `seed_group` that builds its answer could be freed and have its address reused. In a tight loop it certainly can: five successive temporaries all came back with the same `id`. I held a reference and wrote a test.

The test passed with the fix removed. In this loop's actual shape the temporaries do not collide, so the test proved nothing and I have taken it out. The reference is still held, because keying on an `id` you have not retained is fragile whether or not I could make it bite today, but the comment now says it is defensive rather than pretending it fixes something I observed.

Local: ruff clean, `pylint rocketpy/ tests/ docs/` exit 0, 2043 passed and 44 skipped, and the documented example still executes out of the `.rst`.

thc1006:
Three more in `7fb0133`, and the description is rewritten, since it still described the first version of this.

**The documented example filled a cache it threw away.** `reset_seed` generated 1000 pairs, and with per-index seeding that reset happens once per simulation, so a study built on this example produced a thousand pairs and used one, every time:

```
after construction       1000 pairs generated, none asked for
5 reseeds                5000 more generated, 0 used
                         0.099 ms each, about 10 s over 100k simulations
```

The cache starts empty now and `top_up` fills the shortfall on first use. Executed out of the `.rst`: 0 pairs after construction, 4000 after 4000 draws, correlation 0.6986 against the covariance's 0.6981.

**The group reset went through a member.** It dispatched to the first sampler in the group rather than to the group, which assumes every member resets identically and holds nothing of its own. The group owns the shared generator, so it is reset directly when it knows how, with the member as the fallback.

**`_sampler_seed` sorts the names itself.** The caller sorts today; a future one that forgets would hand a single group two different seeds.

I have also removed an earlier comment of mine here that said the conflict with #1054 was one line of imports. That was true of `bffa1fac` and stopped being true once this grew a `seed_group` contract and two signature changes, and it named a `zlib` import that no longer exists. The accurate version is in the description: the two overlap in the sampler lifecycle, they merge cleanly today, and I ran the merged tree rather than trusting that.

Local: ruff clean, `pylint rocketpy/ tests/ docs/` exit 0, 2049 passed and 44 skipped, sixteen tests with a mutation each.

thc1006:
Pushed `74a6ba02` with the ownership rules on `seed_group`, which you asked for and which I had left implicit.

Two ways to break it, both silent, both measured:

Identity has to be stable across calls. Returning from a property makes it easy to build the answer each time, and then every member has a different identity and goes back into a group of its own. A two-member group drops from one reset to two:

```
stable identity, 2 members   -> group reset 1 time
unstable identity, 2 members -> group reset 2 times
```

A group belongs to one model. Declaring the same generator on two models has them both seed it and the later one wins, which is the overwrite the grouping is here to prevent.

I did not add tests for either. The first is a consequence of `test_a_shared_group_is_seeded_once_between_its_members`, which already fails if grouping stops working. The second is a caveat rather than a behaviour worth freezing, and a test asserting the last model wins would fail on anyone who later makes cross-model groups work properly.

I also considered detecting an unstable group and warning, and decided against it. It degrades to per-member seeding, which is the default anyway, so nothing goes wrong that would not have gone wrong without the property.

The wind pair in the docs returns a stored attribute, so the example already teaches the stable form.

Local: ruff clean, pylint 10.00/10 exit 0, stochastic unit suite 39 passed.

thc1006:
Added the changelog entry directly here, as you suggested, since the automation cannot run on a fork pull request until #1101 is fixed. Head is `10a0eeec`.

It records both migrations: fixed-seed `CustomSampler` baselines move, and a sampler built on the legacy `RandomState` has to move to `default_rng` because the seed it now receives is 128 bits wide.

On the merge order, that matches what I found independently: #1104 first, then this branch updated so the Sphinx warnings-as-errors job actually runs against the `custom_sampler.rst` change. I will rebase once #1104 lands rather than now, so the run is against the real merge candidate.

On `StochasticModel(seed=SeedSequence(...))`: agreed it is not a blocker, and agreed the half-accepting behaviour is the part worth avoiding. The docstring says `int`, and `_sampler_seed` builds `SeedSequence(entropy=seed, spawn_key=...)`, so a `SeedSequence` passed in would be rejected there while `default_rng` would have taken it. I would rather reject it early with a clear message than extend the state, since accepting it means deciding what a caller-supplied `spawn_key` composes with. Happy to add that guard here if you want it in scope, otherwise it belongs with the same question in #1054's root seed handling.

rocketpy/stochastic/stochastic_model.py
```diff
@@ -448,31 +486,68 @@ def _validate_positive_int_list(self, input_name, input_value):
...
+                shared[0].append(input_name)
+
+        for names, sampler, group in groups.values():
+            # The group itself when it can be reset, since it is the thing that
+            # holds the shared state. Going through one member instead assumes
+            # every member resets the same way and keeps nothing of its own.
+            resetter = group if hasattr(group, "reset_seed") else sampler
+            try:
+                resetter.reset_seed(_sampler_seed(seed, names))
+            except Exception as error:
```
Gui-FernandesBR, line 519: Broad exception handler. We should avoid it
```python
            except Exception as error:
```

rocketpy/stochastic/stochastic_model.py
```diff
@@ -448,31 +486,68 @@ def _validate_positive_int_list(self, input_name, input_value):
...
             If the input is not in a valid format.
         """
-        try:
-            sampler.reset_seed(seed)
-        except RuntimeError as e:
-            raise RuntimeError(
-                f"An error occurred in the 'reset_seed' method of {input_name} CustomSampler"
-            ) from e
-
+        assert isinstance(sampler, CustomSampler), (
```
Gui-FernandesBR: we should avoid using "assert" in production code. The assert operator was created meaning for tests only

Gui-FernandesBR:
Merged into `develop` as 9593b1de, with your commits and authorship intact.

Done by hand rather than through the button, and worth saying why: the `CHANGELOG.md` entry conflicted with #1103, which landed a few minutes earlier. GitHub's *Update branch* refuses a conflicted head (`422 merge conflict between base and head`), and resolving it on my side means pushing to your fork, which I would rather not do uninvited. So the resolution — both entries kept, yours above #1103's — went in on the merge commit instead. Nothing else was touched: `git diff` against the previous `develop` is exactly your five files, 469 insertions.

What I checked before pushing:

- `pytest tests/unit` on the merge result — 1891 passed, 16 skipped
- The scoping question I had, since seeding moved out of the validation loop into its own pass: `exception_list` is `initial_solution`, `terminate_on_apogee`, `date`, `ensemble_member`, none of which can be a sampler, so `_reset_custom_samplers` walking the whole `__stochastic_dict` covers the same set the old loop did and no more.
- `_sampler_seed` keyed by sorted names, so the two-wrapper wind pair in the docs lands on one seed whichever order it is declared in — your `test_a_shared_generator_lands_on_the_same_seed_whatever_the_order` is the one that would have caught me if I had gotten this wrong.

The behaviour change is real and the changelog says so: fixed-seed `CustomSampler` baselines move, and a sampler built on the legacy `RandomState` now has to move to `default_rng` because the seed carries 128 bits. That is the right trade for two samplers no longer drawing the same deviate.

This PR will show as closed rather than merged, since the merge commit does not come from your branch. It is in `develop` regardless.

Gui-FernandesBR:
Correction, and the miss was mine.

I merged `e7758461` — the head as of when I fetched your branch. You pushed `23be0bab` sixteen seconds before my own push landed, and I did not re-check the head before closing. So the merge went in one commit short: `develop` had the `assert isinstance(sampler, CustomSampler)` you had just replaced, which is the exact `python -O` point I had raised on the review and you had already fixed.

Now applied as `cbafd728`, cherry-picked with your authorship intact. Content identical to `23be0bab`; only the SHA differs.

Verified rather than assumed, since the whole point of that commit is a failure mode the plain test does not see:

- `test_the_refusal_survives_python_dash_o` against the old `assert` — fails, as it should. Against your `raise` — passes. It tests what it claims.
- `pytest tests/unit` — 1897 passed, 16 skipped. Six more than before, which is your two new tests plus the four the earlier merge had already brought in.
- `ruff check` and `ruff format --check` clean.

Two things worth saying plainly. Landing this by hand is what let a stale head through; the button would have refused, and that would have been the better outcome. And on the thirteen other asserts in `stochastic_model.py` — yes, send them separately. Same treatment, and there is a module-level TODO already asking for it.

Nothing was lost, and there was nothing wrong on your end.

thc1006:
Thanks for merging it by hand, and for saying why. The scoping check you did on `exception_list` is the one I would have wanted someone to make.

On your two review points, I went to fix the assert and found something that changes what is worth doing.

**The one you flagged is inert.** Both call sites guard the type immediately before calling:

```python
elif isinstance(input_value, CustomSampler):
    attr_value = self._validate_custom_sampler(input_name, input_value)   # line 141
...
elif isinstance(input_value, CustomSampler):
    return self._validate_custom_sampler(input_name, input_value)         # line 358
```

So it can only fire if the private method is called directly, which is what my own test was doing. Under `python -O` it disappears and nothing changes. I had told myself it was the last thing keeping a non-sampler out of the model; that was wrong, and I only noticed when I went looking for what it protected.

**Thirteen others in the same file are not inert.** They validate what the user passes in, and they do vanish:

```
StochasticModel(obj, mass=("not a number", 0.5))

  python      AssertionError
  python -O   accepted, mass = ('not a number', 0.5, <bound method Generator.normal>)
```

That is a malformed model that will fail somewhere much later with something unhelpful. `rocketpy/` has 22 of these, all in `stochastic/`: 14 in `stochastic_model.py`, 3 each in `stochastic_environment.py` and `stochastic_flight.py`, 1 in `stochastic_aero_surfaces.py`, and 1 left in `stochastic_parachute.py` that #1103 did not reach.

Happy to send all 22 as one pull request, keeping `AssertionError` so nothing that catches it today breaks. That seems more useful than converting the one that cannot fire. Say if you would rather have it split per file, or scoped to `stochastic_model.py` first.

**On the broad handler.** That one I would defend, and I want to give you the reason rather than just leave it. It catches to name the sampler and re-raises with `from error`, so nothing is swallowed and the original stays as `__cause__`. It was `RuntimeError` before, and the reason for widening is in the tests: a sampler built on the legacy `RandomState` raises `ValueError` on a 128-bit seed, and a bare one of those says nothing about which of the model's inputs produced it.

If the convention is firm regardless, the narrow version is `except (RuntimeError, ValueError, TypeError)`, which covers the cases I know about and loses the name for anything else a user's `reset_seed` does. I will send whichever you prefer.

PR #1103, opened 2026-08-08
BUG: accept the callable parachute triggers the docstring promises

thc1006:
Addresses #1095. Not `Closes`, because the keyword only fires when a pull request targets the default branch and this targets `develop`.

## Pull request type

## Checklist

`pylint` exits 0, `pytest tests/unit tests/integration` is 2058 passed, 44 skipped.

## Current behavior

`StochasticParachute._validate_trigger` refused the callable triggers its own docstring promised:

```python
isinstance(member, (str, int, float) or callable(member))
```

The `or` sat inside the `isinstance` call. `(str, int, float)` is a non-empty tuple and therefore truthy, so the expression short-circuited to the tuple and `callable(member)` was never evaluated. The check reduced to `isinstance(member, (str, int, float))`, and a callable is none of those.

Fixing that turned up four more things the check should never have accepted:

```
["banana"]   passed here, then Parachute raised ValueError at create time
[True]       passed as a height of one metre, since bool is an int
[]           passed, because all([]) is True
numpy.int64  refused, while numpy.float64 was accepted
```

## New behavior

The check now accepts exactly what `Parachute` accepts, and nothing else:

```python
def _is_a_trigger(member):
    if callable(member):
        return True
    if isinstance(member, str):
        return member.lower() == "apogee"
    return isinstance(member, (int, float)) and not isinstance(member, bool)
```

`(int, float)` rather than `numbers.Real`, deliberately. `Parachute` uses that exact check, so widening here would let `numpy.int64` past validation and into `create_object`, which is the same defect as accepting `"banana"`: not permissiveness, but moving the failure somewhere less obvious. `bool` is excluded because `Parachute` would otherwise read `True` as one metre without complaint.

The whole contract, checked against both sides in one test:

```
value              Parachute   this check
callable           accepts     accepts
"apogee"/"APOGEE"  accepts     accepts
"banana"           refuses     refuses
800 / 800.0        accepts     accepts
numpy.float64      accepts     accepts
numpy.float32      refuses     refuses
numpy.int64        refuses     refuses
None / {} / ()     refuses     refuses
True / False       accepts     refuses
```

The last row is the one deliberate gap, in the safe direction.

The check is also raised from an `if` rather than asserted. It stays an `AssertionError`, so nothing catching it has to change, but `python -O` strips an `assert` outright and this is the only thing between those triggers and a `Parachute` that either refuses them later or misreads them.

The docstring promised a tuple form that was never implemented, and now describes what the code does.

## Tests

Twenty-five. Five mutations, one per rule, each taking only its own case:

| mutation | fails |
|---|---|
| callable branch removed | the callable and mixed cases |
| accept any string | `["banana"]` |
| drop the bool exclusion | `[True]` |
| drop the non-empty check | `[]` |
| `(int, float)` widened to `numbers.Real` | `numpy.int64` |
| `raise` back to `assert` | the `-O` test |

There is also the end-to-end case that was missing: a callable has to survive `create_object` and be what gets called, rather than merely be accepted by the constructor.

```
built.trigger is _at_apogee
triggerfunc(0.0, 100.0, descending, [], [])  -> True
triggerfunc(0.0, 100.0, ascending,  [], [])  -> False
```

## Breaking change

- [x] No

It accepts callables, which the docstring already promised, and refuses four forms that either failed later or were silently misread.

## Additional information

`Parachute`'s own numeric check is asymmetric: `numpy.float64` passes because it subclasses `float`, `numpy.int64` does not because it subclasses neither, and `True` is taken as a height. This follows that contract rather than trying to improve on it, and the asymmetry is filed separately as #1106, where widening would not strand anything downstream.

Found while reviewing #1054, and unrelated to it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ Attention: The newest changes should be on top -->
 
 ### Fixed
 
+- BUG: Accept the callable parachute triggers `StochasticParachute` documents, and reject the ones it cannot mean. An invalid string, an empty list or a boolean now fails during validation instead of reaching `Parachute` or becoming a one-metre height trigger. [#1103](https://github.com/RocketPy-Team/RocketPy/pull/1103)
 - BUG: rocket with a late-starting thrust curve never leaves the rail [#1085](https://github.com/RocketPy-Team/RocketPy/pull/1085)
 
 ## [v1.13.0] - 2026-07-21
--- a/rocketpy/stochastic/stochastic_parachute.py
+++ b/rocketpy/stochastic/stochastic_parachute.py
@@ -120,8 +120,11 @@ class StochasticParachute(StochasticModel):
         - a tuple that will be further validated in the StochasticModel class
         """
         if trigger is not None:
+            # The `or` belongs beside the isinstance, not inside it. A non-empty
+            # type tuple is truthy, so `X or callable(member)` short-circuited to
+            # X and the callables this docstring promises were refused.
             assert isinstance(trigger, list) and all(
-                isinstance(member, (str, int, float) or callable(member))
+                isinstance(member, (str, int, float)) or callable(member)
                 for member in trigger
             ), "`trigger` must be a list of callables, string 'apogee' or ints/floats"
 
--- a/rocketpy/stochastic/stochastic_parachute.py
+++ b/rocketpy/stochastic/stochastic_parachute.py
@@ -1,8 +1,24 @@
 """Defines the StochasticParachute class."""
 
+from numbers import Real
+
 from rocketpy.rocket import Parachute
 
 from .stochastic_model import StochasticModel
+
+
+def _is_a_trigger(member):
+    """One of the three forms ``Parachute`` accepts.
+
+    ``Real`` rather than ``(int, float)``, which took ``numpy.float64`` but not
+    ``numpy.int64``. ``bool`` is excluded because it is an ``int``, and
+    ``Parachute`` would read ``True`` as a height of one metre.
+    """
+    if callable(member):
+        return True
+    if isinstance(member, str):
+        return member.lower() == "apogee"
+    return isinstance(member, Real) and not isinstance(member, bool)
 
 
 class StochasticParachute(StochasticModel):
@@ -114,19 +130,26 @@ class StochasticParachute(StochasticModel):
         )
 
     def _validate_trigger(self, trigger):
-        """Validates the trigger input. If the trigger input argument is not
-        None, it must be:
-        - a list of callables, string "apogee" or ints/floats
-        - a tuple that will be further validated in the StochasticModel class
+        """Validates the trigger input. If not None, it must be a non-empty
+        list whose members are each a callable, the string "apogee", or a
+        height. One of those is chosen per simulation.
         """
-        if trigger is not None:
-            # The `or` belongs beside the isinstance, not inside it. A non-empty
-            # type tuple is truthy, so `X or callable(member)` short-circuited to
-            # X and the callables this docstring promises were refused.
-            assert isinstance(trigger, list) and all(
-                isinstance(member, (str, int, float)) or callable(member)
-                for member in trigger
-            ), "`trigger` must be a list of callables, string 'apogee' or ints/floats"
+        if trigger is None:
+            return
+
+        valid = (
+            isinstance(trigger, list)
+            and bool(trigger)
+            and all(_is_a_trigger(member) for member in trigger)
+        )
+        # Raised rather than asserted: `python -O` strips an assert, and this
+        # is the only thing standing between a bad trigger and a Parachute
+        # that either refuses it much later or reads True as a height of 1.
+        if not valid:
+            raise AssertionError(
+                "`trigger` must be a non-empty list whose members are "
+                "callables, the string 'apogee', or heights"
+            )
 
     def _validate_noise(self, noise):
         """Validates the noise input. If the noise input argument is not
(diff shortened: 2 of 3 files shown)


thc1006:
All four taken, in `ba9b543`. Each one reproduced before I touched it, and one of them is worse than the review says.

**Strings.** `["banana"]` passed this check and then `Parachute.__init__` raised `ValueError`. So the wrapper was not being permissive, it was moving the failure to create time, which is the opposite of what a validator is for. Only "apogee" now, case-insensitively, since `Parachute` accepts `APOGEE` and `Apogee` too.

**Booleans.** This is the one I would call a real bug rather than a loose check. `bool` is an `int`, so `[True]` passed and `Parachute` took it as a height of one metre. No error anywhere, just a parachute that deploys at 1 m.

**Empty list.** `all([])` is True, so `[]` passed here and failed later.

**Numeric types.** While checking the above I found `(int, float)` was inconsistent rather than merely narrow:

```
[numpy.float64(800)]   accepted     float64 subclasses float
[numpy.int64(800)]     refused      int64 subclasses neither
```

`numbers.Real` takes both, with `bool` excluded explicitly.

**`python -O`.** Kept as `AssertionError` so nothing catching it has to change, but raised from an `if` rather than asserted. A test reads the source and fails if an `assert` statement comes back, since the behaviour it guards is only visible under `-O`.

**The docstring.** It promised "a tuple that will be further validated in the StochasticModel class". No tuple has ever reached that path; the check has always required a list. Rewritten to say what the code does, and `(800,)` is in the refusal list.

## Tests

21 now. Five mutations, one per rule, each taking only its own case:

| mutation | fails |
|---|---|
| drop the bool exclusion | `[True]` |
| accept any string | `["banana"]` |
| drop the non-empty check | `[]` |
| `numbers.Real` back to `(int, float)` | `numpy.int64(800)` |
| `raise` back to `assert` | the `-O` test |

There is also the end-to-end case that was missing: the callable now has to survive `create_object` and be what gets called, not merely be accepted by the constructor.

```
built.trigger is _at_apogee
triggerfunc(0.0, 100.0, descending, [], [])  -> True
triggerfunc(0.0, 100.0, ascending,  [], [])  -> False
```

Local: ruff clean, `pylint rocketpy/ tests/ docs/` exit 0, 2054 passed and 44 skipped.

thc1006:
Good catch, and it is the same mistake I spent the last round criticising. Fixed in `0b21b2e`.

I widened the height check to `numbers.Real` because `(int, float)` looked arbitrarily narrow. `Parachute` uses that exact check, and the two numpy types land on opposite sides of it:

```
numpy.float64(800)   accepted    subclasses float
numpy.int64(800)     ValueError  subclasses neither
```

So `numbers.Real` let `numpy.int64` past the wrapper and into `create_object`, where `Parachute` raised. That is precisely what I said was wrong about accepting `"banana"`: not permissiveness, but moving the failure to a later and less obvious place. I wrote the principle down in the same commit that broke it.

The test made it worse. `test_a_height_is_a_height_whatever_numeric_type_it_arrives_as` asserted `numpy.int64` was accepted, so it was pinning the defect rather than catching it.

Back to `(int, float)` now, matching `Parachute` rather than improving on it, and the tests are rewritten around the property instead of a list of types:

- what the wrapper accepts, `Parachute` also accepts, asserted on both in the same test
- `numpy.int64` and `numpy.int32` are refused at both ends, with the reason in the docstring

Putting it back to `numbers.Real` now fails those.

The asymmetry itself is real and is `Parachute`'s rather than this wrapper's, so I have filed it as #1106 instead of carrying it here. A height read out of an array or a pandas-parsed config arrives as `numpy.int64` whenever it has no decimal point, so it is reachable without anyone reaching for a numpy scalar deliberately. Widening the check there would let both ends agree, and this wrapper follows whatever that one decides.

Local: ruff clean, `pylint rocketpy/ tests/ docs/` exit 0, 2058 passed and 44 skipped.

thc1006:
This was true, and it is what `0b21b2e` fixed a few hours ago. Since the two contracts are easier to check than to argue about, here is the whole table on the current head rather than the numeric row alone:

```
value              Parachute   wrapper
callable           accepts     accepts
"apogee"           accepts     accepts
"APOGEE"           accepts     accepts
"banana"           refuses     refuses
800                accepts     accepts
800.0              accepts     accepts
numpy.float64      accepts     accepts
numpy.float32      refuses     refuses
numpy.int64        refuses     refuses
numpy.int32        refuses     refuses
None               refuses     refuses
{}                 refuses     refuses
True               accepts     refuses
False              accepts     refuses
```

Twelve of fourteen agree. The two that do not are `True` and `False`, and that gap is on purpose: `bool` is an `int`, so `Parachute` takes `True` as a height of one metre without complaint. Refusing it in the wrapper is the safe direction, and a test pins it.

What is left is the asymmetry inside `Parachute` itself, which takes `numpy.float64` because it subclasses `float` and refuses `numpy.int64` because it subclasses neither. The wrapper now mirrors that rather than trying to improve on it, so the two are consistently narrow instead of inconsistently so. Widening belongs in `Parachute`, where nothing downstream would be stranded by it, and that is #1106.

One smaller thing from the same round, in `71a6688e`: the trigger tests built a 14-element state and `Flight` passes 13. Only `y[5]` is read so both worked, but a test whose job is to document the contract should not misstate it.

thc1006:
Added the changelog entry here rather than waiting on the automation, which cannot run on a fork pull request until #1101 is fixed. Head is `8b165129`.

Your wording on the scope is better than mine, so the entry follows it: no documented API break, but an invalid string, an empty list or a boolean now fails during `StochasticParachute` validation instead of reaching `Parachute` construction or silently becoming a one-metre height trigger.

Issue #1099, opened 2026-08-08
Documentation workflow does not run for pull requests targeting develop


thc1006:
**Describe the bug**

The Documentation workflow only runs for `master`:

```yaml
on:
  pull_request:
    types: [opened, synchronize, reopened, ready_for_review]
    branches: [master]
    paths:
      - "docs/**"
      - "rocketpy/**"
      ...
  push:
    branches: [master]
```

Contributions do not go to `master`. Of the last 30 pull requests, 25 target `develop` and 3 target `master`. So the build that runs with `-W --keep-going` sees almost nothing before it is merged, and a docs change reaches it only once `develop` is promoted, in a batch, away from the change that caused it.

The path filter already limits this to work that can affect the docs, and `rocketpy/**` is in that filter, so the gap is not only about `docs/`. A docstring edit in a PR to `develop` can break the autodoc reference and nothing says so.

**To Reproduce**

#1097 changes `docs/user/custom_sampler.rst` and targets `develop`. Its only check is the auto-assign workflow. No documentation build ran.

**Expected behavior**

Add `develop` to both branch filters:

```yaml
    branches: [master, develop]
```

The job already sets `DOCS_SKIP_EXECUTE: "1"`, so it renders `jupyter-execute` cells as static blocks and makes no network calls. The last six runs took between 2m18s and 2m31s, and the path filter keeps it off PRs that cannot affect the docs.

If running it on every `develop` PR is more than you want, the `pull_request` filter alone would still close most of the gap, since that is where the feedback is useful.

**Additional context**

Found while opening #1097, where the reStructuredText I added had nothing behind it but a local `sphinx-build`. I would rather this were checked by CI than by whether I remembered to run it.

Happy to send the one-line PR if you want it.

Verified against `develop` at `1691119`.

Signed-off-by: thc1006 <84045975+thc1006@users.noreply.github.com>

Gui-FernandesBR:
Fixed on `develop` in a5867508 (#1104). The Documentation workflow now triggers on `branches: [master, develop]`, so a docs change meets `-W --keep-going` in its own pull request instead of in the release batch.

The trigger exercised itself on the way in: `.github/workflows/docs.yml` is inside the workflow's own `paths` list, so the change fired the build, and it passed against the merge ref.

PR #1104, opened 2026-08-08
CI: build the docs for pull requests into develop as well

thc1006:
Addresses #1099. Not `Closes`, because the closing keyword only fires when the pull request targets the default branch, and this one targets `develop`. #1094 and #1096 are both still open for exactly that reason.

## Pull request type

## Checklist

One workflow file changes and nothing under `rocketpy/` or `tests/` does, so no code test was added. The Tests, Linters and Documentation workflows all pass on the current head.

## Current behavior

```yaml
  pull_request:
    branches: [master]
```

`branches` filters on the pull request's base, and almost nothing here targets `master`. Of the last 30 pull requests, 25 went to `develop` and 3 to `master`. So the build that runs with `-W --keep-going` first sees a docs change when `develop` is promoted, in a batch, a long way from whatever caused it.

#1097 is the case that prompted this. It changed `docs/user/custom_sampler.rst`, targeted `develop`, and its only check was the auto-assign workflow. My local `sphinx-build` was the only thing standing behind the reStructuredText, which is not where that check belongs.

The path filter includes `rocketpy/**`, because docstrings feed the autodoc reference, so this is not only about `docs/`. A docstring edit merged into `develop` can break the API pages with nothing to say so.

## New behavior

```yaml
    branches: [master, develop]
```

`push` is deliberately left on `master` alone. The pull request check is where the feedback is useful, and running both would pay twice for a second opinion on the same commit.

## Breaking change

- [x] No

## Additional information

Two things worth knowing before adding a `-W` build to every docs-affecting pull request. A check rather than a gate: whether it can block a merge is a branch-rule decision, not this file's, and a workflow with top-level `paths` filters is the wrong shape to make required as it stands.

**It is not slow.** The job sets `DOCS_SKIP_EXECUTE: "1"`, so notebooks are not executed and `jupyter-execute` cells render as static blocks. The documentation examples therefore make no live weather or external-data calls, which is the expensive and flaky part; the job still installs pandoc and the Python dependencies like any other. The run on this pull request took 2m27s.

**`develop` builds clean today**, so this does not switch on a check that is already red. This pull request proves it rather than arguing it: `.github/workflows/docs.yml` is inside the workflow's own `paths` list, so the change triggered itself. `Documentation / build-docs (3.12)` ran against the merge ref and the warnings-as-errors build finished clean.

The workflow asks for `contents: read` and uses no repository or organization secrets, so a fork pull request runs it with a read-only token, subject to whatever approval policy the repository applies to outside contributors. That is where it is needed most.

It has to stay on `pull_request` rather than `pull_request_target`: it checks out the merge tree and then installs and imports the package, `docs/conf.py` and the Sphinx extensions from it, all of which the pull request controls.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,9 +1,6 @@
 name: Documentation
 
 on:
-  # PRs into either long-lived branch, and pushes to master. `branches` filters
-  # on the PR's base, and almost every PR is opened against develop, so master
-  # alone meant the build first saw a docs change in the release batch.
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
     branches: [master, develop]
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,10 +1,12 @@
 name: Documentation
 
 on:
-  # Only PRs targeting master (base branch = master) and pushes to master.
+  # PRs into either long-lived branch, and pushes to master. `branches` filters
+  # on the PR's base, and almost every PR is opened against develop, so master
+  # alone meant the build first saw a docs change in the release batch.
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
-    branches: [master]
+    branches: [master, develop]
     paths:
       - "docs/**"
       - "rocketpy/**" # docstrings feed the autodoc API reference


thc1006:
The new trigger has now exercised itself: `.github/workflows/docs.yml` is inside the workflow's own `paths` list, so this change fired it, and `Documentation / build-docs (3.12)` ran against the merge ref and finished clean with warnings as errors. Nothing but the base-branch filter changes; paths, permissions and build steps are as they were.

I have corrected four things in the description that claimed more than the change delivers.

`Closes #1099` is now `Addresses #1099`. The keyword only closes when the pull request targets the default branch, and this targets `develop`. I had already watched that happen twice this week, with #1094 and #1096 still open behind merged pull requests, and wrote `Closes` anyway. The issue will need closing by hand.

I had called this "a gate". It produces a check; whether anything blocks on it is a branch rule rather than this file. Worth adding that this workflow is not in a shape to be made required as it stands, because a top-level `paths` filter means a pull request that touches no documentation never starts the check at all rather than passing it.

I had written that "nothing reaches the network". The documentation examples make no live calls, which is the part that matters, but the job still does `apt-get install pandoc` and four `pip install` commands, so as written it was simply wrong.

The validation section described a local build that had to exclude the notebooks for want of pandoc. The real run has since made that redundant, so it now cites that instead.

.github/workflows/docs.yml
```diff
@@ -1,10 +1,12 @@
 name: Documentation
 
 on:
-  # Only PRs targeting master (base branch = master) and pushes to master.
+  # PRs into either long-lived branch, and pushes to master. `branches` filters
+  # on the PR's base, and almost every PR is opened against develop, so master
+  # alone meant the build first saw a docs change in the release batch.
```
Gui-FernandesBR:
```suggestion
```

Gui-FernandesBR (approved):

Issue #1107, opened 2026-08-08
requirements allow a SciPy too old to import the package


thc1006:
**Describe the bug**

`requirements.txt` allows SciPy versions that cannot import the package:

```
numpy>=1.17
scipy>=1.0
```

`rocketpy/simulation/monte_carlo.py` imports at module level:

```python
from scipy.stats import bootstrap
```

Quoting the SciPy 1.7.0 release notes:

> `scipy.stats.bootstrap` has been added to allow estimation of the confidence interval and standard error of a statistic.

So anything from SciPy 1.0 to 1.6 satisfies the declared floor and then fails on `import rocketpy`, because `monte_carlo` is reached from the package `__init__`. Not a degraded feature, an ImportError before anything runs.

The NumPy floor has a milder version of the same problem. `pyproject.toml` says `requires-python = ">=3.10"`, and NumPy did not ship a full set of cp310 wheels until 1.21.3:

```
numpy 1.21.0   no cp310 wheels
numpy 1.21.2   cp310 for manylinux x86_64 and aarch64 only
numpy 1.21.3   adds macOS universal2, macOS arm64 and win_amd64
```

A resolver on 3.10 that lands below that builds NumPy from source, which is slow where it works and a wall of compiler errors where it does not.

**To Reproduce**

```
pip install "scipy==1.6.3" "numpy>=1.17"
python -c "import rocketpy"
```

`import rocketpy` does reach it: the package `__init__` pulls in
`rocketpy.simulation.monte_carlo`, which I checked by listing `sys.modules`
afterwards. Removing the attribute from an installed SciPy gives the shape of
what an old one would produce:

```
ImportError: cannot import name 'bootstrap' from 'scipy.stats'
```

I have not installed SciPy 1.6 to see it for real, since it has no wheels for
the Python versions this package supports, which is rather the point.

**Expected behavior**

The floors describe what the code needs:

```
numpy>=1.21.3
scipy>=1.7.0
```

Both are old. SciPy 1.7.0 is from June 2021 and NumPy 1.21.3 from October 2021, and the package already requires Python 3.10, which is newer than either.

**Additional context**

Found while reviewing the requirements line in #1054. That pull request raises `numpy>=1.13` to `>=1.17`, since `SeedSequence` is a 1.17 feature, and I have noted the wider floor question there rather than widening the change. The SciPy floor is untouched by it and predates it.

[...]

PR #1108, opened 2026-08-08
MNT: declare dependency floors the package can actually run on

thc1006:
Addresses #1107. Not `Closes`, since the keyword only fires when a pull request targets the default branch and this targets `develop`.

## Pull request type

## Current behavior

```
numpy>=1.13
scipy>=1.0
```

Neither is true, and `pyproject.toml` reads this file through `dynamic = ["dependencies"]`, so it is what an installer is told.

**SciPy.** `rocketpy/simulation/monte_carlo.py` imports `scipy.stats.bootstrap` at module level, and the package `__init__` reaches that module. The SciPy 1.7.0 release notes are where the function arrives:

> `scipy.stats.bootstrap` has been added to allow estimation of the confidence interval and standard error of a statistic.

So 1.0 through 1.6 satisfy the floor and then fail on `import rocketpy`, before anything runs.

**NumPy.** The binding constraint here is a sibling rather than NumPy itself. `matplotlib>=3.9.0` is in this same file and requires `numpy>=1.23`, which I found the direct way:

```
ImportError: Matplotlib requires numpy>=1.23; you have 1.21.3
```

That is at 1.21.3, which is where NumPy's own Python 3.10 wheels begin, so the wheel-availability answer is not far enough either.

## New behavior

```
numpy>=1.23  # matplotlib 3.9 requires it
scipy>=1.8   # first to allow numpy 1.23; bootstrap needs 1.7
```

SciPy is 1.8 rather than 1.7 because the two have to compose: `scipy==1.7.2` declares `numpy<1.23.0`, so pinning 1.7 and asking for numpy 1.23 is unsatisfiable. 1.8.0 raised its cap to `<1.25.0`.

Both are old. SciPy 1.8.0 is February 2022 and NumPy 1.23.0 June 2022, against a package that already requires Python 3.10.

## Breaking change

- [x] No

It refuses installs that would have failed anyway, earlier and with a message about versions rather than an ImportError.

## How this was checked

Installed rather than reasoned about, since the constraint that decided the NumPy floor was not the one I expected:

```
python 3.10, numpy 1.23.0, scipy 1.8.0 pinned, everything else current

import rocketpy                                   works
tests/unit/simulation + tests/unit/stochastic     162 passed, 5 skipped
```

And resolved from the file itself, taking the lowest each declaration allows:

```
uv pip install --resolution=lowest-direct -r requirements.txt
-> numpy 1.23.0, scipy 1.8.0, matplotlib 3.9.0
```

## One thing I found and deliberately left alone

`requests`, `pytz`, `simplekml` and `dill` carry no floor at all, and under a lowest-version resolution they pick releases that cannot run on Python 3.10:

```
pytz 2018.3       from collections import Mapping
requests (old)    from collections import MutableMapping
```

`pytz>=2018.4` is the first that imports, if anyone wants the number.

I have not touched them, because that is a different defect. These two lines make a claim that is false; those four make no claim at all, and deciding what they should claim is a policy call rather than a correction. It is also not reachable the way the SciPy one is: an ordinary install takes the newest of an unfloored dependency, whereas a deliberate `scipy==1.6` is a thing someone might really have.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
-numpy>=1.13
-scipy>=1.0
+numpy>=1.23  # matplotlib 3.9 requires it
+scipy>=1.8  # first to allow numpy 1.23; bootstrap needs 1.7
 matplotlib>=3.9.0  # Released May 15th 2024
 netCDF4>=1.6.4
 requests


thc1006:
Heads up on a merge order point, found by merging my open PRs together onto `develop` rather than testing each on its own base.

This PR and #1054 both raise the numpy floor off the same `numpy>=1.13`, so they conflict textually in `requirements.txt`:

```
develop   numpy>=1.13
#1054     numpy>=1.17   SeedSequence, added in 1.17.0
#1108     numpy>=1.23   matplotlib 3.9 requires it
```

Whichever lands second hits the conflict. The resolution is to take `1.23`, which already covers what #1054 needs, so nothing is lost either way and no code has to change.

I have not rebased either branch, since resolving it now would only move the conflict to whichever of us merges first. Happy to rebase this one onto #1054 if you would rather have it ready to go.

On the combined tree, `develop` + #1054 + #1102 + #1103 + this, with the floor at 1.23: ruff clean, 2198 passed and 45 skipped.

PR #1112, opened 2026-08-08
CI: run the changelog job for pull requests from forks

### LINKED ISSUE #905: MNT: Automate CHANGELOG.md updates via GitHub Actions
Currently, the `CHANGELOG.md` file is manually updated by contributors in every Pull Request. This process has several downsides:
1.  **Human Error:** Contributors often use the wrong format, forget to add links, or place entries in the wrong section.
2.  **Merge Conflicts:** Since almost every PR touches the `[Unreleased]` section of the changelog, it is a frequent source of merge conflicts that require manual resolution.
3.  **Inconsistency:** The descriptions often vary in style (e.g., starting with "ENH", "BUG", or just plain text).

I would like to implement a GitHub Action (or a bot) that automatically updates the `CHANGELOG.md` file when a PR is merged.

The automation should:
1.  **Read PR Metadata:** Extract the PR title, number, and URL.
2.  **Categorize:** Use GitHub Labels (e.g., `Enhancement`, `Bug`, `Refactor`, `Documentation`) to decide which section of the Changelog the entry belongs to.
3.  **Format:** Automatically prepend the standard prefixes (`ENH`, `BUG`, `DOC`) and append the PR link `[#123](url)`.
4.  **Write:** Append the new entry to the `[Unreleased]` section of `CHANGELOG.md`.

**Proposed Logic for Categorization**
The automation should map GitHub Labels to "Keep a Changelog" sections:

* **Label:** `Enhancement` or `Feature` $\rightarrow$ **Section:** `### Added`
* **Label:** `Bug` $\rightarrow$ **Section:** `### Fixed`
* **Label:** `Refactor` or `Maintenance` $\rightarrow$ **Section:** `### Changed`
* **Label:** `Documentation` $\rightarrow$ **Section:** `### Added` or `### Changed` (depending on configuration)

**Potential Tools**
We should investigate tools such as:
* **Release Drafter:** Creates a draft release with the changes, which can then be pasted into the Changelog.
* **Git Cliff:** A highly customizable changelog generator.
* **Towncrier:** (Used by pandas/scipy) Uses fragment text files to generate the changelog at release time to avoid merge conflicts completely.

**Additional context**
The current format of `CHANGELOG.md` adheres to [Keep a Changelog](http://keepachangelog.com/). The automation must respect this format.

Example of a desired entry generation:
* **PR Title:** "Add new drag model"
* **PR Number:** 895
* **Label:** Enhancement
* **Result:** `- ENH: Add new drag model [#895](https://github.com/RocketPy-Team/RocketPy/pull/895)` under `### Added`.

### LINKED ISSUE #1101: Populate Changelog fails for every pull request from a fork

thc1006:
Replaces #1105, which targets `develop` and therefore cannot take effect. Addresses #1101.

`Populate Changelog` fails for every pull request opened from a fork, and only those:

```
2026-08-08T02:31  success  enh/add-pr-agent-gemini             branch on the repo
2026-08-08T02:22  failure  mnt/pytest-matrix-fail-fast         fork
2026-08-08T02:22  failure  bug/sampled-parachute-geometry      fork
2026-08-08T02:09  failure  bug/rail-takeoff-delayed-thrust     fork
2026-08-08T01:51  failure  ci/software-render-animation-tests  fork
2026-08-08T01:08  failure  doc/custom-sampler-seed             fork
```

The split is exactly fork versus not. GitHub withholds secrets from a `pull_request` run whose head is a fork, so `secrets.RELEASE_TOKEN` is empty and the checkout stops after a few seconds with `Input required and not supplied: token`.

The effect showed up today: #1102, #1103 and #1108 all merged without a changelog entry, and the lines had to be written by hand.

**Why this targets `master`.** A `pull_request_target` workflow is read from the repository's default branch, so the copy that decides whether the event fires is the one on `master`. #1105 changes `develop`'s copy, which means merging it changes nothing at all. I would rather close that one than leave it looking like a fix.

**On the trigger.** `pull_request_target` receives secrets because it runs in the base repository's context, which is also what makes it worth reading carefully. Nothing from the pull request is executed here:

- the checkout is `ref: develop`, not the pull request's head
- the updater is inline in the workflow, so under this trigger it comes from `master` rather than from whatever was just merged
- the title and labels reach Python through the environment, never interpolated into a shell

That last one was already true and already commented; I have not changed it.

**One thing beyond the trigger.** `permissions` drops from `contents: write` to `contents: read`. The job's writes go through `RELEASE_TOKEN`, which the checkout persists for the push step, so `GITHUB_TOKEN` never needed write. Under `pull_request` from a fork that grant was moot because the token was read-only anyway. Under this trigger it would be real, so it seemed the wrong moment to leave it.

Two things I did not do, and would rather you decide:

- pinning `actions/checkout@v4` to a full commit SHA, which GitHub recommends for privileged workflows. Nothing else in the repository pins that way, so it felt like a separate conversation.
- anything about `develop`'s copy of this file, which has since diverged considerably. Once this is on `master`, the two will want reconciling, and the Gemini path there raises its own questions that are not mine to settle.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,12 +1,21 @@
 name: Populate Changelog
 on:
-  pull_request:
+  # `pull_request_target`, not `pull_request`. Secrets are withheld from a
+  # `pull_request` run whose head is a fork, so RELEASE_TOKEN arrived empty and
+  # the checkout below failed for every outside contribution.
+  #
+  # Nothing from the pull request is executed. The checkout is `ref: develop`,
+  # and the updater is inline below, which under this trigger is read from the
+  # default branch rather than from the merged head.
+  pull_request_target:
     types: [closed]
     branches:
       - develop
 
+# Read, because the job's own writes go through RELEASE_TOKEN. This trigger runs
+# with secrets available, so GITHUB_TOKEN should not also carry write.
 permissions:
-  contents: write
+  contents: read
 
 jobs:
   Changelog:


thc1006:
`lint (3.10)` is red here and it is not this change. The failing step is `Ruff (format)`, and the file is `README.md`.

Reproduced against `master` with nothing applied:

```
$ git worktree add --detach /tmp/master-clean upstream/master
$ cd /tmp/master-clean && ruff format --check .
1 file would be reformatted, 297 files already formatted

$ ruff format --diff . | head -2
--- README.md
+++ README.md
```

This pull request touches one file, `.github/workflows/changelog.yml`, so the two do not overlap.

What changed underneath: the linters job installs ruff unpinned, and the run picked up `0.16.2`. Recent ruff formats Python code blocks inside Markdown, so the snippet in `README.md` now differs from how it was committed. Nothing in the repository changed; the tool's scope did.

Two ways to settle it, both outside this pull request:

- run `ruff format README.md` and commit the result, which is a one-line diff to a call in the quickstart snippet
- pin ruff in the linters workflow, the way `pip install ruff` does not today, so the formatter cannot widen under a green branch

I would rather not fold either into a workflow trigger change. Happy to send whichever you prefer as its own pull request.

thc1006:
One thing I should say plainly rather than leave implied: I cannot test the privileged path from outside the repository. A `pull_request_target` run only happens once this is on the default branch, so the first real exercise of it is the next fork pull request that merges into `develop`.

What I did check, since the reasoning has to stand on its own:

- `GITHUB_TOKEN` is not referenced anywhere in the file. The checkout takes `token: ${{ secrets.RELEASE_TOKEN }}`, persists it, and the `git push` at the end uses that. So narrowing `permissions` to `contents: read` cannot reach the push.
- The push path itself works today. The run on `enh/opening-shock-force` at 13:21 succeeded, which is the same steps with the same token from a branch on the repository.

If the narrowing turns out to be wrong in practice, the failure is loud and one line to revert, and it happens on a changelog commit rather than on anything a contributor is waiting for. I would rather that than turn the trigger on and leave a write grant the job does not use.

thc1006:
The `lint (3.10)` failure here clears once #1113 lands. They are the same README formatting, and this pull request cannot fix it without reaching outside what it is for.

Order, if both look right to you: #1113 first, then this one picks up a green `master`.

Worth noting #1113 shows only one check. Its `paths` filters mean a README-only change runs the docs build and nothing else, so the linters job that would have confirmed the fix does not run on the pull request that makes it. It runs here, on the next thing to touch `master`.

thc1006:
Thanks for retargeting and rebasing it. One thing I want to put in front of you rather than leave for the next fork pull request to discover, because it decides when this starts working rather than whether it does.

`pull_request_target` reads its workflow from the default branch:

> This event runs in the context of the default branch of the base repository, rather than in the context of the merge commit, as the `pull_request` event does.

`master` is the default branch here, and its copy still says:

```yaml
on:
  pull_request:
    types: [closed]
    branches:
      - develop
```

So while this sits on `develop`, the trigger that is registered is still `pull_request`, and a fork pull request merging into `develop` will fail exactly as it does today.

It does work eventually. `master` picks this up at the next release sync, the way `1f27358a` and `Rel/v1.13.0` did, and from that point the fork case is fixed. So the question is only whether the gap until then is acceptable.

If it is, nothing to do and I would just note it on #1101 so the next person reading that issue knows the fix is queued rather than missing.

If you would rather close it sooner, the same commit against `master` is what #1105 could not be, and I am happy to open it again. Your call either way, and I may be missing something about how the release sync is timed.

.github/workflows/changelog.yml
Gui-FernandesBR, line 1: does the "read" mode really make sense? We need write permission to push the changes to the branch, no?

ting-hong-shieh:
The `contents: read` permission is consistent with the credentials used by this workflow at head `f16b2c85c1d33bb36a6f413d7f13db56737bf529`.

`actions/checkout@v4` is configured with:

```yaml
token: ${{ secrets.RELEASE_TOKEN }}
```

Checkout persists that credential for later Git commands by default. The final `git push origin HEAD:develop` therefore uses `RELEASE_TOKEN`; the workflow does not reference `GITHUB_TOKEN`. The `permissions` block controls the automatically created `GITHUB_TOKEN`, so keeping it read-only avoids granting a second write credential that the job does not use.

The remaining limitation is the one already noted in the conversation: the real `pull_request_target` path cannot run until the default branch contains this workflow. This is a credential-flow review of the YAML, not a claim that the privileged path has already run.

Current checks on this head: eight passed. Current `develop`: `cb6106a717207dd8fc2dfe1446d80ff75022f21b`.

Gui-FernandesBR (approved):

PR #1113, opened 2026-08-08
Format the README snippets the way ruff now does

thc1006:
`Ruff (format)` fails on `master` today, and nothing in the repository changed to cause it.

```
$ git worktree add --detach /tmp/master-clean upstream/master
$ cd /tmp/master-clean && ruff format --check .
1 file would be reformatted, 297 files already formatted

$ ruff format --diff . | head -2
--- README.md
+++ README.md
```

The linters job installs ruff unpinned, and this run picked up `0.16.2`. Recent versions format Python code blocks inside Markdown, which they did not used to. The quickstart snippets predate that, so they differ from what the formatter now produces: single quotes in one call, a call split across three lines that fits on one, and two blocks indented two spaces rather than four.

Applied with `ruff format README.md` under 0.16.2, the version CI installed. Whitespace and quotes only. `git diff -w` leaves exactly two things, a quote style change and a rejoined call, and no snippet behaves differently when run.

**This will happen again.** The job takes whatever ruff is newest, so the next time the formatter widens its scope it lands the same way: a branch that was green goes red without a commit. Pinning ruff in `linters.yml` would be the fix, and it is a separate change from this one. Happy to send it if you want it.

Noticed while opening #1112, whose `lint (3.10)` is red for this reason and not for anything in that pull request.

### DIFF
--- a/rocketpy/stochastic/stochastic_model.py
+++ b/rocketpy/stochastic/stochastic_model.py
@@ -545,9 +545,13 @@ class StochasticModel:
         AssertionError
             If the input is not in a valid format.
         """
-        assert isinstance(sampler, CustomSampler), (
-            f"`{input_name}` must be a CustomSampler, not {type(sampler).__name__}"
-        )
+        # Raised rather than asserted, the same way #1103 handles it: `python -O`
+        # strips an assert, and the documented AssertionError is kept so callers
+        # that already catch it still do.
+        if not isinstance(sampler, CustomSampler):
+            raise AssertionError(
+                f"`{input_name}` must be a CustomSampler, not {type(sampler).__name__}"
+            )
         return sampler
 
     def _validate_airfoil(self, airfoil):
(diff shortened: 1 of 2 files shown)


Gui-FernandesBR (approved):

thc1006:
Worth recording why this one was a no-op on `develop`, since it explains itself once you see it: `develop`'s README is already formatted this way. The change only ever applied to `master`.

Checked just now against both branches with the ruff CI installs:

```
```

The diff between the two READMEs is exactly the change this pull request made, which is why retargeting it left nothing to apply.

So `master` still fails `Ruff (format)`, and any pull request into `master` shows a red `lint (3.10)` for a reason that has nothing to do with it. #1112 is where I ran into it.

It resolves itself at the next release sync, the same way `1f27358a` brought `develop` to `master` before. Until then it is a standing red on that branch. Happy to reopen this against `master` if you would rather not carry it, and equally happy to leave it if the next sync is close.

Separately, the reason it drifted at all is that the linters job installs ruff unpinned, and recent versions format Python code blocks inside Markdown. That will happen again on whatever the formatter widens to next. Pinning it is the durable fix and is its own change.

thc1006:
Reopened against `master` as #1114, since that is the branch the change applies to. It takes `develop`'s README rather than reformatting, so the two end up identical and the next release sync has nothing to reconcile.

PR #1115, opened 2026-08-08
CI: pin the linters, so a release cannot turn a branch red

thc1006:
`pip install pylint ruff` takes whatever is newest, and both tools decide whether this job passes.

This is not hypothetical. Recent ruff formats Python code blocks inside Markdown, which it did not use to, so `master`'s README stopped matching `ruff format --check` without anyone touching the file:

```
```

Any pull request into `master` now shows a red `lint (3.10)` for a reason unrelated to it. #1114 fixes the content; this stops the next one.

**pylint has the same shape and a wider blast radius.** `.pylintrc` uses `disable=` with named messages rather than enabling named ones, so a new release's new checks are on by default. And CI reads the exit code, not the score, so a single new convention message fails the job at 10.00/10.

Pinned to what CI installs today, verified against `develop` before pinning rather than after:

```
ruff 0.16.2     All checks passed!
                307 files already formatted
pylint 4.0.6    exit 0
```

**One thing I could not make true, so I am not claiming it.** My first draft of the comment said Renovate would raise the bump. It manages the dependencies declared in `pyproject.toml` and the requirements files, and I found no pull request from it for a pinned version inside a workflow, so I do not think its pip manager reads these. That leaves bumping them a manual step, or a Renovate rule that would need adding. I would rather say that plainly than pin behind a promise that does not hold.

If you would rather not carry the maintenance, `ruff~=0.16.0` narrows the window without closing it, though a patch release can still change formatting. Exact is what I would choose for a formatter, and it is what BalloonPoppingChallenge does for the same reason.

### DIFF
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -28,7 +28,12 @@ jobs:
         python -m pip install --upgrade pip
         pip install .[all]
         pip install .[tests]
-        pip install pylint ruff
+        # Pinned, because both decide whether this job passes and both change
+        # what they report between releases. Unpinned, the formatter widening
+        # its scope or pylint adding a check turns a branch red with no commit
+        # to explain it. Bumping these is a deliberate change, made when the
+        # tree is ready for what the new version says.
+        pip install "pylint==4.0.6" "ruff==0.16.2"
     - name: Ruff (lint)
       run: ruff check --output-format=github .
     - name: Ruff (format)


Gui-FernandesBR:
not ok as well, if a linter is updated, we should update our codebase accordingly

thc1006:
That is a better position than mine, and I want to say why rather than just agree.

I framed the red as the tool interfering: a branch goes red with no commit, so stop the tool moving. Your framing is that the red is the tool telling you something about the code, and pinning turns that off. The ruff case is actually an example of yours working rather than mine. What it reported was real, the fix was to update the README, and `develop` had already done it. Nothing was wrong except that `master` had not caught up.

The one thing that stays true is that the message arrives on whichever pull request happens to be open when the new release lands, which is how #1112 ended up carrying a red about a README it does not touch. That is a cost of keeping the tools current, not an argument against it, and it is yours to weigh.

Closing it here. Thanks for taking the time to say why rather than just declining.

PR #1118, opened 2026-08-10
Fix operating_temperature default init values across sensor classes

vapsik:
## Issue
Docstring states that the sensor default `operating_temperature` which is in unit K is assumed to be 25 C = 25+273.15 = 298.15 K. The initialization value, however, was set to 25 across sensor classes and never converted from C to K creating a contradiction.

## Summary
Set operating_temperature default init values from `operating_temperature=25` to `operating_temperature=298.15` across Sensor classes `sensor.py`, `gyroscope.py`, `barometer.py`, and `accelerometer.py`.


Gui-FernandesBR:
we never merge against master, the heading branch should be the develop branch

Gui-FernandesBR (approved):
LGTM

PR #1120, opened 2026-08-11
BUG: accept numpy integer parachute height triggers (#1106)

### LINKED ISSUE #1106: Parachute refuses a numpy integer trigger but accepts a numpy float

thatrandomasiandev:
## Pull request type

## Checklist

## Current behavior

`isinstance(trigger, (int, float))` rejects `numpy.int32` / `numpy.int64` height triggers while accepting `numpy.float64` (which subclasses `float`).

Fixes #1106

## New behavior

Numeric altitude triggers use `numbers.Real` excluding `bool`, so numpy integer/float scalars are accepted. Bools still raise `ValueError`.

## Breaking change

- [x] No

## Additional information

### DIFF
--- a/rocketpy/rocket/parachute.py
+++ b/rocketpy/rocket/parachute.py
@@ -1,4 +1,5 @@
 from inspect import Parameter, signature
+from numbers import Real
 
 import numpy as np
 
@@ -350,8 +351,8 @@ class Parachute:
             self.triggerfunc = _make_wrapper(trigger)
             return
 
-        # Numeric altitude trigger
-        if isinstance(trigger, (int, float)):
+        # Numeric altitude trigger (accept numpy integers/floats; reject bool)
+        if isinstance(trigger, Real) and not isinstance(trigger, bool):
             self._trigger_falling_only = True
 
             def triggerfunc(p, h, y, sensors, u_dot):  # pylint: disable=unused-argument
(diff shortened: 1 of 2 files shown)


Gui-FernandesBR:
duplicated #1116

PR #1121, opened 2026-08-11
BUG: evaluate parachute triggers once per time node (#1086)

### LINKED ISSUE #1086: Parachute triggers are evaluated twice per time node in Flight.__simulate()
Every parachute time node is evaluated twice during `Flight.__simulate()`: once by the inline loop over `node.parachutes`, and again immediately afterwards by `__check_and_handle_parachute_triggers()`, which loops over the same parachutes and repeats the same work.

On `develop` the two paths sit at the same level of the node loop with nothing between them:

- `flight.py:711`, the `for parachute in node.parachutes:` loop inside `__simulate()`
- `flight.py:776`, the call to `__check_and_handle_parachute_triggers(node, phase, phase_index, node_index)`
- `flight.py:920`, the helper itself, whose own loop over `node.parachutes` is at `flight.py:941`

Both are inside `for node_index, node in self.time_iterator(phase.time_nodes)` (line 691), so both run for every node.

### Reproducing it

Using the existing fixtures, with a trigger that only counts and never deploys:

```python
from collections import Counter


def test_trigger_is_evaluated_once_per_node(calisto_robust, example_plain_env):
    from rocketpy import Flight

    calls = []

    def counting_trigger(pressure, height, state):
        calls.append(round(float(height), 6))
        return False

    calisto_robust.parachutes.clear()
    calisto_robust.add_parachute(
        "probe", cd_s=10.0, trigger=counting_trigger, sampling_rate=100, lag=0.0
    )
    Flight(
        rocket=calisto_robust,
        environment=example_plain_env,
        rail_length=5.2,
        inclination=85,
        heading=0,
        time_overshoot=False,
    )

    counts = Counter(calls)
    assert len(calls) == len(counts)
```

That gives 10550 calls for 5270 distinct heights, a ratio of 2.00.

The duplication is also visible in a public attribute. Adding `noise=(0, 8.3, 0.5)` to the same parachute and inspecting it after the flight:

```
clean_pressure_signal: 10550 rows, 5275 distinct timestamps, every timestamp appearing exactly twice
```

### What it affects

- `clean_pressure_signal` and `noise_signal` get two entries per timestamp.
- The parachute noise RNG is drawn twice per node, so the two evaluations see different noisy pressures.
- A user trigger that keeps state or counts is called twice at the same simulation time.
- A trigger that returns `False` and then `True` for the same instant deploys earlier than it should.
- For a trigger with a probabilistic component, the effective chance per node is doubled.

### Possible fix

The inline block and the helper look like the same logic, so removing the inline loop and keeping only the helper call would leave one evaluation per node. I have not sent a patch because I am not sure whether the two were meant to diverge at some point, and the fix wants a regression test pinning one trigger call and one signal row per timestamp. Happy to put that together if the single-helper reading is the intended one.

Found while working on a fork of RocketPy; the behaviour reproduces on `develop` with nothing of ours involved.

thatrandomasiandev:
## Pull request type

## Checklist

## Current behavior

Each time node evaluated parachute triggers twice: an inline loop in `Flight.__simulate` and again in `__check_and_handle_parachute_triggers`.

Fixes #1086

## New behavior

The inline loop is removed; only `__check_and_handle_parachute_triggers` runs per node. Regression test asserts one call per sampled height.

## Breaking change

- [x] No

## Additional information


ting-hong-shieh:
I ran the same never-deploying 10 Hz trigger on current `develop` (`cb6106a717207dd8fc2dfe1446d80ff75022f21b`) and on this head (`d062c37f6c7f5d3e8037f4673744784183323d4b`). The flight inputs were identical.

| Measurement | `develop` | PR head |
| --- | ---: | ---: |
| trigger calls | 600 | 300 |
| unique rounded node heights | 300 | 300 |
| clean pressure rows | 600 | 300 |
| noise rows, including the initial row | 601 | 301 |

This confirms that the duplicate path was evaluating the trigger and appending pressure data twice at each node. The head keeps one evaluation and one pressure row per node.

The committed regression file also passes independently: `pytest tests/unit/test_parachute_triggers.py -q` → **4 passed**.

Environment: Python 3.12.6; NumPy 2.5.2; pytest 9.1.1; macOS 26.5.2 arm64.

Gui-FernandesBR (approved):

PR #1125, opened 2026-08-11
BUG: write MonteCarlo input and output rows atomically (#1110)

### LINKED ISSUE #1110: MonteCarlo writes the inputs and outputs of one simulation non-atomically
**Describe the bug**

A simulation's inputs and outputs go to two files in two separate appends, with nothing tying them together:

```python
with open(self.input_file, "a", encoding="utf-8") as f:
    f.write(inputs_json)
with open(self.output_file, "a", encoding="utf-8") as f:
    f.write(outputs_json)
```

If the second write fails, the first is already on disk. The run then has an inputs row with no matching outputs row, and the two files no longer line up by index.

Both run paths do this, `monte_carlo.py:292-295` in serial and `monte_carlo.py:434-437` in parallel, on `develop` at `235dc6e`.

**To Reproduce**

The shape on its own, with the second write refused:

```python
def record(inputs_json, outputs_json, fail_second=False):
    with open(inp, "a", encoding="utf-8") as f:
        f.write(inputs_json)
    if fail_second:
        raise OSError("no space left on device")
    with open(out, "a", encoding="utf-8") as f:
        f.write(outputs_json)

record('{"index": 0}\n', '{"index": 0}\n')
record('{"index": 1}\n', '{"index": 1}\n', fail_second=True)
```

```
inputs  2 lines
outputs 1 line
```

A disk filling up mid-run is the ordinary way to reach this. It does not need a crash.

**Expected behavior**

Either both rows land or neither does. A run that stops part way through should leave a checkpoint that `append=True` can continue from, rather than one that has to be trimmed by hand or thrown away.

**Additional context**

Found while reviewing #1054, and it predates that branch. Worth being precise about what does and does not change there, since the two are easy to confuse.

#1054 adds a completeness check that compares the two files after the run and refuses to report success when they disagree, so the damage is detected rather than silent. It does not make the pair atomic, and the checkpoint is still unusable afterwards. That branch is scoped to sampled-input reproducibility, so this was deliberately left out of it.

Two ways to close it, in increasing order of work:

- write to temporary files and `os.replace()` them into place once both are complete, which fits the `append=False` case;
- have workers send complete records to the parent and let one writer commit them, which also removes the shared lock the workers currently take around the pair.

The second is the one that also fixes a torn row from a worker killed mid-write, which the lock cannot prevent.

thatrandomasiandev:
## Pull request type

## Checklist

## Current behavior

MonteCarlo appends inputs then outputs in two separate `open(..., "a")` blocks (serial and parallel). If the second write fails, the inputs file keeps an orphan row and the files no longer line up by index.

Fixes #1110

## New behavior

`_append_simulation_record` records the inputs file size, writes inputs then outputs, and truncates inputs back on outputs failure so either both rows land or neither does. Serial and parallel paths both use the helper.

## Breaking change

- [x] No

## Additional information

Focused unit test mocks a failing outputs append and asserts the inputs file was rolled back. Full slow suite not run in this contribution pass.

### DIFF
--- a/rocketpy/simulation/monte_carlo.py
+++ b/rocketpy/simulation/monte_carlo.py
@@ -268,6 +268,39 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
         except OSError as error:
             raise OSError(f"Error creating files: {error}") from error
 
+    def _append_simulation_record(self, inputs_json, outputs_json):
+        """Append one simulation's inputs and outputs as a paired record.
+
+        Writes the inputs row first, then the outputs row. If the outputs write
+        fails, the inputs file is truncated back to its size before this call so
+        the two files do not drift out of alignment.
+
+        Parameters
+        ----------
+        inputs_json : str
+            Serialized inputs row, including its trailing newline.
+        outputs_json : str
+            Serialized outputs row, including its trailing newline.
+        """
+        input_path = self.input_file
+        output_path = self.output_file
+
+        try:
+            previous_input_size = os.path.getsize(input_path)
+        except OSError:
+            previous_input_size = 0
+
+        with open(input_path, "a", encoding="utf-8") as f:
+            f.write(inputs_json)
+
+        try:
+            with open(output_path, "a", encoding="utf-8") as f:
+                f.write(outputs_json)
+        except Exception:
+            with open(input_path, "rb+") as f:
+                f.truncate(previous_input_size)
+            raise
+
     def __run_in_serial(self):
         """
         Runs the monte carlo simulation in serial mode.
@@ -290,10 +323,7 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
                 inputs_json = self.__evaluate_flight_inputs(sim_monitor.count)
                 outputs_json = self.__evaluate_flight_outputs(flight, sim_monitor.count)
 
-                with open(self.input_file, "a", encoding="utf-8") as f:
-                    f.write(inputs_json)
-                with open(self.output_file, "a", encoding="utf-8") as f:
-                    f.write(outputs_json)
+                self._append_simulation_record(inputs_json, outputs_json)
 
                 sim_monitor.print_update_status()
 
@@ -432,10 +462,7 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
 
                         break
 
-                    with open(self.input_file, "a", encoding="utf-8") as f:
-                        f.write(inputs_json)
-                    with open(self.output_file, "a", encoding="utf-8") as f:
-                        f.write(outputs_json)
+                    self._append_simulation_record(inputs_json, outputs_json)
 
                     sim_monitor.print_update_status()
                 finally:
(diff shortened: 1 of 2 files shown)


Gui-FernandesBR:
what is the impact in the final simulation time?

thatrandomasiandev:
no effect on each flight's simulated t_final. Wall-clock time only adds one os.path.getsize per sim on the happy path. I didnt change the physics and the ODE, and the truncate only runs on write failure.

thatrandomasiandev:
Lint CI failure was **not** a Pylint config/fatal problem. Pylint exit code \`4\` is the **warning** bitflag (score can still print \`10.00/10\` with \`fail-under=10\`).

From the job log the PR-introduced warning was:

\`\`\`
tests/unit/simulation/test_monte_carlo.py:107:4: W1113: Keyword argument before variable positional arguments list in the definition of failing_output_open function (keyword-arg-before-vararg)
\`\`\`

Fixed by rewriting the mock as \`def failing_output_open(*args, **kwargs)\` so it no longer puts a keyword arg before \`*args\`. Pushed in a follow-up commit on this branch.

(The \`I0021\` useless-suppression notes on \`motor.py\` / \`sensor.py\` are pre-existing on develop and did not set this exit code.)

ting-hong-shieh:
The direct output-open failure covered by this PR passes on head `b9ad9b5587f3eaf193765335066d293aaaef8d2a`; the complete Monte Carlo unit file reports **36 passed**.

I also exercised writes that append four bytes and then raise, which exposes two remaining inconsistent states:

| Injected failure | Input file after the call | Output file after the call |
| --- | --- | --- |
| partial input write | original row plus `{"in` | original row |
| partial output write | original row | original row plus `{"in` |

The output-side case rolls the input file back but leaves the partially written output row. The input-side case is outside the current `try` block, so no rollback runs.

The smallest complete form I can see is to record both original sizes, wrap both appends in one exception boundary, and truncate both files back to those sizes after any open/write/flush failure. If rollback itself fails, preserving or chaining the original write exception would keep the initiating failure visible. Tests for partial input, partial output, and rollback failure would cover those paths.

I have not opened a replacement PR or changed this branch. Environment: Python 3.12.6; pytest 9.1.1; macOS 26.5.2 arm64. Current `develop`: `cb6106a717207dd8fc2dfe1446d80ff75022f21b`.

Gui-FernandesBR:
Answering my own question about simulation time: the added cost is one `os.path.getsize()` per simulation — a single `stat`, microseconds — against a flight simulation that is orders of magnitude slower, and it happens in a region that already opens and writes two files. Not measurable.

I also checked the thing that worried me more about a rollback-by-truncation approach: whether two parallel producers could interleave a `getsize` / `write` / `truncate` sequence and have one worker's rollback eat another worker's row. They cannot — in `__sim_producer` both writes are inside `mutex.acquire()` / `finally: mutex.release()`, so `_append_simulation_record` always runs with the lock held. Serial mode is trivially fine.

Good to merge.

Gui-FernandesBR (approved):

PR #1127, opened 2026-08-11
TST: harden flaky VTK flight animation tests (#1078)

### LINKED ISSUE #1078: Flaky `test_flight_animation_export_gif`: intermittent bus error in VTK off-screen rendering

thatrandomasiandev:
## Summary
- Mark the three PyVista/VTK flight animation tests with `flaky_vtk` and register the marker in `pyproject.toml` / `tests/conftest.py`.
- Drive CI isolation with `-m "not flaky_vtk"` / `-m flaky_vtk` instead of hard-coded `--deselect` paths; bump native-crash retries from 3 to 5.
- Soft-skip `test_flight_animation_export_gif` only on `OSError` when off-screen rendering is unavailable; assertion failures still fail the test.

`develop` already isolates these tests from the main integration job and retries SIGBUS/SIGSEGV (135/138/139). This PR makes that flake explicit via a marker and a slightly safer export path.

Fixes #1078

## Test plan
- [x] `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest tests/integration/test_plots.py -m flaky_vtk --collect-only` collects exactly the three animation tests
- [x] `-m "not flaky_vtk"` deselects those three and keeps the compare tests
- [ ] CI `Pytest` matrix: integration step skips `flaky_vtk`; VTK step retries up to 5 times on 135/138/139

### DIFF
--- a/.github/workflows/test_pytest.yaml
+++ b/.github/workflows/test_pytest.yaml
@@ -65,24 +65,21 @@ jobs:
           pip install numpy --upgrade
           pytest rocketpy --doctest-modules --cov=rocketpy --cov-append
 
+      # Isolate flaky_vtk (#1078): native VTK/OpenGL crashes (SIGBUS/SIGSEGV)
+      # kill the pytest process, so keep them out of the main integration job
+      # and retry them in the dedicated step below.
       - name: Run Integration Tests
         run: |
           pytest tests/integration \
-            --deselect tests/integration/test_plots.py::test_flight_animations_run_off_screen \
-            --deselect tests/integration/test_plots.py::test_flight_animations_render_all_scene_options \
-            --deselect tests/integration/test_plots.py::test_flight_animation_export_gif \
+            -m "not flaky_vtk" \
             --cov=rocketpy --cov-append
 
       - name: Run VTK animation tests
         run: |
-          tests=(
-            tests/integration/test_plots.py::test_flight_animations_run_off_screen
-            tests/integration/test_plots.py::test_flight_animations_render_all_scene_options
-            tests/integration/test_plots.py::test_flight_animation_export_gif
-          )
-          attempts=3
+          # flaky_vtk marker selects the three PyVista flight animation tests.
+          attempts=5
           for attempt in $(seq 1 "$attempts"); do
-            if pytest "${tests[@]}" --cov=rocketpy --cov-append; then
+            if pytest tests/integration -m flaky_vtk --cov=rocketpy --cov-append; then
               exit 0
             else
               status=$?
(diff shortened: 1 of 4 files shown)


Gui-FernandesBR:
Thanks for picking this up, @thatrandomasiandev — and sorry for the outcome, this one is timing rather than the work itself.

Closing as not needed: the crash #1078 describes is already fixed on `develop`, by two PRs that landed after the issue was opened.

- #1084 — `fail-fast: false` on the matrix (one platform no longer cancels the other five), Mesa software rendering on Linux (`LIBGL_ALWAYS_SOFTWARE` / `GALLIUM_DRIVER=llvmpipe`), and the three animation tests moved out of the integration job into a dedicated retrying step.
- #1100 — the retry now matches `135|138|139` instead of `138` alone, which was letting 8 of the 10 measured crashes straight through.

@thc1006, who reported it, confirmed on 2026-08-08 that macOS 3.14 has been green since, on a leg that had been crashing all week.

Two things in the diff also argue against merging it as it stands.

**The `try/except OSError` cannot address this flake, and it hides real failures.** SIGBUS/SIGSEGV kills the interpreter, so no Python `except` ever runs — it does nothing for #1078. What it does do is turn genuine errors into green skips: `_validate_animation_inputs` raises `FileNotFoundError` (a subclass of `OSError`) at `rocketpy/plots/flight_plots.py:179` when the 3D model file is missing, and that STL ships via `[tool.setuptools.package-data]` while CI runs `pip install .`. A packaging regression would skip quietly instead of failing the job — exactly what this integration test exists to catch. Same for `imageio`'s `open_gif` and PIL's `save` on a write error.

**The `--deselect` → marker move is only half applied.** `test-pytest-slow.yaml` still carries the same hard-coded list of nodeids, so the repo would be left running both conventions at once.

That said, the marker itself is a genuine improvement and stands on its own merits. I checked the assumption behind it: `--deselect` with a nodeid that does not exist is *silently* ignored — pytest collects everything, with no warning at all. So renaming any of those three tests today would quietly put the flake back into the main integration job. If you would like to send that part as a standalone maintenance PR, it would be welcome:

- the marker only, no change to the test body;
- applied to **both** `test_pytest.yaml` and `test-pytest-slow.yaml`;
- registered in one place, not in both `pyproject.toml` and `tests/conftest.py` (the latter already registers `slow`, so that entry would be duplicated);
- `attempts` left at 3 — there is no evidence 3 was insufficient, and the data points the other way since #1084;
- `pytest tests/integration/test_plots.py -m flaky_vtk` rather than collecting the whole integration package to run three tests.

Framed as CI maintenance, not as a fix for #1078. Thanks again for the time you put into it.

PR #1130, opened 2026-08-11
ENH: plot flight center-of-pressure evolution (#954)

### LINKED ISSUE #954: ENH: Flight CP evolution plot
Through the Discord server, a member requested a plot of the CP position through time in the flight class

thatrandomasiandev:
## Summary
- Add `Flight.plots.center_of_pressure()` to plot CP position vs time, evaluated at the flight Mach number at each step
- Overlay center of mass on the same axis and static margin on a twin axis, with out-of-rail / burnout / apogee markers
- Wire the new plot into `Flight.plots.all()` and cover show/save paths in unit tests

Fixes #954

## Test plan
- [x] `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 MPLBACKEND=Agg python -m pytest tests/unit/test_plots.py::test_flight_center_of_pressure_plot -q`
- [ ] Manual: `flight.plots.center_of_pressure()` on a Calisto flight shows CP/CM/static-margin curves through first event time


Gui-FernandesBR:
we need to see one or two examples of the new plot before merging it please

thatrandomasiandev:
@Gui-FernandesBR here are two examples from this branch:

1. Full-ish ascent (rail → burnout → apogee window), truncated at `first_event_time` like other stability plots:
cp evolution

2. `terminate_on_apogee=True`:
cp to apogee

Both generated with `flight.plots.center_of_pressure(filename=...)` under `MPLBACKEND=Agg`. Happy to tweak axis limits / twin-axis layout if you want something closer to the existing stability plots.

Gui-FernandesBR:
the example plots don't really make sense

Gui-FernandesBR:
Following up on my "the example plots don't really make sense" comment — I found why, and it is not the plotting code.

`static_margin` is only ever discretized over the burn:

```python
self.static_margin.set_discrete(lower=0, upper=self.motor.burn_out_time, samples=200)
```

With `extrapolation="constant"`, every query past `burn_out_time` returns the value frozen at the edge of that range. So for the whole coast, apogee and descent the curve is not a computed static margin at all — it is one number repeated, which is exactly the flat/implausible shape in the plots you posted.

For a flight-long quantity, use `Flight.stability_margin`, which is a function of Mach *and* time and is evaluated over the whole flight. If you really want the static margin specifically, it has to be re-discretized over the flight's own time span before plotting.

Two other notes:

- The workflow runs on this PR had been stuck in `action_required` and never actually executed, so this PR has never had real CI. I approved them, so you will get results now.
- Once the curve is right, please post the regenerated plots again so we can compare.

Gui-FernandesBR (commented):
no need to commit docs/static/pr_examples/cp_evolution_calisto_like.png file

Gui-FernandesBR (approved):

PR #1131, opened 2026-08-11
DOC: add SIL parachute ejection integration example (#524)

### LINKED ISSUE #524: DOC: Software in a Loop (SIL) example with parachute ejection systems

Due to its callbacks functions, RocketPy's `Flight` class allows to easily integrate with custom codes that are often used to detect the correct instant to eject the parachutes. 

- Create an example on our documentation showing how to integrate the ejection detection algorithm to the Flight class. 
- This is an useful resource that could be used as an inspiration: https://github.com/RocketPy-Team/RocketPaper/tree/main/Valetudo_Monte_Carlo

**Additional context**

Extra bonus if you at least mention how to create binary

thatrandomasiandev:
## Summary
- Add `docs/user/sil_parachute_ejection.rst` showing Software-in-the-Loop parachute ejection detection wired through `Flight` / parachute trigger callables.
- Link the page from the user guide toctree and from the parachute triggers "See Also" section.
- Include a simplified stateful barometric detector, dual-deploy attachment pattern, notes on wrapping compiled recovery firmware, and a cautious HIL mention; points to the [Valetudo Monte Carlo](https://github.com/RocketPy-Team/RocketPaper/tree/main/Valetudo_Monte_Carlo) inspiration.

Fixes #524

## Test plan
- [ ] Confirm toctree entry renders under Special Case Simulations
- [ ] Skim `sil_parachute_ejection.rst` for Sphinx link resolution (`:doc:`, `:ref:`, `:class:`)
- [ ] Optional: `cd docs && make html` if the docs env is available

### DIFF
--- a/docs/user/index.rst
+++ b/docs/user/index.rst
@@ -26,6 +26,7 @@ RocketPy's User Guide
    Compare Flights Class<compare_flights.rst>
    Flight Comparator Class <flight_comparator.rst>
    Parachute Triggers (Acceleration-Based) <parachute_triggers.rst>
+   Software-in-the-Loop Parachute Ejection <sil_parachute_ejection.rst>
    Deployable Payload <deployable.rst>
    Controllers <controllers.rst>
    Air Brakes Example <airbrakes.rst>
--- a/docs/user/parachute_triggers.rst
+++ b/docs/user/parachute_triggers.rst
@@ -274,6 +274,7 @@ a lower altitude:
 See Also
 --------
 
+- :doc:`Software-in-the-Loop Parachute Ejection <sil_parachute_ejection>`
 - :doc:`Parachute Class Reference </reference/classes/Parachute>`
 - :doc:`Flight Simulation </user/flight>`
 - :doc:`Sensors </notebooks/sensors>`
--- a/docs/user/sil_parachute_ejection.rst
+++ b/docs/user/sil_parachute_ejection.rst
@@ -1,0 +1,245 @@
+.. _sil-parachute-ejection:
+
+Software-in-the-Loop Parachute Ejection
+=======================================
+
+RocketPy can run a **Software-in-the-Loop (SIL)** recovery system by calling
+your ejection-detection algorithm from the parachute trigger (and, when useful,
+from a discrete :doc:`controller <controllers>`). During ``Flight``, RocketPy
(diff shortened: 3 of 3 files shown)


Gui-FernandesBR (approved):

Gui-FernandesBR:
Two things and then this can go in.

**1. The `build-docs` failure is not yours.** It was an invalid RST sublist in `StochasticModel.dict_generator`'s docstring that reached `develop` through another PR:

```
rocketpy/stochastic/stochastic_model.py:docstring of ...dict_generator:13: ERROR: Unexpected indentation. [docutils]
```

I fixed that directly on `develop`. Merge `develop` in and the docs job goes green — nothing to change on your side.

**2. The "Dual path: trigger vs controller callback" section uses a Markdown table inside an `.rst` file.** RST has no pipe-table syntax: a leading `|` is parsed as a *line block*, so this renders as a column of ragged lines rather than a table. Sphinx does not error on it, which is why CI never caught it — but the published page would be wrong. Please convert it to an RST `list-table` (or a simple table), e.g.

```rst
.. list-table::
   :header-rows: 1

   * - Approach
     - When it fires
   * - ``trigger``
     - ...
   * - controller callback
     - ...
```

Gui-FernandesBR:
I pushed the table fix to your branch myself (`67c262f9`) so this doesn't have to wait on another round trip — your rebase had already cleared the `build-docs` failure, and the table was the only thing left.

What changed: the "Dual path: trigger vs controller callback" section is now a `list-table`, matching the two-column explanatory tables already in `docs/user/flight.rst` (`:header-rows: 1`, `:widths: 30 70`). Nothing else touched — one file, +15/-5.

For the record, since Sphinx never warned about this and it would have shipped silently, I built the page both ways with Sphinx 8.1.3 under `-W --keep-going`:

```
before:  0 <table>,  1 line-block   -> pipes and the "---------" row rendered as ragged text
after :  1 <table class="docutils"> with a real header row
```

No new warnings either way, which is exactly why CI stayed green on it.

I also scanned the rest of the 245-line file for other Markdown leakage — no fenced code blocks, no `[text](url)` links, no `#` headings, no other pipe tables. The one table was the only thing.

`Documentation` is green on the new head and this is still approved and mergeable, so it's ready to go. Thanks for the write-up — it's a genuinely useful page.

PR #1132, opened 2026-08-11
ENH: add Folium interactive flight trajectory map (#963)

### LINKED ISSUE #963: Add Folium support for interactive trajectory maps
**Description:**
I'd like to add Folium as an optional visualization tool for Flight objects.

**What problem does this solve?**
Currently, trajectory visualization is limited. Folium would allow users to see launch and landing sites on interactive real-world maps.

**Proposed solution:**
Add a method `Flight.plot_trajectory_on_map()` that:
- Shows the full trajectory path
- Marks launch and landing sites
- Creates an interactive HTML map

**Why Folium?**
- Lightweight and easy to use
- Works directly in Jupyter/Colab
- Open-source and well-maintained

**Checklist before submitting:**
- [x] I have searched for existing issues about this
- [x] I am willing to implement this feature

thatrandomasiandev:
## Summary
- Add `Flight.plots.trajectory_on_map()` using optional Folium to render an interactive HTML map of the ground track with launch and landing markers.
- Introduce a `maps` optional dependency extra (`folium`) and include it in `rocketpy[all]`.
- Add unit tests that mock Folium when absent and exercise real HTML export when Folium is installed.

Fixes #963

## Test plan
- [x] `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest tests/unit/test_flight_trajectory_map.py`
- [ ] Smoke: `flight.plots.trajectory_on_map(filename="traj.html")` opens in a browser with path + markers

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,7 +77,16 @@ animation = [
     "imageio-ffmpeg>=0.5"
 ]
 
-all = ["rocketpy[env-analysis]", "rocketpy[monte-carlo]", "rocketpy[animation]"]
+maps = [
+    "folium>=0.14",
+]
+
+all = [
+    "rocketpy[env-analysis]",
+    "rocketpy[monte-carlo]",
+    "rocketpy[animation]",
+    "rocketpy[maps]",
+]
 
 
 [tool.coverage.report]
--- a/rocketpy/plots/flight_plots.py
+++ b/rocketpy/plots/flight_plots.py
@@ -144,6 +144,76 @@ class _FlightPlots:
         ax1.set_box_aspect(None, zoom=0.95)  # 95% for label adjustment
         show_or_save_plot(filename)
 
+    def trajectory_on_map(self, *, filename=None):
+        """Create an interactive Folium map of the flight trajectory.
+
+        Draws the ground-track path from ``flight.latitude`` /
+        ``flight.longitude`` and marks the launch and landing sites.
+        Requires the optional ``folium`` dependency
+        (``pip install folium`` or ``pip install rocketpy[maps]``).
+
+        Parameters
+        ----------
+        filename : str | None, optional
+            Path to save the map as an HTML file. If None, the map is not
+            written to disk. Default is None.
+
+        Returns
+        -------
+        folium.Map
+            The interactive map object. In Jupyter, displaying the return
+            value renders the map.
+        """
+        folium = import_optional_dependency("folium")
+        flight = self.flight
+
+        latitudes = np.asarray(flight.latitude[:, 1], dtype=float)
+        longitudes = np.asarray(flight.longitude[:, 1], dtype=float)
+        path = list(zip(latitudes.tolist(), longitudes.tolist()))
+        if not path:
+            raise ValueError("Flight has no latitude/longitude samples to plot.")
+
+        launch = path[0]
+        landing = path[-1]
+        center = [
+            float(0.5 * (launch[0] + landing[0])),
+            float(0.5 * (launch[1] + landing[1])),
+        ]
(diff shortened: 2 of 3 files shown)


Gui-FernandesBR:
There are many other ways of implementing this one. We need first to reflect on that.

Gui-FernandesBR:
Hi @thatrandomasiandev — thanks for the patience on this one, and sorry for the long silence after my "there are many other ways of implementing this" comment. That was a vague thing to say without following up, so let me make it concrete.

Back in #963 I mentioned we had internal tooling that already does Folium maps. That's a module we use to produce the impact maps for EuRoC, and rather than keep asking you to guess at our conventions, **I've pushed the port directly onto your branch** (fast-forward, nothing of yours rewritten). Please pull before you continue.

### What I changed

Your structure was right — the optional-dependency handling, the `maps` extra, and the `plots.trajectory_on_map()` placement all match how the codebase does things. What it was missing was the rocketry-specific part:

- **Satellite imagery.** Plain OpenStreetMap can't answer the question you actually ask a map of a rocket flight — what's the terrain, where are the tree lines, is there water. The map now ships OpenStreetMap *and* Esri World Imagery behind a `LayerControl`, with satellite active by default.
- **Apogee marker**, labelled with apogee AGL, between launch and landing. Skipped when apogee was never detected, since `Flight.apogee_time` keeps its initial value of `0` and the marker would otherwise land on the pad.
- **`safety_radii`** — range safety circles around the pad, in their own toggleable layer. This is the feature our internal tooling exists for.
- **`time_step`**, for parity with `Flight.export_kml`: resample the ground track instead of drawing every integration step.
- **`color`** and **`title`** (the title is HTML-escaped before injection).

Ground track over satellite imagery

Launch (green), apogee ground position (blue), landing (red).

Range safety circles

`safety_radii=[2500, 5000]`. Two bugs turned up only by looking at this render: the fitted viewport ignored the circles, so asking for a 5 km ring gave you a screen with no ring on it; and the launch/landing pins were clipped by the map border because the bounding box has no padding. Both fixed.

### Other gaps closed

- **Docs.** This was the real blocker. The `maps` extra was in `pyproject.toml` but documented nowhere — `installation.rst` and `flight.rst` both have sections for the `animation` extra and had none for this. Both now do, including a parameter table and a cross-reference to `export_kml` for the 3D case a 2D map can't cover.
- **Branch was 40 commits behind `develop`.** Merged, no conflicts.
- **Tests: 4 → 10**, covering the tile layers, the apogee marker and its skip path, the safety circles and their framing, title escaping, and `time_step` resampling.

### On the red `build-docs (3.12)`

That check is stale, not a real failure — the current `docs.yml` only triggers for PRs based on `master`, and this one targets `develop`. The logs have expired (HTTP 410) so I couldn't read the original. I ran the docs build locally anyway, exactly as CI does, since this reaches `master` eventually.

### Verification

`sphinx-build -W --keep-going` exits 0 with zero warnings, 10/10 unit tests pass including the real-Folium HTML export, `ruff` clean, `pylint` 10.00/10. The maps above are the actual rendered output, checked in a browser.

### Still open, and worth your call

The bigger prize from the internal module is the **Monte Carlo impact map** — numbered markers coloured by outcome (nominal / ballistic / main-at-apogee), a dynamic legend, and a distance grid. That belongs on `MonteCarlo.plots`, not on `Flight`, so it should be a separate issue rather than growing this PR. Happy to open it if you'd like to take it too.

From my side this is ready to come out of draft. Let me know if you disagree with any of the choices above — particularly satellite-as-default-layer, which is a judgement call.

Gui-FernandesBR (approved):

PR #1137, opened 2026-08-11
ENH: cache downloaded atmosphere netCDF datasets (#654)

### LINKED ISSUE #654: ENH: Cache netcdf4 datasets to avoid multiple downloads
When setting na atmospheric model to "Forecast" or "Ensemble", rocketpy will usually download data from servers on the internet. Currently, by executing your code multiple times, the code would be downloading the same data again and again. We need to create a system to cache the datasets that we download from the internet and re-use it if the atmospheric model is reset.

If the dataset is in the netcdf4 format, we want to save it on a .nc format. But some sources (like windy) may provide .json files instead.

Additionally, I would like to see a "no_cache" option to force the code to download new data again if wanted.

thatrandomasiandev:
## Summary
- Cache Forecast/Ensemble (and Reanalysis remote URLs) location-time profiles as compact `.nc` files under `~/.rocketpy_cache/atmosphere` (override with `ROCKETPY_CACHE`).
- Cache Windy JSON responses on the same path; second identical `set_atmospheric_model` calls reuse disk.
- Add `no_cache=False` to `Environment.set_atmospheric_model` (and Windy processing) so callers can force a fresh download without breaking existing callers.

## Notes / TODOs
OPeNDAP "Best" aggregations are virtual catalogs, not downloadable files, so this PR caches the **extracted profiles** RocketPy already materializes after the first fetch (same end-user outcome: reruns skip the network). Remaining follow-ups:
- Open-Meteo / Meteomatics JSON response caching
- Optional full-file download cache when a remote source is a concrete HTTP `.nc` (non-OPeNDAP)

Fixes #654

## Test plan
- [x] `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest tests/unit/environment/test_atmosphere_cache.py`
- [ ] Confirm Forecast shortcut reuses disk on rerun with same lat/lon/date
- [ ] Confirm `no_cache=True` forces re-download


Gui-FernandesBR:
Not ready to merge yet. Three items, the third being the one I care most about.

**1. Three of this PR's own new tests fail on all six platforms.**

```
FAILED tests/unit/environment/test_environment.py::test_set_atmospheric_model_normalizes_shortcut_case_for_forecast
FAILED tests/unit/environment/test_environment.py::test_forecast_shortcut_and_dictionary_are_case_insensitive[AIGFS]
FAILED tests/unit/environment/test_environment.py::test_forecast_shortcut_and_dictionary_are_case_insensitive[HRRR]

IndexError: too many indices for array: array is 0-dimensional, but 2 were indexed
  rocketpy/environment/environment.py:1637
```

**2. pylint fails (exit 8):**

```
rocketpy/environment/atmosphere_cache.py:158:0: R0915: Too many statements (30/25)
rocketpy/environment/atmosphere_cache.py:270:0: R0915: Too many statements (35/25)
rocketpy/environment/environment.py:1237:4: R0912: Too many branches (28/25)
```

Please split those into smaller functions rather than adding `# pylint: disable`.

**3. Cache invalidation — the design point I want resolved before this lands.**

The cache is enabled by default, and the key contains neither a TTL nor the forecast model run/cycle. So a user who runs a GFS forecast today and again tomorrow silently gets yesterday's forecast, with no indication anything is stale. For forecast products that is a wrong-answer bug, not a stale-file annoyance, and it is the kind of thing that is very hard for a user to notice.

What I would want:

- the model run / cycle timestamp as part of the cache key, or a TTL, so a new cycle can never be served from an old file;
- caching **off** by default until we have confidence in the invalidation;
- and it is worth distinguishing product classes: reanalysis/archive data (ERA5 and friends) is immutable and can be cached indefinitely, forecasts cannot.

Happy to revisit once those are addressed.

Gui-FernandesBR (approved):

PR #1138, opened 2026-08-11
ENH: model unbonded solid-motor grain CM shift (#340)

### LINKED ISSUE #340: ENH: Support for solid motor designs without glued/bonded grains
**Description:**

RocketPy currently assumes that the solid rocket motor design follows the BATES grain configuration, where the ends of each individual grain burn simultaneously as the core burns radially. In this configuration, assuming the grains positions remain fixed, the center of mass of the propellant does not change since the center of mass of each grain remains fixed.

However, there are many solid rocket motor designs that do not glue or bond the grains to the outer liner or casing. During flight, these unglued grains are pushed and shift their position towards the nozzle due to inertial forces, causing the center of mass position of the total propellant to change. This behavior can affect the stability of the rocket and is not currently contemplated in RocketPy.

To address this issue, I suggest adding an option inside the SolidMotor class for motor designs without glued/bonded grains. This would involve modeling the movement of unglued grains during flight and updating the center of mass position of the total propellant accordingly.

**Additional context**

Some research on the matter led me to:

* There appears to be not much of an agreement between the instructions of the usual commercial APCP motors (from companies such as Aerotech, Cesaroni, Loki) regarding this matter. By reading a sample of different assembly instruction manuals, it seems that that the larger and more recent COTS designs are more likely to include a step to glue the grains to the liner (usually with epoxy or some other strong adhesive) - but many do not, and even mention that it is normal for the grains to "rattle" around inside the casing.
* By asking in a group of amateurs who work mostly with rocket candy (KNSU/KNSB/KNDX) motors, there appears to not be much of a consensus either. Heard reports of teams which bond the grains with stuff such as RTV silicone, while others consider a 'snug fit' inside the casing to be good enough. In these cases, the motivation for bonding the grains seems more related to avoid spitting out casting tubes/inhibitors during the burn, rather than stability considerations.

Overall, I believe that this is an important new feature to be considered. It will enable RocketPy to provide more accurate simulations for a seemingly wide range of rockets that utilize freestanding grain motor designs.

thatrandomasiandev:
## Summary
- Adds opt-in `SolidMotor(grains_bonded=True)` (default preserves current fixed-grain / BATES behavior).
- When `grains_bonded=False`, applies a first-order inertial packing model: grains pack against the aft (nozzle-side) face of the initial grain stack, so propellant CM shifts toward the nozzle as `grain_height` regresses: `CM(t) = grains_center_of_mass_position - _csys * (n/2) * (h0 - h(t))`. Parallel-axis `propellant_I_11` uses the packed pitch.
- Documents that this is **not** a DEM / rigid-body grain dynamics model (no friction, rattling, or acceleration threshold). Inter-grain `grain_separation` (spacers) is retained; only grain-height loss shortens the stack. With `only_radial_burn=True`, height is fixed so CM does not shift.

Fixes #340

## Follow-ups (out of scope)
- Acceleration-dependent settling / ignition transient before packing completes
- Full multi-body / DEM grain contact model
- Optional collapse of soft spacers (`grain_separation → 0` when packing)

## Test plan
- [x] `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest tests/unit/motors/test_solidmotor.py`
- [x] Bonded path matches prior fixed CM
- [x] Unbonded multi-grain CM moves aft vs bonded mid-burn; ignition CM matches
- [x] `grains_bonded` round-trips through `to_dict` / `from_dict`

### DIFF
--- a/rocketpy/motors/solid_motor.py
+++ b/rocketpy/motors/solid_motor.py
@@ -543,6 +543,7 @@ class SolidMotor(Motor):
             return (pitch**2) * sum_sq_index
         pitch = self.grain_height + self.grain_separation
         return (pitch**2) * sum_sq_index
+
     # pylint: disable=too-many-statements
     def evaluate_geometry(self):
         """Calculates grain inner radius and grain height as a function of time
@@ -790,7 +791,10 @@ class SolidMotor(Motor):
 
         # Parallel-axis term from grain COM offsets about the propellant COM.
         # Bonded: fixed initial pitch. Unbonded: packed pitch tracks grain_height.
-        I_11 = grain_number * grain_inertia11 + grain_mass * self._grain_pitch_squared_sum()
+        I_11 = (
+            grain_number * grain_inertia11
+            + grain_mass * self._grain_pitch_squared_sum()
+        )
 
         return I_11
 
--- a/rocketpy/motors/solid_motor.py
+++ b/rocketpy/motors/solid_motor.py
@@ -196,6 +196,10 @@ class SolidMotor(Motor):
     SolidMotor.only_radial_burn : bool
         If True, grain regression is restricted to radial burn only (inner radius growth).
         Grain length remains constant throughout the burn. Default is False.
+    SolidMotor.grains_bonded : bool
+        If True (default), grain axial positions stay fixed at the assembled layout
+        (BATES bonded/glued grains). If False, a first-order packing model shifts the
+        propellant center of mass toward the nozzle as grain height regresses.
     """
 
     # pylint: disable=too-many-arguments
@@ -221,6 +225,7 @@ class SolidMotor(Motor):
         coordinate_system_orientation="nozzle_to_combustion_chamber",
         reference_pressure=None,
         only_radial_burn=False,
+        grains_bonded=True,
     ):
         """Initialize Motor class, process thrust curve and geometrical
         parameters and store results.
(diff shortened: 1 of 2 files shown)


Gui-FernandesBR:
I don't believe the solution could be that simple... But we can give it a try

Gui-FernandesBR:
Following up on my "I don't believe the solution could be that simple" comment — I worked through it and it does hold up. Recording the check so it is on the record:

**The CM expression is exactly derivable.** With grains packed against the fixed aft face and instantaneous pitch `p = h + s`, grain *k*'s centre sits at `z_k = z_aft + h/2 + k*(h + s)`. Averaging over `k = 0..N-1`:

```
mean(z_k)          = z_aft + h/2  + (h  + s)*(N-1)/2
mean(z_k) bonded   = z_aft + h0/2 + (h0 + s)*(N-1)/2
difference         = (h - h0)/2 + (h - h0)*(N-1)/2 = -(N/2) * (h0 - h)
```

which is precisely the implemented `grains_center_of_mass_position - _csys * (grain_number/2) * (grain_initial_height - grain_height)`. The instantaneous pitch in the parallel-axis term is the right companion to that, and `propellant_I_22` returns `self.propellant_I_11`, so the transverse components stay consistent for free.

**The default path is unchanged.** For `grains_bonded=True`, `_grain_pitch_squared_sum()` returns `(h0 + s)**2 * sum(offsets**2)`, which is algebraically the same as the old `np.sum(d**2)` with `d = linspace(-iv, iv, n) * (h0 + s)`. So existing motors get bit-identical numbers.

The remaining caveats are modelling choices rather than errors, and the docstring already states them: `grain_separation` is retained while the stack shortens, and there is no acceleration-dependent settling — aft packing is what happens under boost, and after burnout `grain_height` stops regressing so the CM freezes anyway, which lands on the right answer for the wrong reason. Fine as a documented first-order model.

Gui-FernandesBR (approved):

PR #1156, opened 2026-08-14
TST: add Defiance example flight acceptance test

### LINKED ISSUE #1057: Proposal: Migrate the 6-DOF Simulation Core to Rust for Next-Level Performance
# Proposal: Migrate the 6-DOF Simulation Core to Rust for Next-Level Performance

## Context and Motivation
Recent PRs (like the excellent #1049) have drastically optimized our `Function` class, but as noted in the benchmarks, the overall simulation speedup is fundamentally bounded by the architecture. 

If we profile a typical RocketPy simulation, the major bottleneck is the "boundary crossing" between Python and the underlying ODE solver (like `scipy`'s LSODA). Every time the solver takes a step, it must acquire the Python GIL, convert arrays, and call our Python-based `u_dot_generalized` (and aerodynamic lookups) thousands of times per flight second. 

To break this barrier and achieve orders-of-magnitude speedups, we need to remove the Python interpreter from the hot path entirely.

## Proposal: The "Shell and Engine" Architecture
I propose we explore migrating the core mathematical and physical models to **Rust**, while keeping the user-friendly Python API exactly as it is. 

We would adopt a hybrid architecture:
1. **The Python Shell (`rocketpy`)**: Continues handling the user API, object parsing, file reading, `matplotlib` plotting, and printing. 
2. **The Rust Engine (`rocketpy-core` / `_rocketpy_rs`)**: A pure-math backend handling the 6-DOF equations of motion, rotation matrices, aerodynamic interpolations, and the numerical integration loop. 

When a user calls `Flight(...)`, the Python shell extracts the raw numbers/arrays from the Python objects and hands them to a single Rust function. Rust takes over, runs the entire flight integration instantly (without the GIL), and returns the final trajectory array back to Python for plotting.

## Why Rust? (vs. C++ or Cython)
The Python ecosystem is undergoing a massive shift towards Rust. Major projects like **Pydantic (V2)**, **Ruff**, **Polars**, and **Cryptography** have all recently migrated their cores to Rust. The advantages are clear:
- **Zero-Setup Cross-Platform Builds**: Dealing with MSVC/GCC/Clang compilation across Windows/Linux/macOS in C++ is famously painful. Rust’s `cargo` and the Python tool [Maturin](https://github.com/PyO3/maturin) make cross-platform wheel building in GitHub Actions almost effortless. 
- **The `pip install` Experience**: Users will simply run `pip install rocketpy`. They will download pre-compiled `.whl` binaries and will **not** need any Rust compiler installed on their machines.
- **Memory Safety**: If it compiles in Rust, it is virtually immune to the *Segmentation Faults* and memory leaks that haunt C/Cython extensions.
- **PyO3**: The PyO3 crate provides an incredibly elegant and safe way to write Python bindings.

## A Pragmatic Approach to Numerical Integrators (Keeping LSODA)
One valid concern is losing SciPy's robust LSODA solver and its automatic stiff/non-stiff adaptive time-stepping. Rewriting LSODA from scratch in pure Rust would be an enormous, error-prone academic exercise. 

Instead, the most pragmatic engineering approach is to **migrate our equations of motion to Rust, and wrap an existing, battle-tested C-implementation of LSODA using Rust's FFI.**
By using Rust's `build.rs` and the `cc` crate, we can automatically compile a C-version of LSODA alongside our Rust code during the wheel-building process. This means our pure-Rust physics engine can call the C-based LSODA solver in nanoseconds without ever touching the Python GIL. We get the safety and speed of Rust, the maturity of LSODA, and the user still just runs `pip install rocketpy` seamlessly.

## Massive Impact on Monte Carlo (Dispersion)
Currently, our Monte Carlo analysis relies on Python's `multiprocessing`, which incurs heavy overhead (process spawning, memory duplication, pickle serialization).

By migrating to Rust, we unlock **true multithreading**. We could eventually pass a list of 10,000 flight parameters to Rust, and using crates like `Rayon`, run the simulations concurrently across all CPU cores with zero Python overhead. This would make dispersion analysis nearly instantaneous.

## Proposed Phased Approach
We shouldn't rewrite everything at once. A safe migration path could be:

1. **Phase 1 (Proof of Concept)**: Set up a `maturin` project workspace. Write a simple 2D ballistic trajectory integrator in Rust, import it into Python, and benchmark it against our current SciPy implementation to validate the speedup.
2. **Phase 2 (Core Migration)**: Migrate `u_dot_generalized` and aerodynamic lookups to Rust. Bind it to a wrapped C-based LSODA solver via FFI. 
3. **Phase 3 (Monte Carlo)**: Expose a vectorized/multithreaded entry point in Rust specifically for dispersion analysis.

ting-hong-shieh:
## Summary

- Reproduce the deterministic parts of the 2024 Defiance example with repository-local thrust and drag data.
- Compare the simulated apogee above ground level with the measured value published in the flight examples.
- Keep the test independent of live weather services and external downloads.

This is the example-based acceptance-test prerequisite suggested in #1057. It does not introduce a Rust proof of concept or change the flight solver.

## Before and after

| Result | Before | After |
| --- | ---: | ---: |
| Defiance example acceptance checks | 0 | 1 |
| Acceptance tests passing | 16 | 17 |
| Simulated apogee AGL | not checked | 9242.183 m |
| Measured apogee AGL | not checked | 9308.320 m |
| Relative apogee error | not checked | 0.711% |
| Acceptance threshold | not checked | less than 1% |

Base SHA: `62aa0f9be32eeccafbc1aefd6ac3d90306b257ce`

Head SHA: `c1f331b5d37032342fc02a5bda69e3618da74696`

## Validation

```text
$ python -m pytest -q tests/acceptance
17 passed, 9 warnings in 1.54s

$ ruff check tests/acceptance/test_defiance_rocket.py
All checks passed!

$ ruff format --check tests/acceptance/test_defiance_rocket.py
1 file already formatted
```

The warnings come from existing weather mappings, stability checks, interpolation inputs, and the Defiance example's parachute pressure-signal construction. This PR does not change or suppress them.

## Local environment

| Component | Version |
| --- | --- |
| Python | 3.12.6 |
| RocketPy | 1.13.0 |
| NumPy | 2.5.2 |
| SciPy | 1.18.0 |
| pytest | 9.1.1 |
| OS | macOS 26.5 (Darwin 25.5.0), arm64 |

## Scope

This test covers one published output from the Defiance example under its fixed custom atmosphere. It does not validate every example flight, benchmark 6-DOF performance, or define numerical tolerances for a future Rust implementation.

Related to #1057.

### DIFF
--- a/tests/acceptance/test_defiance_rocket.py
+++ b/tests/acceptance/test_defiance_rocket.py
@@ -1,0 +1,109 @@
+"""Acceptance test for the 2024 Defiance example flight."""
+
+from rocketpy import Environment, Flight, Rocket
+from rocketpy.motors import CylindricalTank, Fluid, HybridMotor
+from rocketpy.motors.tank import MassFlowRateBasedTank
+
+
+MEASURED_APOGEE_AGL = 9308.32
+MAX_RELATIVE_APOGEE_ERROR = 0.01
+
+
+def _build_defiance_flight():
+    """Build the deterministic Defiance example flight."""
+    environment = Environment(
+        latitude=47.966527,
+        longitude=-81.87413,
+        elevation=1383.4,
+        date=(2024, 8, 24, 0),
+    )
+    environment.set_atmospheric_model(type="custom_atmosphere", wind_v=1.0, wind_u=-2.9)
+
+    liquid_oxidizer = Fluid(name="N2O_l", density=960)
+    gaseous_oxidizer = Fluid(name="N2O_g", density=1.9277)
+    oxidizer_tank = MassFlowRateBasedTank(
+        name="oxidizer_tank",
+        geometry=CylindricalTank(radius_function=0.0665, height=1.79),
+        flux_time=6.5,
+        liquid=liquid_oxidizer,
+        gas=gaseous_oxidizer,
+        initial_liquid_mass=17,
+        initial_gas_mass=0,
+        liquid_mass_flow_rate_in=0,
+        liquid_mass_flow_rate_out=17 / 6.5,
+        gas_mass_flow_rate_in=0,
+        gas_mass_flow_rate_out=0,
+    )
+
+    motor = HybridMotor(
+        thrust_source="data/rockets/defiance/Thrust_curve.csv",
+        dry_mass=13.832,
+        dry_inertia=(1.801, 1.801, 0.0305),
+        center_of_dry_mass_position=0.780,
+        grain_number=1,
+        grain_separation=0,
+        grain_outer_radius=0.0665,
+        grain_initial_inner_radius=0.061,
+        grain_initial_height=1.25,
+        grain_density=920,
+        nozzle_radius=0.0447,
+        throat_radius=0.0234,
+        grains_center_of_mass_position=0.377,
+        coordinate_system_orientation="nozzle_to_combustion_chamber",
+    )
+    motor.add_tank(tank=oxidizer_tank, position=2.2)
+
+    rocket = Rocket(
+        radius=0.07,
+        mass=37.211,
+        inertia=(94.14, 94.14, 0.09),
+        center_of_mass_without_motor=3.29,
+        power_off_drag="data/rockets/defiance/DragCurve.csv",
+        power_on_drag="data/rockets/defiance/DragCurve.csv",
+        coordinate_system_orientation="tail_to_nose",
+    )
+    rocket.add_motor(motor, position=0.2)
+    rocket.add_nose(length=0.563, kind="vonKarman", position=4.947)
+    rocket.add_trapezoidal_fins(
+        n=3,
+        span=0.115,
+        root_chord=0.4,
+        tip_chord=0.2,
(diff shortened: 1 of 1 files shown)


Gui-FernandesBR:
only apogee? could we set a few more tests? max_speed, impact coordinates, max_acceleration, etc

ting-hong-shieh:
I reran the acceptance suite at commit `fa63fb7`: `17 passed`, with `9 warnings`. The deterministic flight produced `apogee_agl=9242.183488 m`, `max_speed=444.243900 m/s`, `x_impact=1625.548498 m`, `y_impact=81.781992 m`, and `max_acceleration=10400.763520 m/s²`. Repeating the same flight reproduced these values exactly.

The repository currently provides a measured reference only for apogee. The other values are simulation outputs, so assertions for them would be deterministic regression checks rather than independent validation against measured flight data. We can add those checks if that is the intended scope.

Gui-FernandesBR (approved):

PR #1161, opened 2026-08-14
BUG: refuse to run a Monte Carlo over a results file it cannot write

thc1006:
## Pull request type

## Checklist

- [x] Lint (`ruff check` / `ruff format --check`) has passed locally
- [x] All tests (`pytest tests/unit tests/integration`) have passed locally

## Current behavior

`import_outputs()` reads `.txt`, `.csv` and `.json`, sets `output_file` to the path it read, and its own docstring said the imported file could be used "to continue a simulation".

`simulate()` only writes JSONL, and `__setup_files()` opens the file with `w+` when `append=False`. Importing a CSV and running again therefore empties it and writes JSON objects into a file still named `.csv`:

```python
analysis.import_outputs("results.csv")   # output_file is now results.csv
analysis.simulate(number_of_simulations=100)
```

Measured on `develop`: `results.csv` goes from 31 bytes to 0, then fills with JSONL. The extension, the reader that would be picked for it, and the bytes on disk all disagree, and the imported data is gone.

`append=True` is the same mismatch without the truncation, appending JSONL rows after a CSV header.

## New behavior

`simulate()` checks all three log paths before anything is opened, and refuses anything that is not `.txt`. The message names which of `input_file`, `output_file` or `error_file` has to change, and says CSV and JSON stay importable for analysis. Uppercase `.TXT` is accepted, since it is the same file to the filesystem.

The note under `import_outputs` no longer offers any saved file as something to resume from, since that is what led a `.csv` into `output_file` in the first place.

This does not narrow what can be imported. It only stops `simulate()` writing over a file it cannot read back.

## Breaking change

- [x] No

A run that pointed a log at `.csv` or `.json` was destroying or corrupting that file, so nothing was relying on it.

## Additional information

Six tests. Four drive the real `simulate()` through the `monte_carlo_calisto` fixture over `{.csv, .json} x {append=False, append=True}` and assert both that it raises and that `read_bytes()` is unchanged. The others cover the per-path naming and the uppercase suffix.

Removing the call from `simulate()` turns only the four integration cases red, while the direct check stays green; making the check accept everything turns both red. That is what separates the wiring from the check.

Local run against `develop` at 1d04bcc3: ruff clean, pylint exit 0, `pytest tests/unit tests/integration` 2219 passed, 51 skipped.

A format-aware log backend would be the larger fix. This is the preflight that stops data being lost in the meantime.

### DIFF
--- a/rocketpy/environment/environment.py
+++ b/rocketpy/environment/environment.py
@@ -5,8 +5,8 @@ import logging
 import os
 import re
 import warnings
-from collections.abc import Mapping
 from collections import namedtuple
+from collections.abc import Mapping
 from datetime import datetime
 
 import netCDF4
--- a/rocketpy/environment/environment.py
+++ b/rocketpy/environment/environment.py
@@ -5,8 +5,8 @@ import logging
 import os
 import re
 import warnings
-from collections import namedtuple
 from collections.abc import Mapping
+from collections import namedtuple
 from datetime import datetime
 
 import netCDF4
--- a/rocketpy/simulation/monte_carlo.py
+++ b/rocketpy/simulation/monte_carlo.py
@@ -1292,7 +1292,9 @@ class MonteCarlo:  # pylint: disable=too-many-public-methods
         -----
         Notice that you can import the outputs, inputs, and errors from a
         file without the need to run simulations. You can use previously saved
-        files to process analyze the results or to continue a simulation.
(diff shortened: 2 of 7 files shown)


Gui-FernandesBR (approved):

Gui-FernandesBR:
Approved on content, but this went conflicting after #1164 was merged. Please merge `develop` in and push — once CI is green I will merge this.

PR #1166, opened 2026-08-15
Fix spurious ValueError from floating-point roundoff at exact tank depletion

arocketcat:
Pull request type

[x] Code changes (bugfix, features)
[x] Code maintenance (refactoring, formatting, tests)

Checklist

[x] Tests for the changes have been added (if needed)
[ ] Docs have been reviewed and added / updated
[ ] Lint (`black rocketpy/ tests/`) has passed locally
[ ] All tests (`pytest tests -m slow --runslow`) have passed locally
[ ] `CHANGELOG.md` has been updated (if relevant)

Current behavior

Tank subclasses (`MassFlowRateBasedTank`, `UllageBasedTank`, `LevelBasedTank`, `MassBasedTank`) raise a `ValueError` when a tank's liquid mass, volume, or height lands within floating-point roundoff of an exact physical boundary most commonly, exact depletion or exact fill.
This happens with any tank drained by a constant (linear) mass flow rate, which is a common and physically realistic model for regulated or pump-fed propellant feed systems. Because floating-point arithmetic doesn't guarantee an exact `0.0` at the instant of depletion, the computed mass lands at something like `-1e-15 kg` instead being physically zero, but numerically negative.
Minimal repro:
```python
from rocketpy import CylindricalTank, Fluid, MassFlowRateBasedTank

tank = MassFlowRateBasedTank(
    name="oxidizer tank",
    geometry=CylindricalTank(radius_function=0.1, height=1.2, spherical_caps=False),
    flux_time=5.0,
    initial_liquid_mass=32.0,
    initial_gas_mass=0,
    liquid_mass_flow_rate_in=0,
    liquid_mass_flow_rate_out=32.0 / 5.0,  # constant drain -> mass hits exactly 0 at t=5
    gas_mass_flow_rate_in=0,
    gas_mass_flow_rate_out=0,
    liquid=Fluid(name="N2O_l", density=1220),
    gas=Fluid(name="N2O_g", density=1.9277),
)
tank.liquid_mass(5.0)
```
```
ValueError: The tank oxidizer tank is underfilled. The liquid mass is
negative given the mass flow rates.
    Try increasing the initial liquid mass, or reducing the mass flow rates.
    The liquid mass is -0.000 kg at 5.0 s.
```
Note the message itself reports `-0.000 kg`  the check is triggering on numerical noise, not a real physical underfill.
A second, related failure occurs one level downstream: `liquid_height` and `gas_height` call `Function.compose()`, which performs its own strict domain check and raises `ValueError: Input Function image (...) must be within the domain of the Function (...)` when the composed volume is marginally outside `[0, tank_volume]` for the same roundoff reason.
Root cause:
`MassFlowRateBasedTank.liquid_mass` checks `if (liquid_mass < 0).any()` with no tolerance, unlike the neighboring `gas_mass` property just below it, which already uses `if (gas_mass < -1e-6).any():  # -1e-6 is to avoid numerical errors`. The tolerance fix was applied to `gas_mass` but never carried over to `liquid_mass`.
Six separate call sites across the four tank subclasses call `some_function.compose(other_function)` with no tolerance for the input function's values sitting marginally outside the target domain, so the same class of roundoff error surfaces again in `liquid_height`/`gas_height` even after the mass check is fixed.

New behavior

Added the same `-1e-6` tolerance already used for `gas_mass` to `liquid_mass` in `MassFlowRateBasedTank`.
Added a small shared helper, `_compose_clipped(outer, inner)`, that clips `inner`'s values to `outer`'s valid domain before composing. This is applied at all six existing `.compose()` call sites across `MassFlowRateBasedTank`, `UllageBasedTank`, `LevelBasedTank`, and `MassBasedTank`. Physically meaningful over/underfill conditions are still caught by each class's own bounds checks immediately afterward; this only absorbs floating-point noise at the domain boundary.
No public API changes. `_compose_clipped` is a private module-level helper.

Testing:
Added `test_mass_flow_rate_tank_exact_depletion` to `tests/unit/motors/test_tank.py`, which builds a tank drained via a constant flow rate that reaches exact depletion at `flux_time`, and asserts `liquid_mass`, `liquid_height`, and `gas_height` all remain well-defined and non-negative through and past that point.
Confirmed this test fails against `master` with the exact error message above.
Confirmed this test passes with this fix applied.
Ran the full existing `tests/unit/motors/` and `tests/integration/motors/test_tank.py` suites (95 tests) against the patched code: all pass, no regressions. This includes the existing `sample_empty_mass_flow_rate_tank` and `sample_empty_ullage_tank` fixtures, which already cover empty-tank states.

Breaking change

[ ] Yes
[x] No

Additional information

Draining a tank to exact depletion via constant flow rate is a natural way to define a `LiquidMotor`/`HybridMotor` — arguably more common than an asymptotic decay curve that never fully empties, which is what the existing `LiquidMotor` documentation example uses (and which incidentally avoids this bug, since `exp(-kt)` never reaches exactly zero in finite time).


Gui-FernandesBR (approved):

Gui-FernandesBR:
Good find, and the diagnosis is right: `gas_mass` has carried the `-1e-6` tolerance for a long time and `liquid_mass` never got it. I ran `tests/unit/motors/test_tank.py` on your branch — 13 passed, including your new regression test.

Three things before this can land.

**1. Please retarget this PR to `develop`.** RocketPy takes contributions on `develop`; `master` only ever receives release merges. This also fixes your lint failure for free: `master`'s own `README.md` currently fails `ruff format --check` (recent ruff versions format Python inside Markdown code fences), so *any* PR based on `master` has a red `lint` job regardless of its own contents. I verified `develop` is clean — 314 files already formatted — while `master` reports `README.md` would be reformatted.

**2. `ruff format` on your two files.** Two `_compose_clipped(...)` call sites go past 88 columns, and the new test needs two blank lines before it:

```
rocketpy/motors/tank.py         lines 991 and 1793
tests/unit/motors/test_tank.py  line 494 (and a trailing blank line at the end)
```

**3. Minor robustness point on `_compose_clipped`.** It reads `inner.x_array` and `outer.x_initial`, which only exist for array-sourced Functions, whereas `Function.compose` deliberately handles the callable-source case with a lambda. I tried to reach that path from the public API and could not — `discretize=None` with a callable input already raises in `_discretize_fluid_inputs` on `master`, so this is latent rather than a live bug — but the guard is cheap and keeps the helper a drop-in replacement for `compose`:

```python
if outer._source_type is not SourceType.ARRAY or inner._source_type is not SourceType.ARRAY:
    return outer.compose(inner)
```

Also, minor: the PR template you filled in is the old one (it still mentions `black`); `develop` has an updated version with the current lint commands.

Gui-FernandesBR:
This is ready to merge — all 9 checks green. I retargeted it and pushed the remaining fixes to your branch (`feaeb6e4`) rather than send you round again.

**Retargeted `master` → `develop`.** That's where RocketPy takes contributions, and it also cleared most of the lint failure on its own: `master`'s own `README.md` currently fails `ruff format --check` (recent ruff versions format Python inside Markdown fences), so the job was red for a reason that had nothing to do with your diff. `develop` is clean. I checked `master` is an ancestor of `develop` first, so the retarget didn't drag anything extra into the diff. I also merged `develop` in, since the branch was cut from the v1.13.0 tag.

**`ruff format`** on the two files — the two long `_compose_clipped(...)` call sites and the blank lines around the new test.

**`_compose_clipped` is now a real drop-in for `compose`.** It read `x_array` / `y_array` and the domain bounds, which only exist on array-sourced Functions, so a callable source raised `AttributeError: 'Function' object has no attribute 'x_array'` exactly where plain `compose` had worked. It now defers to `outer.compose(inner)` unless both sides are arrays — `Function.compose` already handles that case with a lambda and does no bounds check there, so there's no spurious error to absorb and nothing to clip. Used the public `is_array_source()`, which `motors/motor.py` and `environment/` already use. Four tests pin it; removing the guard turns them red with that AttributeError.

**On the substance, two things I checked rather than assumed:**

Your diagnosis is right, and I confirmed the asymmetry you spotted — `gas_mass` has carried the `-1e-6` tolerance for a long time and `liquid_mass` never got it.

The thing that worried me was the unbounded `np.clip`, because `UllageBasedTank.liquid_height` has no bounds check of its own and previously depended on `compose` raising. If clipping swallowed *gross* over/underfill, that would be a real loss of a safety check. It doesn't: I measured develop against your branch with an ullage 1.5× the tank volume and an ullage of −0.5×, and both raise the identical `ValueError` from `Tank`'s own overfill/underfill checks, which fire independently of the composition. So the clipping genuinely only absorbs boundary noise, as your description claims.

Verified before pushing: `tests/unit` 2161 passed / 17 skipped, `tests/unit/motors` + `tests/integration/motors` 116 passed, ruff clean, pylint 10.00/10.

Two notes for next time, no action needed here: the PR template you filled in is the old one that still mentions `black` — `develop` has a current version with the real lint commands. And a tank drained to exact depletion by a constant flow rate is a genuinely common way to define a motor, so this was a good catch.

PR #1174, opened 2026-08-16
BUG: reject live RNG objects as Sensor seeds

### LINKED ISSUE #1087: Sensor accepts a SeedSequence seed but cannot serialize it

myungjunlee:
## Summary

`Sensor.__init__` hands the seed to `numpy.random.default_rng`, which also accepts `Generator` and `BitGenerator` objects. The sensor constructs, `to_dict()` emits the object verbatim, and the failure surfaces later at `json.dumps()`:

```python
import json
import numpy as np
from rocketpy.sensors import Barometer
from rocketpy._encoders import RocketPyEncoder

sensor = Barometer(sampling_rate=10, seed=np.random.default_rng(5))  # accepted
json.dumps(sensor.to_dict(), cls=RocketPyEncoder)
# TypeError: Object of type Generator is not JSON serializable
```

This is the case #1124 leaves behind. #1124 closed the `SeedSequence` half of #1087 by teaching `RocketPyEncoder` to write one out, which works because a `SeedSequence` is defined by its entropy and spawn key and still describes the stream after a round trip.

A `Generator` has no such description. Its state advances on every draw, so what `to_dict()` wrote would depend on when it ran, and a restored copy would not reproduce the stream the sensor used. There is nothing to serialize it *to*, so it is rejected at the constructor instead, where the caller can still see which argument was wrong.

## What changes

- `Generator` and `BitGenerator` seeds raise `TypeError` at construction.
- `seed` is annotated as `int | Sequence[int] | np.random.SeedSequence | None` on every constructor that takes one — the type this issue names in its own wording — so the contract is stated where the argument is declared and not only in the docstring. The docstrings now match.
- Nothing else. Ints, numpy ints, `SeedSequence`, `None`, and the sequences of ints `default_rng` accepts all behave exactly as they do on develop today.

## Test plan

- `pytest tests/unit/sensors/ -q` → 63 passed, including #1124's `test_seedsequence_sensor_seed_is_json_serializable`
- `pytest tests/unit -q` → no new failures (the 4 `test_sensitivity.py` errors are a missing `statsmodels` in my environment and reproduce on a clean `develop`)
- Added: `Generator`/`BitGenerator` rejected · `SeedSequence` and a sequence of ints explicitly still accepted · `None`/`0`/`int`/`np.int64`/`2**128-1` still accepted and round trip

### DIFF
--- a/rocketpy/sensors/sensor.py
+++ b/rocketpy/sensors/sensor.py
@@ -17,18 +17,24 @@ SeedLike = int | np.integer | Sequence[int] | np.ndarray | np.random.SeedSequenc
 
 
 def _is_int_array_like(value):
-    """Whether ``value`` is an int or a (possibly nested) sequence of ints.
+    """Whether ``value`` is an int or a flat sequence of ints.
 
     This is the half of ``numpy.random.default_rng``'s seed contract that
     ``RocketPyEncoder`` can write out: integers keep their value across a JSON
     round trip, so a seed read back names the same stream it named before.
+
+    Flat, because that is all ``SeedSequence`` takes as entropy: a nested
+    sequence raises ``TypeError`` and a multi-dimensional array ``ValueError``,
+    neither of them naming the seed. Older NumPy let some of those through,
+    and RocketPy pins no upper bound, so they are refused here to keep one
+    answer across versions.
     """
     if isinstance(value, (int, np.integer)):
         return True
     if isinstance(value, np.ndarray):
-        return np.issubdtype(value.dtype, np.integer)
+        return np.issubdtype(value.dtype, np.integer) and value.ndim <= 1
     if isinstance(value, Sequence) and not isinstance(value, (str, bytes)):
-        return all(_is_int_array_like(item) for item in value)
+        return all(isinstance(item, (int, np.integer)) for item in value)
     return False
 
 
--- a/rocketpy/sensors/sensor.py
+++ b/rocketpy/sensors/sensor.py
@@ -2,6 +2,7 @@ import json
 import logging
 import warnings
 from abc import ABC, abstractmethod
+from collections.abc import Sequence
 
 import numpy as np
 
@@ -62,7 +63,7 @@ class Sensor(ABC):
         temperature_bias=0,
         temperature_scale_factor=0,
         name="Sensor",
-        seed=None,
+        seed: int | Sequence[int] | np.random.SeedSequence | None = None,
     ):
         """
         Initialize the accelerometer sensor
@@ -112,15 +113,25 @@ class Sensor(ABC):
             meaning no temperature scale factor is applied.
         name : str, optional
             The name of the sensor. Default is "Sensor".
-        seed : int, optional
+        seed : int, Sequence[int], numpy.random.SeedSequence, optional
             Seed for the random number generator that draws the measurement
             noise. If given, the noise becomes reproducible and independent of
-            the process-global NumPy RNG. Default is None, meaning the noise is
-            seeded from fresh entropy per instance.
+            the process-global NumPy RNG. A ``numpy.random.SeedSequence`` is
+            also accepted and round trips through ``RocketPyEncoder``. The
+            ``Generator`` and ``BitGenerator`` objects that
+            ``numpy.random.default_rng`` takes are rejected here, because their
+            state advances as noise is drawn and so cannot be represented in
+            the dictionary returned by ``to_dict()``. Default is None, meaning
+            the noise is seeded from fresh entropy per instance.
 
         Returns
         -------
         None
+
+        Raises
+        ------
+        TypeError
+            If ``seed`` is a ``Generator`` or a ``BitGenerator``.
 
         See Also
         --------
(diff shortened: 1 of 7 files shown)


thc1006 (commented):
Requesting changes for two correctness gaps:

1. `RandomState` is accepted by `default_rng` on NumPy >= 2.2, so the original “constructs successfully, fails later at JSON serialization” bug is still reachable on a supported RocketPy dependency version.
2. Accepted mutable seed descriptors are stored by reference. A later mutation can make `to_dict()` serialize a seed that no longer describes the stream used to initialize the sensor. This affects list/ndarray seeds and `SeedSequence` instances backed by mutable entropy.

I would also align the annotation and error text with the actual array-like integer contract (`np.integer` and ndarray are accepted today), and make the round-trip tests compare the generated noise stream rather than only the stored seed/state.

The constructor-level rejection is the right direction; these changes would make the boundary complete rather than covering only two currently known live RNG classes.

rocketpy/sensors/sensor.py
```diff
@@ -151,6 +162,24 @@ def __init__(
...
 
+        # default_rng() also accepts Generator and BitGenerator objects, which
+        # are not a description of a stream but a stream already in progress:
+        # their state advances on every draw, so what to_dict() writes depends
+        # on when it ran. #1124 taught RocketPyEncoder to serialize a
+        # SeedSequence, which stays reproducible because it is defined by its
+        # entropy and spawn key; a live generator has no such description.
+        # Without this check the sensor builds fine and only fails at
+        # json.dumps(), far from the call that caused it.
+        if isinstance(seed, (np.random.Generator, np.random.BitGenerator)):
```
thc1006: `np.random.default_rng` also accepts `np.random.RandomState` starting with NumPy 2.2. RocketPy supports `numpy>=1.23` with no upper bound, so this still leaves the same late-failure path on current NumPy: the sensor constructs, the RandomState remains live state, and RocketPyEncoder cannot encode it.

Could we reject `RandomState` here as well and add it to the parametrized rejection test? More generally, validating against the stable seed-descriptor contract (`SeedSequence` entropy inputs) would be less brittle than blacklisting whichever live RNG types `default_rng` happens to accept today.

myungjunlee:
Thanks — the `RandomState` gap is real, and it took the PR's own claim with it. Both correctness points reproduce here on numpy 2.4.6. Pushed two commits.

**`RandomState`, and the shape of the check.** You're right that blacklisting live types is the brittle half of the contract, and this is a good demonstration of why: `RandomState` is not a subclass of either `Generator` or `BitGenerator`, so the `isinstance` list walked straight past it.

```
default_rng(RandomState)         -> Generator
Accelerometer(seed=RandomState)  -> constructs
json.dumps(...)                  -> TypeError: Object of type RandomState is not JSON serializable
```

So the check now accepts descriptors instead of naming live types: `None`, an int (including `np.integer`), an array_like of ints (nested, ndarray of integer dtype, empty included), or a `SeedSequence`. Everything else is refused at the constructor. The live-RNG branch is kept ahead of it purely for the error message, since "your seed's state advances as noise is drawn" is more useful than "invalid seed type".

Worth noting for anyone tightening this later: `default_rng` accepts an empty sequence and `bool`, and both serialize, so the descriptor check has to let them through.

**Annotation.** Widened to what the check actually takes, via a `SeedLike` union in `sensors/sensor.py`. It is a real `types.UnionType`, so `help()` and `inspect.signature()` still expand it to `int | numpy.integer | Sequence[int] | numpy.ndarray | SeedSequence | None` — the alias only keeps the seven signatures inside the 88-column limit. Docstrings and the `Raises` section follow.

**Round-trip tests.** Agreed, comparing the stored seed proves too little. Added a test that draws from the restored sensor and compares the sequence against a fresh sensor built from the same seed, parametrized over int, `np.int64`, int sequence and `SeedSequence`. It goes through `RocketPyDecoder`, since a `SeedSequence` comes back as a plain dict without it.

**Mutable seed descriptors.** Confirmed, and it is worse than only the seed field:

```python
seed = [1, 2, 3]
sensor = Accelerometer(sampling_rate=10, seed=seed)
seed[0] = 999
sensor.to_dict()["seed"]   # [999, 2, 3]
```

`SeedSequence` backed by mutable entropy does the same, and its serialized `signature.hash` stays at the old value while `entropy` changes, so the record is not merely stale but internally inconsistent.

I've left it out of this PR deliberately. `self._seed = seed` predates it and is untouched here, and the failure is a different one — this PR is about a seed that cannot be written down at all and fails loudly at `json.dumps()`, whereas aliasing writes a value that succeeds and is quietly wrong. A copy at assignment would fix it, but that changes what `to_dict()["seed"] is seed` means for every caller, which seems worth its own issue and its own tests rather than riding along here.

Local run: `tests/unit/sensors` 81 passed, ruff and pylint clean. CI on this fork is still waiting on workflow approval.

myungjunlee:
`develop` has moved two commits since I branched — neither touches a file this PR does, and it still merges clean, so there's no rebase to do.

No workflow has run on this PR at all — not on the review commits, not on the first one. The three check suites are still sitting at `action_required` with zero runs.

Gui-FernandesBR (approved):

PR #1175, opened 2026-08-16
Fixed sensor.accelerometer gravity sign for `consider_gravity = True` case

vapsik:
## Issue
In the case of `consider_gravity = True` the current `sensor.accelerometer` sensor accounts for gravity vector incorrectly — $$a_{ \text{inertial} } = a_{\text{ENU}} + g$$ but since the gravity vector has components $g = (0,0,-9.8)$ in ENU frame then it must be $$a_{ \text{inertial} } = a_{\text{ENU}} - g$$. 

A simple example showcasing the current implementation being incorrect is with a stationary state (1) and airless free-falling rocket (2) - both with ENU-aligned accelerometers: 1) during stationary phase where $a_{\text{ENU}} = 0$ the  `consider_gravity = True` accelerometer reading shows $a_{ \text{inertial} } = (0,0,-9.8)$ where as real IMU would show $a_{ \text{inertial} } = (0,0,+9.8)$ as normal force upward from the ground must negate the effect of gravity for achieving the stationary state. 2) as a rocket goes into (airless) free fall real IMU would show inertial acceleration of 0 (weightless) but with current implementation in free fall or near apogee before parachute opening the `consider_gravity = True` accelerometer shows total acceleration of nearly 19.6 showcasing the wrong gravity sign usage.

## Changes
In `accelerometer.py` changed `inertial_acceleration = Vector(u_dot[3:6]) + gravity` into `inertial_acceleration = Vector(u_dot[3:6]) - gravity`.

## Test file showcasing the bugs (made with Claude)
[repro_gravity_bug.py](https://github.com/user-attachments/files/31115443/repro_gravity_bug.py)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ Attention: The newest changes should be on top -->
 
 ### Fixed
 
+- BUG: Correct the gravity sign an `Accelerometer` applies when `consider_gravity=True`. The gravitational field was added to the inertial acceleration instead of subtracted from it, so the sensor reported the negative of the proper acceleration along the vertical: one at rest read -g rather than +g. Recorded accelerometer data taken with `consider_gravity=True` changes sign in that term. [#1175](https://github.com/RocketPy-Team/RocketPy/pull/1175)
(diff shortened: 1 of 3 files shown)


Gui-FernandesBR:
@vapsik tests not passing on CI... Could you take a look pleasE?

Gui-FernandesBR (approved):

Issue #953, opened 2026-04-19
Free form fin in the stochastic method


laurapgp:
Through Discord server, a member has suggested adding a free-form fin in the stochastic method for Monte Carlo simulations

PR #1180, opened 2026-08-16
DOC: backfill the Unreleased changelog entries the automation missed

### LINKED ISSUE #1173: BUG: the changelog workflow stopped running, and CHANGELOG.md is 37 merged pull requests behind

ting-hong-shieh:
Part of #1173.

## What this does

`CHANGELOG.md` was last touched by `5eae273b` on 12 August and the highest pull request it records is #1140. Thirty-nine commits have landed on `develop` since, so `[Unreleased]` no longer describes the branch. #1173 notes that those entries will not appear on their own and suggests writing them in one pass from the merge list; this is that pass.

Twenty-eight entries, `CHANGELOG.md` only:

| Section | Count |
| --- | ---: |
| Added | 6 ENH, 2 DOC |
| Changed | 1 ENH, 1 DOC |
| Fixed | 16 BUG, 2 DOC |

## How they were chosen

From `git log 5eae273b..develop`, following the rules in the file's own header comment: `Added` for new features, `Changed` for changes in existing functionality, `Fixed` for bug fixes, and tests and GitHub maintenance left out. Seven `TST:` commits and four CI or maintenance commits fall in the gap and are therefore not listed.

Each entry uses the pull request title and link, as the header comment asks, with the issue linked alongside where the commit names one. `772480d` is a direct commit to `develop` with no pull request, so that entry links the commit.

Newest on top within each section, as the `[Unreleased]` comment asks.

## Verification

Every link was resolved against the API rather than assembled by hand:

```
41 pull links, 19 issue links (16 distinct issues)
mismatched label vs url: none
```

Every `pull/N` link resolves to a pull request, every `issues/N` link resolves to an issue and not to a pull request, and each entry's text was checked against the pull request's own title. That check caught one wrong description — #1142 is "Document angle-of-attack drag inputs", not the multivariable inputs its squash-commit subject suggests — and twelve smaller drifts from the titles, all corrected.

## What this does not do

It does not fix the automation, and the next merge will fall out of the file again. Both causes #1173 identifies are outside what a pull request into `develop` can reach:

- a `pull_request_target` workflow is read from the default branch, and `master` still carries the older `changelog.yml` with the `pull_request` trigger, so neither trigger matches anything: `pull_request` events read the file from the pull request's branch, where only `pull_request_target` is declared, and `pull_request_target` events read it from `master`, where only `pull_request` is declared;
- `RELEASE_TOKEN` was already resolving to empty before that, which is why the four runs before #1112 all died on `actions/checkout`.

I confirmed the first against `upstream/master:.github/workflows/changelog.yml` and the run history; the second needs someone with access to the repository secrets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,11 +51,7 @@ Attention: The newest changes should be on top -->
 ### Changed
 
 - ENH: Compute the rocket static margin lazily [#1135](https://github.com/RocketPy-Team/RocketPy/pull/1135) [#780](https://github.com/RocketPy-Team/RocketPy/issues/780)
-- CI: Measure coverage on develop, not only on pull requests [#1162](https://github.com/RocketPy-Team/RocketPy/pull/1162)
-- MNT: Let pylint use every core it is given [#1163](https://github.com/RocketPy-Team/RocketPy/pull/1163)
 - DOC: Tighten the comments that came with the sampler seed groups [#1154](https://github.com/RocketPy-Team/RocketPy/pull/1154)
-- CI: Run the changelog job for pull requests from forks [#1112](https://github.com/RocketPy-Team/RocketPy/pull/1112)
-- CI: Keep per-matrix coverage artifacts and fail Codecov errors [#1123](https://github.com/RocketPy-Team/RocketPy/pull/1123) [#1088](https://github.com/RocketPy-Team/RocketPy/issues/1088)
 - CI: make the Gemini PR reviewer actually review [#1140](https://github.com/RocketPy-Team/RocketPy/pull/1140)
 - MNT: declare dependency floors the package can actually run on [#1108](https://github.com/RocketPy-Team/RocketPy/pull/1108)
 - CI: build the docs for pull requests into develop as well [#1104](https://github.com/RocketPy-Team/RocketPy/pull/1104)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,14 @@ Attention: The newest changes should be on top -->
 
 ### Added
 
+- ENH: Support fixed-time parachute deployment triggers [#1133](https://github.com/RocketPy-Team/RocketPy/pull/1133) [#437](https://github.com/RocketPy-Team/RocketPy/issues/437)
+- DOC: Add SIL parachute ejection integration example [#1131](https://github.com/RocketPy-Team/RocketPy/pull/1131) [#524](https://github.com/RocketPy-Team/RocketPy/issues/524)
+- ENH: List NOAA atmosphere datasets and fetch latest [#1136](https://github.com/RocketPy-Team/RocketPy/pull/1136) [#660](https://github.com/RocketPy-Team/RocketPy/issues/660)
+- ENH: Model unbonded solid-motor grain CM shift [#1138](https://github.com/RocketPy-Team/RocketPy/pull/1138) [#340](https://github.com/RocketPy-Team/RocketPy/issues/340)
+- ENH: Add `Flight` post-step callback across all phases [#1128](https://github.com/RocketPy-Team/RocketPy/pull/1128) [#758](https://github.com/RocketPy-Team/RocketPy/issues/758)
+- DOC: Document angle-of-attack drag inputs [#1142](https://github.com/RocketPy-Team/RocketPy/pull/1142)
+- ENH: Create ensembles from user-defined profiles [#1141](https://github.com/RocketPy-Team/RocketPy/pull/1141)
+- ENH: Add tube-fin aerodynamic surface [#1144](https://github.com/RocketPy-Team/RocketPy/pull/1144)
 - ENH: StochasticFreeFormFins for Monte Carlo simulations [#1117](https://github.com/RocketPy-Team/RocketPy/pull/1117)
 - ENH: `StochasticFreeFormFins`, so free-form fin sets can be used in Monte Carlo simulations. The outline is randomized as a block, since a shape is only meaningful as a complete set of points: every coordinate is perturbed by its own draw, the fin root is held on the body line, and a list of candidate outlines can have a different number of points in each. [#953](https://github.com/RocketPy-Team/RocketPy/issues/953)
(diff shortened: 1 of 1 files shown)


thc1006 (commented):
Two numbers in the PR description look stale:

* `Added` contains 6 ENH entries and 2 DOC entries, not 7 ENH and 2 DOC.
* At this head, `[Unreleased]` contains 45 pull links and 20 issue links, not 17 issue links.

The total of 32 new entries and the “no mismatched label vs URL” result are both correct.

thc1006:
One policy point before this lands: the changelog header says tests and GitHub maintenance should not be listed, but this backfill excludes TST commits while including #1112, #1123, #1162 and #1163. The updater script also currently inserts CI, TST and MNT entries, so the repository has two conflicting rules.

Which rule should this backfill follow?

Also, because #1173 is still unfixed, could this branch be refreshed immediately before merge? Otherwise #1177 and any other PR merged after `4263fa9` will begin the next gap straight away.

ting-hong-shieh:
Both numbers were wrong. Corrected in the description.

- `Added` gets 6 ENH and 2 DOC. The seventh ENH is #1135, which went to `Changed`, and I counted it in both rows. The section totals are 8 / 6 / 18, which is the 32 I claimed.
- 45 pull links and 20 issue links. My check deduplicated before counting and I reported the distinct total as the raw one; #953 alone appears three times. It now reads "45 pull links, 20 issue links (17 distinct issues)".

Thanks — the second one means my verification was reporting a number I had not actually looked at.

## Which rule the backfill should follow

There are three, not two, and they disagree in different directions:

1. **The header comment in `CHANGELOG.md`** — no tests, no GitHub maintenance, no merge commits.
2. **`.github/scripts/update_changelog.py`** — `fallback_section_and_prefix` maps `tests`/`TST` to `### Changed`, `MNT` to `### Changed`, and falls through to `CI` in `### Changed`. There is no skip path; every merged pull request gets an entry.
3. **What the file actually contains** — 117 `ENH:`, 75 `BUG:`, 42 `DOC:`, 39 `MNT:`, 4 `CI:`, 2 `TST:`.

I followed (3), which is why `MNT` and `CI` are in and `TST` is out. That was the wrong thing to follow, because it describes the past rather than what happens next.

The backfill should match (2). Whatever the script does is what every future entry will look like once #1173 is fixed, and if the two disagree then this 32-entry stretch becomes the only part of `[Unreleased]` with no `TST` lines — a gap that reads like an oversight rather than a rule. On that basis I should add the seven `TST` commits from the gap, in `### Changed`, which takes the backfill to 39.

The alternative is to make (1) win and give the script a skip path for `tests` and `github maintenance`. That is a real design choice — `CI:` and `MNT:` entries are arguably both "GitHub maintenance", and dropping them would remove things like #1108's dependency floors that users can be affected by. It also changes 39 existing `MNT:` entries from correct to accidental.

It is your call which one is the rule. Say which and I will regenerate the backfill to match, and open the script change separately if it needs one.

## Refreshing before merge

Agreed, and it costs nothing to do. `develop` is still at `4263fa9` right now, so this branch is current as of this comment. I will rebase and add whatever has landed immediately before you merge — ping me and I will refresh within the hour.

## Unrelated, while we are here

The two `TST:` entries in the file are `#914` in v1.12.0 and `#862` in v1.11.0, and #862 is "remove remaining files after test session". That is the same class of problem as the three stray files you caught on #1177: `monte_carlo_calisto` uses `filename="monte_carlo_test"`, so a test run drops `monte_carlo_test.{inputs,outputs,errors}.txt` in the repository root and `.gitignore` does not cover them. Happy to open that as a small separate pull request if it is wanted.

thc1006:
Thx, this makes the conflict clear. My preference is to make the CHANGELOG header the source of truth rather than adding the seven TST entries because the current updater happens to lack a skip path.

Gui-FernandesBR (approved):

PR #1181, opened 2026-08-17
BUG: accept the seed type a Monte Carlo worker is handed

thc1006:
Parallel Monte Carlo does not run on `develop` right now. It hangs, and I think it has since 12 August.

## Pull request type

## Checklist

- `CHANGELOG.md` — no action needed; an LLM workflow auto-updates it after merge. Worth knowing that it has not run since #1112, which I wrote up in #1173.

## Current behavior

A two-worker run over a real `Flight`, bisected:

```
d21abde6^   the commit before #1117     1 passed in 2.32s
develop     4263fa95                    did not finish; killed at 90s, 180s and 900s
develop     4263fa95, serial            1 passed in 3.86s
```

`__run_in_parallel` spawns a `SeedSequence` per worker and passes it down, so `_set_stochastic` receives one of those rather than an int:

```
monte_carlo.py:472   seeds = np.random.SeedSequence().spawn(n_workers)
monte_carlo.py:536   self.environment._set_stochastic(seed)
```

`_sampler_seed` then hands it to `SeedSequence(entropy=...)`, which takes an int or a sequence of ints, so the first worker dies before it draws anything:

```
TypeError: SeedSequence expects int or sequence of ints for entropy
```

Two things kept this quiet. `_sampler_seed` is mine, from #1102, and until recently it was only reached from the custom sampler reset loop, so a model without custom samplers never went near it. #1117 added the list-choice generator at `stochastic_model.py:157`, which every model goes through on every reseed. That turned a path almost nobody took into the one all of them take. Nothing wrong with the #1117 change; my function should have taken the type it can now be given.

The hang on top of the error is a separate problem in the worker's own handler, and I have that in a follow-up rather than in here.

Nothing in the test suite calls `_set_stochastic` with a `SeedSequence`, and no pull request job runs a real parallel Monte Carlo, which is why the matrix stayed green through all of it.

## New behavior

`_sampler_seed` normalizes what it is given before building anything from it.

The value is folded through `generate_state` rather than read off `.entropy`. The children of one root share their entropy and differ only by `spawn_key`, so reading the entropy would have quietly put every worker on the same sampler stream, which is worse than the crash it replaces:

```
np.random.SeedSequence(7).spawn(2)
  same .entropy on both      True
  spawn_key                  (0,)  (1,)
```

An int or `None` seed goes through untouched, so no fixed-seed baseline moves. `stochastic_calisto` under seed 42 still reads `mass=14.906007947 radius=0.063501935`, same as `develop`.

The fold itself lives in `rocketpy.tools`. The component streams in #1170 need the same one and the per-index seeding will too, and three copies would drift on width and word order.

`tests/unit/simulation/test_monte_carlo_parallel_runs.py` runs a real serial and parallel Monte Carlo. Both together take under six seconds and it is not marked slow, so a pull request gets the signal that was missing here. It builds its own `MonteCarlo` on `tmp_path` rather than retargeting the fixture's: `filename` is a plain attribute and the three log paths are settled in `__init__`, so assigning it would leave the test writing into the working directory.

## Breaking change

- [x] No

## Additional information

```
pytest tests/unit                     2168 passed
pytest rocketpy --doctest-modules       48 passed
pytest tests/integration               154 passed
pytest tests/acceptance                 18 passed
ruff check . / ruff format --check .   clean
pylint rocketpy/ tests/ docs/          10.00/10, exit 0
```

The four `tests/unit/test_sensitivity.py` failures on my machine are a missing `statsmodels` and fail the same way on an untouched `develop`.

Each mechanism is pinned by a mutation, and each leaves a control standing:

|---|---|---|
| the seed is not normalized | five | the integer baseline test |
| the fold becomes `return seed.entropy` | exactly one, `test_two_workers_do_not_share_a_sampler_stream` | the other five |

The second row is the reason that test exists. Reading `.entropy` unbreaks everything else and silently collapses the workers onto one stream, so the suite would have gone green on a fix that is worse than the bug.

Merged with #1169, #1170 and my follow-up into a throwaway tree on `develop` and run there as well, since green on separate bases says nothing about the combination.

### DIFF
--- a/rocketpy/stochastic/stochastic_model.py
+++ b/rocketpy/stochastic/stochastic_model.py
@@ -8,7 +8,7 @@ import numpy as np
 from rocketpy.mathutils.function import Function
 from rocketpy.stochastic.custom_sampler import CustomSampler
 
-from ..tools import get_distribution
+from ..tools import _seed_sequence_to_int, get_distribution
 
 
 def _names_as_spawn_key(input_names):
@@ -41,6 +41,18 @@ def _format_number(value):
     return f"array of shape {np.shape(value)}"
 
 
+def _seed_as_entropy(seed):
+    """A seed as something ``SeedSequence`` will take as entropy.
+
+    A parallel run is handed a ``SeedSequence``, which it will not take. Any
+    other seed goes through untouched, so the stream an int reaches stays where
+    it was.
+    """
+    if not isinstance(seed, np.random.SeedSequence):
+        return seed
+    return _seed_sequence_to_int(seed)
+
+
 def _sampler_seed(seed, input_names):
     """Derive a seed for one sampler, or for one group that shares a generator.
 
@@ -54,10 +66,10 @@ def _sampler_seed(seed, input_names):
     # Sorted here rather than trusting the caller, so a future call site cannot
     # give one group two different seeds by listing its members another way.
     root = np.random.SeedSequence(
-        entropy=seed, spawn_key=_names_as_spawn_key(tuple(sorted(input_names)))
+        entropy=_seed_as_entropy(seed),
+        spawn_key=_names_as_spawn_key(tuple(sorted(input_names))),
     )
-    words = root.generate_state(4, dtype=np.uint32)
-    return sum(int(word) << (32 * position) for position, word in enumerate(words))
+    return _seed_sequence_to_int(root)
 
 
 # TODO: Stop using assert in production code. Use exceptions instead.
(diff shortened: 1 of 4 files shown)


thc1006:
Pushed 044ffd90, wording only. Calling the result the int a `SeedSequence` can be rebuilt from reads as a round trip of the entropy and spawn key, and it is neither: the helper derives a 128-bit seed and the state it came from cannot be read back out. No behaviour change.

Gui-FernandesBR (approved):

Issue #1186, opened 2026-08-25
BUG: u_dot_generalized applies the center of mass lever arm with the wrong sign


thc1006:
**Describe the bug**

`Flight.u_dot_generalized` reads `r_CM` and `r_NOZ` with the opposite orientation to the one the formulas around them assume, so the angular solve ends up with the wrong lever arm. A lateral force applied at the center of dry mass pitches the rocket the other way from what Newton-Euler gives.

`Rocket.evaluate_com_to_cdm_function` and `Rocket.evaluate_nozzle_to_cdm` both carry a leading minus, so they run CM to CDM and NOZ to CDM. The [Equations of Motion v1](https://docs.rocketpy.org/en/latest/technical/equations_of_motion_v1.html) page defines `r_CM` the other way. Its linear equation is

```
m (v' + w' x r_CM + w x (w x r_CM) + r_CM'' + 2 w x r_CM') = ...
```

and that left hand side is `m a_CM` only when `r_CM` runs CDM to CM. `Rocket.__evaluate_single_surface_cp_to_cdm` has no leading minus, so `surfaces_cp_to_cdm` already matches the documented orientation and the aerodynamic moments come out about the CDM as the derivation expects.

The documented block system is the usual rigid body pair for a body fixed reference point, and the two coupling matrices are transposes of one another:

```
M v' + [m r]x^T w' = T20
I w' + [m r]x   v' = T21
```

In the code the same one shows up in both:

```python
w_dot = I_CM.inverse @ (T21 + (T20 ^ r_CM))
v_dot = K @ (T20 / total_mass - (r_CM ^ w_dot)) - 2 * (w_earth ^ v)
```

Taking `r_CM` as the property actually returns it, the `v_dot` line is the consistent one and `w_dot` is not. I mention that because flipping the sign in `v_dot` is the change the algebra suggests first, and on its own it makes things worse.

After the cancellations the difference is

```
w_dot_code - w_dot_true = 2 * I_CM^-1 * (r_CM x (thrust + aerodynamic force))
```

Thrust is always axial here, so `r_CM x thrust` vanishes and the aerodynamic part is what survives. The lever from the center of mass to a center of pressure comes out as `cp + r_CM` rather than `cp - r_CM`. On `calisto` at ignition `r_CM` is 0.132 m against a 0.261 m arm, so the restoring moment acts through roughly twice the distance it should.

**To Reproduce**

[...]

Issue #1185, opened 2026-08-25
BUG: second derivative of com_to_cdm_function diverges at the burn start time


thc1006:
**Describe the bug**

`Flight.u_dot_generalized` builds the second time derivative of the CM offset like this:

```python
r_CM_ddot = Vector([0, 0, r_CM_z.differentiate(t, order=2)])
```

`com_to_cdm_function` has a kink in its first derivative at the burn start time: flat before ignition, sloped after. When its source is a callable, `Function.differentiate(x, dx=1e-6, order=2)` falls back to `(f(x+dx) - 2 f(x) + f(x-dx)) / dx**2`, and across a kink of slope `s` that evaluates to about `s / dx`. It grows without bound as the step shrinks.

Whether it bites depends on how `com_to_cdm_function` ends up being sourced, which in turn depends on the motor:

| fixture | source | `r_CM''` at `burn_start_time` | `- m * r_CM''` |
|---|---|---|---|
| `calisto` | ndarray, linear | 0 | 0 N |
| `calisto_robust` | ndarray, linear | 0 | 0 N |
| `calisto_liquid_modded` | ndarray, linear | 0 | 0 N |
| `calisto_hybrid_modded` | callable | -3.749668e+04 m/s² | **+1.92e+06 N** |

That last row is a 51.1 kg rocket, so the spurious force is about 3800 times its weight.

The array cases are not really fine either. For a linearly interpolated array the evaluator's second derivative is identically zero, so the `- total_mass * r_CM_ddot` term from the documented equations is silently dropped everywhere rather than being wrong at one point.

**To Reproduce**

Against the fixtures:

```python
def test_probe(calisto_hybrid_modded):
    rocket = calisto_hybrid_modded
    t0 = rocket.motor.burn_start_time
    print(rocket.com_to_cdm_function.differentiate(t0, order=2))
    # -37496.68...
```

And without any fixtures, showing the `1/dx` growth on a bare kink:

```python
from rocketpy.mathutils.function import Function

slope = -3.1134e-4
kink = Function(lambda t: 0.0 if t <= 1.01 else slope * (t - 1.01))
for dx in (1e-5, 1e-6, 1e-7):
    print(dx, kink.differentiate(1.01, dx=dx, order=2))

# 1e-05 -31.13...
# 1e-06 -311.3...
# 1e-07 -3113....
```

**Expected behavior**

[...]

PR #1188, opened 2026-08-25
BUG: correct the nozzle gyration tensor parallel axis term

thc1006:
This patch corrects the parallel axis term in `Rocket.evaluate_nozzle_gyration_tensor`. It is currently a quarter of what the exit plane integral gives. The unit stated in the docstring is fixed too.

## Pull request type

## Checklist

- [x] Lint has passed locally

Some notes so the boxes are not overstated.

The lint line still mentions `black`, so to be precise about what I ran: `ruff check .`, `ruff format --check .` and `pylint rocketpy/ tests/ docs/`, matching the Linters workflow. The only docs change is prose inside an existing numpydoc docstring, with no directives or references, so I did not build the docs locally.

I left the last box unticked rather than claim something I did not see finish. Of the 35 slow tests, 34 passed. `test_hrrr_atmosphere` failed once on a forecast download and passes on re-run, on this branch and on `develop` alike; its own docstring notes that the HRRR latest model sometimes lacks 24 hours of forecast. `test_monte_carlo_simulate[True]`, the parallel case, does not finish inside 40 minutes on this machine, so I cannot report it either way. The sequential case of the same test passed, and the two run the same simulations.

## Current behavior

`evaluate_nozzle_gyration_tensor` builds the tensor as:

```python
S_noz_33 = 0.5 * self.motor.nozzle_radius**2
S_noz_11 = S_noz_22 = 0.5 * S_noz_33 + 0.25 * self.nozzle_to_cdm**2
```

`S_noz` is the second moment of the nozzle exit plane per unit area, taken about the center of dry mass. For a uniform disk of radius `R` centred on the body axis at a distance `d` from that point:

```
S = (1/A) * integral over A of (|x|^2 * I - x x^T) dA
  = diag(R^2/4 + d^2,  R^2/4 + d^2,  R^2/2)
```

The axial component already matches. The two lateral ones should carry `d^2` rather than `d^2/4`.

The quickest way to see it is the point limit. Let `R` go to zero and the exit plane collapses to a single mass flux location at distance `d`, so `S` has to tend to `diag(d^2, d^2, 0)`. Today it returns a quarter of that, and nothing about that limit depends on the rest of the model.

The older solid propulsion path already uses the full squared distance for the same quantity:

```python
# Flight.u_dot
- mass_flow_rate_at_t * ((nozzle_radius / 2) ** 2 + (c - b * mu / rocket_dry_mass) ** 2)
```

One smaller thing in the same function: the docstring says the tensor is in `kg*m²`. Since `T05 = mdot * S - I_dot` has to be kg·m²/s and `mdot` is kg/s, `S` is in m². I reworded that line and the tolerance comment in the existing test, which read `1e-3 * 1e-2 * 1e-2  # Equivalent to 1g * 1cm^2`.

## New behavior

The parallel axis term becomes `self.nozzle_to_cdm**2`. On the `calisto` fixture `S_11` goes from 0.3940207 to 1.5752660 m².

`S_noz` reaches the trajectory only through `T05 @ w` in `u_dot_generalized`, so it is scaled by mass flow rate and body rate. How much of it shows up depends on the rocket. On `flight_calisto_custom_wind`, a solid motor burning for 3.9 s, it is small. At apogee:

| | before | after |
|---|---|---|
| aerodynamic moment M1 | -0.509418 N·m | -0.519917 N·m |
| velocity x | -14.826366 m/s | -14.593411 m/s |
| apogee above ground | 3423.2700 m | 3423.8760 m |

On the Defiance example it is not small. That is a hybrid burning for 6.5 s, and the impact point moves from (1625.57, 81.79) to (1609.40, 87.03), roughly 17 m over a 1626 m range. That takes `test_defiance_rocket_matches_reference_flight_metrics` out of its band, and I re-recorded the two impact constants. There is more on that in the comments, including the check against the measured apogee, which the change moves closer rather than further.

Three recorded values in `tests/unit/simulation/test_flight.py` move past their 5e-3 tolerance and are updated. I re-recorded each tuple from the same run so it stays internally consistent, which is why the roll moment reads `-1.005368e-18` where it used to read `-2.089120e-14`. Both are numerical zero, and that component is unchanged in substance. `test_accelerations` stays inside tolerance and is left alone. Everything else in `tests/unit` and `tests/integration` gives the same result it does on `develop`.

## Breaking change

- [x] No

No API change. Simulation results do shift slightly, as above.

## Additional information

Two tests are added next to the existing one in `tests/unit/rocket/test_rocket.py`:

- `test_evaluate_nozzle_gyration_tensor_matches_the_exit_disk` is parameterized over `calisto`, `calisto_liquid_modded` and `calisto_hybrid_modded`, and checks all three diagonal entries against `R^2/4 + d^2` and `R^2/2` built from each rocket's own geometry, plus the off-diagonal zeros. It does not restate the recorded number.
- `test_evaluate_nozzle_gyration_tensor_parallel_axis_term` subtracts the disk contribution and checks what is left equals `nozzle_to_cdm**2`, which is the coefficient this patch changes.

I put the one line back to `0.25 * ...` to check the tests really hold it. All five nozzle tests fail, including the parameterized ones on the liquid and hybrid motors, and the other 61 tests in that file still pass.

While reading `u_dot_generalized` I ran into two other things that look independent of this change, so I filed them as #1185 and #1186 rather than fold them in here.


thc1006:
CI caught something I missed locally. I ran `tests/unit` and `tests/integration` but not `tests/acceptance`, and `test_defiance_rocket_matches_reference_flight_metrics` goes red.

Here is what this change does to that example:

| | `develop` | this PR |
|---|---|---|
| max_speed | 444.2545 | 444.2364 |
| max_acceleration | 10400.7640 | 10400.7640 |
| x_impact | 1625.5656 | 1609.4019 |
| y_impact | 81.7854 | 87.0258 |

`y_impact` lands outside its 3 m band. `x_impact` stayed inside, but only because `pytest.approx` takes the larger of `rel` and `abs`, so the band there works out at 16.26 m against a move of 16.15 m. A 0.1 m margin is not something I want to leave behind on a matrix that includes Windows and macOS.

The four `REFERENCE_*` constants agree with what `develop` produces to five significant figures, and the test's docstring says it guards the deterministic example's metrics, so I read them as recorded from the model rather than measured. The measured quantity in that file is `MEASURED_APOGEE_AGL = 9308.32`, and against that one the change goes the right way:

| | apogee AGL | error vs measured |
|---|---|---|
| `develop` | 9242.3701 m | 0.709 % |
| this PR | 9243.8350 m | 0.693 % |

The other 17 acceptance tests pass untouched, including the measured-flight comparisons for Bella Lui, NDRT 2020 and Prometheus.

I pushed 2953fd43 re-recording the two impact constants. Re-recording a guard someone else set is your call rather than mine, so if you would rather keep them pinned and take this a different way, say so and I will drop that commit.

While I have the numbers: on this rocket the correction is worth about 17 m of impact point, not the sub-metre it is worth on `flight_calisto_custom_wind`. Defiance is a hybrid burning for 6.5 s, so the mass flow rate stays large for longer and `T05 @ w` has more to act on. My first description understated that and I have fixed it.

thc1006:
Closing the one loose end from the description. I said `test_monte_carlo_simulate[True]` did not finish inside 40 minutes here and that I could not report it either way. I have since run the same test on unmodified `develop` on this machine and it behaves identically: the sequential case passes, the parallel one is still going at the 40 minute mark. So it is this machine rather than anything in the patch.

thc1006:
Pushed 22363439, a follow-up on the tolerance comment I added. `np.allclose` defaults to `rtol=1e-5`, which for a value near 1.575 works out at roughly 157 times the `1e-7` the comment describes, so the check was far looser than it read. It passes `rtol=0` now, and the recorded value carries one more digit, which leaves 61 times the margin rather than 3.

Gui-FernandesBR (approved):
code looks clean

PR #1190, opened 2026-09-04
Creation of the magnetometer sensor

aitorperezgrau-sys:
## Pull request type

## Checklist

## New behavior
The backbone of this PR is the creation of the magnetometer sensor. 

Unlike other sensors, the measured value (magnetic field) is not a result of the flight simulation, as a result I used the WMM (World Magnetic Model) to obtain the magnetic field, and use frame changes to adjust the magnetic field reading to the sensor position. 
When it comes to the noise and distortions included we can divide them into: 

1. **Magnetic distortion:**
To represent the magnetic distortions (hard iron distortion, soft iron distortion and power interference) I created auxiliary classes `Plate` and  `Wire`. Their distortions are applied in the Body Axis Coordinate System frame. Moreover, print and plot methods to validate their geometric position and its characteristics have been included together with new methods to satisfy their physical requirements.  

2. **Sensor Noise**
The noise related to the apparatus limitations has been covered using the existing `InertialSensor` and `Sensor` noise model and included in the measured value when the magnetic field is in the sensor frame. 

Furthermore, the documentation has been improved to include the new features and testing of every new element has been performed successfully. 

## Breaking change
- [ ] Yes
- [x] No

## Additional information
- This PR requires pywmm Python library to evaluate the World Magnetic Model (WMM). It has been added to requirements.txt.

- A change in the accelerometer sensor has been made, due to existing coordinate transformation error. 

- I opted for a software architecture that raises error when input parameters are incorrect because this approach allows users to know the cause of the error rather than seeing the error in an inner layer of the code. The obvious disadvantage is that for personal/private usage of the library there is no room for other parameters if desired; nonetheless, I have accounted for these having more flexible arguments. This is illustrated with the `ignition_wire_type`, which allows any type of strings, so that It can be used for HIL testing. 

- One of the functions implemented that were required to ensure a physically accurate model was the `general_radius` method. This method belongs to the rocket and returns the radius of the rocket, without the fins, for any value along the longitudinal axis. Also methods for adding plates and wires to the rocket have been implemented `add_wire` and `add_plate`. 

- The distortion created by wires is calculated using the Biot-Savart expression for finite wire, while the soft iron distortion modeled by the Plates is the sum of the distortions of the points that form the plate. 

- The current model supports plates that turns around the rocket, are bigger than the size of the rocket, have radius changes. All of these plate geometry characteristics are illustrated in the following image: 
nose_cone_with_plate

Other images that illustrate the creation of the plate are the following
image
image

- Focusing on code efficiency, the calculation of the magnetic distortion (by the wire, and the plate) is calculated once, when it is needed, and then stored in a dictionary with keys the position for which it has been calculated and value the magnetic field or soft iron distortion matrix computed. 

- Aiming to have a coherent naming style and better readability, almost all the variables/attributes that start with _ are Vector instances. 

- Plots of all plates and wires attached to the rocket have also been implemented in `_RocketPlot `(`draw_wires`, `draw_plates`). Due to code repetition in this class, a function `_rocket_shape_plot` that creates the outline of the rocket has been added. 
Example of `draw_wires` 
image
Example of `draw_plates`
image

- Supporting features were included in `sensor.py`, `nose_cone.py`, `tail.py`, `components.py`....

- The user defined coordinate system orientation and the center of dry mass are required for the changes in coordinate systems in `Wire` and `Plate` and the rocket instance to which they are attached is necessary for the corresponding plotting associated classes. As a result, the `_rocket_belonging` method is implemented in `Wire` and `Plate` and called when added to the rocket with `add_wire` `add_plate`. I am aware that cross referencing is not the best software practice, and it reduces the conceptual understanding of the inner library; however, I find it the best solution for this necessity.  

- All plotting methods have been implemented using Matplotlib

- The test "tests/unit/environment/test_environment_analysis.py::test_values" failed during local test, yet it is an atmospheric test unrelated to the new features. Outputs of Monte Carlo testing `.../monte_carlo_analysis/monte_carlo_analysis_outputs/mrs.inputs.txt` and `.../monte_carlo_analysis/monte_carlo_analysis_outputs/mrs.outputs.txt ` have been automatically included in the current version.


docs/reference/index.rst
Gui-FernandesBR, line 1: e and plates? what do those mean?

aitorperezgrau-sys: In rockets, magnetometers face three magnetic distortions beyond the sensor's intrinsic limitations. These are

- **Soft iron distortion**: caused by materials with higher magnetic permeability than vacuum. As a result, the magnetic field lines are forced to pass through these materials, altering the magnetic reading. To account for this magnetic distortion, I created the `Plate` class, which simulates a material sheet that is responsible for this bending of the magnetic field lines.

- **Power interference**: caused by wires in the avionics bay, which, due to current flow, generate a magnetic field, affecting the magnetic field reading. The power interference is, in turn, divided into communication interference, which comprises the magnetic field generated by wires with constant electron flow; and activation signal interference, which includes only the magnetic field generated by wires which fulfill some activation/pyrotechnic function (such as the one that activates the engine, or the one used to deploy the parachute). To account for this power interference, I created the `Wire` class.

- **Hard iron distortion**: caused by fixed components in the avionics bay, and it is simulated by a vector addition.

There is additional information about the formulation and usage of the `Wire` and `Plate` classes in the PR description and in their docstrings, let me know if you have more questions or if you would like me to clarify something.

PR #1144, opened 2026-08-13
ENH: Add tube-fin aerodynamic surface

ting-hong-shieh:
Closes #816.

## Pull request type

- [x] Code changes (bugfix, features)
- [x] Code maintenance (refactoring, formatting, tests)
- [x] ReadMe, Docs and GitHub updates

## Checklist

- [x] Tests for the changes have been added
- [x] Docs have been reviewed and updated
- [x] Lint has passed locally with Ruff and Pylint
- [ ] All tests have passed locally; four existing sensitivity tests require the optional `statsmodels` dependency, which is not installed in the local test environment
- `CHANGELOG.md` — no action needed; an LLM workflow auto-updates it after merge

## Current behavior

RocketPy has no built-in aerodynamic model for tube fins. Users must supply custom coefficients through `GenericSurface`, which requires external aerodynamic data and does not expose tube-fin geometry directly.

## New behavior

This PR adds a `TubeFins` aerodynamic surface and `Rocket.add_tube_fins`. The first implementation follows the Ribner ring-airfoil normal-force slope used by OpenRocket and includes:

- a normal-force coefficient derived from tube count, length, inner radius, outer radius, and rocket radius;
- a 20-degree angle-of-attack cap;
- a fixed quarter-chord center of pressure for Mach 0.5 and below;
- validation for at least three evenly distributed, uncanted tubes that touch the rocket body and both neighboring tubes;
- JSON serialization, top-level exports, rocket integration, prints, plots, tests, and user and technical documentation.

The model intentionally does not calculate Mach-dependent center of pressure above Mach 0.5, tube friction or pressure drag, cant and roll behavior, side force and yaw, or separated and overlapping tube layouts. Tube-fin drag must remain included in the rocket's power-on and power-off drag curves.

## Validation

- `18 passed` in the focused `TubeFins` tests.
- `200 passed, 7 skipped` across aerodynamic-surface, Rocket, and encoding tests.
- `2170 passed, 59 skipped` in the complete test suite; four sensitivity tests failed because the optional `statsmodels` dependency is not installed.
- Ruff formatting and lint checks passed.
- Pylint rated the new implementation and tests `10.00/10`.
- The new technical page and API reference passed targeted Sphinx builds with warnings treated as errors.

## Breaking change

- [x] No

PR #1194, opened 2026-09-14
Updates master with develop before releasing v1.14.0

Gui-FernandesBR:
It seems like we already have enough content to prepare a new release!
PRs #1187 and #1190 will also be included in this release.
All other PRs must wait for v1.15 or higher versions.


MateusStano:
I went through all the PRs. There is a LOT to fix, breaking changes included. Some should be reverted, some should be reverted but are still useful. There are a million changes to the Monte Carlo/Stochastic code and they really need some improvements.

Listing everything here:

### Revert before the release

- **#1128**: adds a new argument to the `Flight` constructor and conflicts heavily with `enh/events`. The events branch already covers this use case, so lets not add something we would have to deprecate later
- **#1129**: this is the exact design that was proposed and dropped in #558,  it is making the class components more complex, it technically a breaking change because of the interface with `rocket.aerosurface`, and it is adding/fixing nothing. The way it was on master was fine
- **#1133**: useful but the design is weird because it adds a new way of passing the trigger. Anyway, the trigger just needs the time as an input, and that is easy to do with the refactors on `enh/events` (just have to do `trigger = lambda context: context["time"] > 8`)
- **#1134**: parachute noise is already deprecated on `enh/events` (Sensors do noise much better and current implementation of parachute noise makes flight simulation slower), and this PR adds a new public `seed=` argument to `Parachute` just for it. We would have to deprecate it right after releasing
- **#1136** : from what I understand this is 400+ lines of code that nothing in the library uses, isn't exported and isn't documented. And there seems to be repeated code like how `set_atmospheric_model(type="Forecast", file="GFS")` already gets the latest run through THREDDS.
- **#1137** there are a lot of issues with this one
  1. The cache decides if a file is fresh only by its age (6 h by default), not by which model run it came from. The run time isn't even in the file name. So with GFS you can keep getting the previous run for up to 6 more hours after a new one is out, and for HRRR/RAP, which publish every hour, you can be ~6 runs behind. The 6 h limit is the same for every model
  2. Nothing tells you a cached file was used, and the file doesn't record which run it holds. For a launch day go/no-go this is the worst case: the numbers look fine and are just old
  3. It is on by default, so every forecast download now silently writes to `~/.rocketpy_cache/atmosphere`. None of this is in the docs (the folder, the 6 h limit, `ROCKETPY_CACHE`/`ROCKETPY_CACHE_TTL`). The only mention is the `no_cache` parameter description
  4. It is a lot of lines of code

  `Environment.to_dict`/`from_dict` already store and restore the atmosphere without downloading again, so maybe there is a much simpler way to do this. If we ever want automatic caching, it needs at least the model run time in the file name, a limit per model (1 h for HRRR/RAP, 6 h for GFS) and a message whenever a cached file is used...

- **#1144**: i would really want to see some validation of the equations to have this in our code. All other aero surfaces have their coefficinents validated in several papers/articles. I didn´t search for it but I don´t know if this tube fins does. I checked the equations, and they are only valid for rings much wider than they are long. Tube fins are usually about as long as they are wide, so there are geometrical restrictions to the formulas. The formulas are also only valid for mach < 0.5. There is also a stall limit in this implementation, which I think just comes from copying all the formulas from open rocket, but we don´t have stall at all in rocketpy so it really shouldn´t be here. Open rocket also computes roll damping for tube fins, which this implementation is not doing. It would be nice to have tube fins implemented, but it needs more work. Also this would conflict with enh/aero-refactor so I would just leave it for later.

### Keep, but fix/improve before the release

- **#1138**: physics is fine and the default didn't change, but:
  1. Missing the new parameter in MonteCarlo
  2. The `grains_bonded` docstring needs a rewrite in simpler language:
     - remove `_csys` from the formula (private attribute), just say "toward the nozzle"
     - remove "DEM contacts" and the "Follow-ups may add..." sentence
     - don't call bonded grains "BATES-style". BATES is a grain shape, a BATES motor can be bonded or not
     - say that `grain_separation` is treated as a solid spacer, not an air gap
  3. `docs/user/motors/solidmotor.rst` doesn't mention the option at all. Add a short section with an example and saying that the effect is usually small (~0.03 calibers of static margin on Calisto)
- **#1166**: the fix works, but `_compose_clipped`  could be removed and the fix should just be in the Function class:
  1. `_compose_clipped` clips any amount, not only roundoff, and it only covers the 6 places in `tank.py` that call it. The same roundoff problem can show up in any other `compose` call
  2. in `Function.compose`, let the inner values be outside the domain by a tiny tolerance (1e-9 of the domain size) and clip only that. Anything bigger still raises the error as it does today
  3. delete `_compose_clipped` and go back to plain `compose` in those 6 calls
 
- **#1176**: The new fixture is good, but the tests are now back in the default run and add ~45 s to every full test run, just to parse 20 years of reanalysis data. This is just making tests much slower for no reason since very rarely we ever touch EnvironmentAnalysis. Id prefer to just keep them as `run slow`, but since we are here I think we can improve things further to actually fix the slowness issue:
  1. commit a small cut of the two EuRoC `.nc` files with only what the fixture uses (2019-10-23 to 2021-10-23, 6h to 18h, the grid points around the launch site) and point the fixture to them. The expected values in the tests shouldn't change
  2. if that doesn't get the tests time down significantly, put the `slow` marks back and keep only the fixture change
  3. without the optional dependencies installed, 5 of the 10 tests now fail instead of being skipped. Add `pytest.importorskip` for `windrose`, `ipywidgets` and `jsonpickle`, and set `timezone="UTC"` in the fixture (`test_values` gives a different number when `timezonefinder` is missing)
- #1141: Works, but:
  1. The file only loads for the date and hour the Environment had when you created it. Change the launch time by one hour and loading fails with "Chosen launch time is not available in the provided file". That check makes sense for a real forecast, but our own profiles usually don't have a date at all (what-if winds, old soundings), so the file can't really be reused. 
  2. For the same reason it raises an error if the Environment has no date, even though the profiles don't need one
  3. It always saves a file (custom_ensemble.nc in the current folder by default), so running the same notebook cell twice gives FileExistsError. Would be nice to have an option to not save anything and keep it in memory
  4.  Below the lowest point of the profile the values are just held constant. That is ok, but the docs should tell people to start their profiles at or below ground level

- **#1142** : the docs are correct. They need three short notes in the new section of `rocket_usage.rst`:
  - the Reynolds number uses the rocket **diameter** as the reference length. Lots of tools use the length...
  - alpha and beta are separate signed angles (pitch and yaw planes), neither is the total angle of attack. Drag that depends on the total angle should combine both, like `alpha**2 + beta**2`. The example only uses alpha
  - the CSV needs every combination of the grid filled. If a point is missing the file still loads, but a less accurate interpolation is used and the values change

### After the release

- **#1161**: importing a results file still changes where `simulate()` writes. The PR blocks the bad case, but the real fix is that importing shouldn't touch the output file

### Monte Carlo / Stochastic (fix before the release)

There are 1800 new lines in the lib and ~5000 in tests relating to Monte Carlo. It is mostly around seeding, which makes me thing something is wrong because seeding is not something complex to implement. Most of the code is way more complicated than it needs to be and I am not even sure some of these features are needed for anything. I am pointing out some problems and some solutions here, but the more I looked at it the more issues I saw so I stoped here:

**Problems**

1. #1187 is a breaking change, and the changelog only tells part of it:
   - a seed given to a Stochastic object (`StochasticRocket(..., seed=42)`) does nothing in `MonteCarlo` now, because everything is reseeded at every simulation from `random_seed`. A seeded serial run was reproducible in v1.13.0 and now it isn't, unless you pass `random_seed`. There is no warning and it isn't in the changelog
   - every result with a fixed seed changes
   - `simulate(append=True)` on files from v1.13.0 raises an error, so you can't continue an old study
   - serial runs are numbered `0..n-1` now, it was `1..n`
   - every row has a new `run_root` field
2. `run_root` ends up in `mc.results` as a list of strings. `plots.all()` makes a boxplot of every result, and boxplot fails on strings. I only tested the matplotlib call, not a full run, but I think `mc.plots.all()` crashes on any new run
3. `simulate_convergence` doesn't have `random_seed`, so it can't be seeded
4. `stochastic.rst` still says that serial runs never reseed and that parallel runs fail to. Not true anymore
5. The parallel code from #1182 is ~450 lines. Each worker writes the files itself, so now we need a shared lock, checking exit codes every 0.2 s, asking/terminating/killing stuck workers, and reading both files again at the end to make sure nothing was lost. All of this only exists because a worker can die in the middle of writing
6. #1187 saves the seed as four numpy internal values in every input row, plus a SHA hash in every output row, plus ~100 lines just to read it back safely. Also the same files are now read by three different functions
7. #1169 added ~120 lines (and 617 of tests) to stop the nominal values from drifting. The real cause is that `_set_stochastic` validates everything again at every reseed, and it reads the nominal from an object that `create_object` already changed. There is a TODO in the code saying exactly that
8. There are four helper functions just to convert seeds between `SeedSequence` and int (two of them are the same function). The seeds are 128 bit ints now, which breaks any user `CustomSampler` that uses `RandomState`
9. #1161 blocks `.csv`/`.json` log files, but the cause is still there: `import_outputs("x.csv")` changes where the next `simulate()` writes. It also added a few checks that nobody will ever hit (same file by inode, `indent`, `separators`)
10. #1111 changed every `assert x, msg` to `if not x: raise AssertionError(msg)`. These should be `ValueError`/`TypeError`
11. `StochasticSolidMotor` doesn't pass `grains_bonded` (see #1138 above)

**What to change**

1. Workers only run the simulation, the main process writes the files. Since #1187 a simulation only depends on its index, so this can be a simple pool: each task seeds, runs and returns the two rows, and the main process saves them. If a worker fails we get its error, and if it is killed the pool raises. Almost everything from problem 5 can be deleted, and most of the four `test_monte_carlo_worker_*`/`run_completeness` test files (~940 lines) go with it. Serial and parallel can also share the same loop
2. Use one integer as the seed. `random_seed` is an int, and if it is None we draw one and print it. The seed of simulation i is `np.random.SeedSequence([random_seed, i])`. Save `"seed": <int>` in the input rows only, and keep it (and `index`) out of `results`. To append, read the seed from the last row and continue from the number of rows. Appending to old files should give a warning, not an error. One function to read the files instead of three
3. Validate only once, in `__init__`. A reseed should only replace the random generators. For that we store the name of the distribution (`(nominal, std, "normal")`) instead of a function tied to a generator. Then everything from problem 7 can be deleted, and the reseed gets faster, which matters now that it happens at every simulation
4. Convert the seed to a `SeedSequence` once and pass it down (`default_rng` accepts it). Give a normal 32 bit int to `CustomSampler.reset_seed`, so `RandomState` samplers keep working
5. Importing a file should never change where `simulate()` writes, and should never create files. With that fixed, the checks from #1161 aren't needed
6. Warn when a Stochastic object has a seed and `random_seed` is None, add `random_seed` to `simulate_convergence`, fix `stochastic.rst`, and write a clear breaking changes note in the changelog
7. `ValueError`/`TypeError` instead of `raise AssertionError`
8. Clean the comments. About a quarter of the new lines are long comments telling the history of the code ("used to...", "#1070 added..."). One line saying why is enough

### Related, not merged yet

- **#1195** should go in before the release. `solid_propulsion` EOMs were zeroing the angular rates, so there is no aerodynamic damping at all. With canted fins the roll rate grows without bound and the simulation basically hangs

MateusStano:
And I just saw #920. Id say its best to revert that and implement it together with enh/events if that is not a problem

tests/acceptance/test_defiance_rocket.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+MEASURED_APOGEE_AGL = 9308.32
+MAX_RELATIVE_APOGEE_ERROR = 0.01
+REFERENCE_MAX_SPEED = 444.24
+REFERENCE_MAX_ACCELERATION = 10400.76
+REFERENCE_IMPACT_X = 1609.40
+REFERENCE_IMPACT_Y = 87.03
+REFERENCE_METRIC_RELATIVE_TOLERANCE = 0.01
+REFERENCE_IMPACT_ABSOLUTE_TOLERANCE = 3.0
```
Gui-FernandesBR: only pending double-checking these values

CHANGELOG.md
Gui-FernandesBR: Before merging onto develop I think we should verify if the CHANGELOG is in a perfect condition. 

1. Check if all the PRs are included here
2. Check if every line here is a PR not a commit message

rocketpy/sensors/accelerometer.py
```diff
@@ -232,7 +239,7 @@ def measure(self, time, **kwargs):
         gravity = (
             Vector([0, 0, -gravity]) if self.consider_gravity else Vector([0, 0, 0])
         )
-        inertial_acceleration = Vector(u_dot[3:6]) + gravity
+        inertial_acceleration = Vector(u_dot[3:6]) - gravity
```
Gui-FernandesBR: Please double check this signal change. Ther should be somewhere else in the code telling whether this was a right or a bad modification. I'm not sure.

rocketpy/stochastic/custom_sampler.py
```diff
@@ -8,6 +8,23 @@
 class CustomSampler(ABC):
     """Abstract subclass for user defined samplers"""
 
+    @property
+    def seed_group(self):
```
Gui-FernandesBR: I wonder how important it is to have a method returning "self". I don't really see the need to it... Maybe we could simply delete these lines.

tests/integration/environment/test_environment_analysis.py
Gui-FernandesBR: I wonder how much execution time we are adding to every test suite execution by marking these slow tests as not slow.

rocketpy/motors/solid_motor.py
```diff
@@ -728,14 +789,12 @@ def propellant_I_11(self):
...
-        d = d * (self.grain_initial_height + self.grain_separation)
-
-        # Calculate inertia for all grains
-        I_11 = grain_number * grain_inertia11 + grain_mass * np.sum(d**2)
+        # Parallel-axis term from grain COM offsets about the propellant COM.
+        # Bonded: fixed initial pitch. Unbonded: packed pitch tracks grain_height.
+        I_11 = (
+            grain_number * grain_inertia11
+            + grain_mass * self._grain_pitch_squared_sum()
+        )
```
Gui-FernandesBR: pending a validation on the inertia moment of this implementation

rocketpy/sensors/sensor.py
```diff
@@ -2,13 +2,55 @@
...
 
 logger = logging.getLogger(__name__)
 
+# The seed kinds that describe a stream, and so survive a to_dict() round trip.
+# Named once because every concrete sensor repeats it in its signature; it is a
+# union, so help() and inspect.signature() still show the members in full.
+SeedLike = int | np.integer | Sequence[int] | np.ndarray | np.random.SeedSequence
+
+
+def _is_int_array_like(value):
```
Gui-FernandesBR: the `_is_int_array_like` is a simple implementation, however it is so simple that I wonder if numpy (or similar) already have it implememnted.

It could also be positioned in other file as well, for example the tools.py module.

rocketpy/sensors/sensor.py
```diff
@@ -151,6 +206,40 @@ def __init__(
...
+        # snapshot of the moment it ran rather than the stream the sensor used.
+        # #1124 taught RocketPyEncoder to serialize a SeedSequence, which stays
+        # reproducible because it is defined by its entropy and spawn key; a
+        # live generator has no such description to write.
+        #
+        # The check names the descriptors instead of the live types because the
+        # descriptors are the stable half of that contract: a seed kind numpy
+        # starts accepting later is refused here rather than reaching
+        # json.dumps(). Without it the sensor builds fine and only fails at
+        # serialization, far from the call that caused it.
```
Gui-FernandesBR: complex long comment alert

rocketpy/_encoders.py
```diff
@@ -146,6 +160,32 @@ def object_hook(self, obj):
             return obj
 
 
+def rebuild_minimal_flight(class_, obj, hash_):
```
Gui-FernandesBR: I would like to understand better how the `rebuild_minimal_flight` is called in the rest of the code.

rocketpy/_encoders.py
```diff
@@ -110,14 +121,17 @@ def object_hook(self, obj):
...
 
+                if class_ is np.random.SeedSequence:
+                    # Cython __init__ has no __code__, so the generic kwargs
+                    # path cannot rebuild SeedSequence; restore from state.
+                    return np.random.SeedSequence(
+                        entropy=obj.get("entropy"),
+                        spawn_key=tuple(obj.get("spawn_key", ())),
+                        pool_size=obj.get("pool_size", 4),
+                        n_children_spawned=obj.get("n_children_spawned", 0),
+                    )
```
Gui-FernandesBR: Although it is probably true, we don't have any tests to hold these comment lines. Implementing tests to cover a Cython behavior might be almost impossible, right?

rocketpy/environment/environment.py
Gui-FernandesBR: This file is adding 1400 lines and removing only 40.

We should be careful because it's a really high addition.

Wouldn't be better if we split the file into smaller ones?

rocketpy/prints/environment_prints.py
```diff
@@ -102,7 +102,17 @@ def atmospheric_model_details(self):
             f"{model_type} Maximum Height: "
             f"{self.environment.max_expected_height / 1000:.3f} km"
         )
-        if model_type in ["Forecast", "Reanalysis", "Ensemble"]:
+        # set_atmospheric_model accepts the type case-insensitively and stores it
+        # as the user spelled it, so compare in lower case while still printing
+        # the original spelling above.
```
Gui-FernandesBR:
```suggestion
```

docs/static/flight/trajectory_on_map.jpg
Gui-FernandesBR: is this file being loaded in the documentation? where?

if not we should simply delete it.

.github/scripts/test_update_changelog.py
Gui-FernandesBR: How important it is to have a test for the changelog? If we just remove it and let the changelog file fail, wouldn't it work the same way?

docs/user/aerodynamics/surfaces.rst
```diff
@@ -178,6 +181,31 @@ Parameters:
...
+- Lift is capped at an absolute angle of attack of 20 degrees.
+- At least three uncanted tubes must be distributed evenly around the body.
+  Every tube must touch the body and both adjacent tubes. Separated and
+  overlapping tube layouts are rejected.
+- The model does not calculate tube-fin friction drag, pressure drag, roll,
+  side force, or yaw. Include tube-fin drag in the rocket's power-on and
+  power-off drag curves.
+
+For the equations and geometry constraint, see
+:doc:`Tube Fin Aerodynamics </technical/aerodynamics/tube_fins>`.
```
Gui-FernandesBR: the limitations are listed at least. 

We just need to double check whether they are limitations from our implementation itself or if what we have implemented is already the best available option out there.

rocketpy/environment/fetchers/base.py
Gui-FernandesBR: We have the same "MAX_RETRY_DELAY_SECONDS "  value for every single API fetcher. 
Wouldn't it be better if we use different values for different APIs?

